PR #7, opened 2017-07-25
Addition of custom launchers documentation

omlins:


writing_checks/README.md
```diff
@@ -468,6 +468,38 @@ whereas if it run on a system with ALPS it will be translated to
 ddt --offline aprun ...
 ```
 
+## Custom Launchers
+
+ReFrame permits the simple addition of custom scheduler launchers. A launcher that invokes `mpirun` can for example be added as follows:
```
vkarak: Put each sentence in a separate line. This way we avoid conflicts in the future, or we can resolve them faster if they occur.

vkarak: I think we should explain what a launcher is. If I'm not mistaken, we're not saying anything yet.

writing_checks/README.md
```diff
@@ -468,6 +468,38 @@ whereas if it run on a system with ALPS it will be translated to
...
+class MpirunLauncher(JobLauncher):
+    @property
+    def executable(self):
+        return 'mpirun'
+
+    @property
+    def fixed_options(self):
+        return [ '-np %s' % self.job.num_tasks ]
+```
+It will be translated to `mpirun -np ...`. While the definition of the property `executable` is obviously mandatory, the definition of `fixed_options` is optional; it defaults to no options. Note that the launcher class may use the information contained in `self.job`.
```
vkarak: It will be translated to `mpirun -np <num_tasks>`

vkarak: Here the reference to `self.job` is a bit misleading. You should somehow explain that a launcher is associated with a job descriptor, from which it can extract the job submission information.

writing_checks/README.md
```diff
@@ -468,6 +468,38 @@ whereas if it run on a system with ALPS it will be translated to
...
+
+```python
+class MpirunTest(RegressionTest):
+    ...
+    def setup(self, system, environ, **job_opts):
+        super().setup(system, environ, **job_opts)
+        self.job.launcher = MpirunLauncher(self.job)
+```
+
+ReFrame will provide a collection of custom launchers (added uppon request). Currently it provides only one: a launcher for VisIt. The VisIt launcher can be used in a regression test the same way as the above `MpirunLauncher`; only the following import statement is required to make it available:
```
vkarak: I would convert the the VisIt to a link to the project's site. I would also add "Currently it provides only one: a launcher for the VisIt visualisation software".

writing_checks/README.md
omlins: OK, done.

omlins: Yes, that might be unclear to some readers. 
The first occurrence of the word 'launcher' is in the subsection `## Executable Options`, so one might could add it there. However, as it is a frequent word in the whole section `# Customising the Run of a Test`, such that I think it could be better to add it right there (after `ReFrame offers several other options for customizing the behavior of regression tests.` and with a proper transition to it).
What do you think?

writing_checks/README.md
omlins: OK, done.

writing_checks/README.md
omlins: OK, done.

writing_checks/README.md
vkarak: I think the description (a single sentence could be enough) is better to be put here, since here is the point where (a) we elaborate a bit more on launchers and (b) the user here will need to know what a launcher really is.

vkarak: Until we introduce the versioning of the documentation, you should add here "From version 2.5 onward, Reframe [...]"

omlins: OK, done.

PR #8, opened 2017-07-27
Document execution modes

vkarak:


running/README.md
```diff
@@ -590,6 +590,49 @@ This can be achieved by setting the `max_jobs` property of a partition in the `s
 
 ```
 
+## Execution modes
+
+Reframe 2.5 permits you to define different *execution modes* of the framework.
```
victorusu: Would it be more precise if we state something like:
Starting from version 2.5, ReFrame ...

What do you think?

running/README.md
```diff
@@ -590,6 +590,49 @@ This can be achieved by setting the `max_jobs` property of a partition in the `s
...
+        ...
+    },
+    'environments' : {
+        ...
+    },
+    'modes' : {
+        *' : {
+            'maintenance' : [
+                '--exec-policy=async',
+                '--output=$APPS/UES/$USER/regression/maintenance',
```
victorusu: Should we avoid to use out own paths here?
Why not:
`$APPS/regression/maintenance'`

The $USER/$USER is not needed. The same applies to `--logdir` and `--stage`
What do you think?

running/README.md
vkarak: Sure.

running/README.md
vkarak: For the first part, yes. For the second part I think it's good to have an environment variable somehow, so that we show that they are treated correctly. We don't say it in the text, but if they are there they, it is at least implied. I could though add it explicitly to the text.

PR #13, opened 2017-09-14
Addition of doc for sourcesdir equals None

omlins:


reference/README.md
```diff
@@ -112,7 +112,7 @@ If `propagate` is set to `False`, no flags will be passed to `make`.
 | `reference` | The set of reference values for this test. | `{}` | scoped dictionary | 2.2 |
 | `sanity_patterns` | The set of sanity patterns to look for for this test. | `None` | See [Output parsing](#output-parsing-and-performance-assessment) | 2.2 |
 | `sourcepath` | The path to the source file or source directory of the test. If not absolute, it is resolved against the `sourcesdir` directory. If it refers to a regular file, this file will be compiled (its language will be automatically recognized) and the produced executable will be placed in the test's stage directory. If it refers to a directory, this will be copied to the test's stage directory and `make` will be invoked in that. | `''` | string | 2.2 |
-| `sourcesdir` | The directory containing the test's resources. | `<test-prefix>/src/` | string | 2.2 |
+| `sourcesdir` | The directory containing the test's resources. In a `RunOnlyRegressionTest`, you can set this field to `None` if you do not need any resources at all. | `<test-prefix>/src/` | string | 2.2 |
```
vkarak: You should also add a note in the "NOTES" section later that the behaviour has changed from version 2.5 onward.

writing_checks/README.md
```diff
@@ -435,12 +435,17 @@ def setup(self, system, environ, **job_opts):
...
 ```python
 def __init__(self, **kwargs):
     ...
     self.sourcesdir = '/apps/input/folders'
 ```
+In a `RunOnlyRegressionTest`, you can set 
+```python
+self.sourcesdir = None
+```
+if you do not need any resources at all.
```
vkarak: Better format it as follows:

In a `RunOnlyRegressionTest`, you can set `sourcesdir` to `None` if you do not need any resources at all:
```python
self.sourcesdir = None
```

omlins: Done.

reference/README.md
omlins: I added a note.

vkarak:
Thanks @omlins!

PR #14, opened 2017-09-15
Correcting the position of the EOF-check in the parser example

kraushm:
Fixes Gitlab issue #336


vkarak:
Good catch, since it was the example that it was actually wrong!

PR #123, opened 2018-02-20
Add prefix 'rfm_' to the given job name

### LINKED ISSUE #122: Adopt a test naming convention for jobs submitted by ReFrame
We could add the `rfm_` prefix to the jobs' names.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/161

teojgo:
* Make minimal unittest changes to reflect the change
  in the job name.

Closes #122

### DIFF
--- a/unittests/test_launchers.py
+++ b/unittests/test_launchers.py
@@ -128,8 +128,8 @@ class TestSrunallocLauncher(_TestLauncher):
         return ('srun '
                 '--job-name=rfm_fake_job '
                 '--time=0:10:0 '
-                '--output=rfm_fake_job.out '
-                '--error=rfm_fake_job.err '
+                '--output=fake_job.out '
+                '--error=fake_job.err '
                 '--ntasks=1 '
                 '--foo '
                 'ls -l')
--- a/unittests/test_launchers.py
+++ b/unittests/test_launchers.py
@@ -102,7 +102,7 @@ class TestSrunallocLauncher(_TestLauncher):
     @property
     def expected_command(self):
         return ('srun '
-                '--job-name=fake_job '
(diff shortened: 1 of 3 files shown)


unittests/test_launchers.py
```diff
@@ -126,10 +126,10 @@ def expected_command(self):
     @property
     def expected_minimal_command(self):
         return ('srun '
-                '--job-name=fake_job '
+                '--job-name=rfm_fake_job '
                 '--time=0:10:0 '
-                '--output=fake_job.out '
-                '--error=fake_job.err '
+                '--output=rfm_fake_job.out '
+                '--error=rfm_fake_job.err '
```
vkarak: Only the job name should be prefixed.

vkarak: Basically, you have to change the unit tests and pass explicitly the `stdout` and `stderr` arguments to the `FakeJob`. We need to test the possibility for these being different, because this is how it is called from the `RegressionTest`.

teojgo: I think that passing explicitly `stdout` and `stderr` is what `self.job` does. In case of the `self.minimal_job` only the required arguments are given and thus the `stdout` and `stderr` are named according to the job name.

vkarak: They should be passed as well to emulate the desired behaviour. We are not expecting the stdout/stderr to be prefixed.

teojgo: Then the passed name and not the actual `job` field should be passed to the `stdout/stderr` if they are `None` in the setup.

teojgo:
@jenkins-cscs retry daint leone kesch monch

PR #128, opened 2018-02-21
Submit jobs from the stage directory

### LINKED ISSUE #35: The job script generators change too late to the working directory
Quoting from @bcfriesen:

> Actually, now there is a new issue related to this.

> When using DataWarp, the scheduler creates a new temporary directory on the DataWarp file system which is pointed to by an environment variable. For "striped" access, the environment variable is $DW_JOB_STRIPED. The data specified by the #DW stage_in header in the batch script is automatically copied to $DW_JOB_STRIPED. The remainder of the batch script typically involves copying the executable into $DW_JOB_STRIPED, because the value of that environment variable changes with every run (so it can't be hard-coded into the batch script, or into the code itself).

> I tried to address this by writing a simple bash script which copies the executable into $DW_JOB_STRIPED:

> ```#!/bin/bash
> EXE=a.out
> cp $EXE $DW_JOB_STRIPED
> cd $DW_JOB_STRIPED
> ```
> and putting it into the self.job.pre_run stage as described here. The issue is that the auto-generated Slurm script will execute that bash script, but then follow it immediately with cd /path/to/stagedir, as specified by the stagedir in the _site_configuration. So the test never actually tests the functionality of DataWarp because it reads everything from stagedir, which is not on the DataWarp file system.

Yes, this is indeed a problem of our job script generator. The `cd` to the working directory should be done at the very beginning or we should try using the `--workdir` option of Slurm instead.

### LINKED ISSUE #91: Submit jobs from the stage directory
Currently, we submit jobs from the directory where ReFrame is launched and we insert a `cd /to/stagedir` in the generated script. This creates some complications as described in #35 and leads to a bit cumbersome tests and, more importantly, to not easily reproducible problems. Submitting jobs from the stage directory will make the generation of the job script more straightforward and the job script itself more readable. Also, with the deprecation of the `Job`'s `pre_run` attribute, there is no way to obtain and set a variable using the stage directory, unless the `self.stagedir` becomes a deferrable.

vkarak:
This PR also adds a tutorial example for using `pre_run` and `post_run`.

Fixes #91.
Fixes #35.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -398,23 +398,21 @@ Specifically, we would like to execute the following shell script and check that
 
 .. code-block:: bash
 
-  #!/usr/bin/env bash
+   #!/usr/bin/env bash
 
-  for i in {1..100}
-  do
-      echo Random: $((RANDOM%($UPPER+1-$LOWER)+$LOWER))
-  done
+   if [ -z $LOWER ]; then
+       export LOWER=90
+   fi
+
+   if [ -z $UPPER ]; then
+       export UPPER=100
+   fi
+
+   for i in {1..100}; do
+       echo Random: $((RANDOM%($UPPER+1-$LOWER)+$LOWER))
+   done
 
 The above script simply prints 100 random integers between the limits given by the variables ``LOWER`` and ``UPPER``.
-For this example the above limits are exported as environment variables by the ``limits.sh`` script as follows:
-
-.. code-block:: bash
-
-  #!/usr/bin/env bash
-
-  export LOWER=90
-  export UPPER=100
-
 In the corresponding regression test we want to check that all the random numbers printed lie between 90 and 100 ensuring that the script executed correctly.
 Hence, a common sanity check has to be applied to all the printed random numbers.
 In ReFrame this can achieved by the use of :func:`map <reframe.utility.sanity.map>` sanity function accepting a function and an iterable as arguments.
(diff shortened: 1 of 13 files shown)


docs/advanced.rst
```diff
@@ -483,8 +476,92 @@ The full syntax for the :attr:`sanity_patterns` is the following:
...
-In this case, we make sure that the ``limits.sh`` file is sourced before executing the ``advanced_example6.sh`` script.
+
+Customizing the Generated Job Script
+------------------------------------
+
+It is often the case that you must run some commands before and/or after the parallel launch of your executable.
+This can be easily achieved by using the :attr:`pre_run <reframe.core.pipeline.RegressionTest.pre_run>` and :attr:`post_run <reframe.core.pipeline.RegressionTest.post_run>` attributes of :class:`RegressionTest`.
+
+The following example is a slightly modified version of the previous one.
+The lower and upper limits for the random numbers are now set inside a helper shell script in ``scripts/limits.sh`` and we want also to print the word ``FINISHED`` after the parallel job has finished.
```
teojgo: Maybe refer to the job as 'submitted' instead of 'parallel'.

vkarak: Here, I mean the executable, so "submitted job" is more confusing, I think. I'll see how I can rephrase it.

vkarak:

PR #129, opened 2018-02-21
Fix emission of Slurm's --exclusive flag

### LINKED ISSUE #127: Slurm's --exclusive is always set in the job scripts
The problem comes from how we emit the job options in preamble. We emit them iff the corresponding `Job` attributes are not `None`. However, this option is always set to `False` (i.e., not `None`) and, thus, it is always emitted.

vkarak:
Also added the associated unit tests to avoid future regressions.

Fixes #127.


vkarak:

PR #132, opened 2018-02-22
Remove 'pre_run' and 'post_run' access from 'job'

### LINKED ISSUE #112: Remove use of `self.job.pre_run` and `self.job.post_run` in regression tests
Internal issue: https://madra.cscs.ch/scs/reframe/issues/579

### LINKED ISSUE #136: Adapt CDO/NCO tests to the new Slurm behaviour
All the CDO/NCO tests that require empty output are now failing due to the statistics information that it is now printed out in the standard output by `srun`.

teojgo:
* Remove the `setup` method when used only to access the `pre_run`
  and `post_run` from `job`.

Fixes #112 
Fixes #136

### DIFF
--- a/cscs-checks/tools/io/cdo.py
+++ b/cscs-checks/tools/io/cdo.py
@@ -32,9 +32,13 @@ class CDOBaseCheck(RunOnlyRegressionTest):
         self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc',
                               'kesch:pn', 'kesch:cn']
         self.valid_prog_environs = ['PrgEnv-gnu']
-        self.modules  = ['CDO']
         self.maintainers = ['SO']
         self.tags = {'production'}
+
+    def setup(self, partition, environ, **job_opts):
+        cdo_name = 'cdo' if self.current_system.name == 'kesch' else 'CDO'
+        self.modules = [cdo_name]
+        super().setup(partition, environ, **job_opts)
 
 
 # Check that the netCDF loaded by the CDO module supports the nc4 filetype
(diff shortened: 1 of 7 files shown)


teojgo:
@jenkins-cscs retry daint monch leone kesch

cscs-checks/tools/io/cdo.py
```diff
@@ -170,11 +168,13 @@ def __init__(self, **kwargs):
...
-                                'test_echam_spectral-deflated_wl.nc4c',
-                                'test_echam_spectral-deflated_ws.nc4c',
-                                'test_echam_spectral-deflated_wind10_wl_ws.nc4c']
+        self.executable_opts = [
+            '-O', '-z', 'zip', 'merge',
+            'test_echam_spectral-deflated_wind10.nc4c',
+            'test_echam_spectral-deflated_wl.nc4c',
+            'test_echam_spectral-deflated_ws.nc4c',
+            'test_echam_spectral-deflated_wind10_wl_ws.nc4c'
+        ]
```
vkarak: The old formatting was also correct. I suggest leaving it as it was.

vkarak:
NCO/CDO tests are failing due to #136. You can either fix them in this PR or with a separate one. It's up to you.

teojgo:
@vkarak There is also a failure due to the difference modules names for `nco` and `cdo` on kesch. I have fixed that and will fix #136 as well.

cscs-checks/tools/io/cdo.py
teojgo: The last list item exceeded 79 characters that's why I did it.

teojgo:
@jenkins-cscs retry daint leone monch kesch

cscs-checks/tools/io/nco.py
```diff
@@ -69,15 +73,17 @@ def __init__(self, **kwargs):
         super().__init__('cdo_module_compat', **kwargs)
         self.descr = ('verifies compatibility with the CDO module')
         self.sourcesdir = None
-        self.executable = 'echo'
+        output_file = 'nco_cdo_check.out'
+        self.executable = 'echo > %s' % output_file
         self.sanity_patterns = sn.all([
-            sn.assert_not_found(r'.+', self.stdout),
+            sn.assert_not_found(r'.+', output_file),
```
vkarak: This is an identity, cos `output_file` will always be empty. The original intent of the test was to just check that the CDO can be loaded without conflict (see comment above). Since `module`  commands print to the standard error, it's sufficient to check that only this is empty or it does not contain the work `ERROR`.

@omlins Reading again the comment now, I wonder why this test should run on the compute nodes as well. If it was to run only on the login nodes, there would be no need for the `pre_run` and `echo` trick, would it?

omlins: I think it should be run on the compute node to make sure that there it works, because there is where NCO and CDO must be run normally. The compute nodes and login nodes are supposed to be in sync, but sometimes differences may appear...

vkarak: Personally, I find it a bit redundant to run this on the compute nodes, cos it's sth completely related to the environment. Modules and homes are shared on the compute nodes, so they cannot be unsynced, unless there is some more fundamental problem which would come up earlier. Anyways, I've just proposed to run on the login nodes, just to make test simpler.

vkarak:
@omlins @teojgo I'm gonna merge this one.

PR #133, opened 2018-02-22
Module mapping file and corresponding command line option

### LINKED ISSUE #45: Command line option for trying different module versions
Sysadmins usually come up with requests, like, does the regression works with the new (non-default) version of sth? Currently, there is no easy way to test this. One would need to modify the affected user checks or change the default module.

This issue proposes the addition of a new command line option, e.g., `--try-module modname=new_modname` that would load `new_modname` instead of `modname`. For example, if we need to try a new version of NAMD, say NAMD/2.12, we would do it easily as following:

```
./run_regression.py --try-module NAMD=NAMD/2.12 -n namd_gpu_check
```

Internal issue: https://madra.cscs.ch/scs/reframe/issues/158

### LINKED ISSUE #76: Module mapping file and corresponding command line option
A way should be defined for specifying module mappings. The most straightforward is to have files, that the user could specify in the command line, in the following form:

```
gromacs GROMACS/2016.1
amber Amber/16
```

Internal issue: https://madra.cscs.ch/scs/reframe/issues/552

rsarm:
Closes #76.
Closes #45.


rsarm:
It's still needed a fix on `load_mapping_from_file` to consider the case when there is a space between two modules instead of a coma:
```
m0: m1 m2
```

reframe/core/modules.py
```diff
@@ -4,6 +4,7 @@
 
 import abc
 import itertools
+import sys
```
vkarak: This seems not to be needed.

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
     def _is_module_loaded(self, name):
         return self._backend.is_module_loaded(Module(name))
 
+    def load_mapping_from_file(self, module_mapping_file):
+        """Update ``self.module_map`` with a mapping defined in a file."""
```
vkarak: * Do not use `self.module_map` in the docstring. Better write "Update the internal module mapping [...]".
* `module_mapping_file` -> `filename`.

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
...
+        module_mapping = {}
```
vkarak: Why don't you set directly the internal mapping?

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
...
+                        key, value = re.split(r'\s*:\s*', line.strip(':'))
+                        values_list = re.findall(
+                            r'(?P<value>[A-Za-z0-9_\-/\+\.]+)', value)
+                        module_mapping.setdefault(key, []).extend(values_list)
+                    except ValueError as e:
+                        raise ConfigError(
+                            'Incorrect format of the'
+                            'module mapping file: %s' % e) from e
+        except OSError as e:
+            raise ConfigError from e
```
vkarak: `OSError`s should be propagated as is. They are not configuration errors. So you don't have to do a try/except here.

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
     def _is_module_loaded(self, name):
         return self._backend.is_module_loaded(Module(name))
 
+    def load_mapping_from_file(self, module_mapping_file):
+        """Update ``self.module_map`` with a mapping defined in a file."""
+        module_mapping = {}
+        try:
+            with open(module_mapping_file, 'r') as f:
```
vkarak:
```python
with open(filename) as fp:
```

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
...
+                for line in f.readlines():
```
vkarak: You can iterate over the lines directly using the file pointer:

```python
for line in fp:
```

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
...
+                    line = re.split(r'\s*#\s*', line.strip())[0]
```
vkarak: No need for regexes here. You can simply split the line at `#`. Everything at the left is your line of interest.

```python
line = line.split('#')[0].strip()
```

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
...
+                    if not line:
+                        continue
+
+                    try:
+                        key, value = re.split(r'\s*:\s*', line.strip(':'))
+                        values_list = re.findall(
+                            r'(?P<value>[A-Za-z0-9_\-/\+\.]+)', value)
+                        module_mapping.setdefault(key, []).extend(values_list)
+                    except ValueError as e:
```
vkarak: I don't think a `ValueError` can ever be thrown here.

reframe/core/modules.py
```diff
@@ -207,6 +208,30 @@ def is_module_loaded(self, name):
...
+            with open(module_mapping_file, 'r') as f:
+                for line in f.readlines():
+                    line = re.split(r'\s*#\s*', line.strip())[0]
+                    if not line:
+                        continue
+
+                    try:
+                        key, value = re.split(r'\s*:\s*', line.strip(':'))
+                        values_list = re.findall(
+                            r'(?P<value>[A-Za-z0-9_\-/\+\.]+)', value)
```
vkarak: Regexes are not very useful here. They complicate the code and restrict the syntax (in this particular case). The idea is simple:

1. Split the line at `:` once (with the string's `split()` method)
2. If any of the left or right parts are empty, that's an error (here you can throw a `ConfigError`)
3. Your left part is the key
4. Split the right part at `,`, the resulting list contains your mappings for the key.
5. Set `self.module_map[key] = value` and there you are.

vkarak: When throwing a `ConfigError` to denote problems in the file, it's good to keep track of the line numbers. You can achieve this as follows:

```python
for lineno, line in enumerate(fp):
```

reframe/frontend/cli.py
```diff
@@ -84,6 +84,9 @@ def main():
         help='Load checks recursively')
 
     # Select options
+    select_options.add_argument(
+        '--try-module', action='append', dest='module_map', default=[],
+        help='Map a module name to a different one')
```
vkarak: This is not the option described in #76.

vkarak: It should accept a mapping file and should not append its values. Perhaps we should call it `--module-file`.

reframe/frontend/cli.py
```diff
@@ -367,6 +370,9 @@ def main():
 
         # Act on checks
 
+        if options.module_map:
+            get_modules_system().load_mapping_from_file(options.module_map[0])
+
```
vkarak: This should move up. Here is the part where we "act on checks". The treatment of this option should be at the beginning after the current system is set up.

unittests/test_modules.py
```diff
@@ -402,3 +402,45 @@ def test_mapping_deep_cycle(self):
         }
         self.assertRaisesRegex(EnvironError, 'm0->m1->m3->m4->m2->m1',
                                self.modules_system.load_module, 'm0')
+
+    def test_mapping_from_file_simple(self):
+        with open('module_mappings_file.txt', 'w') as f:
```
vkarak: You should create a temporary file for that. Also use `fp` instead of `f`, since this is the "convention" elsewhere in the code.

unittests/test_modules.py
```diff
@@ -402,3 +402,45 @@ def test_mapping_deep_cycle(self):
...
+                          'module_mappings_file.txt')
+
+    def test_maping_from_file_missing_file(self):
+        self.assertRaises(ConfigError,
+                          self.modules_system.load_mapping_from_file,
+                          'module_mappings_file.txt')
+
+    def tearDown(self):
+        if os.path.isfile('module_mappings_file.txt'):
+            os.remove('module_mappings_file.txt')
```
vkarak: `tearDown()` will be called after every test. So perhaps, it's better to create the temporary module mapping file in the `setUp()` method.

unittests/test_modules.py
```diff
@@ -402,3 +402,45 @@ def test_mapping_deep_cycle(self):
...
+
+    def test_mapping_from_file_simple(self):
+        with open('module_mappings_file.txt', 'w') as f:
+            print('m1:m2, m3'
+                  '\nm2 : m4',
+                  '\n  #m5: m6',
+                  '\n',
+                  '\nm1: m7, # An inline comment after the coma',
+                  '\nm1: m7 # An inline comment with no coma',
+                  file=f)
```
vkarak: Do not use the `print()` here. Write it as follows:

```python
fp.write('m1:m2, m3\n'
         'm2 : m4\n'
         '  #m5: m6\n'
         '\n'
         'm1: m7, # An inline comment after the comma\n'
         'm1: m7 # An inline comment with no comma\n')
```

_Notice the lack of commas at the end of each line. Python will concatenate the strings._

Also the comment is still safe and sound and never got in a [coma](https://en.wikipedia.org/wiki/Coma). 😂 Make sure to spell it as comma.

unittests/test_modules.py
```diff
@@ -402,3 +402,45 @@ def test_mapping_deep_cycle(self):
...
+        reference_map = {
+            'm1': ['m2', 'm3', 'm7', 'm7'],
```
vkarak: That's an interesting case. I think the loader should make sure to remove duplicates from the lists.

unittests/test_modules.py
```diff
@@ -402,3 +402,45 @@ def test_mapping_deep_cycle(self):
...
+            'm2': ['m4'],
+        }
+        self.modules_system.load_mapping_from_file('module_mappings_file.txt')
+        self.assertEqual(reference_map, self.modules_system.module_map)
+
+    def test_maping_from_file_missing_key_separator(self):
+        with open('module_mappings_file.txt', 'w') as f:
+            print('m1 m2', file=f)
```
vkarak: Leave a blank line after the `with` block

unittests/test_modules.py
```diff
@@ -402,3 +402,45 @@ def test_mapping_deep_cycle(self):
...
+        self.assertRaises(ConfigError,
+                          self.modules_system.load_mapping_from_file,
+                          'module_mappings_file.txt')
```
vkarak: You should also add a test where the `:` is present but there's no key specified.

unittests/test_modules.py
```diff
@@ -402,3 +402,45 @@ def test_mapping_deep_cycle(self):
...
+                  '\nm5: m6',
+                  file=f)
+        self.assertRaises(ConfigError,
+                          self.modules_system.load_mapping_from_file,
+                          'module_mappings_file.txt')
+
+    def test_maping_from_file_missing_file(self):
+        self.assertRaises(ConfigError,
+                          self.modules_system.load_mapping_from_file,
+                          'module_mappings_file.txt')
```
vkarak: That's confusing. Use a filename like `/foo` or sth similar that's unlikely to exist and will immediately strike your eye.

vkarak:
@rsarm We will need to document this feature, too.

reframe/core/modules.py
```diff
@@ -203,6 +204,27 @@ def is_module_loaded(self, name):
...
+                line = line.strip().split('#')[0]
+                if not line:
+                    continue
+
+                try:
+                    self.load_mapping(line)
+                except Exception as e:
+                    raise ConfigError('Line %d of "%s": '
+                                      'Invalid format of the module mapping'
+                                      'file.\n %s' % (lineno, filename, e))
```
vkarak: The proper way of reraising this is the following:

```python
try:
    may_throw_op()
except ConfigError as e:
    raise ConfigError('%s:%s' % (filename, lineno)) from e
```

This will print out `/path/to/map/file:55: message from e`.

reframe/frontend/cli.py
```diff
@@ -1,21 +1,24 @@
 import os
 import socket
 import sys
+import traceback
```
vkarak: I don't think this is needed.

reframe/frontend/cli.py
```diff
@@ -1,21 +1,24 @@
...
 
 import reframe
 import reframe.core.logging as logging
 import reframe.utility.os as os_ext
-from reframe.core.exceptions import (EnvironError, ReframeError,
+
```
vkarak: Please remove this blank line. The correct grouping of imports is the following:

```python
import standard_library_stuff
from standard_library_stuff import names

import reframe_stuff
from reframe_stuff import rfm_names
```

The order of the statements for each subgroup is first the `import` statements, then the `from .. import ..` ones, and each of these subsubgroups *must* be sorted alphabetically.

reframe/frontend/cli.py
```diff
@@ -81,6 +84,12 @@ def main():
         help='Load checks recursively')
 
     # Select options
+    select_options.add_argument(
+        '--modules-mapping', action='store', dest='module_map_file',
+        help='Map a module name to a different one')
```
vkarak: The help message is wrong.

reframe/frontend/cli.py
```diff
@@ -81,6 +84,12 @@ def main():
         help='Load checks recursively')
 
     # Select options
+    select_options.add_argument(
+        '--modules-mapping', action='store', dest='module_map_file',
```
vkarak: I think `--module-map` is a better name. Also set the `metavar=FILE` so as to have a better description in the help message.

reframe/frontend/cli.py
```diff
@@ -81,6 +84,12 @@ def main():
...
+        help='Map a module name to a different one')
+    select_options.add_argument(
+        '--try-module', action='append', dest='module_map_cl', default=[],
```
vkarak: @victorusu @teojgo Do you agree with the `--try-module` name of the option?

@rsarm The `dest` name should have a better name, e.g., `module_mappings` or `explicit_module_mappings`. You should also set `metavar='MODULE MAPPING'`

reframe/frontend/cli.py
```diff
@@ -81,6 +84,12 @@ def main():
...
+        help='Map a module name to a different one')
```
vkarak: If we keep the `--try-module` name for the option, we should  perhaps adapt a bit this help message.

reframe/frontend/cli.py
```diff
@@ -229,6 +238,17 @@ def main():
     # Init modules system
     init_modules_system(system.modules_system)
 
+    if options.module_map_file:
+        get_modules_system().load_mapping_from_file(options.module_map_file)
```
vkarak: This call may throw.

reframe/frontend/cli.py
```diff
@@ -229,6 +238,17 @@ def main():
...
+
+    if options.module_map_cl:
+        for m in options.module_map_cl:
+            try:
+                get_modules_system().load_mapping(m)
+            except Exception as e:
+                raise ConfigError('Invalid format of the command-line'
+                                  'module mapping: %s\n %s' % (m, e))
```
vkarak: In case of error you should exit.

vkarak: - Generally, both of the above calls (loading from file and from the command line arguments) you could put them in a `try/except` block.
- The error message should read as (_lowercase_): `"could not set module mapping: %s" % e`

unittests/test_modules.py
```diff
@@ -4,6 +4,7 @@
 import reframe.core.modules as modules
 from reframe.core.environments import EnvironmentSnapshot
 from reframe.core.exceptions import ConfigError, EnvironError
+from tempfile import NamedTemporaryFile
```
vkarak: This import is at the wrong place. See my comment above.

unittests/test_modules.py
```diff
@@ -199,6 +200,7 @@ class TestModuleMapping(unittest.TestCase):
     def setUp(self):
         self.modules_activity = ModulesSystemEmulator()
         self.modules_system = modules.ModulesSystem(self.modules_activity)
+        self.fp = NamedTemporaryFile('wt', delete=False)
```
vkarak: Here you should not use `fp`. It's like using `i` as a variable outside the scope of a for loop in C/C++... Better call it sth more meaningful, e.g., `mapping_file`.

unittests/test_modules.py
```diff
@@ -401,3 +403,55 @@ def test_mapping_deep_cycle(self):
         }
         self.assertRaisesRegex(EnvironError, 'm0->m1->m3->m4->m2->m1',
                                self.modules_system.load_module, 'm0')
+
+    def test_mapping_from_file_simple(self):
+        self.fp.write('m1:m2  m3   m3\n'
+                      'm2 : m4 \n'
+                      ' #m5: m6\n'
+                      '\n'
+                      'm5: m6 # Inline comment')
```
vkarak: Having these two lines with `m5: m6`, you cannot tell which one was added. Please, use different names, like `m7: m8`.

unittests/test_modules.py
```diff
@@ -401,3 +403,55 @@ def test_mapping_deep_cycle(self):
...
+        reference_map = {
+            'm1': ['m2', 'm3'],
+            'm2': ['m4'],
+            'm5': ['m6']
+        }
+        self.fp.close()
+        self.modules_system.load_mapping_from_file(self.fp.name)
+        self.assertEqual(reference_map, self.modules_system.module_map)
+
+    def test_mapping_from_file_repeated_key(self):
```
vkarak: I think this test can be merged with the above. Simply, add a duplicate key in the above dictionary and update your `reference_map`.

unittests/test_modules.py
```diff
@@ -401,3 +403,55 @@ def test_mapping_deep_cycle(self):
...
+                      'm2 : m4 \n'
+                      ' #m5: m6\n'
+                      '\n'
+                      'm5: m6 # Inline comment')
+        reference_map = {
+            'm1': ['m2', 'm3'],
+            'm2': ['m4'],
+            'm5': ['m6']
+        }
+        self.fp.close()
```
vkarak: Wouldn't this work if you moved the `close()` to the `tearDown()`?

unittests/test_modules.py
```diff
@@ -401,3 +403,55 @@ def test_mapping_deep_cycle(self):
...
+        self.fp.write(' : m4')
+        self.fp.close()
+        self.assertRaises(ConfigError,
+                          self.modules_system.load_mapping_from_file,
+                          self.fp.name)
+
+    def test_maping_from_file_missing_file(self):
+        self.assertRaises(OSError,
+                          self.modules_system.load_mapping_from_file,
+                          'foo')
```
vkarak: If you don't close the file here, you leak the file descriptor.

unittests/test_modules.py
vkarak: If it doesn't work, then it's better to do the following:

```python
with self.mapping_file:
    self.mapping_file.write(...)
```

reframe/core/modules.py
```diff
@@ -203,6 +204,27 @@ def is_module_loaded(self, name):
...
 
+    def load_mapping(self, mapping):
+        key, values = mapping.strip().strip(':').split(':')
+        self.module_map[key.strip()] = list(
+            OrderedDict.fromkeys(values.split()))
+
+    def load_mapping_from_file(self, filename):
+        """Update the internal module mapping from a file."""
+
+        with open(filename, 'r') as fp:
```
vkarak: `'r'` is the default mode, no need to specify it.

reframe/core/modules.py
```diff
@@ -203,6 +204,27 @@ def is_module_loaded(self, name):
     def _is_module_loaded(self, name):
         return self._backend.is_module_loaded(Module(name))
 
+    def load_mapping(self, mapping):
+        key, values = mapping.strip().strip(':').split(':')
```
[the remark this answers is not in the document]
rsarm: To throw an Exception in the case of an empty mapping `m0:`. If ':' is striped before splitting, then the `split(:)` will return a list with one element and the `key, error = ` will fail.

reframe/core/modules.py
```diff
@@ -203,6 +204,27 @@ def is_module_loaded(self, name):
...
+        self.module_map[key.strip()] = list(
+            OrderedDict.fromkeys(values.split()))
+
+    def load_mapping_from_file(self, filename):
+        """Update the internal module mapping from a file."""
+
+        with open(filename, 'r') as fp:
+            for lineno, line in enumerate(fp):
+                line = line.strip().split('#')[0]
```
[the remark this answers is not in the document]
rsarm: Because if the commented line has a space before, `split('#')` will return a list with the first item an space `[' ', '#...']` and I preffered an empty line, just to skip it the same way that an empty line in the file is skipped.

reframe/core/modules.py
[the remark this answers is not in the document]
vkarak: Well, it's a bit cryptic... I prefer the more verbose way:

```python
try:
    key, values_str = mapping.split(':', maxsplit=1)
except ValueError:
    raise ConfigError('module mapping not properly formatted: %s' % mapping) from None

key = key.strip()
if not values_str:
    raise ConfigError('no target modules found for module: %s' % key)

values = values_str.split()
```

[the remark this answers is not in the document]
rsarm: To raise a `ConfigError` in the case of `m0 : m1 : m2`, is it a good idea to use `key, values_str = mapping.split(':')` without `maxsplit` so it returns a list with more than two items?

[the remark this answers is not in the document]
vkarak: Good point. Actually, there is an even better way to write the whole thing, without having to care about the `ValueError` at all:

```python
key, *rest = mapping.split(':')
if len(rest) != 1:
    raise ConfigError('invalid mapping syntax: %s' % mapping)

key = key.strip()
values = rest[0].strip()
```

You can even be more precise with the error messages. If `rest` is empty, then no values were specified, if it's larger than one, then multiple `:` have been specified. But, I prefer it simple as above.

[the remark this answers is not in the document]
rsarm: Nice!

reframe/frontend/cli.py
rsarm: Is it too long `--module-map-from-file` or `--module-maps-file` for the file of mappings? Then maybe we could have `--module-map` for the string mappings, instead of `--try-module`? What do you think?

vkarak: The problem with naming `--module-map` the current `--try-module` is that it is not a map. It simply defines a mapping.

What about `--module-mappings=FILE` for the one taking the file as argument and `--map-module=MAPPING` the one taking a mapping? I don't think is necessary to include `file` as part of the option name, because you can nicely show in the help message that this option expects  a file by setting `metavar='FILE'`.

vkarak: Having a `--module-map` and a `--map-module` option can be confusing. That's why I prefer `--module-mappings` for the first one. Your help message is fine for this option. For the second one, I find the "to the corresponding module" confusing, because I can't see where this "corresponding module" is. I'd suggest sth like "Apply a single module mapping".

reframe/core/modules.py
```diff
@@ -8,6 +8,7 @@
 
 import reframe.core.fields as fields
 import reframe.utility.os as os_ext
+from collections import OrderedDict
```
vkarak: Please have a look in the [coding style](https://github.com/eth-cscs/reframe/wiki/Coding-Style-Guide#imports) regarding the import statements.

reframe/core/modules.py
```diff
@@ -203,6 +204,32 @@ def is_module_loaded(self, name):
...
+    def load_mapping(self, mapping):
+        """Updates the internal module mapping with a single mapping"""
+        key, *rest = mapping.split(':')
+        if len(rest) != 1:
+            raise ConfigError('invalid mapping syntax: %s' % mapping)
+
+        key = key.strip()
+        values = rest[0].split()
+        if not key or not values:
+            raise ConfigError('invalid mapping syntax: %s' % mapping)
```
vkarak: I would split the check here in two:

```python
if not key:
    raise ConfigError('no key found in mapping: %s' % mapping)

if not values:
    raise ConfigError('no mappings defined for module: %s' % key)
```

reframe/core/modules.py
```diff
@@ -203,6 +204,32 @@ def is_module_loaded(self, name):
...
+
+        self.module_map[key] = list(OrderedDict.fromkeys(values))
+
+    def load_mapping_from_file(self, filename):
+        """Update the internal module mapping from mappings in a file."""
+        with open(filename) as fp:
+            for lineno, line in enumerate(fp):
```
vkarak: You should pass `start=1` to `enumerate` so as to count the lines in a human readable way.

reframe/frontend/cli.py
```diff
@@ -5,9 +5,10 @@
 import reframe
 import reframe.core.logging as logging
 import reframe.utility.os as os_ext
-from reframe.core.exceptions import (EnvironError, ReframeError,
+from reframe.core.exceptions import (EnvironError, ReframeError, ConfigError,
```
vkarak: Please have a look at the [import guidelines](https://github.com/eth-cscs/reframe/wiki/Coding-Style-Guide#imports).

reframe/frontend/cli.py
```diff
@@ -81,6 +82,14 @@ def main():
...
 
     # Select options
+    select_options.add_argument(
+        '--module-mappings', action='store', metavar='FILE',
+        dest='module_map_file',
+        help='Apply module mappings defined in FILE')
+    select_options.add_argument(
+        '--map-module', action='append', metavar='MAPPING',
+        dest='module_mappings', default=[],
+        help='Apply a single module mapping')
```
vkarak: These two options should be part of the `misc_options`.

reframe/frontend/cli.py
```diff
@@ -81,6 +82,14 @@ def main():
         help='Load checks recursively')
 
     # Select options
+    select_options.add_argument(
+        '--module-mappings', action='store', metavar='FILE',
+        dest='module_map_file',
+        help='Apply module mappings defined in FILE')
+    select_options.add_argument(
+        '--map-module', action='append', metavar='MAPPING',
```
vkarak: I would like to have also a `-M` short option for this one.

reframe/frontend/cli.py
```diff
@@ -229,6 +238,20 @@ def main():
     # Init modules system
     init_modules_system(system.modules_system)
 
+    if options.module_map_file:
+        try:
+            get_modules_system().load_mapping_from_file(
+                options.module_map_file)
+        except Exception as e:
+                raise ConfigError('could not set module mapping: %s' % e)
```
vkarak: Here you are at the top-level of the code. If you raise any exception, it will fail badly. You should do the following:

```python
try:
   # Here you can process both options; no need for two try/except
except ConfigError, OSError as e:
    printer.error('could not load module mappings: %s' % e)
    sys.exit(1)
```

unittests/test_modules.py
```diff
@@ -401,3 +403,57 @@ def test_mapping_deep_cycle(self):
...
                                self.modules_system.load_module, 'm0')
+
+    def test_mapping_from_file_simple(self):
+        with self.mapping_file:
+            self.mapping_file.write('m1:m2  m3   m3\n'
+                                    'm2 : m4 \n'
+                                    ' #m5: m6\n'
+                                    '\n'
+                                    ' m2: m7 m8\n'
+                                    'm9: m10 # Inline comment')
```
vkarak: Please leave a blank line after all the `with` blocks.

unittests/test_modules.py
```diff
@@ -401,3 +403,57 @@ def test_mapping_deep_cycle(self):
...
+        self.assertRaises(ConfigError,
+                          self.modules_system.load_mapping_from_file,
+                          self.mapping_file.name)
+
+    def test_maping_from_file_missing_file(self):
+        self.assertRaises(OSError,
+                          self.modules_system.load_mapping_from_file,
+                          'foo')
+
+    def tearDown(self):
```
vkarak: Please move the `tearDown()` just after the `setUp()`.

unittests/test_modules.py
```diff
@@ -401,3 +403,57 @@ def test_mapping_deep_cycle(self):
...
+        if not self.mapping_file.closed:
```
vkarak: No need to check if the file is closed. It is handled automatically by `close()`.

reframe/frontend/cli.py
vkarak: Here, you should better check for a `ReframeError` rather than `ConfigError`. You should remember to remove it also from the `import` above.

rsarm: Isn't it necessary one `try/except` for `if options.module_map_file` and other for `if options.module_mappings`?

vkarak: It's not bad how you have it and it is generally a good practice not to have very large try/except blocks. But here both options are eventually calling the same methods, so I see it more as redundant code.

reframe/frontend/cli.py
```diff
@@ -229,6 +238,18 @@ def main():
...
+        if options.module_map_file:
+            get_modules_system().load_mapping_from_file(
+                options.module_map_file)
+
+        if options.module_mappings:
+            for m in options.module_mappings:
+                get_modules_system().load_mapping(m)
+
+    except (ReframeError, OSError) as e:
+        raise ReframeError('could not set module mapping: %s' % e)
```
vkarak: You should not raise an exception here. You are just handling them here and present the error to the user. Have you tried the output by passing an invalid mapping. It's like crashing. You should do instead:

```python
    except (ReframeError, OSError) as e:
        printer.error('could not load module mappings: %s' % e)
        sys.exit(1)
```

PR #143, opened 2018-02-27
Make '--exclude' work with flexible allocation

### LINKED ISSUE #104: Make flexible node allocations work with the --exclude flag
The scenario is having a test with `num_tasks = 0` and calling ReFrame with `--reservation=foo --exclude=[nodelist]`.

teojgo:
* Check that the node name is not included in the excluded
  node list when counting compatible nodes.

* Create a unittest class to test the flexible node allocation
  combined with the `--exclude` option.

Fixes #104

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -184,15 +184,15 @@ class SlurmJob(sched.Job):
     def _get_reservation_nodes(self):
         command = 'scontrol show res %s' % self.sched_reservation
         completed = os_ext.run_command(command, check=True)
-        node_match = reservation_nodes = re.search('(Nodes=\S+)',
-                                                   completed.stdout)
+        node_match = re.search('(Nodes=\S+)', completed.stdout)
         if node_match:
             reservation_nodes = node_match[1]
         else:
             raise JobError("could not extract the nodes names for "
                            "reservation '%s'" % self.sched_reservation)
+
         completed = os_ext.run_command(
-            'scontrol show -o -a %s' % reservation_nodes, check=True)
+            'scontrol show -o %s' % reservation_nodes, check=True)
         node_descriptions = completed.stdout.splitlines()
         return (SlurmNode(descr) for descr in node_descriptions)
 
@@ -200,7 +200,7 @@ class SlurmJob(sched.Job):
         if not self.sched_exclude_nodelist:
             return set()
 
-        command = 'scontrol show -o -a %s' % self.sched_exclude_nodelist
+        command = 'scontrol show -o node %s' % self.sched_exclude_nodelist
         try:
             completed = os_ext.run_command(command, check=True)
         except SpawnedProcessError as e:
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -157,8 +157,7 @@ class SlurmJob(sched.Job):
         constraints = set()
         partitions = (set(self.sched_partition.split())
                       if self.sched_partition else set())
-        excluded_node_names = ({n.name for n in self._get_excluded_nodes()}
-                               if self.sched_exclude_nodelist else set())
+        excluded_node_names = self._get_excluded_nodes()
 
         if self.options:
             for optstr in self.options:
@@ -198,15 +197,18 @@ class SlurmJob(sched.Job):
         return (SlurmNode(descr) for descr in node_descriptions)
 
     def _get_excluded_nodes(self):
-        command = 'scontrol show -o -a %s' % self.sched_exclude_nodelist
-        completed = os_ext.run_command(command, check=True)
+        if not self.sched_exclude_nodelist:
+            return set()
 
-        node_descriptions_output = completed.stdout()
-        if node_descriptions == '':
-            raise JobError("could not extract the node description for "
-                           "excluded nodes '%s'" % self.sched_exclude_nodelist)
-        else:
-            return (SlurmNode(descr) for descr in node_descriptions)
+        command = 'scontrol show -o -a %s' % self.sched_exclude_nodelist
+        try:
+            completed = os_ext.run_command(command, check=True)
+        except SpawnedProcessError as e:
+            raise JobError('could not retrieve the node description '
+                           'of nodes: %s' % self.sched_exclude_nodelist) from e
+        node_descriptions_output = completed.stdout
+        slurm_nodes = (SlurmNode(descr) for descr in node_descriptions)
+        return {n.name for n in slurm_nodes}
 
     def _update_state(self):
         """Check the status of the job."""
@@ -383,6 +385,4 @@ class SlurmNode:
                            "node description" % attr_name)
 
     def __str__(self):
-        return ('Slurm node {{name: {0}, partitions: {1}, '
-                'active features: {2}}}').format(
-            self._name, self._partitions, self._active_features)
+        return self._name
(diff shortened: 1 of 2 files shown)


unittests/test_schedulers.py
```diff
@@ -481,6 +481,55 @@ def prepare_job(self):
         self.testjob.prepare(self.builder)
 
 
+class TestSlurmFlexibleNodeAllocationExclude(TestSlurmFlexibleNodeAllocation):
```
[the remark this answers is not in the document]
teojgo: The expected number of tasks in not the same for all the tests. Also there are tests that the `TestSlurmFlexibleNodeAllocationExclude` raises exceptions while the parent one does not.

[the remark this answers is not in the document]
vkarak: I know, that's why each test should set it separately.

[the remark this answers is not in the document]
teojgo: Even in the parent class the expected_num_tasks is not uniform so each of the tests on the parent class has to set it. Thus calling the parent class would override any setting set by the child class.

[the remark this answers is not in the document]
vkarak: You can do sth like the following:

```python
class TestSlurmFlexibleNodeAllocation(unittest.TestCase):
    def test_valid_constraint(self, expected_tasks=8):
        self.testjob._sched_reservation = 'Foo'
        self.testjob.options = ['-C f1']
        self.prepare_job()
        self.assertEquals(self.testjob.num_tasks, expected_tasks)


class TestSlurmFlexibleNodeAllocationExclude(TestSlurmFlexibleNodeAllocation):
    def setUp(self):
        super().setUp()
        self.testjob._sched_exclude_nodelist = 'nid00001'

    def test_valid_constraint(self):
        super().test_valid_constraint(expected_tasks=4)

```

reframe/core/schedulers/slurm.py
```diff
@@ -157,6 +157,8 @@ def _count_compatible_nodes(self, nodes):
         constraints = set()
         partitions = (set(self.sched_partition.split())
                       if self.sched_partition else set())
+        excluded_nodes = (set(self.sched_exclude_nodelist.split())
```
vkarak: As discussed offline, the `split()` here is restrictive. The `sched_exclude_nodelist` can get any value that Slurm accepts.

vkarak: Also, for debugging reasons, please implement the `__str__()` method of the `SlurmNode`. It could simply return the node name.

reframe/core/schedulers/slurm.py
```diff
@@ -157,6 +157,8 @@ def _count_compatible_nodes(self, nodes):
         constraints = set()
         partitions = (set(self.sched_partition.split())
                       if self.sched_partition else set())
+        excluded_node_names = ({n.name for n in self._get_excluded_nodes()}
+                               if self.sched_exclude_nodelist else set())
```
vkarak: Why don't you incorporate this logic in the `_get_excluded_nodes()` so as to make this statement much simpler?

reframe/core/schedulers/slurm.py
```diff
@@ -193,6 +197,17 @@ def _get_reservation_nodes(self):
         node_descriptions = completed.stdout.splitlines()
         return (SlurmNode(descr) for descr in node_descriptions)
 
+    def _get_excluded_nodes(self):
+        command = 'scontrol show -o -a %s' % self.sched_exclude_nodelist
+        completed = os_ext.run_command(command, check=True)
+
+        node_descriptions_output = completed.stdout()
```
vkarak: The `stdout` is a property of `completed`, not a function. I don't know how this has worked... Also you can use it directly in the following if.

reframe/core/schedulers/slurm.py
```diff
@@ -193,6 +197,17 @@ def _get_reservation_nodes(self):
...
+        if node_descriptions == '':
+            raise JobError("could not extract the node description for "
+                           "excluded nodes '%s'" % self.sched_exclude_nodelist)
```
vkarak: What does the empty output mean? If anything happens during the `scontrol show` an exception will be thrown from the `run_command` because you have the `check=True`. I suggest catching the `SpawnedProcessError` and reraising it as a `JobError` if you like:

```python
try:
    os_ext.run_command(...)
except SpawnedProcessError as e:
    raise JobError('could not retrieve the node description of nodes: %s' self.sched_exclude_nodelist) from e
```

reframe/core/schedulers/slurm.py
```diff
@@ -193,6 +197,17 @@ def _get_reservation_nodes(self):
...
+        else:
```
vkarak: No need for an `else` here.

reframe/core/schedulers/slurm.py
```diff
@@ -366,3 +381,8 @@ def _extract_attribute(self, attr_name, node_descr):
         else:
             raise JobError("could not extract attribute '%s' from "
                            "node description" % attr_name)
+
+    def __str__(self):
+        return ('Slurm node {{name: {0}, partitions: {1}, '
+                'active features: {2}}}').format(
+            self._name, self._partitions, self._active_features)
```
vkarak: I would simply return the node name here, because this would should up nicely and naturally in '%s'.
If you want debug information, you could simply implement the `__repr__` and use the utility we provide from the framework that does print all the internal fields:

```python

def __repr__(self):
    return debug.repr(self)
```

unittests/test_schedulers.py
```diff
@@ -514,3 +548,14 @@ def test_attributes(self):
                          {'p1', 'p2'})
         self.assertEqual(self.node.active_features,
                          {'f1', 'f2'})
+
+    def test_str(self):
```
vkarak: I suggest simply returning the node name in `__str__`, so this has now meaning.

reframe/core/schedulers/slurm.py
```diff
@@ -193,6 +196,20 @@ def _get_reservation_nodes(self):
...
+            return set()
+
+        command = 'scontrol show -o -a %s' % self.sched_exclude_nodelist
+        try:
+            completed = os_ext.run_command(command, check=True)
+        except SpawnedProcessError as e:
+            raise JobError('could not retrieve the node description '
+                           'of nodes: %s' % self.sched_exclude_nodelist) from e
+        node_descriptions_output = completed.stdout
+        slurm_nodes = (SlurmNode(descr) for descr in node_descriptions)
```
vkarak: Use directly `completed.stdout` here and please leave a blank line after the `try/except` block.

teojgo: These is also a bug here since I do not split the lines.

reframe/core/schedulers/slurm.py
```diff
@@ -193,6 +196,21 @@ def _get_reservation_nodes(self):
         node_descriptions = completed.stdout.splitlines()
         return (SlurmNode(descr) for descr in node_descriptions)
 
+    def _get_excluded_node_names(self):
+        if not self.sched_exclude_nodelist:
+            return set()
+
+        command = 'scontrol show -o -a %s' % self.sched_exclude_nodelist
```
vkarak: The command should be `scontrol show node -o -a [nodelist]`.

teojgo: I assumed that we are getting the nodes with `scontrol show -o -a Nodes=[nodelist]` which also works.

PR #135, opened 2018-02-22
Fix occasional unit test failure in async policy.

vkarak:
Fixes #78.

Issue #78, opened 2018-02-15
Unit test failure of the TestAsynchronousExecutionPolicy


vkarak:
The failure happens when running the unittest suite on a local machine (laptop) with 2 Codes - 4 Threads. The failure is the following:
```
FAIL: test_kbd_interrupt_in_wait_with_limited_concurrency (unittests.test_policies.TestAsynchronousExecutionPolicy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/manitart/reframe_dev_local/unittests/test_policies.py", line 310, in test_kbd_interrupt_in_wait_with_limited_concurrency
    self._run_checks(checks, 2)
  File "/home/manitart/reframe_dev_local/unittests/test_policies.py", line 289, in _run_checks
    self.assertEqual(4, self.runner.stats.num_cases())
AssertionError: 4 != 3
```

Internal issue: https://madra.cscs.ch/scs/reframe/issues/582

vkarak:
This happens sporadically and appeared also once during running by Travis CI.

vkarak:
This has become recurring. We should address it asap.

vkarak:
I spotted the problem and it appears to be caused by the (wrong) assumption the unit tests make that none of the submitted tasks would finish at the time when the partition task limit is reached. When this limit is reached, all the currently running tasks are polled. If any of them has finished (in this case the `KeyboardInterruptCheck`), it will be waited for, and this is where the `KeyboardInterrupt` is raised. At this point not all of the tasks may have been submitted leading to this unit test failure. Here is a bit of code to always trigger the failure:

```diff
diff --git a/reframe/frontend/executors/policies.py b/reframe/frontend/executors/policies.py
index 3f4d4a7..e8bc481 100644
--- a/reframe/frontend/executors/policies.py
+++ b/reframe/frontend/executors/policies.py
@@ -189,6 +189,9 @@ class AsynchronousExecutionPolicy(ExecutionPolicy, TaskEventListener):

             self._failall(e)
             raise
+        finally:
+            import time
+            time.sleep(.5)

     def _poll_tasks(self):
         """Update the counts of running checks per partition."""
```

vkarak:
This seems not to be fixed entirely. It still happens from time to time on Travis CI. Perhaps we need to increase the sleep time of the `KeyboardInterruptCheck`.

vkarak:
The problem currently is a bit different. The failure we get is the following:

```
======================================================================
[...]

PR #154, opened 2018-02-28
Fix failure of the asynchronous execution policy's unit tests

vkarak:


teojgo:
In my local computer it works fine. But it is still failing in Travis and the error is `AssertionError: 4 != 2` instead of `AssertionError: 4 != 3`.

vkarak:
Do you see the `4 != 3` error? In which test? Because a see another error in another test:

```
FAIL: test_kbd_interrupt_in_wait_with_concurrency (unittests.test_policies.TestAsynchronousExecutionPolicy)
----------------------------------------------------------------------
  File "/home/travis/build/eth-cscs/reframe/unittests/test_policies.py", line 299, in test_kbd_interrupt_in_wait_with_concurrency
    self._run_checks(checks, 4)
  File "/home/travis/build/eth-cscs/reframe/unittests/test_policies.py", line 289, in _run_checks
    self.assertEqual(4, self.runner.stats.num_failures())
AssertionError: 4 != 2
----------------------------------------------------------------------
```

vkarak:
I'm gonna fix this test, too. The fix is the same as for the other one failing.

teojgo:
Before the fix when the error was thrown, the assertion was `4 !=3`.

vkarak:
@teojgo It's crucial to see which test was failing exactly, because until now there three different types of failure!

teojgo:
I am referring to the failures like the ones in https://github.com/eth-cscs/reframe/pull/143. This PR does not contain the fix.

vkarak:
@teojgo Wait a sec, #143 does not contain the fix apparently, cos this one is not merged with the master. So you will be getting the `4 != 3` error, which btw is different from the first similar error, that #135 fixed. Check the last comment on #78 to see the difference. Now in *this* PR there is another unit test failing (because of the same reason that the async unit tests with limited concurrency were failing).

teojgo:
Yes I see, indeed this is another test that is failing.

teojgo:
@jenkins-cscs retry dom

vkarak:
I've rerun the Travis CI several times and the fix seems to work. @omlins @teojgo Could you guys, approve this PR (if you don't have any other objections, of course).

teojgo:
I tested many times locally and it works fine.

PR #169, opened 2018-03-02
Allow upper thresholds >= 1.0 in 'assert_reference'

### LINKED ISSUE #61: Function `assert_reference` is too restrictive
There are cases where the expected values differ more than 100% from the reference one. Currently, the limit values have to be between 0.0 and 1.0 thus restricting use of `assert_reference` in the cases mentioned above.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/575

teojgo:
Fixes #61

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -824,7 +824,7 @@ The first occurrence will be used as the reference value of the ``perf`` perform
 In our example, the ``perf`` key will be resolved in the ``daint:gpu`` scope giving us the reference value.
 
 Reference values in ReFrame are specified as a three-tuple comprising the reference value and lower and upper thresholds.
-Thresholds are specified as decimal fractions of the reference value. For nonnegative reference values, the lower threshold must lie in the [-1,0], whereas the upper threshold must lie in the [0,inf] interval.
+Thresholds are specified as decimal fractions of the reference value. For nonnegative reference values, the lower threshold must lie in the [-1,0], whereas the upper threshold may be any positive real number or zero.
 In our example, the reference value for this test on ``daint:gpu`` is 50 Gflop/s ±10%. Setting a threshold value to :class:`None` disables the threshold.
 
 Combining It All Together
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -824,7 +824,7 @@ The first occurrence will be used as the reference value of the ``perf`` perform
 In our example, the ``perf`` key will be resolved in the ``daint:gpu`` scope giving us the reference value.
 
 Reference values in ReFrame are specified as a three-tuple comprising the reference value and lower and upper thresholds.
-Thresholds are specified as decimal fractions of the reference value. The lower threshold must lie in the [-1,0] interval, whereas the upper threshold must be lie in the [0,1] interval.
+Thresholds are specified as decimal fractions of the reference value. The lower threshold must lie in the [-1,0] interval, whereas the upper threshold must be lie in the [0,inf] interval.
 In our example, the reference value for this test on ``daint:gpu`` is 50 Gflop/s ±10%. Setting a threshold value to :class:`None` disables the threshold.
 
 Combining It All Together
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -824,7 +824,7 @@ The first occurrence will be used as the reference value of the ``perf`` perform
 In our example, the ``perf`` key will be resolved in the ``daint:gpu`` scope giving us the reference value.
 
 Reference values in ReFrame are specified as a three-tuple comprising the reference value and lower and upper thresholds.
-Thresholds are specified as decimal fractions of the reference value. The lower threshold must lie in the [-1,0] interval, whereas the upper threshold must be lie in the [0,inf] interval.
+Thresholds are specified as decimal fractions of the reference value. For nonnegative reference values, the lower threshold must lie in the [-1,0], whereas the upper threshold must lie in the [0,inf] interval.
 In our example, the reference value for this test on ``daint:gpu`` is 50 Gflop/s ±10%. Setting a threshold value to :class:`None` disables the threshold.
 
 Combining It All Together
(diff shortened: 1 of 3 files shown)


unittests/test_sanity_functions.py
```diff
@@ -371,16 +371,17 @@ def test_assert_reference(self):
...
                                'invalid low threshold value: 0\.2',
                                evaluate, sn.assert_reference(0.9, 1, 0.2, 0.1))
         self.assertRaisesRegex(SanityError,
                                'invalid low threshold value: 1\.2',
                                evaluate, sn.assert_reference(0.9, 1, 1.2, 0.1))
 
+        # Check upper threshold values greater than 1
+        self.assertTrue(sn.assert_reference(30.0, 10.0, None, 3.0))
+        self.assertTrue(sn.assert_reference(50.0, 20.0, None, 2.0))
+
```
vkarak: This part should before all the `assertRaises`.

reframe/utility/sanity.py
```diff
@@ -488,7 +488,7 @@ def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
 
     if upper_thres is not None:
         try:
-            evaluate(assert_bounded(upper_thres, 0, 1))
+            evaluate(assert_bounded(upper_thres, 0, None))
```
vkarak: Extending only the upper bound won't work correctly for negative reference values. It should be `None` both ways.

unittests/test_sanity_functions.py
vkarak: You should also check for negative references here.

reframe/utility/sanity.py
```diff
@@ -488,7 +488,7 @@ def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
 
     if upper_thres is not None:
         try:
-            evaluate(assert_bounded(upper_thres, 0, 1))
+            evaluate(assert_bounded(upper_thres, None, None))
```
vkarak: As discussed offline, this is not enough. For positive references, the upper threshold must be in [0, inf], whereas for negative references, the lower threshold must be unbounded, [-inf, 0].

vkarak:

reframe/utility/sanity.py
```diff
@@ -471,26 +471,40 @@ def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
...
+        if ref >= 0.0:
+            try:
+                evaluate(assert_bounded(lower_thres, -1, 0))
+            except SanityError:
+                raise SanityError('invalid low threshold value: %s' % lower_thres)
+        else:
+            try:
+                evaluate(assert_bounded(lower_thres, None, 0))
+            except SanityError:
+                raise SanityError('invalid low threshold value: %s' % lower_thres)
```
vkarak: Too much code duplication here. Please rewrite it as follows:

```python
lower_thres_limit = -1 if ref >= 0 else None
try:
    evaluate(assert_bounded(lower_thres, lower_thres_limit, 0))
except SanityError:
    raise SanityError('invalid low threshold value: %s' % lower_thres) from None
```

Same thing for the `upper_thres`.

unittests/test_sanity_functions.py
```diff
@@ -340,6 +340,12 @@ def test_assert_reference(self):
         self.assertTrue(sn.assert_reference(-0.9, -1, upper_thres=0.1))
         self.assertTrue(sn.assert_reference(-0.9, -1))
 
+        # Check upper threshold values greater than 1
+        self.assertTrue(sn.assert_reference(30.0, 10.0, None, 3.0))
+        self.assertTrue(sn.assert_reference(-50.0, -20.0, -2.0, 0.5))
+        self.assertTrue(sn.assert_reference(30.0, 10.0, None, 3.0))
+        self.assertTrue(sn.assert_reference(-50.0, -20.0, -2.0, 0.5))
```
vkarak: These are duplicated.

vkarak: Also do not use always values equals to the limit.

unittests/test_sanity_functions.py
```diff
@@ -340,6 +340,12 @@ def test_assert_reference(self):
         self.assertTrue(sn.assert_reference(-0.9, -1, upper_thres=0.1))
         self.assertTrue(sn.assert_reference(-0.9, -1))
 
+        # Check upper threshold values greater than 1
+        self.assertTrue(sn.assert_reference(30.0, 10.0, None, 3.0))
+        self.assertTrue(sn.assert_reference(-50.0, -20.0, -2.0, 0.5))
```
vkarak: I don't understand why this should succeed. The value should be in [-40, -10] and it's -50.

docs/tutorial.rst
```diff
@@ -824,7 +824,7 @@ The first occurrence will be used as the reference value of the ``perf`` perform
 In our example, the ``perf`` key will be resolved in the ``daint:gpu`` scope giving us the reference value.
 
 Reference values in ReFrame are specified as a three-tuple comprising the reference value and lower and upper thresholds.
-Thresholds are specified as decimal fractions of the reference value. The lower threshold must lie in the [-1,0] interval, whereas the upper threshold must be lie in the [0,1] interval.
+Thresholds are specified as decimal fractions of the reference value. For nonnegative reference values, the lower threshold must lie in the [-1,0], whereas the upper threshold must lie in the [0,inf] interval.
```
vkarak: "may be any positive real number or zero."

unittests/test_sanity_functions.py
teojgo: Because in this case -2.0 means -200 % difference thus the corresponding lower bound is -20.0 -2.0 * 20.0 = -60.0.

reframe/utility/sanity.py
```diff
@@ -471,26 +471,32 @@ def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
...
 
     if upper_thres is not None:
+        upper_thres_limit = None if ref >= 0 else 1
         try:
-            evaluate(assert_bounded(upper_thres, 0, 1))
+            evaluate(assert_bounded(upper_thres, 0, upper_thres_limit))
         except SanityError:
-            raise SanityError('invalid high threshold value: %s' % upper_thres)
+            raise SanityError('invalid high threshold value: %s'
+                              % upper_thres) from None
```
vkarak: We split the lines after the binary operator. [PEP8](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator) prefers splitting before but does not mandate it.

PR #172, opened 2018-03-04
Fix crash when init_modules_system throws

vkarak:
- Catch errors raised by `init_modules_system` in the CLI and report
  nicely to user


vkarak:

PR #173, opened 2018-03-05
Revert to normal leone checkout

teojgo:


teojgo:
@jenkins-cscs retry dom

PR #175, opened 2018-03-05
Version bump

vkarak:


vkarak:
@jenkins-cscs retry daint kesch monch leone

PR #177, opened 2018-03-06
Adapt the G2G and AlltoAll tests

teojgo:


vkarak:
@jenkins-cscs retry dom

PR #194, opened 2018-03-08
Retry failed tests automatically

### LINKED ISSUE #148: Support for multiple retries of failing tests
This is a useful feature for avoiding transient failures in performance tests. We do not have a specific requirement on when to rerun the tests, so implementation complexity for each alternative must be considered. For the beginning, we should opt for the most straightforward implementation. A couple of alternatives are the following:

1. Rerun a failing test immediately after it has failed by re-running a part of its pipeline (from the run phase onward).
2. Rerun all the failed tests at the end after `run_all()` has finished.

We should also have a command-line option for specifying the retries, e.g., `--num-retries`.

omlins:
This is a first implementation for support of retries with the asynchronous execution policy.
The implementation for support with the synchronous execution policy will be added in a second step.

Fixes #148.


vkarak:
Your branch is a bit out-of-date from master. That's the the Travis CI fails. Could you please update your branch?

reframe/frontend/cli.py
```diff
@@ -405,13 +405,15 @@ def main():
             exec_policy.sched_nodelist = options.nodelist
             exec_policy.sched_exclude_nodelist = options.exclude_nodes
             exec_policy.sched_options = options.job_options
-            runner = Runner(exec_policy, printer)
+            # TODO: Get this later from the user; set to 0 by default (get_checks_failed programmed such that can be get per partition)
+            max_retries = 2
```
omlins: This could be the reason for the current failure of the async unit tests in my branch. Once I do not have it hardcoded, here it should pass again, I guess.

reframe/frontend/executors/__init__.py
```diff
@@ -16,17 +16,26 @@ class RegressionTask:
...
         self._failed_stage = None
         self._current_stage = None
         self._exc_info = (None, None, None)
         self._environ = None
         self._listeners = list(listeners)
+        # TODO: if we had a runner accociated, then we could do
+        # self._retry_num = self.runner.retry_num
+        # Would that make sense? The runner could be passed as constructor
+        # argument instead of retry_num.
+        self._retry_num = retry_num
```
omlins: What do you think abou the issue noted in this comment?

reframe/frontend/cli.py
vkarak: I would say to update with the latest master, because otherwise we cannot be sure. I am quite confident that the old failure is fixed properly.

omlins:
Example of output of the current version:

```
Command line: /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async
Reframe version: 2.10
Launched by user: omlins
Launched on host: daint104
Reframe paths
=============
    Check prefix      : 
    Check search path : '/users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py'
    Stage dir prefix  : /scratch/snx3000/omlins/0_temp/stage/
    Output dir prefix : /scratch/snx3000/omlins/0_temp/output/
    Logging dir       : /scratch/snx3000/omlins/0_temp/logs
[==========] Started on Thu Mar  8 16:13:50 2018 

[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] retrying (1/2) debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] retrying (1/2) debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[       OK ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] retrying (2/2) debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[  FAILED  ] Ran 2 test case(s) from 2 check(s) (1 failure(s) after 2 retries; 1 test cases passed in retries)
[==========] Finished on Thu Mar  8 16:16:06 2018 
SUMMARY OF FAILURES
FAILURE INFO for debug_write_check2_snx3000 (for the last of 2 retries)
  * System partition: daint:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray
  * Job type: batch job (id=6310898)
  * Maintainers: []
  * Failing phase: sanity
  * Reason: sanity error: pattern `^Max Write: ' not found in `/scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray/debug_write_check2_snx3000.out'

Process finished with exit code 1
```

reframe/frontend/executors/__init__.py
```diff
@@ -268,6 +308,9 @@ def __init__(self):
         # Task event listeners
         self.task_listeners = []
 
+        # Associated runner
+        self.runner = None
```
vkarak: The only reason for this dependency is just the printing of the retry message. I would rather have the `runner` print a more generic message like "Retrying XX check(s) that have failed before" and leave the policy untouched.

reframe/frontend/statistics.py
```diff
@@ -8,82 +8,88 @@ def __init__(self, tasks=[]):
...
-            num_cases += len(self._tasks_bypart[partition])
-        else:
-            # count all failures
-            for tclist in self._tasks_bypart.values():
-                num_cases += len(tclist)
-
-        return num_cases
-
-    def failure_report(self):
+    # The retry_num can be indexed from the end (e.g. -1 for the last retry)
```
vkarak: I suggest keeping the interface of statistics to the bare minimum. Only the functionality needed in other places of the framework must be implemented + a function that would allow sb to iterate over the finished tasks.

reframe/frontend/statistics.py
```diff
@@ -8,82 +8,88 @@ def __init__(self, tasks=[]):
...
-                    # This shouldn't happen...
-                    report.append('Unknown error.')
+        for tf in (t for t in self._tasks if t.failed and
+                                             t.retry_num == retry_num):
+            check = tf.check
+            partition = check.current_partition
+            partname = partition.fullname if partition else 'None'
+            environ_name = (check.current_environ.name
+                            if check.current_environ else 'None')
+            retry_info = ('(for the last of %s retries)' % tf.retry_num
```
vkarak: This message needs a bit of fine-tuning.

reframe/frontend/executors/__init__.py
```diff
@@ -141,24 +150,38 @@ class Runner:
...
     @property
     def policy(self):
         return self._policy
 
     @property
+    # This property function should in general be preferred over self._stats,
+    # because it makes sure that the stats are updated.
     def stats(self):
+        # Always update statistics
+        self._stats = self._policy.getstats()
```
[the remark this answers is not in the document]
omlins: Done.

reframe/frontend/executors/__init__.py
omlins: Done.

omlins:
Current output:
```
/apps/dom/UES/jenkins/6.0.UP04/gpu/easybuild/software/Python-bare/3.6.2/bin/python3.6 /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 2
Command line: /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 2
Reframe version: 2.11
Launched by user: omlins
Launched on host: daint101
Reframe paths
=============
    Check prefix      : 
    Check search path : '/users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py'
    Stage dir prefix  : /scratch/snx3000/omlins/0_temp/stage/
    Output dir prefix : /scratch/snx3000/omlins/0_temp/output/
    Logging dir       : /scratch/snx3000/omlins/0_temp/logs
[==========] Started on Wed Mar 21 19:07:36 2018 

[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[       OK ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[----------] Retrying the 1 failed check(s) (retry 1/2)
[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[----------] Retrying the 1 failed check(s) (retry 2/2)
[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[  FAILED  ] Ran 2 test case(s) from 2 check(s) (1 failure(s) after 2 retries; 0 test case(s) passed in retries)
[==========] Finished on Wed Mar 21 19:09:02 2018 
SUMMARY OF FAILURES
FAILURE INFO for debug_write_check2_snx3000 (for the last of 2 retries)
  * System partition: daint:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray
  * Job type: batch job (id=6524189)
  * Maintainers: []
  * Failing phase: sanity
  * Reason: sanity error: pattern `^Max Write: ' not found in `/scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray/debug_write_check2_snx3000.out'

Process finished with exit code 1
```

reframe/frontend/executors/__init__.py
```diff
@@ -197,6 +213,24 @@ def _environ_supported(self, check, environ):
         else:
             return ret and check.supports_environ(environ.name)
 
+    def _retry(self, checks, system):
+        while (self._stats.num_failures() and self._try_num <
+                                             self._max_retries):
+            self._try_num += 1
+            self._stats.next_try(self._try_num)
+            check_names_failed = self._stats.check_names_failed(try_num=-2)
```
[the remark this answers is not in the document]
vkarak: I would rewrite the whole loop as follows (more or less):

```python
while self._retry_count < self._num_retries:
    failed_checks = list(filter(lambda c: c if self._stats.has_failed(c.name)))
    if not failed_checks:
        break

    self._retry_count += 1
    self._stats.nextrun()
    self._printer.separator(...)
    self._runall(failed_checks, system)
```

reframe/frontend/executors/policies.py
```diff
@@ -304,4 +299,4 @@ def exit(self):
                 raise
 
         self.printer.separator('short single line',
-                               'all spawned checks have finished')
+                               'all spawned checks have finished\n')
```
vkarak: If the new line is to accommodate the retries feature, it should not be added here.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
+from itertools import chain
+
 import reframe.core.debug as debug
+import reframe.core.exceptions as exc
```
vkarak: Just import the exception you want.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
...
+
+    # The try_num can be indexed from the end (e.g. -1 for the last retry)
+    def _convert_try_num(self, try_num):
+        return self._try_num + try_num + 1 if try_num < 0 else try_num
+
+    def num_success_all_retries(self):
+        tasks_retries = list(chain.from_iterable(self._tasks[1:]))
+        return len([t for t in tasks_retries if not t.failed])
+
+    def num_success(self, try_num=0):
```
vkarak: This is not needed.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
...
+        try_num = self._convert_try_num(try_num)
+        return len([t for t in self._tasks[try_num] if not t.failed])
+
+    def num_failures(self, try_num=0):
+        try_num = self._convert_try_num(try_num)
+        return len([t for t in self._tasks[try_num] if t.failed])
+
+    def num_failures_stage(self, stage, try_num=0):
```
vkarak: Not needed.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
...
-        else:
-            # count all failures
-            for tclist in self._tasks_bypart.values():
-                num_cases += len(tclist)
-
-        return num_cases
-
-    def failure_report(self):
+    def next_try(self, try_num):
+        # TODO: RAISE DIFFERENT error?
```
vkarak: I disagree with this interface. You're giving too much power to the callers, without even needing it. Then you have to deal with bad inputs. This should be very very simple:

```python
def addrun(self):
    self._tasks.append([])
    self._current_run += 1
```

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
...
-        # Store test cases per partition internally
-        self._tasks_bypart = {}
-        for t in tasks:
-            partition = t.check.current_partition
-            partname = partition.fullname if partition else 'None'
-            tclist = self._tasks_bypart.setdefault(partname, [])
-            tclist.append(t)
+    def __init__(self):
+        self._tasks = []
+        self._try_num = -1
```
vkarak: This variable should hold the current run (name it also accordingly). Just that. It will be used by the `add_task` to append a task to the right list.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
...
-
-        # Store test cases per partition internally
-        self._tasks_bypart = {}
-        for t in tasks:
-            partition = t.check.current_partition
-            partname = partition.fullname if partition else 'None'
-            tclist = self._tasks_bypart.setdefault(partname, [])
-            tclist.append(t)
+    def __init__(self):
+        self._tasks = []
```
vkarak: This variable is also confusing. It's no more a list of tasks. It holds the runs per task. Honestly, I would prefer a dictionary. It would make the code much easier to read and would facilitate the implementation. I could understand the list of lists solution in a different language, but here I would definitely go for a dictionary.

```python
self._tasks_per_run = {}
```

Then you would initialise list of tasks simply as follows:

```python
self._task_per_run[self._current_run] = []
```

It can also be an `OrderedDict`, so as to keep the order of the keys (i.e., the runs).

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
...
+            raise exc.ReframeFatalError('try_num out of sync')
+
+        self._try_num = try_num
+        self._tasks.append([])
+
+    def add_task(self, task):
+        self._tasks[self._try_num].append(task)
+
+    # The try_num can be indexed from the end (e.g. -1 for the last retry)
+    def _convert_try_num(self, try_num):
```
vkarak: I don't see the purpose of this method. Python supports backward indexing with negative indexes natively.

reframe/frontend/executors/__init__.py
```diff
@@ -168,14 +180,18 @@ def runall(self, checks, system):
...
             self._runall(checks, system)
+            if self._max_retries > 0:
+                self._retry(checks, system)
+
         finally:
-            # Always update statistics and print the summary line
-            self._stats = self._policy.getstats()
-            num_failures = self._stats.num_failures()
+            # Print the summary line
+            num_failures_last_retry = self._stats.num_failures(try_num=-1)
```
vkarak: The `num_failures` should return the failures of the last retry by default.

reframe/frontend/executors/__init__.py
```diff
@@ -168,14 +180,18 @@ def runall(self, checks, system):
...
             self._printer.status(
-                'FAILED' if num_failures else 'PASSED',
-                'Ran %d test case(s) from %d check(s) (%d failure(s))' %
-                (self._stats.num_cases(), len(checks), num_failures),
+                'FAILED' if num_failures_last_retry else 'PASSED',
+                'Ran %d test case(s) from %d check(s) (%d failure(s) after '
+                '%d retries; %d test case(s) passed in retries)' %
+                (self._stats.num_cases(), len(checks), num_failures_last_retry,
+                 self._try_num, self._stats.num_success_all_retries()),
```
vkarak: I would keep the old message as it was and add a `retry_report()` method in the statistics. This report would print "No regression test has been retried." if none was retried otherwise it should print sth like the following:

```
RETRY SUMMARY
- Test blah_blah_test has been retried 1 time(s) and passed.
- Test blouh_blouh_test has been retried 2 time(s) and failed.
```

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,98 @@
...
+
+    def num_success_all_retries(self):
+        tasks_retries = list(chain.from_iterable(self._tasks[1:]))
+        return len([t for t in tasks_retries if not t.failed])
+
+    def num_success(self, try_num=0):
+        try_num = self._convert_try_num(try_num)
+        return len([t for t in self._tasks[try_num] if not t.failed])
+
+    def num_failures(self, try_num=0):
```
vkarak: The second argument of this method and all similar ones must be named as `run`. It should be by default `-1` so as to give you always the results of the last run. **BUT** you must check for index/key errors and raise an error accordingly:

```python
def _get_tasks(run):
    try:
        return self._tasks[run]
    except KeyError: # if you go for dictionaries
        raise StatisticsError('no such run: %s' % run)
```

vkarak:
@omlins We should also add a unit test for this feature and a couple of lines in the documentation. I forgot to mention these in my review.

reframe/frontend/cli.py
```diff
@@ -153,6 +153,10 @@ def main():
     run_options.add_argument(
         '--mode', action='store', help='Execution mode to use'
     )
+    run_options.add_argument(
+        '--max-retries', metavar='MAX', action='store', default=0,
+        help='Specify the maximum number of retries per regression test.'
```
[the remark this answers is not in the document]
omlins: Done.

reframe/frontend/cli.py
```diff
@@ -153,6 +153,10 @@ def main():
     run_options.add_argument(
         '--mode', action='store', help='Execution mode to use'
     )
+    run_options.add_argument(
+        '--max-retries', metavar='MAX', action='store', default=0,
```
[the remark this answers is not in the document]
omlins: I believe that `--num-retries` is a bit ambigous: one could think that ReFrame retries failed tests the number of times given by `--num-retries`, no matter whether the tests passed in the retries or not; one could imagine that ReFrame reports how many retries failed and how many passed... 
However, what we do is to stop retrying as soon as a test has passed, i.e. the given argument is the maximum number of retries.
What do you think?

reframe/frontend/cli.py
```diff
@@ -429,13 +433,14 @@ def main():
             exec_policy.sched_nodelist = options.nodelist
             exec_policy.sched_exclude_nodelist = options.exclude_nodes
             exec_policy.sched_options = options.job_options
-            runner = Runner(exec_policy, printer)
+            max_retries = int(options.max_retries)
```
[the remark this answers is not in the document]
omlins: Done.

reframe/frontend/cli.py
```diff
@@ -429,13 +433,14 @@ def main():
...
+            runner = Runner(exec_policy, printer, max_retries)
             try:
                 runner.runall(checks_matched, system)
             finally:
-                # always print a report
                 if runner.stats.num_failures():
-                    printer.info(runner.stats.failure_report())
+                    # always print a report (if retries, for the last retry)
+                    printer.info(runner.stats.failure_report(try_num=-1))
```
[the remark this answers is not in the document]
omlins: Done.

reframe/frontend/executors/__init__.py
```diff
@@ -141,18 +142,29 @@ class Runner:
...
     def __repr__(self):
         return debug.repr(self)
 
+    @property
+    def max_retries(self):
+        return self._max_retries
+
+    @property
+    def try_num(self):
+        return self._try_num
```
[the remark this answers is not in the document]
omlins: They are not needed anymore in the current design. Removed.

reframe/frontend/executors/__init__.py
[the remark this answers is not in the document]
omlins: I adapted it. However, I don't think that replacing a proper while condition with a break statement inside the while loop makes it more clear...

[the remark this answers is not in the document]
omlins: You noted "I would prefer sth more primitive. You get the failed tasks and the caller should do the filtering. After all, statistics care only about and manage the tasks". Therefore I created a function `self.tasks_failed` in `statistics` rather than `self.has_failed`, which would take a check name as argument and therefore the `statistics` would again have to deal with checks and not only tasks.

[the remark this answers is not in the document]
omlins: Done.

reframe/frontend/executors/policies.py
omlins: I think it should in general be here, independent of the retries feature.

reframe/frontend/statistics.py
omlins: Done.

reframe/frontend/statistics.py
omlins: Removed.

reframe/frontend/statistics.py
omlins: This is called multiple times in `TestSerialExecutionPolicy`.

reframe/frontend/statistics.py
omlins: No, my intention is not to give power to the caller, I just try to ensure that the counter the `statistics` realizes if his counter is out of sync with the runners counter (as the runner has the real counter of the retries and in statistics we have only a duplication of it).

reframe/frontend/statistics.py
omlins: I now initialize with `self._try_num = 0`. Done.

reframe/frontend/statistics.py
omlins: It does not hold the runs per task but the tasks ordered by (re-)try number. The (re-)try number is an integer (not a string); that's why I store it as a list of lists. 
The code has become much simpler compared to the previous implementation with dictionaries.

reframe/frontend/statistics.py
omlins: Yes, with the new design, it is not anymore needed everywhere, but only needed in `failure_report()`. Adapted.

reframe/frontend/executors/__init__.py
omlins: I had put `try_num=0` as default for all functions in stats for not to interfere with the calls from elsewhere in ReFrame as e.g. the unit tests in `TestSerialExecutionPolicy`. Should I make `try_num=-1` as default for all functions and assure that also in the rest of ReFrame it is called consistently?

reframe/frontend/executors/__init__.py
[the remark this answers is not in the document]
vkarak: > I don't think that replacing a proper while condition with a break statement inside the while loop makes it more clear...

I disagree. There are cases that `break` statements gotos even make the code more clear and more efficient. The reason I used `break` is that I wanted to go over the tests once per loop iteration and not twice (count the failures first, then retrieve the failed tests). Instead, I retrieve once the failed tests and if none exists, I simply break.

reframe/frontend/executors/policies.py
vkarak: No, it's on purpose without a newline.

reframe/frontend/statistics.py
vkarak: No, keep the interface simple. The `Runner` should make sure to call correctly this method and whenever necessary.

reframe/frontend/statistics.py
vkarak: Then perhaps it should be a function in the unit tests only.

reframe/frontend/statistics.py
omlins: The word `run` is already used in different contexts with different meanings. Thus, I would rather go for something like `retry` or `try` or anything similar but not `run`. I think (re-)try makes it pretty clear what we are talking about, while `run` is so general that it could mean many things.

reframe/frontend/cli.py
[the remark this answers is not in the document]
omlins: OK, I used `NUM` as `metavar`. Done.

reframe/frontend/executors/__init__.py
```diff
@@ -141,18 +142,29 @@ class Runner:
     """Responsible for executing a set of regression tests based on an execution
     policy."""
 
-    def __init__(self, policy, printer=None):
-        self._printer = printer or PrettyPrinter()
+    def __init__(self, policy, printer=None, max_retries=0):
         self._policy = policy
+        self._printer = printer or PrettyPrinter()
+        self._max_retries = max_retries
```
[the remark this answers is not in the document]
omlins: Keeping `max_retries` following the discussion in your first comment.

reframe/frontend/executors/__init__.py
```diff
@@ -141,18 +142,29 @@ class Runner:
...
+        self._try_num = 0
```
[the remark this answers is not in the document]
omlins: In a comment below you suggested `run` in basically the same context. I think we should decide for one term and use it everywhere consistently. I started the term discussion could be done in the comment where you suggested `run`...

omlins:
Output before changes:
```
/apps/dom/UES/jenkins/6.0.UP04/gpu/easybuild/software/Python-bare/3.6.2/bin/python3.6 /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 2
Command line: /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 2
Reframe version: 2.11
Launched by user: omlins
Launched on host: daint101
Reframe paths
=============
    Check prefix      : 
    Check search path : '/users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py'
    Stage dir prefix  : /scratch/snx3000/omlins/0_temp/stage/
    Output dir prefix : /scratch/snx3000/omlins/0_temp/output/
    Logging dir       : /scratch/snx3000/omlins/0_temp/logs
[==========] Started on Wed Mar 21 19:07:36 2018 

[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[       OK ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[----------] Retrying the 1 failed check(s) (retry 1/2)
[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[----------] Retrying the 1 failed check(s) (retry 2/2)
[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[  FAILED  ] Ran 2 test case(s) from 2 check(s) (1 failure(s) after 2 retries; 0 test case(s) passed in retries)
[==========] Finished on Wed Mar 21 19:09:02 2018 
SUMMARY OF FAILURES
FAILURE INFO for debug_write_check2_snx3000 (for the last of 2 retries)
  * System partition: daint:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray
  * Job type: batch job (id=6524189)
  * Maintainers: []
  * Failing phase: sanity
  * Reason: sanity error: pattern `^Max Write: ' not found in `/scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray/debug_write_check2_snx3000.out'

Process finished with exit code 1
```

reframe/frontend/statistics.py
vkarak: I don't like the word `try`, because we are introducing a new term for sth that we already have one. The `Runner` _runs_ as a set of checks and we have a _run action_ as well to describe the same this action. So, since now we can run multiple times some tests, I see pretty much natural to count the number of _runs_ you do. The number of retries is always `num_runs - 1`. So, in summary, I would either count the _runs_ or the _retries_.

omlins:
New output with statistics on failures and retries at the end:

/apps/dom/UES/jenkins/6.0.UP04/gpu/easybuild/software/Python-bare/3.6.2/bin/python3.6 /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 3
Command line: /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 3
Reframe version: 2.11
Launched by user: omlins
Launched on host: daint102
Reframe paths
=============
    Check prefix      : 
    Check search path : '/users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py'
    Stage dir prefix  : /scratch/snx3000/omlins/0_temp/stage/
    Output dir prefix : /scratch/snx3000/omlins/0_temp/output/
    Logging dir       : /scratch/snx3000/omlins/0_temp/logs
[==========] Started on Thu Apr  5 10:54:36 2018 

[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[----------] Retrying 2 failed check(s) (retry 1/3)
[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[----------] Retrying 2 failed check(s) (retry 2/3)
[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[----------] Retrying 2 failed check(s) (retry 3/3)
[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[       OK ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray

[  FAILED  ] Ran 2 test case(s) from 2 check(s)
[==========] Finished on Thu Apr  5 10:59:16 2018 
SUMMARY OF FAILURES
  * Number of failed tests: 1
  * Number of retries: 3
  * Number of test cases passed after retries: 1
  * Failed tests: debug_write_check2_snx3000
  * Tests passed after 3 retries (1): debug_write_check_snx3000
FAILURE INFO for debug_write_check2_snx3000 (for the last of 3 retries)
  * System partition: daint:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray
  * Job type: batch job (id=6724313)
  * Maintainers: []
  * Failing phase: sanity
  * Reason: sanity error: pattern `^Max Write: ' not found in `/scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray/debug_write_check2_snx3000.out'

Process finished with exit code 1

reframe/frontend/executors/__init__.py
omlins: I organised the output as follows (the points in parenthesis are not printed if there are no retries):

SUMMARY OF FAILURES
1. global stats on failures
(2. global stats on retries)
3. List of failed test
(4. List of tests passed in retry x (only printed if any test passed in retry x))
5. FAILURE INFO for each failed test

reframe/frontend/executors/__init__.py
vkarak: The most reasonable thing is to always report from the last run. What unit tests do, it should not affect the design. Also, I don't understand the "elsewhere in ReFrame." This method is only called in the `Runner` and in the failure report.

reframe/frontend/executors/__init__.py
vkarak: I would like to keep the reports separate as in my original proposal:

```
[The output of the runs as before + the additional runs]
SUMMARY OF RETRIES
    - Test foo_check on daint:gpu using PrgEnv-gnu was retried 1 time(s) and passed.
    - Test foo_check on daint:gpu using PrgEnv-cray was retried 2 time(s) and failed.
SUMMARY OF FAILURES
[as before]
```
The failure summary should only be printed if there are failures (as of now). The string `foo_check on daint:gpu using PrgEnv-gnu` may be easily retrieved by calling the `info()` method of the check.

reframe/frontend/executors/__init__.py
omlins: OK, I made the last run to be the default for all functions. Done.

reframe/frontend/executors/__init__.py
omlins: OK. One thing though: at no place appears the number of failed test cases. How about, I add it 
in this line
```
[  FAILED  ] Ran 2 test case(s) from 2 check(s)
```
e.g. something like
```
[  FAILED  ] Ran 2 test case(s) from 2 check(s) with 1 failure(s)
```
?

reframe/frontend/statistics.py
omlins: OK, `_current_run` seems perfect to me. Done.

reframe/frontend/executors/__init__.py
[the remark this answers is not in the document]
omlins: Used `_current_run`. Done.

reframe/frontend/statistics.py
omlins: OK, in agreement with the variable `_current_run`, I use here now `run`. Done.

reframe/frontend/executors/__init__.py
vkarak: This status line should be unchanged, exactly as before. It contained already the failure count information. What I am suggesting here is to put the retry information in a separate section.

omlins: I added the separate "SUMMARY OF RETRIES" section following your example. However, if there are no retries at all, then I would rather completely omit the SUMMARY OF RETRIES" section instead of printing "No regression test has been retried.". Do you agree?

omlins:
New output if there are no failures:
```
/apps/dom/UES/jenkins/6.0.UP04/gpu/easybuild/software/Python-bare/3.6.2/bin/python3.6 /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 3
Command line: /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --exec-policy async --max-retries 3
Reframe version: 2.11
Launched by user: omlins
Launched on host: daint102
Reframe paths
=============
    Check prefix      : 
    Check search path : '/users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py'
    Stage dir prefix  : /scratch/snx3000/omlins/0_temp/stage/
    Output dir prefix : /scratch/snx3000/omlins/0_temp/output/
    Logging dir       : /scratch/snx3000/omlins/0_temp/logs
[==========] Started on Tue Apr 10 16:45:03 2018 

[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[       OK ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray

[  PASSED  ] Ran 1 test case(s) from 1 check(s) (0 failure(s))
[==========] Finished on Tue Apr 10 16:45:13 2018 

Process finished with exit code 0
```

reframe/frontend/statistics.py
omlins: I have deleted this function and created instead one in `TestSerialExecutionPolicy`. Done.

omlins:
Output for the serial execution policy:
```
/apps/dom/UES/jenkins/6.0.UP04/gpu/easybuild/software/Python-bare/3.6.2/bin/python3.6 /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --max-retries 3
Command line: /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files -P debug --max-retries 3
Reframe version: 2.11
Launched by user: omlins
Launched on host: daint102
Reframe paths
=============
    Check prefix      : 
    Check search path : '/users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py'
    Stage dir prefix  : /scratch/snx3000/omlins/0_temp/stage/
    Output dir prefix : /scratch/snx3000/omlins/0_temp/output/
    Logging dir       : /scratch/snx3000/omlins/0_temp/logs
[==========] Started on Wed Apr 11 15:39:36 2018 

[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[----------] Retrying 2 failed check(s) (retry 1/3)
[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[       OK ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[----------] Retrying 1 failed check(s) (retry 2/3)
[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[----------] Retrying 1 failed check(s) (retry 3/3)
[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[  FAILED  ] Ran 2 test case(s) from 2 check(s) (1 failure(s))
[==========] Finished on Wed Apr 11 15:42:20 2018 
SUMMARY OF RETRIES
  * Test debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray was retried 3 time(s) and failed.
  * Test debug_write_check_snx3000 on daint:gpu using PrgEnv-cray was retried 1 time(s) and passed.
SUMMARY OF FAILURES
FAILURE INFO for debug_write_check2_snx3000 (for the last of 3 retries)
  * System partition: daint:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray
  * Job type: batch job (id=6795334)
  * Maintainers: []
  * Failing phase: sanity
  * Reason: sanity error: pattern `^Max Write: ' not found in `/scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray/debug_write_check2_snx3000.out'

Process finished with exit code 1

```

reframe/frontend/executors/__init__.py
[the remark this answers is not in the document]
omlins: > The reason I used break is that I wanted to go over the tests once per loop iteration and not twice (count the failures first, then retrieve the failed tests). 
I see your point. However, if you want to do optimization, I would rather do it inside the `statistics`. There one could make sure that everything is computed in an optimal way. Yet, I doubt that the computation of the statistics is at all relevant for the performance of ReFrame. Running x tests should always take *much* more time than reading from x jobs whether it failed etc. So I would not trade in simplicity and clarity for unneeded performance optimizations.

omlins:
@vkarak: could you have a look again? I think it can be merged soon now...

reframe/frontend/executors/__init__.py
```diff
@@ -168,15 +175,16 @@ def runall(self, checks, system):
             self._printer.timestamp('Started on', 'short double line')
             self._printer.info()
             self._runall(checks, system)
+            if self._max_retries:
+                self._retry_failed(checks, system)
+
         finally:
-            # Always update statistics and print the summary line
-            self._stats = self._policy.getstats()
-            num_failures = self._stats.num_failures()
```
vkarak: Keep this line so as to avoid the double call to `num_failures`. It will also facilitate the formatting.

reframe/frontend/executors/__init__.py
```diff
@@ -168,15 +175,16 @@ def runall(self, checks, system):
...
+            # Print the summary line
             self._printer.status(
-                'FAILED' if num_failures else 'PASSED',
+                'FAILED' if self._stats.num_failures() else 'PASSED',
                 'Ran %d test case(s) from %d check(s) (%d failure(s))' %
-                (self._stats.num_cases(), len(checks), num_failures),
-                just='center'
+                (self._stats.num_cases(run=0), len(checks),
```
vkarak: You may also move this call out of this statement.

reframe/frontend/executors/__init__.py
```diff
@@ -197,6 +205,21 @@ def _environ_supported(self, check, environ):
         else:
             return ret and check.supports_environ(environ.name)
 
+    def _retry_failed(self, checks, system):
+        while (self._stats.num_failures() and self._current_run <
+                                              self._max_retries):
```
vkarak:
```python
while (self._stats.num_failures() and
       self._current_run < self._max_retries):
```

reframe/frontend/executors/__init__.py
```diff
@@ -197,6 +205,21 @@ def _environ_supported(self, check, environ):
...
+            self._current_run += 1
+            self._stats.next_run(self._current_run)
```
vkarak: I disagree passing the current run to the `next_run`. `next_run` should simply advance the run count and take no argument. If sb wants to check if everything is ok, should do by himself, i.e., statistics should offer read-only access to their internal run count and any caller could validate the status.

What you are doing here is basically a `set_run`, but I really don't see the purpose for that.

vkarak: You should also advance the statistics run counter after you get the failures so as not to have to pass `run=-2` later.

reframe/frontend/executors/policies.py
vkarak: Please remove the extra newline, because it keeps consistency of the output in exceptional cases (cancellation, serial execution etc.)

reframe/frontend/executors/__init__.py
```diff
@@ -197,6 +205,21 @@ def _environ_supported(self, check, environ):
...
+            check_names_failed = set([t.check.name for t in
+                                      self._stats.tasks_failed(run=-2)])
+            checks_failed = [c for c in checks if c.name in check_names_failed]
```
vkarak: You can rewrite this with a single list comprehension:

```python
            failed_checks = [
                c for c in checks if c.name in
                set(t.check.name for t in self._stats.tasks_failed())
            ]
```

vkarak: I prefer also the name `failed_checks` over the `checks_failed`.

reframe/frontend/executors/policies.py
```diff
@@ -27,6 +22,7 @@ def run_check(self, check, partition, environ):
         )
         task = RegressionTask(check)
         self._tasks.append(task)
+        self.stats.add_task(task)
```
vkarak: This should go in the base class.

reframe/frontend/executors/policies.py
```diff
@@ -121,6 +117,7 @@ def __init__(self):
 
         self.task_listeners.append(self)
 
+
```
vkarak: Remove this line, it's against the coding style.

reframe/frontend/executors/policies.py
```diff
@@ -154,13 +151,11 @@ def enter_partition(self, c, p):
...
         self._max_jobs.setdefault(p.fullname, p.max_jobs)
 
-    def getstats(self):
-        return TestStats(self._tasks)
-
     def run_check(self, check, partition, environ):
         super().run_check(check, partition, environ)
         task = RegressionTask(check, self.task_listeners)
         self._tasks.append(task)
+        self.stats.add_task(task)
```
vkarak: Please move this to the base class.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,103 @@
...
-        num_cases = 0
-        if partition:
-            num_cases += len(self._tasks_bypart[partition])
-        else:
-            # count all failures
-            for tclist in self._tasks_bypart.values():
-                num_cases += len(tclist)
-
-        return num_cases
+    def next_run(self, current_run):
```
vkarak: As I said before, this should not take an argument and the check below should be removed.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,103 @@
...
-
-        # Store test cases per partition internally
-        self._tasks_bypart = {}
-        for t in tasks:
-            partition = t.check.current_partition
-            partname = partition.fullname if partition else 'None'
-            tclist = self._tasks_bypart.setdefault(partname, [])
-            tclist.append(t)
+    def __init__(self):
+        self._tasks = [[]]
```
vkarak: Please add a small comment as to what this nested list keeps.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,103 @@
...
+        self._tasks.append([])
+
+    def add_task(self, task):
+        self._tasks[self._current_run].append(task)
+
+    def _get_tasks(self, run):
+        try:
+            return self._tasks[run]
+        except IndexError:
+            raise StatisticsError('no such run: %s' % run)
```
vkarak: `from None`

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,103 @@
 import reframe.core.debug as debug
+from reframe.core.exceptions import (ReframeFatalError, StatisticsError)
```
vkarak: Leave a blank line before this one.

reframe/frontend/statistics.py
vkarak: You can instead add a property that reads out the `current_run`.

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,103 @@
...
+
+        line_width = 78
+        report = [line_width * '=']
+        report.append('SUMMARY OF RETRIES')
+        report.append(line_width * '-')
+        messages = {}
+        for run in range(1, len(self._tasks)):
+            for t in self._get_tasks(run):
+                # Overwrite entry from previous run if available
+                messages[t.check.info()] = (
```
vkarak: You should not rely on `check.info()` to hash the messages! This may be changed by the user.

vkarak: You should manually create a key out of the check name and the partition name as follows:

```python
key = '%s:%s:%s' % (t.check.name, t.check.current_partition.fullname, t.check.current_environ.name)
```

reframe/frontend/statistics.py
```diff
@@ -1,89 +1,103 @@
...
+                    '  * Test %s was retried %s time(s) and %s.' %
+                    (t.check.info(), run, 'failed' if t.failed else 'passed')
+                )
+
+        for key in sorted(messages.keys()):
```
vkarak: What's the purpose of sorting here?

unittests/test_policies.py
```diff
@@ -34,15 +34,20 @@ def setUp(self):
     def tearDown(self):
         shutil.rmtree(self.resourcesdir, ignore_errors=True)
 
+    def _num_failures_stage(self, stage):
+        stats = self.runner.stats
+        return len([t for t in stats.tasks_failed() if t.failed_stage ==
+                                                         stage])
```
vkarak: Fix the line split here. Move the whole `if` statement in the next line.

vkarak:
I have a couple more comments regarding the output:

A matching line (like a closing box) should be added after each retry. This way you won't have to append a new line after the `all spawned checks have finished` message.

```
[----------] Retrying 2 failed check(s) (retry 1/3)
[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[       OK ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] Finished retrying failed checks
```

Perhaps each retry output box should be denoted with `=`, i.e., a `short double line`.

reframe/frontend/executors/policies.py
omlins: Done.

reframe/frontend/cli.py
```diff
@@ -153,6 +153,10 @@ def main():
     run_options.add_argument(
         '--mode', action='store', help='Execution mode to use'
     )
+    run_options.add_argument(
+        '--max-retries', metavar='NUM', action='store', default=0,
+        help='Specify the maximum number of times a failed regression test may'
+             ' be retried (default: 0)')
```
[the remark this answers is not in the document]
omlins: Done.

docs/running.rst
```diff
@@ -332,6 +332,9 @@ They are summarized below:
 
     ./bin/reframe -c /path/to/my/check.py -p PrgEnv-gnu --skip-prgenv-check -r
 
+* ``--max-retries NUM``: Specify the maximum number of times a failed
+                        regression test may be retried (default: 0).
```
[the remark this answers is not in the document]
omlins: Done.

reframe/frontend/executors/__init__.py
```diff
@@ -141,13 +142,19 @@ class Runner:
...
-    def __init__(self, policy, printer=None):
-        self._printer = printer or PrettyPrinter()
+    def __init__(self, policy, printer=None, max_retries=0):
         self._policy = policy
+        self._printer = printer or PrettyPrinter()
+        try:
+            self._max_retries = int(max_retries)
+        except ValueError:
+            raise ConfigError('--max-retries is not a valid integer: %s' %
+                              max_retries) from None
```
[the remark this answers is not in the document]
omlins: Yes, I prefer as well first solution very much. Done.

reframe/frontend/cli.py
```diff
@@ -429,12 +433,17 @@ def main():
             exec_policy.sched_nodelist = options.nodelist
             exec_policy.sched_exclude_nodelist = options.exclude_nodes
             exec_policy.sched_options = options.job_options
-            runner = Runner(exec_policy, printer)
+            max_retries = options.max_retries
+            runner = Runner(exec_policy, printer, max_retries)
```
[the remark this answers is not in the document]
omlins: Kept separate as int casting and exception handling was moved here (see comment below). Done.

reframe/frontend/cli.py
```diff
@@ -429,12 +433,17 @@ def main():
...
             try:
                 runner.runall(checks_matched, system)
             finally:
-                # always print a report
+                if runner.stats.num_failures(run=0):
```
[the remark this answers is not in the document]
omlins: I thought that it would be clearer this way that the retry report is only printed if there were any retries (= if any failures occurred in run 0). Do you still want to remove the if statement?

reframe/frontend/executors/__init__.py
omlins: I really don't think that we should try to reduce the number of calls to functions of statistics, because (1) I do not believe that it is at all relevant for performance and (2) if it turned out to be performance relevant, then we should optimise the class statistics such that it stores once computed results internally to avoid recomputing them.

reframe/frontend/executors/__init__.py
omlins: Done.

reframe/frontend/executors/__init__.py
omlins: OK, I will remove the `current_run` argument from `next_run`. However, if you think about giving read-only access to the internal run count of the `statistics`, then I would say that we should rather give read-only access to the internal run count of the `Runner`. In fact, the runner has the original and always correct count; so it would make much more sense that this is made accessible to others. Yet, you did not like that in my previous versions; so I guess, we should just remove any counter verification...

reframe/frontend/cli.py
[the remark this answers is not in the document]
vkarak: You can leave it as it is and put a comment saying sth like "print a retry report if we had failures in the first run".

reframe/frontend/executors/__init__.py
vkarak: My argument is mostly for clarity and easier formatting.

reframe/frontend/executors/__init__.py
omlins: Done.

reframe/frontend/executors/policies.py
omlins: I am not sure if I remember right, but didn't you say that the policies are not necessarily required to use tasks?

reframe/frontend/executors/policies.py
omlins: Done.

reframe/frontend/executors/__init__.py
vkarak: I am not against the counter verification, and it's good to have such guards generally (btw, it should raise `AssertionError` since it denotes a bug). What I am against to is the interface. Looking at the interface, it's like having a counter and an increment operation on it, but instead of letting the counter increment itself, you pass it's new value (the one *you * count) and the counter then checks this against its own counter. Doesn't this sound weird? It would be much more natural to let the counter count the way it knows and you also count your own way and compare your values afterwards for sanity.

reframe/frontend/statistics.py
vkarak: I would vote for `# Tasks per run stored as follows: [[run0_tasks], [run1_tasks], ...]`

reframe/frontend/statistics.py
omlins: Then, I should also use these explicit properties instead of `check.info()` to create the message?

reframe/frontend/statistics.py
omlins: To have the output sorted by the test name (and the other elements that appear in the key) as
```
Test A ...
Test A ...
Test A ...
Test C ...
Test D ...
Test D ...
```
 Else we could have an output like:
```
Test D ...
Test A ...
Test C ...
Test A ...
Test D ...
Test A ...
```

unittests/test_policies.py
omlins: Done.

reframe/frontend/statistics.py
vkarak: No, for the message to print you don't need to care. I've made the comment because you were hashing the message, which may not be unique.

reframe/frontend/statistics.py
vkarak: Ok, leave it sorted then.

reframe/frontend/statistics.py
omlins: Done.

reframe/frontend/cli.py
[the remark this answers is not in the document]
omlins: Done.

reframe/frontend/executors/__init__.py
omlins: Sorry, I thought you were still talking about the optimisation. Yes, of course it looks better with the additional line `num_failures = self._stats.num_failures()`. Done.

reframe/frontend/executors/__init__.py
omlins: Done.

reframe/frontend/statistics.py
omlins: Done.

reframe/frontend/statistics.py
omlins: Done.

reframe/frontend/statistics.py
omlins: Done.

reframe/frontend/executors/__init__.py
vkarak: If we adopt what you propose for a `set_run()`, we are violating the notion of setting a property. Theoretically, with a `set_run` one would expect to be able to set any run number, but here it would be just an increment. It is exactly as the example with the counter I gave you before; it's like calling `counter.inc(new_value)`, where the only allowed value for `new_value` would be the current plus one. It doesn't make any sense, because in this case the caller does the counting, and the counter does the sanity checking! It should be the other way around. So I definitely prefer the first one.

omlins: OK, done.

reframe/frontend/statistics.py
omlins: Done.

omlins:
...and the corresponding new output for the sync execution policy:
```
/apps/dom/UES/jenkins/6.0.UP04/gpu/easybuild/software/Python-bare/3.6.2/bin/python3.6 /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files --max-retries 2 -P debug
Command line: /users/omlins/2_reframe/current/reframe/reframe.py -c /users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py -r --prefix /scratch/snx3000/omlins/0_temp --keep-stage-files --max-retries 2 -P debug
Reframe version: 2.11
Launched by user: omlins
Launched on host: daint101
Reframe paths
=============
    Check prefix      : 
    Check search path : '/users/omlins/2_reframe/current/reframe/cscs-checks/system/io/ior_mini_check.py'
    Stage dir prefix  : /scratch/snx3000/omlins/0_temp/stage/
    Output dir prefix : /scratch/snx3000/omlins/0_temp/output/
    Logging dir       : /scratch/snx3000/omlins/0_temp/logs
[==========] Started on Thu Apr 19 16:30:43 2018 

[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[==========] Retrying 2 failed check(s) (retry 1/2)
[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[==========] Retrying 2 failed check(s) (retry 2/2)
[----------] started processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check_snx3000 (Debug check (/scratch/snx3000))

[----------] started processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))
[ RUN      ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[     FAIL ] debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray
[----------] finished processing debug_write_check2_snx3000 (Debug check (/scratch/snx3000))

[  FAILED  ] Ran 2 test case(s) from 2 check(s) (2 failure(s))
[==========] Finished on Thu Apr 19 16:32:27 2018 
SUMMARY OF RETRIES
  * Test debug_write_check2_snx3000 on daint:gpu using PrgEnv-cray was retried 2 time(s) and failed.
  * Test debug_write_check_snx3000 on daint:gpu using PrgEnv-cray was retried 2 time(s) and failed.
SUMMARY OF FAILURES
FAILURE INFO for debug_write_check_snx3000 (for the last of 2 retries)
  * System partition: daint:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check_snx3000/PrgEnv-cray
  * Job type: batch job (id=6916435)
  * Maintainers: []
  * Failing phase: sanity
  * Reason: sanity error: pattern `^Max Write: ' not found in `/scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check_snx3000/PrgEnv-cray/debug_write_check_snx3000.out'
FAILURE INFO for debug_write_check2_snx3000 (for the last of 2 retries)
  * System partition: daint:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray
  * Job type: batch job (id=6916436)
  * Maintainers: []
  * Failing phase: sanity
  * Reason: sanity error: pattern `^Max Write: ' not found in `/scratch/snx3000/omlins/0_temp/stage/gpu/debug_write_check2_snx3000/PrgEnv-cray/debug_write_check2_snx3000.out'

Process finished with exit code 1
```

reframe/frontend/executors/__init__.py
```diff
@@ -197,6 +202,25 @@ def _environ_supported(self, check, environ):
...
+        while (self._stats.num_failures() and
+               self._current_run < self._max_retries):
+            failed_checks = [
+                c for c in checks if c.name in
+                set([t.check.name for t in self._stats.tasks_failed()])
+            ]
+            self._current_run += 1
+            self._stats.next_run()
+            if self._stats.current_run != self._current_run:
+                    raise AssertionError('current_run variable out of sync')
```
vkarak: You could print the two values here to help debugging.

unittests/resources/frontend_checks.py
```diff
@@ -159,6 +159,19 @@ def __init__(self, sleep_time, **kwargs):
         SleepCheck._next_id += 1
 
 
+class RetriesCheck(BaseFrontendCheck):
+
+    def __init__(self, run_to_pass, file_name, **kwargs):
```
vkarak: I would use `filename` as a single word.

unittests/resources/frontend_checks.py
```diff
@@ -159,6 +159,19 @@ def __init__(self, sleep_time, **kwargs):
...
+        super().__init__(type(self).__name__, **kwargs)
+        self.sourcesdir = None
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
+        self.executable = 'bash'
+        self.executable_opts = ['-c "echo $(($(cat %s) + 1)) > %s "' %
+                                (file_name, file_name)]
```
vkarak: I would use `pre_run` and `post_run` to make it cleaner and also allow numbering to start from `0`:

```python
self.pre_run = ['current_run=$(cat %s)' % filename]
self.executable = ['echo $current_run']
self.post_run = ['((current_run++))', 'echo $current_run > %s' % filename]
```

unittests/test_policies.py
```diff
@@ -115,6 +123,49 @@ def test_system_exit_within_test(self):
...
+        self.assertEqual(0, self.runner.stats.current_run)
+        self.assertEqual(0, self.runner.stats.num_failures())
+
+    def test_pass_in_retries(self):
+        max_retries = 3
+        run_to_pass = 2
+        # Create a file containing the current_run; Run 0 will set it to 0,
+        # run 1 to 1 and so on.
+        with tempfile.NamedTemporaryFile(mode='wt', delete=False) as fp:
+            fp.write("-1\n")
```
vkarak: Please use single quotes here. You may also consider my suggestion for the `RetriesCheck` so as to start from the numbering from `0`.

unittests/test_policies.py
```diff
@@ -115,6 +123,49 @@ def test_system_exit_within_test(self):
...
+                               resources=self.resources)]
+        self.runner._max_retries = max_retries
+        self.runner.runall(checks, self.system)
+
+        # Ensure that the test passed after retries in run #run_to_pass.
+        self.assertEqual(1, self.runner.stats.num_cases())
+        self.assertEqual(1, self.runner.stats.num_failures(run=0))
+        self.assertEqual(run_to_pass, self.runner.stats.current_run)
+        self.assertEqual(0, self.runner.stats.num_failures())
+
```
vkarak: You may remove this line.

unittests/resources/frontend_checks.py
omlins: Done.

reframe/frontend/executors/__init__.py
omlins: Done.

unittests/resources/frontend_checks.py
omlins: Great idea! Done.

omlins: (in addition I had to change the sanity check such that it reads again from `stdout` and not from `filename` - that's what you must have had in mind...)

unittests/test_policies.py
omlins: Done.

unittests/test_policies.py
omlins: Done.

vkarak:
Fixes #148.

vkarak:
@jenkins-cscs retry monch

PR #195, opened 2018-03-08
ReFrame upcoming version

vkarak:


vkarak:
@victorusu Do you agree with the version numbering?

PR #196, opened 2018-03-09
Add check for TensorFlow

### LINKED ISSUE #103: Add regression test for TensorFlow
Internal issue: https://madra.cscs.ch/scs/reframe/issues/218

teojgo:
* Add `mnist` and `wide_deep` official model tests for TensorFlow.

 Closes #103

### DIFF
--- a/cscs-checks/apps/tensorflow/tensorflow_check.py
+++ b/cscs-checks/apps/tensorflow/tensorflow_check.py
@@ -1,0 +1,72 @@
+import os
+import itertools
+import reframe.utility.sanity as sn
+
+from reframe.core.pipeline import RunOnlyRegressionTest
+
+
+class TensorFlowTestBase(RunOnlyRegressionTest):
+    def __init__(self, model_name, **kwargs):
+        super().__init__('tensorflow_%s_check' % model_name,
+                         os.path.dirname(__file__), **kwargs)
+
+        self.descr = 'Tensorflow official %s test' % model_name
+        self.strict_check = False
+        self.valid_systems = ['daint:gpu', 'dom:gpu']
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.sourcesdir = 'https://github.com/tensorflow/models.git'
+        self.maintainers = ['TM']
+        self.tags = {'production'}
+        self.num_tasks = 1
+        self.num_gpus_per_node = 1
+        self.modules = ['TensorFlow/1.4.1-CrayGNU-17.08-cuda-8.0-python3']
+
+        # Checkout to the branch corresponding to the module version of
+        # TensorFlow
+        self.pre_run = ['git checkout r1.4.0']
+
+
+class TensorFlowMnistTest(TensorFlowTestBase):
+    def __init__(self, **kwargs):
+        super().__init__('mnist', **kwargs)
+        self.executable = ('python3 ./official/mnist/mnist.py '
+                           '--model_dir "." --export_dir "." --data_dir "."')
+
+        self.sanity_patterns = sn.all([
+            sn.assert_found('INFO:tensorflow:Finished evaluation at',
+                            self.stderr),
+            sn.assert_gt(sn.extractsingle(
+                r"Evaluation results:\s+\{.*'accuracy':\s+(?P<accuracy>\S+)"
+                r"(?:,|\})", self.stdout, 'accuracy', float), 0.99)
+        ])
+
+
+class TensorFlowWidedeepTest(TensorFlowTestBase):
+    def __init__(self, **kwargs):
+        super().__init__('wide_deep', **kwargs)
+
+        train_epochs = 10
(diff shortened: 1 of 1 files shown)


cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -0,0 +1,72 @@
...
+        self.descr = 'Tensorflow official %s test' % model_name
+        self.strict_check = False
+        self.valid_systems = ['daint:gpu', 'dom:gpu']
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.sourcesdir = 'https://github.com/tensorflow/models.git'
+        self.maintainers = ['TM']
+        self.tags = {'production'}
+        self.num_tasks = 1
+        self.num_gpus_per_node = 1
+        self.modules = ['TensorFlow/1.4.1-CrayGNU-17.08-cuda-8.0-python3']
```
victorusu: Why not use as module name `TF/1.4.1-python3` and map the module name to the real module?
Probably this way, we could extend it to get `1.4.1` as a parameter and do the proper `git checkout` and have different TF modules versions mapped.

What do you think?

teojgo: I guess that in order to achieve this a corresponding file with the actual mappings has to be created. This needs also changes in the ci to use the corresponding command line argument. @vkarak what do you think?

vkarak: Decoupling the module names in the tests from the real modules, I see it as a separate project. I don't think we should address it specifically for this test.

cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -0,0 +1,72 @@
+import os
+import itertools
+import reframe.utility.sanity as sn
+
+from reframe.core.pipeline import RunOnlyRegressionTest
+
+
+class TensorFlowTestBase(RunOnlyRegressionTest):
```
vkarak: `TensorFlowBaseTest`

cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -0,0 +1,72 @@
...
+    def __init__(self, model_name, **kwargs):
+        super().__init__('tensorflow_%s_check' % model_name,
+                         os.path.dirname(__file__), **kwargs)
+
+        self.descr = 'Tensorflow official %s test' % model_name
+        self.strict_check = False
```
vkarak: This is relevant only for performance tests.

cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -0,0 +1,72 @@
...
+        # Checkout to the branch corresponding to the module version of
+        # TensorFlow
+        self.pre_run = ['git checkout r1.4.0']
+
+
+class TensorFlowMnistTest(TensorFlowTestBase):
+    def __init__(self, **kwargs):
+        super().__init__('mnist', **kwargs)
+        self.executable = ('python3 ./official/mnist/mnist.py '
+                           '--model_dir "." --export_dir "." --data_dir "."')
```
vkarak: I would use also the `executable_opts` here.

cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self.sanity_patterns = sn.all([
+            sn.assert_found('INFO:tensorflow:Finished evaluation at',
```
vkarak: Use raw strings here.

cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -0,0 +1,72 @@
...
+                            self.stderr),
+            sn.assert_reference(sn.extractsingle(
+                r"Results at epoch %s[\s\S]+accuracy:\s+(?P<accuracy>\S+)" %
+                train_epochs, self.stdout, 'accuracy', float),
+                0.85, -0.05, None)
+
+        ])
+
+        self.pre_run.extend(['mkdir ./official/wide_deep/model_dir',
```
vkarak: `+=` is nicer here.

cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -0,0 +1,72 @@
...
+        # TensorFlow
+        self.pre_run = ['git checkout r1.4.0']
+
+
+class TensorFlowMnistTest(TensorFlowBaseTest):
+    def __init__(self, **kwargs):
+        super().__init__('mnist', **kwargs)
+        self.executable = 'python3 ./official/mnist/mnist.py'
+        self.executable_opts = ['--model_dir', '"."', '--export_dir', '"."',
+                                ' --data_dir', '"."']
```
vkarak: Do you really need the quotes around `.` here?

PR #199, opened 2018-03-12
Adapt Gromacs production

teojgo:


cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -108,10 +108,10 @@ def __init__(self, **kwargs):
         self.tags |= {'production'}
         self.reference = {
             'dom:mc': {
-                'perf': (36.0, -0.15, None)
+                'perf': (38.36, -0.05, None)
             },
             'daint:mc': {
-                'perf': (73.0, -0.25, None)   # (82.0, -0.15, None)
+                'perf': (72.98, -0.50, None)
```
victorusu: The difference of `72.98` and `73.0` is negligible and it `73.0` is easier to parse. I would not change that. What do you think?

cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -108,10 +108,10 @@ def __init__(self, **kwargs):
         self.tags |= {'production'}
         self.reference = {
             'dom:mc': {
-                'perf': (36.0, -0.15, None)
+                'perf': (38.36, -0.05, None)
```
victorusu: I would round to `38.0` so it is easier to parse. What do you think?

cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -76,10 +76,10 @@ def __init__(self, **kwargs):
         self.tags |= {'production'}
         self.reference = {
             'dom:gpu': {
-                'perf': (15.7, -0.15, None)
+                'perf': (32.05, -0.05, None)
             },
             'daint:gpu': {
-                'perf': (55.1, -0.39, None)
+                'perf': (47.71, -0.40, None)
```
victorusu: I would round to `47.5` so it is easier to parse. What do you think?

cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -76,10 +76,10 @@ def __init__(self, **kwargs):
         self.tags |= {'production'}
         self.reference = {
             'dom:gpu': {
-                'perf': (15.7, -0.15, None)
+                'perf': (32.05, -0.05, None)
```
victorusu: I would round to `32.0` so it is easier to parse. What do you think?

vkarak: I would say either `32.0` or `32.1` depending on the real median value.

PR #200, opened 2018-03-12
Add numerical sanity check to cp2k test

### LINKED ISSUE #160: Check numerical sanity in CP2K test

rsarm:
Closes #160


rsarm:
I have set the same references and boundaries both for gpu and cpu. 

`gpu`
Mean = -4404.232288847591  (-1.7260723051323787e-08, 7.0019334434651457e-09)

`cpu`
Mean = -4404.232290105533  (-1.7060044662503173e-08, 7.3283963208809364e-09)

`cpu and gpu together`
Mean = -4404.232289541480 (-1.7188115373848677e-08, 7.200325612658878e-09)

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -13,11 +13,15 @@ def __init__(self, check_name, check_descr, **kwargs):
         self.executable = 'cp2k.psmp'
         self.executable_opts = ['H2O-256.inp']
 
+        energy = sn.extractsingle(r'\s+ENERGY\| Total FORCE_EVAL \( QS \) '
+                                  r'energy \(a.u.\):\s+(?P<energy>\S+)',
```
vkarak: The dots should be escaped here, if you actually mean the `.` character.

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -13,11 +13,15 @@ def __init__(self, check_name, check_descr, **kwargs):
...
+                                  self.stdout, 'energy', float, item=-1)
         self.sanity_patterns = sn.all([
             sn.assert_found(r'PROGRAM STOPPED IN', self.stdout),
             sn.assert_eq(sn.count(sn.extractall(
                 r'(?P<step_count>STEP NUM)',
-                self.stdout, 'step_count')), 10)
+                self.stdout, 'step_count')), 10),
+            sn.assert_reference(energy, -4404.23228954148, -2.e-8, 1e-8)
```
vkarak: Does it make sense to use the `assert_bounded` with `abs` here? What's the meaning of the percentage of the reference energy?

victorusu: > Does it make sense to use the assert_bounded with abs here?
I think so too.

The code is compiled in double precision, so we should at least the double precision digits. I agree with @vkarak , I do not think we should have percentages for reference energies unless the code is compiled with a given precision and it prints more than it can handle. Which I think, it is not the case.

cscs-checks/apps/cp2k/cp2k_check.py
victorusu: Nice catch!

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -13,11 +13,15 @@ def __init__(self, check_name, check_descr, **kwargs):
...
+        energy = sn.extractsingle(r'\s+ENERGY\| Total FORCE_EVAL \( QS \) '
+                                  r'energy \(a\.u\.\):\s+(?P<energy>\S+)',
+                                  self.stdout, 'energy', float, item=-1)
         self.sanity_patterns = sn.all([
             sn.assert_found(r'PROGRAM STOPPED IN', self.stdout),
             sn.assert_eq(sn.count(sn.extractall(
                 r'(?P<step_count>STEP NUM)',
-                self.stdout, 'step_count')), 10)
+                self.stdout, 'step_count')), 10),
+            sn.assert_bounded(energy, -4404.2324, -4404.2322)
```
vkarak: `assert_bounded` is better than `assert_reference` but not as descriptive as we want. Usually, when you check a numerical value, you need the reference value and a sufficiently small value (epsilon) for the acceptable deviation from the reference. The `assert_bounded` doesn't show those two. Here is what I propose:

```python
energy  = sn.extractsingle(...)
energy_diff = sn.abs(energy - REFERENCE)  # Put the reference energy value here
    ...
    sn.assert_lt(energy_diff, 1e-6)    # put a reasonable epsilon value is for this case.
])
```

rsarm:
retest this please

PR #204, opened 2018-03-15
Custom configuration file support

### LINKED ISSUE #65: Support for loading custom settings file
We need a more flexible way to support site configurations and the framework's settings.

This will facilitate also front-end unit tests, since we can provide the settings files as unit test fixtures. It will also allow us to unify the public and private repositories.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/233

kraushm:
The user can pass a custom machine configuration file to ReFrame using the -C argument. NB: some unit tests still fail.

Fixes #65


unittests/test_cli.py
```diff
@@ -71,15 +71,16 @@ def setUp(self):
         self.prefix = tempfile.mkdtemp(dir='unittests')
         self.system = 'generic:login'
         self.checkpath = ['unittests/resources/hellocheck.py']
-        self.environs  = ['builtin-gcc']
```
vkarak: This formatting is also allowed by our coding style. I'm not against the strict PEP8 on that, but I find the alignment nice if you only need to pad with at maximum so many spaces as the length of the following operator (here one).

unittests/test_cli.py
```diff
@@ -71,15 +71,16 @@ def setUp(self):
...
+        self.environs = ['builtin-gcc']
+        self.local = True
         self.action = 'run'
         self.more_options = []
         self.mode = None
 
         # Monkey patch logging configuration
         self.logfile = os.path.join(self.prefix, 'reframe.log')
-        settings._logging_config = {
+        self.settings = config.load_from_file("reframe/settings.py")
```
vkarak: Please use single quotes here.

vkarak: With your feature there is no more need for the monkey-patching of settings. We could provide a custom settings file just for the unit tests (I also want this in my refactoring of the runtime resources and the new style checks). Assigning `settings` to `self` has no effect, that's why the logging unit tests fail. Before, we had to do the monkey-patching, because we had no way of providing alternative settings for the unit tests.

unittests/test_cli.py
```diff
@@ -132,6 +133,11 @@ def _perflog_exists(self, check_name, partitions):
 
     def assert_log_file_is_saved(self):
         outputdir = os.path.join(self.prefix, 'output')
+        print("\n")
+        print(outputdir)
+        print("\n")
+        print(os.path.join(outputdir, os.path.basename(self.logfile)))
+        print("\n")
```
vkarak: Just a reminded to remove these prints.

unittests/test_cli.py
```diff
@@ -284,7 +290,7 @@ def test_same_output_stage_dir(self):
 
     def test_execution_modes(self):
         self.checkpath = []
-        self.environs  = []
+        self.environs = []
```
vkarak: Again, no need to adjust this. I also find it uglier.

kraushm:
@jenkins-cscs retry

vkarak:
@jenkins-cscs retry leone

vkarak:
I will merge this and create a new issue for writing the documentation. I will need it to do the refactoring of the unit tests, because the way we deal with native configurations is a real pain!

PR #207, opened 2018-03-18
Terminate pending task upon abort in serial policy

### LINKED ISSUE #151: The serial execution policy does not cancel the spawned job upon abort

vkarak:
- Also add unit tests to verify this behaviour in both the serial and
  async execution policy.

Fixes #151.


teojgo:
I did a test run watching the squeue and the jobs are successfully removed from the queue once the test is cancelled,

PR #210, opened 2018-03-20
Enforce unique test names

### LINKED ISSUE #50: Enforce unique regression test names
Internal issue: https://madra.cscs.ch/scs/reframe/issues/300

vkarak:
- If two tests have the same name, an error is issued.
- A new command-line option `--ignore-check-conflicts` is added to ignore the conflicted check.
- Unit tests are added.

Fixes #50.


vkarak:
Do you think we should raise an error here instead of a warning?

docs/running.rst
```diff
@@ -184,6 +184,13 @@ The ``-c``\ option accepts also regular files. This is very useful when you are
 
   ./bin/reframe -c /path/to/my/new/test.py -r
 
+.. note::
+   .. versionadded:: 2.12
+
+   The names of the loaded tests must be unique.
+   If you try to load two or more tests with the same name, a warning will be issued and the duplicate will be ignored.
```
vkarak: I should update the documentation here.

PR #225, opened 2018-03-23
Fine tuning of the config file option impl

vkarak:
- Error checking
- Fine tuned generation of the test config file
- Unit tests
- To run the unit tests with the host settings, set the environment
  variable RFM_CONFIG_FILE='/path/to/host/settings.py'. This does not
  yet reproduce entirely the old behaviour.


vkarak:
The unit tests with the host configuration are real pain. This feature seemed quite innocent, but eventually so much was depending on the `settings` module in the unit tests! I will refactor completely this part as soon as this is merged.

PR #227, opened 2018-03-24
Better implementation of the sample program

vkarak:
As soon as this is merged the ReFrame test must be adapted accordingly.


rsarm:
This is better, I adapt then the test to the new C code.

Issue #134, opened 2018-02-22
Parallel execution of unit tests


vkarak:
Originally opened by @teojgo and abandoned due to errors. We were simply missing an option!

```
./test_reframe.py --process-timeout=300 --processes=4
................................................................................SSS...S...................................................SSSSSS.......................................................S...........................................................................S..S................S.S..SS..S.S..SS.........................................................
----------------------------------------------------------------------
Ran 368 tests in 18.155s

OK (SKIP=21)
```

vs single-threaded:

```
....S........................................................SSS...S...................................................................................................SSSSSS..............................S..S.........................................................................................................................S.S..SS..S.S..SS.......................
----------------------------------------------------------------------
Ran 368 tests in 47.144s

OK (SKIP=21)
```

The only thing is what is a reasonable timeout? I would say to put set it to quite a high number, since for the submission unit tests, we might get stuck in the job scheduler for a long time if the target system is too loaded.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/521

teojgo:
Tested it and I get the following error sometimes:
```
FAIL: test_runall (unittests.test_policies.TestSerialExecutionPolicy)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/manitart/reframe_dev_public/unittests/test_policies.py", line 43, in test_runall
    self.assertEqual(5, stats.num_failures())
AssertionError: 5 != 6
```
Maybe it is related to #78

vkarak:
It could be. I've tested only on the branch with the fix of #78.

vkarak:
With minor changes in the unit tests we can migrate to `pytest`, which is widely supported and with plenty of plugins.

vkarak:
This almost works:

```
$ ./test_reframe.py --workers=auto --forked
```

I'm getting two failures only:

```
____________________________________________________________________________________________ test_sanity_of_checks _____________________________________________________________________________________________
run_reframe = <function run_reframe.<locals>._run_reframe at 0x1061f8ae8>, tmp_path = PosixPath('/private/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/pytest-of-karakasv/pytest-75/test_sanity_of_checks0')
logfile = PosixPath('.rfm_unittest.log')

    def test_sanity_of_checks(run_reframe, tmp_path, logfile):
        # This test will effectively load all the tests in the checks path and
        # will force a syntactic and runtime check at least for the constructor
        # of the checks
[...]

PR #231, opened 2018-03-28
Make unittests compatible with pytest

### LINKED ISSUE #111: Consider alternative unit testing frameworks
Internal issue: https://madra.cscs.ch/scs/reframe/issues/521

teojgo:
* Move `unittest.TestCase` inheritance to the actual test
  classes.

Closes #111.
Related to #134


.travis.yml
```diff
@@ -7,4 +7,4 @@ install:
     - pip install -r requirements.txt
 
 script:
-    ./test_reframe.py
```
vkarak: We will need to be able to run the the tests with `./test_reframe.py` passing any options to pytest. We can achieve this by calling `pytest.main()` instead of `nose.main` inside the `test_reframe.py`.

teojgo: Yes, I know but first we must add also pytest to corresponding Python bare module.

teojgo:
@vkarak I have tried with all the possible settings and I always get 387 tests with either pytest or nose.

.travis.yml
vkarak: In any case we should add it to the `requirements.txt`, so that anybody can run the unit tests.

vkarak:
@teojgo I've checked again the current status of this branch and the master and both run the same number of unit tests. So, it's fine from this point of view.

requirements.txt
```diff
@@ -1 +1,3 @@
 nose==1.3.7
```
vkarak: We should remove `nose` now.

teojgo: Ok, I have also a module for python bare ready.

vkarak: As soon as this is fixed, this PR is fine for me. We could though postpone its merging until the `Python-bare` module is fixed. The only thing is that we should perhaps add some additional pytest plugins in `Python-bare`, although we don't use them now, so as not to have to update this module every time.

vkarak:
Moved to the next sprint; it needs a fix from the big PR for new style checks and the unit tests refactoring.

reframe/core/exceptions.py
```diff
@@ -40,11 +40,11 @@ class ReframeSyntaxError(ReframeError):
     """Raised when the syntax of regression tests is not correct."""
 
 
-class TestLoadError(ReframeError):
+class LoadTestError(ReframeError):
```
vkarak: I suggest naming this `RegressionTestLoadError`.

reframe/frontend/loader.py
```diff
@@ -107,8 +107,9 @@ def load_from_module(self, module):
         old_syntax = hasattr(module, '_get_checks')
         new_syntax = hasattr(module, '_rfm_gettests')
         if old_syntax and new_syntax:
-            raise TestLoadError('%s: mixing old and new regression test '
-                                'syntax is not allowed' % module.__file__)
+            raise RegressionTestLoadError('%s: mixing old and new regression '
+                                          'test syntax is not allowed'
+                                          % module.__file__)
```
vkarak: We split lines *after* the operators, but never mind.

PR #232, opened 2018-03-29
Implement a versioning abstraction and the corresponding unit tests

### LINKED ISSUE #228: Implement a versioning abstraction
As discussed with  @vkarak and @victorusu, we need to implement an abstraction layer for handling ReFrame version comparisons inside the framework. 

This is need by #166.

rsarm:
Closes #228

### DIFF
--- a/reframe/utility/versioning.py
+++ b/reframe/utility/versioning.py
@@ -6,8 +6,7 @@ import sys
 class Version:
     def __init__(self, version):
         if version is None:
-            raise ValueError('invalid type for version: %s'
-                             % version) from None
+            raise ValueError('version string may not be None')
 
         base_part, *dev_part = version.split('-dev')
 
--- a/reframe/utility/versioning.py
+++ b/reframe/utility/versioning.py
@@ -1,10 +1,16 @@
 import functools
+import sys
 
 
 @functools.total_ordering
 class Version:
     def __init__(self, version):
+        if version is None:
+            raise ValueError('invalid type for version: %s'
+                             % version) from None
+
         base_part, *dev_part = version.split('-dev')
+
         try:
             major, minor, *patch_part = base_part.split('.')
         except ValueError:
@@ -34,16 +40,18 @@ class Version:
         if not isinstance(other, type(self)):
             return NotImplemented
 
-        if self._value() != other._value():
+        if self._value() > other._value():
             return self._value() > other._value()
-        elif self._dev_number is None and other._dev_number is None:
+
+        if self._value() < other._value():
             return self._value() > other._value()
-        elif self._dev_number is not None and other._dev_number is None:
-            return False
-        elif self._dev_number is None and other._dev_number is not None:
-            return True
-        elif self._dev_number is not None and other._dev_number is not None:
-            return self._dev_number > other._dev_number
+
+        self_dev_number = (self._dev_number if self._dev_number is not None
+                           else sys.maxsize)
+        other_dev_number = (other._dev_number if other._dev_number is not None
+                            else sys.maxsize)
+
+        return self_dev_number > other_dev_number
 
     def __repr__(self):
         return "Version('%s')" % self
--- a/reframe/utility/versioning.py
+++ b/reframe/utility/versioning.py
@@ -1,8 +1,6 @@
 import abc
 import re
 
-from reframe.core.exceptions import ConfigError
-
 
 class CustomBooleanOperations(abc.ABC):
     """Abstract class to customize Boolean operations."""
@@ -27,7 +25,7 @@ class CustomBooleanOperations(abc.ABC):
 
     @abc.abstractmethod
     def _compare(self, other, operation):
-        """Defines the method for comparing versions.
+        """Define the method for comparing versions.
 
         This method returns a bool.
         """
--- a/reframe/utility/versioning.py
+++ b/reframe/utility/versioning.py
@@ -1,15 +1,34 @@
-import abc
-import re
+import functools
+import math
 
 
-class CustomBooleanOperations(abc.ABC):
-    """Abstract class to customize Boolean operations."""
+@functools.total_ordering
+class Version:
+    def __init__(self, version):
+        try:
+            str_base, dev = version.split('-dev')
+            base = str_base.split('.')
+            try:
+                self.dev = int(dev)
+            except ValueError:
+                raise ValueError('invalid version: %s' % version)
 
-    def __lt__(self, other):
-        return self._compare(other, lambda s, o: s < o)
+        except:
+            base = version.split('.')
+            self.dev = None
 
-    def __le__(self, other):
-        return self._compare(other, lambda s, o: s <= o)
+        if len(base) < 2 or len(base) > 3:
+            raise ValueError('invalid version: %s' % version)
+        elif len(base) == 2:
+            base.append('0')
+
+        for i, j in enumerate(base):
+            try:
+                base[i] = int(j)
+            except ValueError:
+                raise ValueError('invalid version: %s' % version)
+
+        self.base = tuple(base)
 
     def __eq__(self, other):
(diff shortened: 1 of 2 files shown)


reframe/utility/versioning.py
```diff
@@ -0,0 +1,104 @@
...
+    def _compare(self, other, operation):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+
+        return operation(self.dev_level, other.dev_level)
+
+
+class Version(CustomBooleanOperations):
+    def __init__(self, version):
+        self._version = version.strip()
```
vkarak: You should not install internally the version string. Instead, you should store a tuple `(M, m, p)`, where `M` is the major version number, `m` the minor version number and `p` the patch level. The optional development part should be stored separately as a number. This should only be considered if two version tuples compare equal. Following this approach, you won't need any of the additional classes you have.

reframe/utility/versioning.py
```diff
@@ -0,0 +1,104 @@
...
+    def tag(self):
+        try:
+            dev_level = self._version.split('-dev')[1]
+            tag = DevelopmentTag(int(dev_level))
+        except:
+            tag = ReleaseTag()
+
+        return tag
+
+    def _is_version_valid(self):
```
vkarak: This could be integrated directly into the constructor.

reframe/utility/versioning.py
```diff
@@ -0,0 +1,104 @@
...
+    def __repr__(self):
+        return "release"
+
+    def _compare(self, other, operation):
+        if isinstance(other, self.__class__):
+            return operation(0, 0)
+        elif isinstance(other, DevelopmentTag):
+            return operation(1, 0)
+        else:
+            return NotImplemented
```
vkarak: I don't see the reason for all this infrastructure here. I don't think you need it. The `Version` class itself should implement the rich comparison operators. Also there is no need to define all of them. Python defines `__ne__`  by call inverting the result of the `__eq__` function. For the rest of the comparison operators, you should better use [functools.total_ordering](https://docs.python.org/3.6/library/functools.html#functools.total_ordering).

reframe/utility/versioning.py
```diff
@@ -0,0 +1,104 @@
...
+        try:
+            dev_level = self._version.split('-dev')[1]
+            tag = DevelopmentTag(int(dev_level))
+        except:
+            tag = ReleaseTag()
+
+        return tag
+
+    def _is_version_valid(self):
+        pattern = re.compile(r'^\d+\.\d+(\.\d){0,1}\d*(-dev\d+){0,1}$')
```
vkarak: Again, I wouldn't use a regex. Given the implementation I described in my other comment, which internally uses a tuple, I would do the following:

1. Split on `-dev` to get the development part if it exists.
2. Split on `.` to get the version parts.
3. If the individual parts are not integral numbers (you can checks if the derive from `numbers.Integral`), raise a `ValueError`.

reframe/utility/versioning.py
```diff
@@ -0,0 +1,104 @@
...
+            raise ValueError('invalid version format')
+
+    def _compare(self, other, operation):
+        if not isinstance(other, self.__class__):
+            return NotImplemented
+
+        return operation((self.base, self.tag), (other.base, other.tag))
+
+    def __repr__(self):
+        return self._version
```
vkarak: The convention for `repr` is to return a string that would create this object. So, it should return `Version(...)`. You should implement `__str__` as well to return just the version number.

reframe/utility/versioning.py
```diff
@@ -0,0 +1,63 @@
...
+        self.base = tuple(base)
+
+    def __eq__(self, other):
+        return self._compare(other, lambda s, o: s == o)
+
+    def __ge__(self, other):
+        return self._compare(other, lambda s, o: s >= o)
+
+    def _compare(self, other, operation):
+        if not isinstance(other, self.__class__):
```
vkarak: The type checking should be as follows:

```python
if not isinstance(other, type(self)):
    return NotImplemented
```

reframe/utility/versioning.py
```diff
@@ -0,0 +1,63 @@
...
+            other_dev = math.inf
+        else:
+            other_dev = other.dev
+
+        return operation((self.base, self_dev), (other.base, other_dev))
+
+    def __repr__(self):
+        return "Version('%s')" % self.__str__()
+
+    def __str__(self):
```
vkarak: This should be adapted to use the major/minor/patchlevel internal variables.

reframe/utility/versioning.py
```diff
@@ -0,0 +1,63 @@
...
+
+        if other.dev is None:
+            other_dev = math.inf
+        else:
+            other_dev = other.dev
+
+        return operation((self.base, self_dev), (other.base, other_dev))
+
+    def __repr__(self):
+        return "Version('%s')" % self.__str__()
```
vkarak: Do not call directly the internal methods. Use instead `str(self)`. In this case particularly, you don't even need to call `str()`, because you use the `%s` modifier.

reframe/utility/versioning.py
```diff
@@ -0,0 +1,63 @@
...
+                base[i] = int(j)
+            except ValueError:
+                raise ValueError('invalid version: %s' % version)
+
+        self.base = tuple(base)
+
+    def __eq__(self, other):
+        return self._compare(other, lambda s, o: s == o)
+
+    def __ge__(self, other):
```
vkarak: You should implement either the `__gt__` or the `__lt__`, since you already have an equality operator. Using the version value trick, the comparison should be simple. First, compare the versions and if they're not equal return the result of the comparison. If they are, then you should compare the dev numbers and return the result.

unittests/test_versioning.py
```diff
@@ -0,0 +1,27 @@
...
+    def test_version_format(self):
+        Version('1.2')
+        Version('1.2.3')
+        Version('1.2-dev0')
+        Version('1.2-dev5')
+        Version('1.2.3-dev2')
+        self.assertRaises(ValueError, Version, '1')
+        self.assertRaises(ValueError, Version, '1.2a')
+        self.assertRaises(ValueError, Version, 'a.b.c')
+        self.assertRaises(ValueError, Version, '1.2.3-dev')
```
vkarak: You should also check if sb calls you by passing `None` or an empty string.

rsarm: The unit test with `None` would through `AttributeError` but I don't know if `Version` should check if the version number that is passed is not an string. Because when once we have the `Field` implemented for the versioning, then Version will always get a string, right?

vkarak: `Version` should check for `None`s and disallow them raising a `ValueError`.

reframe/utility/versioning.py
```diff
@@ -0,0 +1,56 @@
...
+        if self._value() != other._value():
+            return self._value() > other._value()
+        elif self._dev_number is None and other._dev_number is None:
+            return self._value() > other._value()
+        elif self._dev_number is not None and other._dev_number is None:
+            return False
+        elif self._dev_number is None and other._dev_number is not None:
+            return True
+        elif self._dev_number is not None and other._dev_number is not None:
+            return self._dev_number > other._dev_number
```
vkarak: To avoid all these elifs, I would use the trick of converting the dev number to infinity, which you have used originally, but without using `math` module:

```python
if self._value() > other._value():
    return True

if self._value() < other._value():
    return False

# Both versions are equal here; compare the dev numbers
self_dev_number  = self._dev_number  or sys.maxsize
other_dev_number = other._dev_number or sys.maxsize
return self_dev_number > other_dev_number
```

rsarm: Here I checked instead if it's not `None` because the short version gives the wrong result when the `self._dev_numbe` is zero.

vkarak:
@jenkins-cscs retry leone

PR #233, opened 2018-04-03
Support wildcards in valid_prog_environs

### LINKED ISSUE #224: Support wildcards in 'valid_prog_environs'
It would be useful to specify programming environments like that:

```python
self.valid_prog_environs = ['PrgEnv-gnu-*', 'foss/*']
```

teojgo:
* Add support for wildcards in `valid_prog_environs` of a `RegressionTest`.
* Create the corresponding unit tests.
* Update doc for the `valid_prog_environs` field.

Closes #224

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -2,6 +2,7 @@
 # Basic functionality for regression tests
 #
 
+import fnmatch
 import os
 import shutil
 
@@ -54,10 +55,14 @@ class RegressionTest:
     #: :type: Alphanumeric string.
     name = fields.AlphanumericField('name')
 
-    #: List of programming environmets supported by this test.
+    #: List of programming environments supported by this test.
     #:
     #: :type: :class:`list[str]`
     #: :default: ``[]``
+    #:
+    #: .. note::
+    #:     .. versionchanged:: 2.12
+    #:        Programming environments can now be specified using wildcards.
     valid_prog_environs = fields.TypedListField('valid_prog_environs', str)
 
     #: List of systems supported by this test.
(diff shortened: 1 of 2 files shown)


reframe/core/pipeline.py
```diff
@@ -676,10 +681,10 @@ def supports_system(self, partition_name):
         return partition_name in self.valid_systems
 
     def supports_environ(self, env_name):
-        if '*' in self.valid_prog_environs:
-            return True
-
-        return env_name in self.valid_prog_environs
+        for env in self.valid_prog_environs:
+            if fnmatch.fnmatch(env_name, env):
+                return True
```
vkarak: Please leave a blank newline here.

unittests/test_pipeline.py
```diff
@@ -242,6 +242,27 @@ def test_supports_system(self):
...
+
+        test.valid_prog_environs = ['PrgEnv-foo-*']
+        self.assertTrue(test.supports_environ('PrgEnv-foo-version1'))
+        self.assertTrue(test.supports_environ('PrgEnv-foo-version2'))
+        self.assertFalse(test.supports_environ('PrgEnv-boo-version1'))
+
+        test.valid_prog_environs = ['PrgEnv-*']
+        self.assertTrue(test.supports_environ('PrgEnv-foo1'))
+        self.assertTrue(test.supports_environ('PrgEnv-foo2'))
+        self.assertFalse(test.supports_environ('Prgenv-foo'))
```
vkarak: I don't see the reason for this set of checks. They are equivalent to the above block. You can simply add a check for case sensitivity.

reframe/core/pipeline.py
```diff
@@ -676,10 +681,10 @@ def supports_system(self, partition_name):
         return partition_name in self.valid_systems
 
     def supports_environ(self, env_name):
-        if '*' in self.valid_prog_environs:
-            return True
-
-        return env_name in self.valid_prog_environs
+        for env in self.valid_prog_environs:
+            if fnmatch.fnmatch(env_name, env):
```
vkarak: I like this and it's simple. Would it make sense to support wildcards also in the `supports_system()`?

teojgo: It is possible to do so if you want to support also wildcards for systems.

PR #234, opened 2018-04-04
Fix infinite recursion bug in 'copytree_virtual'

### LINKED ISSUE #221: Infinite recursion during copying to stage directory if 'sourcesdir' is a relative path
Here is the scenario, which is a typical scenario for usage of the framework as development CI tool:

#### Directory structure
```
ci-scripts/
    hellocheck.py
src/
    hello.c
```

#### hello.c

```C
#include <stdio.h>

int main()
{
    printf("Hello, World!\n");
    return 0;
}
```

#### hellocheck.py

```python
import os

from reframe.core.pipeline import RegressionTest


class HelloTest(RegressionTest):
    def __init__(self, **kwargs):
        super().__init__('hellocheck', os.path.dirname(__file__), **kwargs)
        self.descr = 'C Hello World test'

        self.valid_systems = ['*']
        self.sourcesdir = os.path.join(self.prefix, '..')
        self.sourcepath = 'hello.c'
        self.tags = {'foo', 'bar'}
        self.sanity_patterns = sn.assert_found(r'Hello, World\!', self.stdout)
        self.maintainers = ['VK']


def _get_checks(**kwargs):
    return [HelloTest(**kwargs)]
```

Run from the top-level dir with

```
reframe -c ci-scripts/hellocheck.py -r
```

Output (full path is not shown):

```
Reframe version: 2.11
Launched by user: karakasv
Launched on host: dom101
Reframe paths
=============
    Check prefix      :
    Check search path : 'ci-scripts/hellocheck.py'
    Stage dir prefix  : /users/karakasv/Devel/reframe/mytests/sourcesdir-relpath/stage/
    Output dir prefix : /users/karakasv/Devel/reframe/mytests/sourcesdir-relpath/output/
    Logging dir       : /users/karakasv/Devel/reframe/mytests/sourcesdir-relpath/logs
[==========] Started on Thu Mar 22 14:06:00 2018

[----------] started processing hellocheck (C Hello World test)
[ RUN      ] hellocheck on dom:login using PrgEnv-cray
[     FAIL ] hellocheck on dom:login using PrgEnv-cray
[ RUN      ] hellocheck on dom:login using PrgEnv-gnu
^C[     FAIL ] hellocheck on dom:login using PrgEnv-gnu
[  FAILED  ] Ran 2 test case(s) from 1 check(s) (2 failure(s))
[==========] Finished on Thu Mar 22 14:06:11 2018
SUMMARY OF FAILURES
FAILURE INFO for hellocheck
  * System partition: dom:login
  * Environment: PrgEnv-cray
  * Stage directory: /users/karakasv/Devel/reframe/mytests/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray
  * Maintainers: ['VK']
  * Failing phase: compile
  * Reason: caught framework exception: virtual copying of files failed: [('/users/karakasv/Devel/reframe/mytests/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-cray/sourcesdir-relpath'")]
FAILURE INFO for hellocheck
  * System partition: dom:login
  * Environment: PrgEnv-gnu
  * Stage directory: /users/karakasv/Devel/reframe/mytests/sourcesdir-relpath/stage/login/hellocheck/PrgEnv-gnu
  * Maintainers: ['VK']
  * Failing phase: compile
  * Reason: cancelled by user
```

If you run with `--prefix` pointing to a different directory, everything runs as expected.

teojgo:
* Check that the source path is not a parent path of
  the destination one.
* Create the corresponding unittest.

Fixes #221.

### DIFF
--- a/reframe/utility/os_ext.py
+++ b/reframe/utility/os_ext.py
@@ -81,6 +81,10 @@ def copytree(src, dst, symlinks=False, ignore=None, copy_function=shutil.copy2,
 
     In this case it will first remove it and then call the standard
     shutil.copytree()."""
+    if src == os.path.commonpath([src, dst]):
+        raise ValueError("cannot copy recursively the parent directory "
+                         "`%s' into one of its descendants `%s'" % (src, dst))
+
     if os.path.exists(dst):
         shutil.rmtree(dst)
 
@@ -104,10 +108,6 @@ def copytree_virtual(src, dst, file_links=[],
     # Work with absolute paths
     src = os.path.abspath(src)
     dst = os.path.abspath(dst)
-
-    if src == os.path.commonpath([src, dst]):
-        raise ValueError("copytree_virtual() failed: "
-                         "`%s' is a parent directory of `%s'" % (src, dst))
 
     # 1. Check that the link targets are valid
     # 2. Convert link targets to absolute paths
--- a/reframe/utility/os_ext.py
+++ b/reframe/utility/os_ext.py
@@ -105,6 +105,10 @@ def copytree_virtual(src, dst, file_links=[],
     src = os.path.abspath(src)
     dst = os.path.abspath(dst)
 
+    if src == os.path.commonpath([src, dst]):
+        raise ValueError("copytree_virtual() failed: "
+                         "`%s' is a parent directory of `%s'" % (src, dst))
+
     # 1. Check that the link targets are valid
     # 2. Convert link targets to absolute paths
     # 3. Store them in a set for quick look up inside the ignore function
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -62,6 +62,15 @@ class TestOSTools(unittest.TestCase):
         shutil.rmtree(dir_src)
         shutil.rmtree(dir_dst)
 
+    def test_copytree_src_parent_of_dst(self):
+        dst_path = tempfile.mkdtemp()
+        src_path = os.path.abspath(os.path.join(dst_path, '..'))
+
+        self.assertRaises(ValueError, os_ext.copytree,
+                          src_path, dst_path)
+
+        shutil.rmtree(dst_path)
+
     def test_inpath(self):
         self.assertTrue(os_ext.inpath('/foo/bin', '/bin:/foo/bin:/usr/bin'))
         self.assertFalse(os_ext.inpath('/foo/bin', '/bin:/usr/local/bin'))
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -234,12 +234,6 @@ class TestCopyTree(unittest.TestCase):
         self.assertRaises(OSError, os_ext.copytree_virtual,
                           self.prefix, self.target, file_links)
 
-    def test_virtual_copy_src_parent_of_dst(self):
-        src_path = os.path.join(self.prefix, 'bar', '..')
-        dst_path = os.path.join(self.prefix, 'bar')
-        self.assertRaises(ValueError, os_ext.copytree_virtual,
-                          src_path, dst_path)
-
     def tearDown(self):
         shutil.rmtree(self.prefix)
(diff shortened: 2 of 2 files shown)


reframe/utility/os_ext.py
```diff
@@ -105,6 +105,10 @@ def copytree_virtual(src, dst, file_links=[],
     src = os.path.abspath(src)
     dst = os.path.abspath(dst)
 
+    if src == os.path.commonpath([src, dst]):
+        raise ValueError("copytree_virtual() failed: "
+                         "`%s' is a parent directory of `%s'" % (src, dst))
```
vkarak: The problem is not in this function, but in the `copytree()` a bit above. This version of `copytree()` differs from the `shutil.copytree()` in that it deletes the destination first, if it exists. Also, I don't know if your fix here is the correct behaviour, because it's pretty much valid and well-defined to do `cp -r . ..`. So, we should get back to the scenario of the bug and see what would be the intended and/or the expected behaviour.

teojgo: Yes but is recursively copying `..` to `.`  valid? i.e. `cp -r .. .`?

teojgo: If you create a directory hierarchy`foo/bar`, `cd` into `foo/bar` and try `cp -r .. .` you get the following error:
` cp: cannot copy a directory, '..', into itself, '.'`
If the dir has lots of files, the whole directory structure will be copied before throwing the error. Therefore I believe that we should somehow not allow that to happen by considering it an error before trying to perform the copying of the tree.

vkarak: After some investigation, your solution seems to be correct. However, it should be put inside the `copytree()` function and not the `copytree_virtual()`. The error message should be a bit different as well: ``"cannot copy recursively the parent directory `%s' into one of its descendants `%s'"``.

unittests/test_utility.py
```diff
@@ -225,6 +234,12 @@ def test_virtual_copy_linkself(self):
         self.assertRaises(OSError, os_ext.copytree_virtual,
                           self.prefix, self.target, file_links)
 
+    def test_virtual_copy_src_parent_of_dst(self):
```
vkarak: I don't think this test is needed, since they way you call it is equivalent to the `copytree`.

PR #241, opened 2018-04-09
Fix default 'settings.py' path

### LINKED ISSUE #238: Default settings are not loaded if ReFrame is invoked from a different directory
```
~/Devel/reframe/bin/reframe -c ci-scripts/hellocheck.py -r
/users/karakasv/Devel/reframe/bin/reframe: could not load settings: could not load configuration file `reframe/settings.py': [Errno 2] No such file or directory: 'reframe/settings.py'
```

The problem is that we are always looking for `reframe/setttings.py` in the current working directory, whereas we should look in the "installation" directory of ReFrame. The solution is to do the same thing we do for the default checks lookup.

teojgo:
* Use `__file__` to get the absolute path of the `cli.py` path and define the default `settings.py` path 
  according to it.

Closes #238

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -1,4 +1,5 @@
 import os
+import pathlib
 import socket
 import sys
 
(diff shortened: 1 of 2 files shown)


reframe/frontend/cli.py
```diff
@@ -178,11 +179,16 @@ def main():
     misc_options.add_argument(
         '--system', action='store',
         help='Load SYSTEM configuration explicitly')
+
+    filepath = pathlib.Path(__file__)
```
vkarak: I know you prefer the pathlib, but for consistency inside the framework, I suggest using `os.path`. I suggest writing it also in a single-line. You could perhaps define the default settings file as a global variable `DEFAULT_SETTINGS_FILE = os.path.abspath('../../reframe/settings.py')` inside the `cli.py`.

teojgo:
@jenkins-cscs retry all

Issue #242, opened 2018-04-10
Fix CDO/NCO compatibility checks that rely on empty stderr


teojgo:
The stderr may contain information which do not constitute failure of the check:

* [nco_module_compat](https://github.com/eth-cscs/reframe/blob/0890e215b6a1abbe2413dc7319ebb376e38db4e7/cscs-checks/tools/io/cdo.py#L75)

* [cdo_module_compat](https://github.com/eth-cscs/reframe/blob/0890e215b6a1abbe2413dc7319ebb376e38db4e7/cscs-checks/tools/io/nco.py#L71)

vkarak:
@teojgo Does this fail now also on Daint?

teojgo:
No, just in dom.

PR #260, opened 2018-04-19
Fix nco/cdo compatibility checks

teojgo:


cscs-checks/tools/io/cdo.py
```diff
@@ -78,8 +78,8 @@ def __init__(self, **kwargs):
         self.descr = ('verifies compatibility with the NCO module')
         self.sourcesdir = None
         self.executable = 'echo'
-        self.sanity_patterns = sn.all([
-            sn.assert_not_found(r'.+', self.stderr)])
+        self.sanity_patterns = sn.assert_not_found(r'(ERROR)|(conflict)',
```
omlins: I think you should use a case insensitive search using `(?i)`, as for example `r'(?i)error|conflict'`. 

The `(?i)` needs to stand only at the beginning of the pattern ( [http://www.rexegg.com/regex-disambiguation.html#modifiers](url)):
```
All popular regex flavors apart from JavaScript support inline modifiers, which allow you to tell the engine, in a pattern, to change how to interpret the pattern. For instance, (?i) turns on case-insensitivity. Except in Python, (?-i) turns it off. 

If a modifier appears at the head of the pattern, it modifies the matching mode for the whole pattern—unless it is later turned off. But (except in Python) a modifier can appear in mid-pattern, in which case in only affects the portion of the pattern that follows. 
```
I have just seen that in the `InfoNC4CCheck` you can replace `r'(?i)unsupported|(?i)error'` with `r'(?i)unsupported|error'`.

omlins: As we try to capture here any possible error message, I think it would make sense to increase the list for keywords that indicate an error, I.e. you could maybe add `unsupported` and `failure` and other words alike to the pattern.

vkarak:
@teojgo Can you add in the description the issue that this PR fixes?

vkarak:
@teojgo Does it fix #242?

PR #275, opened 2018-04-30
More unit tests for the scoped dictionaries

### LINKED ISSUE #206: Add more unit tests to the scoped dictionary
We need to add a couple of unit tests more to the scoped dictionary that would test the (correct) behaviour described in #60. More specifically:

1. Adding a scope with a same name as a key:
```python
scoped['s0:s1'] = 1
scoped['s0:s1:k0'] = 2
assert scoped['s0:s1'] == 1
assert scoped['s0:s1:k0'] == 2
del scoped['s0:s1']
assert scoped['s0:s1:k0'] == 2
scoped['s0:s1']   # -> this should raise `KeyError`
```

2. Nested scopes are not deleted even if the parent scope is deleted.

teojgo:
* Add unittest checking that deletion of parent scope
  does not delete child scopes.

Fixes #206


unittests/test_fields.py
```diff
@@ -669,6 +671,22 @@ def test_delitem(self):
...
+        scoped_dict = fields.ScopedDict({
+            's0': {'s1': 1},
+            's0:s1': {'k0': 2}
+        })
+
+        self.assertEqual(1, scoped_dict['s0:s1'])
+        self.assertEqual(2, scoped_dict['s0:s1:k0'])
+
+        # delete the parent scope
+        del scoped_dict['s0']
```
vkarak: Check my comment in the issue. The desired behaviour is the one described there, so there is a bug in the `__delitem__`. Besides, I think there is already a unit tests for deleting a whole scope.

unittests/test_fields.py
```diff
@@ -669,6 +671,22 @@ def test_delitem(self):
             KeyError, exec, "del scoped_dict['a:k4']", globals(), locals()
         )
 
+    def test_delparent_scope(self):
```
vkarak: This name does not reflect the purpose of this test. The purpose of the test is "Adding a scope with a same name as a key"

teojgo: By nested scope do you mean the `s0:s1` part of `s0:s1:k0`?

vkarak: No, this is not nested! In `s0:s1`, `s1` is a key, not a scope. In the second one `s0` and `s1` are scopes. This test checks what happens if a key and scope have the same name. This test is valid (with the adjustments I proposed) and you should just change its name. The nested scopes are different a different thing and should be in a different test function. Here is an example:

```python
s = ScopedDict()
s['s0:k0'] = 1
s['s0:s1:k0'] = 2
s['*:k0'] = 3
del s['s0']
assert s['s0:k0'] == 3
assert s['s0:s1:k0'] == 2
```

vkarak:
Moving to the next sprint.

reframe/utility/__init__.py
```diff
@@ -1,5 +1,136 @@
...
+            self.data[scope] = {}
+
+        self.data[scope][lookup_key] = value
+
+    def __delitem__(self, key):
+        """Deletes either a key or a scope if key refers to a scope.
+
+        If key refers to scope, the whole scope entry will be deleted. If not,
+        the exact key requested will be deleted. No key resolution will be
+        performed."""
```
vkarak: We should expand the description to cover the case when a key and a scope have the same name. I suggest the following:

```
If the key refers to both a scope and a key, the key will be deleted.
```

unittests/test_utility.py
```diff
@@ -320,3 +321,287 @@ def test_exception_propagation(self):
...
+        self.assertRaises(
+            KeyError, exec, "scoped_dict['a:k3']", globals(), locals()
+        )
+
+        # try to delete a non-existent key
+        self.assertRaises(
+            KeyError, exec, "del scoped_dict['a:k4']", globals(), locals()
+        )
+
+    def test_delkey_in_global_scope(self):
```
vkarak: This test can be merged with the `test_delitem`.

unittests/test_utility.py
```diff
@@ -320,3 +321,287 @@ def test_exception_propagation(self):
...
+        })
+        self.assertEqual(0, scoped_dict['k0'])
+        self.assertEqual(1, scoped_dict['k1'])
+        del scoped_dict['k0']
+        self.assertEqual(1, scoped_dict['k1'])
+        self.assertRaises(
+            KeyError, exec, "scoped_dict['k0']", globals(), locals()
+        )
+
+    def test_addscope_same_as_key(self):
```
vkarak: Rename to `test_scope_key_name_pseudoconflict`.

unittests/test_utility.py
```diff
@@ -320,3 +321,287 @@ def test_exception_propagation(self):
...
+        scoped_dict = reframe.utility.ScopedDict({
+            's0': {'s1': 1},
+            's0:s1': {'k0': 2}
+        })
+
+        self.assertEqual(1, scoped_dict['s0:s1'])
+        self.assertEqual(2, scoped_dict['s0:s1:k0'])
+
+        # delete the parent scope
```
vkarak: This comment is inaccurate; please remove.

unittests/test_utility.py
```diff
@@ -320,3 +321,287 @@ def test_exception_propagation(self):
...
+        scoped_dict_alt = reframe.utility.ScopedDict({'a': {'k1': 3, 'k2': 5}})
+        scoped_dict_alt.update({
+            'a': {'k1': 1, 'k2': 2},
+            'a:b': {'k1': 3, 'k3': 4},
+            'a:b:c': {'k2': 5, 'k3': 6},
+            '*': {'k1': 7, 'k3': 9, 'k4': 10}
+        })
+        self.assertEqual(scoped_dict, scoped_dict_alt)
+
+    def test_delete_parent_keep_nested(self):
```
vkarak: Rename to `test_delete_parent_scope`.

unittests/test_utility.py
```diff
@@ -320,3 +321,287 @@ def test_exception_propagation(self):
...
+            'a:b:c': {'k2': 5, 'k3': 6},
+            '*': {'k1': 7, 'k3': 9, 'k4': 10}
+        })
+
+        # delete key
+        del scoped_dict['a:k1']
+        self.assertEqual(7, scoped_dict['a:k1'])
+
+        # delete a whole scope
+        del scoped_dict['*']
```
vkarak: I think this is equivalent (or almost) to the `test_delete_parent_keep_nested` test. Perhaps, we should merge these two.

teojgo:
@jenkins-cscs retry all

PR #277, opened 2018-05-02
PBS scheduler back-end support

rafa-esco:
Adding PBS backend to ReFrame. This implementation should be mostly general. 

It uses the fairly common qsub resource option `"-lselect=<N/n>:ncpus=<n>:mpiprocs=<n>[:extra options]"`... e.g. to run a regression test with a total of 1024 of cores using 16 cores per node, we will generate the following line in the pbs script:

`#PBS -lselect=64:ncpus=16:mpiprocs=16`

it will optionally add any extra resource option you add in the settings.py file via the "access" array of each partition, e.g.

`'access':  [ 'cpu_type=ivy_bridge', 'mem=120GB' ],`

will produce:

`#PBS -lselect=64:ncpus=16:mpiprocs=16:cpu_type=ivy_bridge:mem=120GB`

The PBS queue (analog to SLURM partitions) can be passed to the job scheduler using the` --partition=<queue>` option of reframe, e.g. running

`reframe -n openfoam_motorbike -r --partition=htc`

will add a this line to the batch script:

`#PBS -q htc`  

The time limit for jobs is taken into account by adding a 

`#PBS -l walltime=0:10:0`

### DIFF
--- a/reframe/core/schedulers/pbs.py
+++ b/reframe/core/schedulers/pbs.py
@@ -1,0 +1,101 @@
+import itertools
+import os
+import re
+import time
+from datetime import datetime
+
+import reframe.core.schedulers as sched
+import reframe.utility.os_ext as os_ext
+from reframe.core.exceptions import (SpawnedProcessError,
+                                     JobBlockedError, JobError)
+from reframe.core.logging import getlogger
+from reframe.core.schedulers.registry import register_scheduler
+from reframe.settings import settings
+
+
+@register_scheduler('pbs')
+class PbsJob(sched.Job):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._prefix  = '#PBS'
+        self._is_cancelling = False
+        # fix for regression tests with compile
+        if os.path.dirname(self.command) is '.':
+            self._command = os.path.join(
+                self.workdir, os.path.basename(self.command))
+
+    def _emit_job_option(self, var, option, builder):
+        if var is not None:
+            if isinstance(var, tuple):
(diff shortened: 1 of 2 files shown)


vkarak:
@jenkins-cscs retry all

reframe/core/schedulers/pbs.py
```diff
@@ -0,0 +1,101 @@
...
+        super().cancel()
+        getlogger().debug('cancelling job (id=%s)' % self._jobid)
+        self._is_cancelling = True
+        jobid, server = self._fulljobid.split(".")
+        if server == 'pbspro':
+            self._run_command('qdel %s' % self._fulljobid,
+                              settings.job_submit_timeout)
+        elif server == 'pbs11':
+            self._run_command('qdel %s@pbs11' % self._fulljobid,
+                              settings.job_submit_timeout)
```
vkarak: @rafa-esco I am adapting your PR in order to prepare it for merging. I am using the Torque wrappers of Slurm to test it and I was wondering, why you need this special treatment here? Couldn't `qdel JOBID` just work?

rafa-esco: Hi!, yep, you are totally right...  to be honest this is quite particular to our installation. We have two pbs servers working side by side: pbspro (the default name) and pbs11. So yes, in a normal installation with only one server, a qdel JOBID should work. So removing the if,elif,else and just go with a 'qdel %s' % self._fulljobid will be general enough for anybody.

vkarak:
@rafa-esco Since I didn't want you to bother more with Git stuff, I've submitted a new PR from my fork with an almost final implementation. So, I will close this one.

PR #280, opened 2018-05-03
Reduce PDF file sizes

vkarak:
We compromise the quality of some figures but the file sizes go down to 1.5MB from 10+MB. Since the files are in the repository, it is crucial to be small enough.


vkarak:
@victorusu Can you have a look at this PR? Do you accept the resulting image quality?

victorusu:
@vkarak, I think the difference in size does justifies the image quality. The image quality is not good, but we can make the compromise. I can understand the slides based on the images. Those that I do not understand are not due to the image itself, but rather due to the lack of the presentation.

vkarak:
@jenkins-cscs retry kesch monch

Issue #247, opened 2018-04-18
Adapt OpenACC+MPI GPU test


vkarak:
MCH will provide an updated version of the test's source code:

https://github.com/eth-cscs/reframe/blob/master/cscs-checks/prgenv/src/gpu_direct_acc.f90

Also the filename is not accurate.

lxavier:
see https://github.com/eth-cscs/reframe/pull/281

vkarak:
Hi @lxavier, I'd suggest putting a reference in the corresponding pull request's description rather than here (Github tracks the references automatically). Also, if your PR fixes this issue, simply put in the description the following text `Fixes #247`, so that Github will automatically close the issue as soon as we merge it.

vkarak:
Quoting from @lxavier about running the test:

> nothing special I would start to run it on one node (e.g. 8 GPUs on Kesch) just did srun -n 8
The important thing is to load the module that we use. Since our new module generator is not yet ready you could use the module file on /project /project/c14/install/kesch/cosmo5/modules_fortran_gpu_double.env

PR #281, opened 2018-05-04
clean up and extend mpi+openacc test to work with n tasks

lxavier:
I only changed the the src file, the python script may need to be adapted since I added a ifdef, for Crar the -eZ option should be added.

I have only tested with Cray on Kesch

Addresses #247.


vkarak:
@jenkins-cscs retry all

lxavier:
yes unless the python compiler flags need to be adapted because of the ifdef (but I don't know where to edit)

vkarak:
@lxavier This question was automatic from Jenkins. I've triggered the build also on our machines. My question though is now, do we need to adapt the corresponding ReFrame test as well?

lxavier:
@vkarak yes please, the test code is ready from my side

lxavier:
How should we proceed with the new codes that we want to add, should the responsible person add then in cscs-checks/prgenv/src in branches, or send you the code sample ? Some of then will contain multiple files (e.g. openacc+cuda).

vkarak:
@lxavier I've replied to you by email.

vkarak:
@lxavier I am about to merge this and then we will start working on updating the corresponding ReFrame test. What we are missing is how you run it?

lxavier:
nothing special I would start to run it on one node (e.g. 8 GPUs on Kesch) just did srun -n 8
The important thing is to load the module that we use. Since our new module generator is not yet ready you could use the module file on /project  /project/c14/install/kesch/cosmo5/modules_fortran_gpu_double.env

vkarak:
Thanks @lxavier

lxavier:
true, I just wanted to implement some easy test where I could verify if ok or not on any number of proc. 
I suggest to start with that for now, but if you want to change it fine for me, you just need to exchange message that you can verify.

PR #287, opened 2018-05-23
Support for version compatibility check

### LINKED ISSUE #166: Provide a way to tests to specify the ReFrame version that they require
Since we are actively developing the framework, adding new features, deprecating and removing older stuff, it would be good to have a way in a regression test to specify the version(s) of ReFrame that can run it. During loading, ReFrame will simply ignore incompatible tests.

@eth-cscs/reframe-devteam What do you think?

rsarm:
Closes #166


reframe/utility/versioning.py
```diff
@@ -61,3 +63,60 @@ def __str__(self):
...
+        except ValueError:
+            raise ValueError('invalid interval: "%s", '
+                             'expecting "min_version, max_version"'
+                             % self.condition.strip()) from None
+
+        return ((Version(version_ref) > min_version) and
+                (Version(version_ref) < max_version))
+
+    def _validate_relation(self, version_ref):
+        version, op = self._parse_condition(self.condition)
```
vkarak: This is too late to parse and validate the condition. It should fail early in the constructor.

reframe/utility/versioning.py
```diff
@@ -61,3 +63,60 @@ def __str__(self):
...
+
+        return self._operations[op](Version(version_ref), version)
+
+    def validate(self, version_ref):
+        try:
+            res = self._validate_interval(version_ref)
+        except ValueError:
```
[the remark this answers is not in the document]
vkarak: The try/except here or an if/else is a bit complicated here. We could go for an alternative, more OOP, design using a factory:

```python
class _ValidatorImpl:
    @abc.abstractmethod
    def validate(version):
        pass

class _IntervalValidator(_ValidatorImpl):
    def __init__(self, condition):
        # parse and validate the interval condition here

    def validate(version):
        # validate version against interval

class _RelationalValidator(_ValidatorImpl):
    def __init__(self, condition):
        # parse and validate the relational condition here

    def validate(version):
        # validate version against the condition

class VersionValidator:
    def __new__(cls, condition):
        if '..' in condition:
            return _IntervalValidator(condition)
        else:
            return _RelationalValidator(condition)
```

reframe/utility/versioning.py
```diff
@@ -61,3 +63,60 @@ def __str__(self):
             return base
 
         return base + '-dev%s' % self._dev_number
+
+
+class VersionValidator:
```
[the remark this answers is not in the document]
vkarak: Using my proposed design, the docstring for this class should separated in the two implementations.

rsarm:
@jenkins-cscs retry daint dom

rsarm:
@jenkins-cscs retry all

rsarm:
@vkarak I wasn't very sure about which message to print when skipping a test and where to put it.

reframe/core/decorators.py
```diff
@@ -83,3 +93,24 @@ def _do_register(cls):
         return cls
 
     return _do_register
+
+
+def required_version(*compat_versions):
+    printer = PrettyPrinter()
+    """Class decorator for skipping version-uncompatible tests."""
```
vkarak: This should be documented properly. It's a public function.

reframe/core/decorators.py
```diff
@@ -83,3 +93,24 @@ def _do_register(cls):
...
+    def _skip_tests(cls):
+        mod = inspect.getmodule(cls)
+        if not any(c.validate(VERSION) for c in conditions):
+            printer.status('SKIP',
+                           'skipping uncompatible class %s' % cls.__name__,
+                           just='center')
+            try:
+                mod.__rfm_skip_tests |= {cls}
+            except AttributeError:
+                mod.__rfm_skip_tests = {cls}
```
vkarak: I'd move this try/except outside this `if`:

```python
mod = inspect.getmodule(cls)
if not hasattr(mod, '__rfm_skip_tests'):
    mod.__rfm_skip_tests = set()

if not any(c.validate(reframe.VERSION) for c in conditions):
    mod.__rfm_skip_tests.add(cls)

...
```

reframe/utility/versioning.py
```diff
@@ -61,3 +65,81 @@ def __str__(self):
...
+            "==": lambda x, y: x == y,
+            "!=": lambda x, y: x != y,
+        }
+        cond_match = re.match(r'(\W{0,2})(\S+)', condition)
+        if not cond_match:
+            raise ValueError("invalid condition: '%s'" % condition)
+
+        self._ref_version = Version(cond_match.group(2))
+        op = cond_match.group(1)
+        if op == '':
```
vkarak:
```python
if not op:
```

reframe/utility/versioning.py
```diff
@@ -61,3 +65,81 @@ def __str__(self):
...
+            op = '=='
+
+        if op not in self._op_actions.keys():
+            raise ValueError("invalid boolean operator: '%s'" % op)
+        else:
+            self._operator = op
+
+    def validate(self, version):
+        return self._op_actions[self._operator](Version(version),
+                                                self._ref_version)
```
vkarak: Better write this as follows:

```python
do_validate = self._op_actions[self._operator]
return do_validate(Version(version), self._ref_version)
```

unittests/test_versioning.py
```diff
@@ -27,3 +27,22 @@ def test_comparing_versions(self):
...
+        self.assertTrue(any(c.validate('3.0') for c in conditions))
+        self.assertFalse(any(c.validate('3.1') for c in conditions))
+        self.assertRaises(ValueError, VersionValidator, '2.0.0..')
+        self.assertRaises(ValueError, VersionValidator, '..2.0.0')
+        self.assertRaises(ValueError, VersionValidator, '1.0.0..2.0.0..3.0.0')
+        self.assertRaises(ValueError, VersionValidator, '=>2.0.0')
+        self.assertRaises(ValueError, VersionValidator, '2.0.0>')
+        self.assertRaises(ValueError, VersionValidator, '2.0.0>1.0.0')
+        self.assertRaises(ValueError, VersionValidator, '=>')
+        self.assertRaises(ValueError, VersionValidator, '>1')
```
vkarak: We will need more unit tests for the new decorator.

unittests/resources/checks_unlisted/required_versions.py
```diff
@@ -0,0 +1,30 @@
...
+@rfm.required_version('1.0..2.0', '<100.0')
+@rfm.simple_test
+class Test4Check(rfm.RegressionTest):
+    pass
+
+
+@rfm.simple_test
+@rfm.required_version('!=2.0')
+class Test5Check(rfm.RegressionTest):
+    pass
```
vkarak: I don't think we need this file at all. We already have the `unittests/resources/checks_unlisted/good.py` that is being loaded by the check loader. We could simply decorate one of the tests in there with a `required_version` that is always true and add another test, which should be skipped.

reframe/core/decorators.py
```diff
@@ -83,3 +93,32 @@ def _do_register(cls):
         return cls
 
     return _do_register
+
+
+def required_version(*versions):
+    """Class decorator for skipping version-uncompatible tests.
```
vkarak: Class decorator for specifying the required ReFrame version for the following test.

reframe/core/decorators.py
```diff
@@ -83,3 +93,32 @@ def _do_register(cls):
...
+
+    The decorated class must derive from
+    :class:`reframe.core.pipeline.RegressionTest`. This decorator is also
+    available directly under the :mod:`reframe` module.
+
+    :arg versions: The versions that are compatible with the test.
+
+    .. versionadded:: 2.13
+
```
vkarak: Remove this blank line.

reframe/core/decorators.py
```diff
@@ -83,3 +93,32 @@ def _do_register(cls):
...
 
     return _do_register
+
+
+def required_version(*versions):
+    """Class decorator for skipping version-uncompatible tests.
+
+    The decorated class must derive from
+    :class:`reframe.core.pipeline.RegressionTest`. This decorator is also
+    available directly under the :mod:`reframe` module.
```
vkarak: I don't think you test for that! And I don't think you should in any case...

reframe/core/decorators.py
```diff
@@ -83,3 +93,32 @@ def _do_register(cls):
...
+
+    :arg versions: The versions that are compatible with the test.
```
vkarak: Please explain a bit the syntax here. Also you should specify the type of the argument. See other part of the documentation on how this is done.

vkarak:
@jenkins-cscs retry daint

vkarak:
@jenkins-cscs retry daint

PR #288, opened 2018-05-23
Including name of the regression test file to the list of checks

### LINKED ISSUE #219: List also the regression test file when a test is listed
We have now hundreds of tests and it would be really useful for the `-l` option to print also the file a regression test is defined, so as to locate it easily.

rsarm:
Closes #219


reframe/core/pipeline.py
```diff
@@ -1081,9 +1082,11 @@ def cleanup(self, remove_files=False, unload_env=True):
 
     def __str__(self):
         return ('%s (%s)\n'
-                '        tags: [%s], maintainers: [%s]' %
+                '        tags: [%s], maintainers: [%s]\n'
+                '        file: [%s]' %
```
vkarak: I prefer an output like the following:

```
  * tensorflow_wide_deep_check (found in /path/to/reframe/cscs-checks/apps/tensorflow/tensorflow_check.py)
        descr: Tensorflow official wide_deep test
        tags: [production], maintainers: [TM]
```

reframe/core/pipeline.py
```diff
@@ -3,6 +3,7 @@
 #
 
 import fnmatch
+import inspect
```
vkarak: This causes a conflict with the master, since now it's already imported.

reframe/core/pipeline.py
vkarak: Also for the tags, I think we can keep the default set formatting with the `{}`.

reframe/core/pipeline.py
```diff
@@ -1095,10 +1095,13 @@ def cleanup(self, remove_files=False, unload_env=True):
             self._current_partition.local_env.unload()
 
     def __str__(self):
-        return ('%s (%s)\n'
-                '        tags: [%s], maintainers: [%s]' %
-                (self.name, self.descr,
-                 ', '.join(self.tags), ', '.join(self.maintainers)))
+        return ('%s (found in %s)\n'
+                '        descr: %s\n'
+                '        tags: {%s}, maintainers: [%s]' %
```
vkarak: Here we could directly print the set and the maintainers, without needing to have `{}` or `[]`.

reframe/core/pipeline.py
```diff
@@ -1095,10 +1095,13 @@ def cleanup(self, remove_files=False, unload_env=True):
...
+                (self.name,
+                 self.prefix + os.path.basename(inspect.getfile(type(self))),
+                 self.descr, ', '.join(self.tags),
+                 ', '.join(self.maintainers)))
```
vkarak: Similarly here, no need to join, just use the pure attribute.

PR #289, opened 2018-05-25
Add a knob for controlling the 'sacct'/'squeue' ratio in Slurm 'sacct' backend

### LINKED ISSUE #283: Very high 'squeue' rate from the Slurm 'sacct' backend
Even the Slurm backend using `sacct` does use `squeue` for retrieving the reason a job is pending, since AFAIK cannot be retrieved by `sacct`. This feature is required to cancel the job if it's blocked in a unrecoverable state and allows ReFrame to progress. However, it creates a high rate of `squeue` commands, which may be undesirable in some environments. I suggest doing this check every 10 or 20 `sacct` polls.

teojgo:
* Reduce the `squeue` rate by introducing the constant
  `SACCT_SQUEUE_RATIO`.

* Fix some warnings regarding regexes.

Fixes #283

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -34,8 +34,11 @@ SLURM_JOB_TIMEOUT     = SlurmJobState('TIMEOUT')
 
 @register_scheduler('slurm')
 class SlurmJob(sched.Job):
-    # The following ratio was introduced in order to reduce the number of
-    # squeue calls during asynchronous execution of ReFrame
+    # In some systems, scheduler performance is sensitive to the squeue poll
+    # ratio. In this backend, squeue is used to obtain the reason a job is
+    # blocked, so as to cancel it if it is blocked indefinitely. The following
+    # variable controls the frequency of squeue polling compared to the
+    # standard job state polling using sacct.
     SACCT_SQUEUE_RATIO = 10
 
     def __init__(self, *args, **kwargs):
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -31,6 +31,9 @@ SLURM_JOB_RUNNING     = SlurmJobState('RUNNING')
 SLURM_JOB_SUSPENDED   = SlurmJobState('SUSPENDED')
 SLURM_JOB_TIMEOUT     = SlurmJobState('TIMEOUT')
 
+# Number of _update_state calls per which _cancel_if_blocked is called
+SACCT_SQUEUE_RATIO = 20
+
 
 @register_scheduler('slurm')
 class SlurmJob(sched.Job):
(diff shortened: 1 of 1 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -61,6 +64,7 @@ def __init__(self, *args, **kwargs):
                                 'ReqNodeNotAvail',  # Inaccurate SLURM doc
                                 'QOSUsageThreshold']
         self._is_cancelling = False
+        self._update_state_count = 0
```
victorusu: Is it state or status?

reframe/core/schedulers/slurm.py
```diff
@@ -226,7 +231,11 @@ def _update_state(self):
             return
 
         self._state = SlurmJobState(state_match.group('state'))
-        self._cancel_if_blocked()
+
+        if self._update_state_count == SACCT_SQUEUE_RATIO:
```
vkarak: I would simply do a module operation here:

```python
if not self._update_state_count % SACCT_SQUEUE_RATIO:
    self._cancel_if_blocked()
```

reframe/core/schedulers/slurm.py
```diff
@@ -31,6 +31,9 @@ class SlurmJobState(sched.JobState):
 SLURM_JOB_SUSPENDED   = SlurmJobState('SUSPENDED')
 SLURM_JOB_TIMEOUT     = SlurmJobState('TIMEOUT')
 
+# Number of _update_state calls per which _cancel_if_blocked is called
+SACCT_SQUEUE_RATIO = 20
```
vkarak: I think this constant should be defined inside the `SlurmJob` class as class attribute, since it is very specific to this implementation.

vkarak: @victorusu Should we start with 10 and see if ops are happy?

reframe/core/schedulers/slurm.py
```diff
@@ -31,6 +31,9 @@ class SlurmJobState(sched.JobState):
 SLURM_JOB_SUSPENDED   = SlurmJobState('SUSPENDED')
 SLURM_JOB_TIMEOUT     = SlurmJobState('TIMEOUT')
 
+# Number of _update_state calls per which _cancel_if_blocked is called
```
vkarak: I don't like this comment very much; it contradicts with the name of the variable. If we use the `SACCT_SQUEUE_RATIO` as a name, I'd like to see a comment describing the rationale behind this instead, because what this variable controls is obvious from its name.

reframe/core/schedulers/slurm.py
vkarak: @victorusu I think 'state' is fine.

PR #305, opened 2018-05-29
Treat correctly relative imports in 'import_module_from_file'

### LINKED ISSUE #302: Unexpected behavior when passing relative paths to command line options
A `ModuleNotFoundError` is thrown when passing check files or directories to the `-c` option if this is done  from a directory that's not the base `~/reframe/`.
```bash
:prgenv > reframe -c ulimit_check.py -l
Command line: /home/sarafael/.local/bin/reframe --config-file /home/sarafael/git/reframe/reframe/settings.py --system generic -c ulimit_check.py -l
Reframe version: 2.13-dev0
Launched by user: sarafael
Launched on host: sarafael-thinkpad
Reframe paths
=============
    Check prefix      : 
    Check search path : 'ulimit_check.py'
    Stage dir prefix  : /home/sarafael/reframe/cscs-checks/prgenv/stage/
    Output dir prefix : /home/sarafael/reframe/cscs-checks/prgenv/output/
    Logging dir       : /home/sarafael/reframe/cscs-checks/prgenv/logs
/home/sarafael/.local/bin/reframe: unexpected error: No module named 'ulimit_check'
  File "/home/sarafael/git/reframe/reframe/frontend/cli.py", line 324, in main
  File "/home/sarafael/git/reframe/reframe/frontend/loader.py", line 181, in load_all
    checks.extend(self.load_from_file(d))
  File "/home/sarafael/git/reframe/reframe/frontend/loader.py", line 150, in load_from_file
    return self.load_from_module(util.import_module_from_file(filename))
  File "/home/sarafael/git/reframe/reframe/utility/__init__.py", line 51, in import_module_from_file
    return importlib.import_module(module_name)
  File "/home/sarafael/software/anaconda3/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 978, in _gcd_import
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load
  File "<frozen importlib._bootstrap>", line 948, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'ulimit_check'
```

Also, when passing to `--config-file` a path that starts with a `.` like `../../../config/cscs.py`, reframe is replacing the `/`s by `.`s and is then looking for `.........config.cscs`.

```bash
../../../bin/reframe: could not load settings: could not load configuration file `../../../config/cscs.py': the 'package' argument is required to perform a relative import for '.........config.cscs'
```

teojgo:
* Use the absolute path of the filename if the normalized
  path is not a directory.

Fixes #302


teojgo:
Maybe this needs a little bit of redesign. Should we allow loading of modules that don't exist?

unittests/test_utility.py
```diff
@@ -280,6 +280,22 @@ def test_load_unknown_path(self):
...
+    def test_load_dir_starts_with_ddot(self):
+        parent_dir = os.path.basename(os.path.abspath(os.pardir))
+        filename = '../%s/reframe' % parent_dir
+        module = util.import_module_from_file(filename)
+        self.assertEqual(reframe.VERSION, module.VERSION)
+        self.assertEqual('reframe', module.__name__)
+        self.assertIs(module, sys.modules.get('reframe'))
+
+    def test_load_file_starts_with_ddot(self):
+        parent_dir = os.path.basename(os.path.abspath(os.pardir))
```
vkarak: Better use `..` instead of `os.pardir` to be clearer and consistent with other usages.

vkarak: Also, I wouldn't write the test this way. I'd instead change temporarily to a directory and do a relative import from there. This is the faulty scenario in any case. You can use the `utility.os_ext.change_dir` context manager for doing the directory change safely.

reframe/utility/__init__.py
```diff
@@ -41,9 +41,13 @@ def import_module_from_file(filename):
     """Import module from file."""
 
     filename = os.path.normpath(os.path.expandvars(filename))
+
     if os.path.isdir(filename):
         filename = os.path.join(filename, '__init__.py')
 
+    if filename.startswith('..'):
+        filename = os.path.abspath(filename)
```
vkarak: I think that this is the only solution. It seems we cannot use the `importlib.import_module()`, because we don't which package to attach to. Am I right?

teojgo: yes, that is correct.

teojgo: I tested and it now works both with settings and checks defined with relative paths.

unittests/test_utility.py
```diff
@@ -280,6 +280,22 @@ def test_load_unknown_path(self):
             self.assertEqual('foo', e.name)
             self.assertEqual('/foo', e.path)
 
+    def test_load_dir_starts_with_ddot(self):
+        with os_ext.change_dir('reframe'):
+            filename = '..%s/reframe' % os.getcwd()
```
vkarak: I don't think that this does the intended thing, because `os.getcwd()` returns an absolute path. I don't think it's needed. In any case the `change_dir()` assumes that we are running the unit tests from top-level directory.

PR #308, opened 2018-05-29
Print filename if a test file contains syntax errors

### LINKED ISSUE #267: Filename information on syntax errors in user test files is missed
Example traceback:

```
/apps/common/UES/jenkins/SLES12/easybuild/software/reframe/2.11/bin/reframe: unexpected error: invalid syntax (<unknown>, line 47)
  File "/apps/common/UES/jenkins/SLES12/easybuild/software/reframe/2.11/reframe/frontend/cli.py", line 339, in main
    checks_found = loader.load_all(system=system, resources=resources)
  File "/apps/common/UES/jenkins/SLES12/easybuild/software/reframe/2.11/reframe/frontend/loader.py", line 135, in load_all
    checks.extend(self.load_from_file(d, **check_args))
  File "/apps/common/UES/jenkins/SLES12/easybuild/software/reframe/2.11/reframe/frontend/loader.py", line 99, in load_from_file
    if not self._validate_source(filename):
  File "/apps/common/UES/jenkins/SLES12/easybuild/software/reframe/2.11/reframe/frontend/loader.py", line 64, in _validate_source
    source_tree = ast.parse(f.read())
  File "/apps/common/UES/jenkins/SLES12/easybuild/software/Python-bare/3.6.2/lib/python3.6/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 47
    'MPICH_RDMA_ENABLED_CUDA': '1'
                             ^
SyntaxError: invalid syntax
```

The problems comes from how we do the AST parsing in `frontend/loader.py`:

```python
        with open(filename, 'r') as f:
            source_tree = ast.parse(f.read())

        validator = RegressionCheckValidator()
        validator.visit(source_tree)
        return validator.valid

```

teojgo:
* Raise a `ReframeSyntaxError` with a descriptive message if parsing of
  a `RegressionTest` fails.

* Create the corresponding unittest

Fixes #267

### DIFF
--- a/reframe/frontend/loader.py
+++ b/reframe/frontend/loader.py
@@ -11,8 +11,7 @@ from importlib.machinery import SourceFileLoader
 import reframe.core.debug as debug
 import reframe.utility as util
 from reframe.core.exceptions import (NameConflictError,
-                                     RegressionTestLoadError,
-                                     ReframeSyntaxError)
+                                     RegressionTestLoadError)
 from reframe.core.logging import getlogger
 
 
@@ -81,12 +80,7 @@ class RegressionCheckLoader:
         further tests and finalizes and validation."""
 
         with open(filename, 'r') as f:
-            try:
-                source_tree = ast.parse(f.read())
-            except SyntaxError as e:
-                message = ('\nRegression check syntax error:\n{0}Line: {1}\n'
-                           'File: {2}'.format(e.text, e.lineno, filename))
-                raise ReframeSyntaxError(message)
+            source_tree = ast.parse(f.read(), filename)
 
         validator = RegressionCheckValidator()
         validator.visit(source_tree)
--- a/reframe/frontend/loader.py
+++ b/reframe/frontend/loader.py
@@ -10,8 +10,7 @@ from importlib.machinery import SourceFileLoader
 
 import reframe.core.debug as debug
 import reframe.utility as util
-from reframe.core.exceptions import (NameConflictError,
-                                     RegressionTestLoadError)
+from reframe.core.exceptions import NameConflictError, RegressionTestLoadError
 from reframe.core.logging import getlogger
 
 
--- a/reframe/frontend/loader.py
+++ b/reframe/frontend/loader.py
@@ -10,7 +10,9 @@ from importlib.machinery import SourceFileLoader
 
 import reframe.core.debug as debug
 import reframe.utility as util
-from reframe.core.exceptions import NameConflictError, RegressionTestLoadError
+from reframe.core.exceptions import (NameConflictError,
+                                     RegressionTestLoadError,
+                                     ReframeSyntaxError)
 from reframe.core.logging import getlogger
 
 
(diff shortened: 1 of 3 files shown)


reframe/frontend/loader.py
```diff
@@ -79,7 +81,12 @@ def _validate_source(self, filename):
         further tests and finalizes and validation."""
 
         with open(filename, 'r') as f:
-            source_tree = ast.parse(f.read())
+            try:
+                source_tree = ast.parse(f.read())
+            except SyntaxError as e:
+                message = ('\nRegression check syntax error:\n{0}Line: {1}\n'
+                           'File: {2}'.format(e.text, e.lineno, filename))
+                raise ReframeSyntaxError(message)
```
vkarak: I don't like very much reraising a real `SyntaxError` as `ReframeSyntaxError`. The purpose of the latter is just to denote errors in the usage of the decorators. I agree the name is not very accurate, but I couldn't come up with a better name. 

The problem here is firstly conceptual: `ReframeSyntaxError`s *are* `ReframeError`s, but a real `SyntaxError` can never be a `ReframeError`. Secondly, reraising it differently may hide useful information from real problems.

The solution to this problem is really simple! Just pass a `filename` argument to `parse()`. Have a look [here](https://docs.python.org/3.6/library/ast.html#ast.parse).

unittests/test_loader.py
```diff
@@ -58,3 +59,8 @@ def test_conflicted_checks(self):
     def test_load_error(self):
         self.assertRaises(OSError, self.loader.load_from_file,
                           'unittests/resources/checks/foo.py')
+
+    def test_load_invalid_sytax(self):
+        self.assertRaises(ReframeSyntaxError, self.loader.load_from_file,
+                          'unittests/resources/checks_unlisted/'
+                          'invalid_syntax_check.py')
```
vkarak: Here you should check that the `filename` attribute of the raised `SyntaxError` is set properly.

reframe/frontend/loader.py
```diff
@@ -10,7 +10,8 @@
 
 import reframe.core.debug as debug
 import reframe.utility as util
-from reframe.core.exceptions import NameConflictError, RegressionTestLoadError
+from reframe.core.exceptions import (NameConflictError,
+                                     RegressionTestLoadError)
```
vkarak: Do not split the line.

unittests/test_loader.py
```diff
@@ -58,3 +58,11 @@ def test_conflicted_checks(self):
...
         self.assertRaises(OSError, self.loader.load_from_file,
                           'unittests/resources/checks/foo.py')
+
+    def test_load_invalid_sytax(self):
+        invalid_check = ('unittests/resources/checks_unlisted/'
+                         'invalid_syntax_check.py')
+        with self.assertRaises(SyntaxError) as e:
+            self.loader.load_from_file(invalid_check)
+
+        self.assertEqual(e.exception.filename, invalid_check)
```
vkarak: I know that `e` is visible outside the `with` statement, but does it make sense to have this assertion outside the `with` block?

teojgo: If you put the assertion inside the with block, it is not reachable since the with block exits once the exception is thrown.

vkarak:
@jenkins-cscs retry dom

PR #310, opened 2018-05-30
Add flexible A2A check

### LINKED ISSUE #180: Add flexible A2A test

rsarm:
Closes #180


cscs-checks/microbenchmarks/osu/flexible_a2a.py
```diff
@@ -0,0 +1,28 @@
+import os
```
vkarak: This import is not needed.

PR #311, opened 2018-06-01
Automatic arrays in compiler check

### LINKED ISSUE #254: Add performance test for Fortran working array (is memory pool implemented by compiler)
MCH will provide the test case.

ajocksch:
Closes #254

### DIFF
--- a/cscs-checks/mch/automatic_arrays.py
+++ b/cscs-checks/mch/automatic_arrays.py
@@ -12,9 +12,11 @@ class AutomaticArraysCheck(rfm.RegressionTest):
         if self.current_system.name in ['daint', 'dom']:
             self.modules = ['craype-accel-nvidia60']
             self._pgi_flags = '-acc -ta=tesla:cc60 -Mnorpath'
+            self._cray_variables = {}
         elif self.current_system.name in ['kesch']:
             self.modules = ['craype-accel-nvidia35']
             self._pgi_flags = '-O2 -ta=tesla,cc35,cuda8.0'
+            self._cray_variables = {'MV2_USE_CUDA': '1'}
 
         self.num_tasks = 1
         self.num_gpus_per_node = 1
@@ -51,6 +53,7 @@ class AutomaticArraysCheck(rfm.RegressionTest):
         if environ.name.startswith('PrgEnv-cray'):
             environ.fflags = '-O2 -hacc -hnoomp'
             key = 'PrgEnv-cray'
+            self.variables = self._cray_variables
         elif environ.name.startswith('PrgEnv-pgi'):
             environ.fflags = self._pgi_flags
             key = 'PrgEnv-pgi'
--- a/cscs-checks/mch/automatic_arrays.py
+++ b/cscs-checks/mch/automatic_arrays.py
@@ -1,13 +1,11 @@
-import os
-
+import reframe as rfm
 import reframe.utility.sanity as sn
-from reframe.core.pipeline import RegressionTest
 
 
-class AutomaticArraysCheck(RegressionTest):
+@rfm.simple_test
+class AutomaticArraysCheck(rfm.RegressionTest):
     def __init__(self, **kwargs):
-        super().__init__('automatic_arrays_check',
-                         os.path.dirname(__file__), **kwargs)
+        super().__init__()
         self.valid_systems = ['daint:gpu', 'dom:gpu', 'kesch:cn']
         self.valid_prog_environs = ['PrgEnv-cray*', 'PrgEnv-pgi*',
                                     'PrgEnv-gnu']
@@ -28,7 +26,7 @@ class AutomaticArraysCheck(RegressionTest):
                                      self.stdout, 'perf', float)
         }
 
-        self.aarrays_reference = {
+        self.arrays_reference = {
             'PrgEnv-cray': {
                 'daint:gpu': {'perf': (1.4E-04, None, 0.15)},
                 'dom:gpu': {'perf': (1.4E-04, None, 0.15)},
@@ -47,18 +45,15 @@ class AutomaticArraysCheck(RegressionTest):
         }
 
         self.maintainers = ['AJ', 'VK']
+        self.tags = {'production'}
 
     def setup(self, partition, environ, **job_opts):
-        if 'PrgEnv-cray' in environ.name:
+        if environ.name.startswith('PrgEnv-cray'):
             environ.fflags = '-O2 -hacc -hnoomp'
-        elif 'PrgEnv-pgi' in environ.name:
+        elif environ.name.startswith('PrgEnv-pgi'):
             environ.fflags = self._pgi_flags
-        elif 'PrgEnv-gnu' in environ.name:
+        elif environ.name.startswith('PrgEnv-gnu'):
             environ.fflags = '-O2'
 
+        self.reference = self.arrays_reference[environ.name]
         super().setup(partition, environ, **job_opts)
-        self.reference = self.aarrays_reference[self.current_environ.name]
-
-
-def _get_checks(**kwargs):
-    return [AutomaticArraysCheck(**kwargs)]
--- a/cscs-checks/mch/automatic_arrays.py
+++ b/cscs-checks/mch/automatic_arrays.py
@@ -1,0 +1,64 @@
+import os
+
+import reframe.utility.sanity as sn
+from reframe.core.pipeline import RegressionTest
+
+
+class AutomaticArraysCheck(RegressionTest):
+    def __init__(self, **kwargs):
+        super().__init__('automatic_arrays_check',
+                         os.path.dirname(__file__), **kwargs)
+        self.valid_systems = ['daint:gpu', 'dom:gpu', 'kesch:cn']
+        self.valid_prog_environs = ['PrgEnv-cray*', 'PrgEnv-pgi*',
+                                    'PrgEnv-gnu']
+        if self.current_system.name in ['daint', 'dom']:
+            self.modules = ['craype-accel-nvidia60']
+            self._pgi_flags = '-acc -ta=tesla:cc60'
+        elif self.current_system.name in ['kesch']:
+            self.modules = ['craype-accel-nvidia35']
+            self._pgi_flags = '-O2 -ta=tesla,cc35,cuda8.0'
+
+        self.num_tasks = 1
+        self.num_gpus_per_node = 1
+        self.num_tasks_per_node = 1
+        self.sourcepath = 'automatic_arrays.f90'
+        self.sanity_patterns = sn.assert_found(r'Result:\s+OK', self.stdout)
+        self.perf_patterns = {
+            'perf': sn.extractsingle(r'Timing:\s+(?P<perf>\S+)',
+                                     self.stdout, 'perf', float)
+        }
+
+        self.aarrays_reference = {
+            'PrgEnv-cray': {
+                'daint:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'dom:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'kesch:cn': {'perf': (2.9E-04, None, 0.15)},
+            },
+            'PrgEnv-gnu': {
+                'daint:gpu': {'perf': (7.4E-03, None, 0.15)},
+                'dom:gpu': {'perf': (7.4E-03, None, 0.15)},
+                'kesch:cn': {'perf': (6.5E-03, None, 0.15)},
+            },
+            'PrgEnv-pgi': {
+                'daint:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'dom:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'kesch:cn': {'perf': (1.4E-04, None, 0.15)},
+            }
+        }
+
+        self.maintainers = ['AJ', 'VK']
+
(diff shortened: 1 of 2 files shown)


cscs-checks/mch/automatic_arrays.py
```diff
@@ -0,0 +1,64 @@
+import os
+
+import reframe.utility.sanity as sn
+from reframe.core.pipeline import RegressionTest
+
+
+class AutomaticArraysCheck(RegressionTest):
+    def __init__(self, **kwargs):
+        super().__init__('automatic_arrays_check',
+                         os.path.dirname(__file__), **kwargs)
```
vkarak: Can you use the new syntax for regression tests? This boilerplate code won't be needed any more, as well as the `_get_checks()` function.

cscs-checks/mch/automatic_arrays.py
```diff
@@ -0,0 +1,64 @@
...
+        self.num_gpus_per_node = 1
+        self.num_tasks_per_node = 1
+        self.sourcepath = 'automatic_arrays.f90'
+        self.sanity_patterns = sn.assert_found(r'Result:\s+OK', self.stdout)
+        self.perf_patterns = {
+            'perf': sn.extractsingle(r'Timing:\s+(?P<perf>\S+)',
+                                     self.stdout, 'perf', float)
+        }
+
+        self.aarrays_reference = {
```
vkarak: Do you mean `arrays_reference`?

cscs-checks/mch/automatic_arrays.py
```diff
@@ -0,0 +1,64 @@
...
+                'daint:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'dom:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'kesch:cn': {'perf': (1.4E-04, None, 0.15)},
+            }
+        }
+
+        self.maintainers = ['AJ', 'VK']
+
+    def setup(self, partition, environ, **job_opts):
+        if 'PrgEnv-cray' in environ.name:
```
vkarak: Why are you using `in` here instead of `==`?

cscs-checks/mch/automatic_arrays.py
```diff
@@ -0,0 +1,64 @@
...
+            environ.fflags = '-O2 -hacc -hnoomp'
+        elif 'PrgEnv-pgi' in environ.name:
+            environ.fflags = self._pgi_flags
+        elif 'PrgEnv-gnu' in environ.name:
+            environ.fflags = '-O2'
+
+        super().setup(partition, environ, **job_opts)
+        self.reference = self.aarrays_reference[self.current_environ.name]
```
vkarak: Since you are already using `environ.name`, you'd better move this before `super().setup(...)` and use `environ.name` here as well, for symmetry.

cscs-checks/mch/automatic_arrays.py
```diff
@@ -0,0 +1,64 @@
...
+                'kesch:cn': {'perf': (6.5E-03, None, 0.15)},
+            },
+            'PrgEnv-pgi': {
+                'daint:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'dom:gpu': {'perf': (1.4E-04, None, 0.15)},
+                'kesch:cn': {'perf': (1.4E-04, None, 0.15)},
+            }
+        }
+
+        self.maintainers = ['AJ', 'VK']
```
vkarak: I think you should tag this test as `production`, too.

cscs-checks/mch/automatic_arrays.py
vkarak: @ajocksch You can check the tutorial examples to see the actual syntax.

cscs-checks/mch/automatic_arrays.py
ajocksch: since it might be PrgEnv-cray/xxxyyy

vkarak: Ok, I got that from your other PR. I think though is better to check `environ.name` starts with `PrgEnv-cray`, because what you have now allows also `xxx-PrgEnv-cray`.

cscs-checks/mch/automatic_arrays.py
ajocksch: I can do it. However a few checks fail and we will have "red: in the ci.

vkarak: No problem for that. We know that the programming environments are not working properly on Kesch. I will merge it as soon as the rest of the systems are "green".

cscs-checks/mch/automatic_arrays.py
ajocksch: done

cscs-checks/mch/automatic_arrays.py
ajocksch: done

cscs-checks/mch/automatic_arrays.py
ajocksch: done

cscs-checks/mch/automatic_arrays.py
ajocksch: done

cscs-checks/mch/automatic_arrays.py
ajocksch: ok

vkarak:
@jenkins-cscs retry dom

cscs-checks/mch/automatic_arrays.py
```diff
@@ -0,0 +1,59 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+@rfm.simple_test
+class AutomaticArraysCheck(rfm.RegressionTest):
+    def __init__(self, **kwargs):
+        super().__init__()
+        self.valid_systems = ['daint:gpu', 'dom:gpu', 'kesch:cn']
```
vkarak: @ajocksch The test fails constantly on Dom and (perhaps) will do so on the updated Daint. This needs investigation. The problem is that the performance checking is hardcoded inside this test, so performance checking in ReFrame has practically no effect (apart from logging) and, besides, we cannot adjust the performance values for other systems. For this reason, I don't think this test is really portable. I see two possible solutions here:

1. Make the test more portable by separating the sanity checking from performance checking. The sanity should make sure that no validation errors occur (see source code). The performance numbers should be adapted for each system.
2. Make this system available only for Kesch.

ajocksch: I think @victorusu made the point in the stand-up meeting: This check should behave differently for the different versions of the compilers. We should check for negative results if expected. Thus * is not working, one needs to specify all programming environments separately.

vkarak: @ajocksch @victorusu Guys, I understand that we need to have different performance values for the different programming environments, but the test itself is not portable. It assumes a single performance value (obtained perhaps only on a single system) and prints a PASS/FAIL based on that. If we want to do proper sanity and performance checking, we should ignore the PASS/FAIL printed by the test and let ReFrame do the performance checking based on the reference we put *per system*. If we go to the direction of putting this test in production for Daint/Dom, we should make it more robust and fix the performance values accordingly. If not, which is also my proposal since we want this test in ASAP, we should only allow it to run on Kesch.

vkarak:
@ajocksch The test fails due to `KeyError` in the `arrays_reference`. Have a look here:

https://jenkins.cscs.ch/blue/organizations/jenkins/ReframeCI/detail/ReframeCI/588/pipeline

ajocksch:
The * in PrgEnv is the problem, since the keys are only for PrgEnv without *

One solution: Run the check for one PrgEnv only.

Another solution: Extend the dictionaries? or allow somehow the * for searches

vkarak:
You can also do the following:

```python
if environ.name.startswith('PrgEnv-pgi'):
    key = 'PrgEnv-pgi'
else:
    key = environ.name

self.reference = self.arrays_reference[key]
```

ajocksch:
PrgEnv-pgi_16 PrgEnv-pgi_17 PrgEnv-pgi_18 fail as expected

also PrgEnv-cray_aj* fails since the mvapich* libraries set only the -I and -L path for mpif90 and not for ftn; this needs to be discussed

vkarak:
@jenkins-cscs retry all

vkarak:
@ajocksch This is the output I am getting:

```
SUMMARY OF FAILURES
FAILURE INFO for AutomaticArraysCheck
  * System partition: kesch:cn
  * Environment: PrgEnv-pgi_16
  * Stage directory: /users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_16
  * Maintainers: ['AJ', 'VK']
  * Failing phase: compile
  * Reason: OS error: [Errno 2] No such file or directory: 'mpif90': 'mpif90'
FAILURE INFO for AutomaticArraysCheck
  * System partition: kesch:cn
  * Environment: PrgEnv-pgi_17
  * Stage directory: /users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_17
  * Maintainers: ['AJ', 'VK']
  * Failing phase: compile
  * Reason: caught framework exception: Command '['mpif90', '-O2', '-ta=tesla,cc35,cuda8.0', '-I/users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_17', '/users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_17/automatic_arrays.f90', '-o', '/users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_17/./AutomaticArraysCheck']' failed with exit code 127:
=== STDOUT ===
=== STDERR ===
/appsmnt/escha/UES/RH7.3_experimental/pgi/18.4/linux86-64/2018/mpi/openmpi-2.1.2/bin/.bin/mpif90: error while loading shared libraries: libpgm.so: cannot open shared object file: No such file or directory

FAILURE INFO for AutomaticArraysCheck
  * System partition: kesch:cn
  * Environment: PrgEnv-pgi_18
  * Stage directory: /users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_18
  * Job type: batch job (id=817530)
  * Maintainers: ['AJ', 'VK']
  * Failing phase: sanity
  * Reason: sanity error: pattern `Result: ' not found in `/users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_18/AutomaticArraysCheck.out'
FAILURE INFO for AutomaticArraysCheck
  * System partition: kesch:cn
  * Environment: PrgEnv-pgi_18_aj
  * Stage directory: /users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-pgi_18_aj
  * Job type: batch job (id=817528)
  * Maintainers: ['AJ', 'VK']
  * Failing phase: performance
  * Reason: sanity error: 0.0001628 is beyond reference value 0.00014 (l=-inf, u=0.00016099999999999998)
FAILURE INFO for AutomaticArraysCheck
  * System partition: kesch:cn
  * Environment: PrgEnv-cray_aj
  * Stage directory: /users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-cray_aj
  * Job type: batch job (id=817483)
  * Maintainers: ['AJ', 'VK']
  * Failing phase: sanity
  * Reason: sanity error: pattern `Result: ' not found in `/users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-cray_aj/AutomaticArraysCheck.out'
FAILURE INFO for AutomaticArraysCheck
  * System partition: kesch:cn
  * Environment: PrgEnv-cray_aj_b
  * Stage directory: /users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-cray_aj_b
  * Job type: batch job (id=817438)
  * Maintainers: ['AJ', 'VK']
  * Failing phase: sanity
  * Reason: sanity error: pattern `Result: ' not found in `/users/karakasv/Devel/reframe/stage/cn/AutomaticArraysCheck/PrgEnv-cray_aj_b/AutomaticArraysCheck.out'
```

You should not also rely on the CI, because it only runs a test if you have changed the test's Python file. In this case, you don't, that's why it does not run it. You should try it manually.

ajocksch:
it is necessary to set the variable MV2_USE_CUDA for the cray compiler and mvapich compiled for gcc, although no gpu-direct is used; otherwise the code hangs in the first OpenACC directives

ajocksch:
@lxavier the performance of the check is not 100% reproducible; there might be a problem with the dynamic adaptation of the clock frequencies of the nodes

vkarak:

lxavier:
> @lxavier it is necessary to set the variable MV2_USE_CUDA for the cray compiler and ..
Interesting I think when we run cosmo on CPU we set MV2_USE_CUDA=0 , but we may use a different mvapich for cpu. Anyone, all this will have to be in the cosmo module files once Hannes is completed. Let it like this for now

lxavier:
> @lxavier the performance of the check is not 100% reproducible
We try to make it long enough so that this should not be an issue. We can increase the threshold. We want to detect if the timing goes completly of.  In addition we wanted to have a graphic to http://jenkins-mch.cscs.ch/view/POMPA/job/cosmo5_performance_benchmark/ so that we can monitor time, so it is ok if it fluctuate a bit.

PR #312, opened 2018-06-04
More reliable check for jobs pending indefinitely

### LINKED ISSUE #303: 'ReqNodeNotAvail' is not always a reason of indefinite blocking of a job

```
```

teojgo:
* Handle the above case which means that the job is pending.

* Minor codestyle fix.

Fixes #303

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -255,8 +255,8 @@ class SlurmJob(sched.Job):
         self._check_and_cancel(completed.stdout)
 
     def _check_and_cancel(self, reason_descr):
-        """Check if blocking reason ``reason_descr`` is unrecoverable and cancel the
-        job in this case."""
+        """Check if blocking reason ``reason_descr`` is unrecoverable and
+        cancel the job in this case."""
 
         # The reason description may have two parts as follows:
         # "ReqNodeNotAvail, UnavailableNodes:nid00[408,411-415]"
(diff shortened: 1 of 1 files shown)


teojgo:
@jenkins-cscs retry daint kesch monch

reframe/core/schedulers/slurm.py
```diff
@@ -266,6 +266,12 @@ def _check_and_cancel(self, reason_descr):
             # no reason details
             reason, reason_details = reason_descr, None
 
+        # Here we handle the case were the UnavailableNodes list is empty,
+        # which actually means that the job is pending
+        if reason == 'ReqNodeNotAvail' and reason_details:
+            if not re.match(r'UnavailableNodes:\S+', reason_details):
```
vkarak: I think this regex is a bit restrictive, cos you require that whatever comes after `:` has no whitespaces. I'd use instead the following:

```python
if re.match(r'UnavailableNodes:$', reason_details):
    return
```

vkarak: This whole `if` should go inside the next one, which checks if the reason is of a cancellation type.

vkarak: Another thing to remove is comment about "inaccurate SLURM documentation" in the definition of the `_cancel_reasons` list.

reframe/core/schedulers/slurm.py
```diff
@@ -267,6 +267,12 @@ def _check_and_cancel(self, reason_descr):
             reason, reason_details = reason_descr, None
 
         if reason in self._cancel_reasons:
+            # Here we handle the case were the UnavailableNodes list is empty,
+            # which actually means that the job is pending
+            if reason == 'ReqNodeNotAvail' and reason_details:
+                if not re.match(r'UnavailableNodes:$', reason_details):
```
vkarak: The `not` is wrong here. This regex matches only when there is nothing after the `:`.

PR #316, opened 2018-06-06
PBS scheduler backend

### LINKED ISSUE #46: Support for PBS submission system
Playing a bit around with Trque, one thing that initially bothered my was that there is now equivalent to `sacct` that we use for reliable job monitoring with Slurm. There is a `tracejob` command which provides the job history, but this is meant for admin usage only. However, there is a good thing with Torque that may allow us to have reliable job monitoring using the `qstat` command (the equivalent of Slurm's `squeue`). The job's standard output/error files are not created immediately in the directory the job was launched in, but rather they are copied there after the job has finished. As a matter of fact there is even a special command (`qpeek`) for looking at the output of a running job. So this feature (of lazy creation of job output) could be used as the regression's synchronisation point.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/215

vkarak:
Fixes #46 and replaces #277.

To test on Dom, you should do the following:

```bash
export PATH=/apps/dom/UES/karakasv/slurm-wrappers/bin:$PATH
./test_reframe.py --rfm-user-config=config/cscs-pbs.py
```
This special configuration is only valid for Dom and uses `pbs+mpiexec`.

Still todo:

- [x] Update user documentation
- [x] Enable CI testing

### DIFF
--- a/cscs-checks/cuda/cuda_checks.py
+++ b/cscs-checks/cuda/cuda_checks.py
@@ -12,7 +12,7 @@ class CudaCheck(RegressionTest):
         self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu']
         self.sourcesdir = os.path.join(self.current_system.resourcesdir,
                                        'CUDA', 'essentials')
-        self.modules = ['cudatoolkit']
+        self.modules = ['craype-accel-nvidia60']
         self.maintainers = ['AJ', 'VK']
         self.num_gpus_per_node = 1
         self.tags = {'production'}
--- a/ci-scripts/ci-runner.bash
+++ b/ci-scripts/ci-runner.bash
@@ -183,6 +183,18 @@ else
 
     checked_exec ./test_reframe.py --rfm-user-config=config/cscs.py
 
+    echo "==================================="
+    echo "Running unit tests with PBS backend"
+    echo "==================================="
+
+    if [[ $(hostname) =~ dom ]]; then
+        PATH_save=$PATH
+        export PATH=/apps/dom/UES/karakasv/slurm-wrappers/bin:$PATH
+        checked_exec ./test_reframe.py --rfm-user-config=config/cscs-pbs.py
+        export PATH=$PATH_save
+    fi
+
+
     # Find modified or added user checks
     userchecks=( $(git log --name-status --oneline --no-merges -1 | \
                    awk '/^[AM]/ { print $2 } /^R0[0-9][0-9]/ { print $3 }' | \
--- a/config/cscs-pbs.py
+++ b/config/cscs-pbs.py
@@ -1,0 +1,161 @@
+#
+# Minimal CSCS configuration for testing the PBS backend
+#
+
+
+class ReframeSettings:
+    _reframe_module = 'reframe'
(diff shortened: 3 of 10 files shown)


vkarak:
@rafa-esco Would you bother testing this version of the backend? Downloading the `pbs.py` file and replacing yours should be enough.

cscs-checks/cuda/cuda_checks.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, name, **kwargs):
         self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu']
         self.sourcesdir = os.path.join(self.current_system.resourcesdir,
                                        'CUDA', 'essentials')
-        self.modules = ['cudatoolkit']
+        self.modules = ['craype-accel-nvidia60']
```
vkarak: That's irrelevant here. I should remove it.

rafa-esco:
@vkarak No problemo, I just had to download a zip of the entire repository (because of the import reframe.core.config failing). It runs but there is a problem when cancelling the jobs... you have to change:

 ```
if self._pbs_server:
            jobid += ':' + self._pbs_server

```
by

 ```
if self._pbs_server:
            jobid += '.' + self._pbs_server

```
a part from that everything seems to work nicely :).

vkarak:
@rafa-esco Thanks for testing! Good catch, that was a typo. I wanted `.` actually!

rafa-esco:
@vkarak I just noticed a small bug in the _emit_lselect_function... If I do not define _num_cpus_per_task I do not get the "ncpus=" value so my job will only get 1 cpu per node. The default value for _num_cpus_per_tasks should be 1, so maybe adding a num_cpus_per_task = self._num_cpus_per_task or 1 and then using this variable in the downstream will always generate the correct ncpus option:

```
 def _emit_lselect_option(self, builder):
        num_tasks_per_node = self._num_tasks_per_node or 1
        num_cpus_per_task = self._num_cpus_per_task or 1
        num_nodes = self._num_tasks // num_tasks_per_node
        ret = '-l select=%s:mpiprocs=%s' % (num_nodes, num_tasks_per_node)
        num_cpus_per_node = num_tasks_per_node * num_cpus_per_task
        ret += ':ncpus=%s' % num_cpus_per_node

        if self.options:
            ret += ':' + ':'.join(self.options)

        self._emit_job_option(ret, builder)`
```

vkarak:
@rafa-esco So, in your setup, the `ncpus` must always be present? For our case, there wasn't a problem of not specifying it at all, if it wasn't specified.

The general idea for the backends is to generate the minimum required fields based on the user input in his regression test. So if the user does not set `self.num_cpus_per_tasks` and `ncpus` is not required by the scheduler, then it is omitted.

vkarak:
@rafa-esco OK, now I get it. You are right. I'll fix that.

vkarak:
@rafa-esco I've pushed the fixes you suggested. Could you test once more?

rafa-esco:
@vkarak Yep, all my tests passed :)

vkarak:
@victorusu I will merge this one. Is this ok?

PR #319, opened 2018-06-07
Filter out tests with no programming environments

### LINKED ISSUE #244: List should exclude tests with empty PrgEnvs
The `-l` cli flag relies on the `self.valid_systems` variable defined in the test.

One can set `self.valid_prog_environs = []`. This way, the test is listed in the prompt, but in fact, does not run on the system. It is skipped.

This is the correct behavior as stated in the documentation.  

However, it does not reflect the reality of the actual list of tests that run on the system.

teojgo:
* Create the corresponding unit test.

Fixes #244

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -351,8 +351,9 @@ def main():
         # Filter checks by prgenv
         if not options.skip_prgenv_check:
             checks_matched = filter(
-                lambda c: c if all(c.supports_environ(e)
-                                   for e in options.prgenv) else None,
+                lambda c: c if (c.valid_prog_environs
+                                and all(c.supports_environ(e)
+                                        for e in options.prgenv)) else None,
                 checks_matched
             )
 
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -350,17 +350,14 @@ def main():
         )
 
         # Filter checks by prgenv
-        if not options.skip_prgenv_check:
+        def filter_prgenv(c):
             if options.prgenv:
-                filter_func = lambda c: c if util.allx(
-                    c.supports_environ(e) for e in options.prgenv) else None
+                return util.allx(c.supports_environ(e) for e in options.prgenv)
             else:
-                # Here we also check that the valid_prog_environs of the check
-                # is an iterable and not empty
-                filter_func = lambda c: c if util.allx(
-                    c.valid_prog_environs) else None
+                return bool(c.valid_prog_environs)
 
-            checks_matched = filter(filter_func, checks_matched)
+        if not options.skip_prgenv_check:
+            checks_matched = filter(filter_prgenv, checks_matched)
 
         # Filter checks further
         if options.gpu_only and options.cpu_only:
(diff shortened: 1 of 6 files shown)


reframe/frontend/cli.py
```diff
@@ -351,8 +351,9 @@ def main():
         # Filter checks by prgenv
         if not options.skip_prgenv_check:
             checks_matched = filter(
-                lambda c: c if all(c.supports_environ(e)
-                                   for e in options.prgenv) else None,
+                lambda c: c if (c.valid_prog_environs
+                                and all(c.supports_environ(e)
+                                        for e in options.prgenv)) else None,
```
vkarak: This lambda gets a bit too complicated. Since this operation is quite useful, I suggest writing a function in utility doing that thing:

```python
def allx(iterable):
    # we should check here specifically if `iterable` is actually an iterable, as real all() does
    if not iterable:
        return False

    return all(iterable)
```

Would it also make sense to provide a deferrable version of this function in this PR, too?

unittests/test_sanity_functions.py
```diff
@@ -55,6 +55,23 @@ def test_all(self):
...
+        expr3 = sn.allx([])
+        expr4 = sn.allx(None)
+        expr5 = sn.allx(i for i in [])
+        expr6 = sn.allx(i for i in [1])
+        self.assertTrue(expr1)
+        self.assertFalse(expr2)
+        self.assertFalse(expr3)
+        self.assertFalse(expr4)
+        self.assertFalse(expr5)
+        self.assertTrue(expr6)
```
vkarak: You can combine these assertions with the expressions above.

reframe/utility/__init__.py
```diff
@@ -54,6 +56,28 @@ def import_module_from_file(filename):
...
+
+    """
+
+    # Check if iterable is an empty generator
+    if isinstance(iterable, types.GeneratorType):
+        try:
+            first_item = next(iterable)
+            return all(itertools.chain([first_item], iterable))
+        except StopIteration:
+            return False
```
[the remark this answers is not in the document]
teojgo: Because we do not want to accept empty generators. Is there any other way to check that a generator is empty?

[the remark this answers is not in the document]
vkarak: Now, I get what you are trying to do. The comment above the code is really misleading...

reframe/utility/__init__.py
```diff
@@ -54,6 +56,28 @@ def import_module_from_file(filename):
...
+
+    .. versionadded:: 2.13
+
+    """
+
+    # Check if iterable is an empty generator
+    if isinstance(iterable, types.GeneratorType):
+        try:
+            first_item = next(iterable)
+            return all(itertools.chain([first_item], iterable))
```
vkarak: Better put this in the `else` block of the try/except. Also rename `first_item` to `first` or `head`.

teojgo:
@jenkins-cscs retry daint monch kesch

teojgo:
@victorusu what should be the expected behavior when a check has an empty prgenv list and we also do not specify any particular prgenv to ReFrame?

vkarak:
@teojgo @victorusu The logical thing to do would be to print the test if it supports *any* programming environment, i.e., `valid_prog_environs` is not empty.

PR #323, opened 2018-06-08
Fix importing module from python file

### LINKED ISSUE #322: ModuleNotFoundError thrown for certain relative paths
For instance, when running an example from the tutorial on the root folder of reframe with
```
reframe -r -c tutorial/example1.py
```
the test runs fine. But when running it inside `reframe/tutorial` a `ModuleNotFoundError: No module named 'example1'` is thrown:
```
$> reframe -r -c example1.py 
Command line: /home/sarafael/.local/bin/reframe --config-file /home/sarafael/git/reframe/reframe/settings.py --system generic -r -c example1.py
Reframe version: 2.13-dev1
Launched by user: sarafael
Launched on host: sarafael-thinkpad
Reframe paths
=============
    Check prefix      : 
    Check search path : 'example1.py'
    Stage dir prefix  : /home/sarafael/git/reframe/tutorial/stage/
    Output dir prefix : /home/sarafael/git/reframe/tutorial/output/
    Logging dir       : /home/sarafael/git/reframe/tutorial/logs
/home/sarafael/.local/bin/reframe: unexpected error: No module named 'example1'
  File "/home/sarafael/git/reframe/reframe/frontend/cli.py", line 327, in main
  File "/home/sarafael/git/reframe/reframe/frontend/loader.py", line 182, in load_all
    checks.extend(self.load_from_file(d))
  File "/home/sarafael/git/reframe/reframe/frontend/loader.py", line 151, in load_from_file
    return self.load_from_module(util.import_module_from_file(filename))
  File "/home/sarafael/git/reframe/reframe/utility/__init__.py", line 54, in import_module_from_file
    return importlib.import_module(module_name)
  File "/home/sarafael/software/anaconda3/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 978, in _gcd_import
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load
  File "<frozen importlib._bootstrap>", line 948, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'example1'

```

However, it works fine if adding a path that starts with `../` before the relative path to the test. For instance, both
```
reframe -r -c ../../reframe/tutorial/example1.py
```
and
```
reframe -r -c ../tutorial/example1.py
```
work inside `~/reframe/tutorial`.

teojgo:
Fixes #322

### DIFF
--- a/reframe/utility/__init__.py
+++ b/reframe/utility/__init__.py
@@ -44,7 +44,7 @@ def import_module_from_file(filename):
     if os.path.isdir(filename):
         filename = os.path.join(filename, '__init__.py')
 
-    if filename.startswith('..'):
+    if filename.startswith('..') or os.path.isfile(filename):
         filename = os.path.abspath(filename)
 
     module_name = _get_module_name(filename)
--- a/reframe/utility/__init__.py
+++ b/reframe/utility/__init__.py
@@ -48,7 +48,7 @@ def import_module_from_file(filename):
         filename = os.path.abspath(filename)
 
     module_name = _get_module_name(filename)
-    if os.path.isabs(filename) or os.path.isfile(filename):
+    if os.path.isabs(filename) or os.path.basename(filename) == filename:
         return _do_import_module_from_file(filename, module_name)
 
     return importlib.import_module(module_name)
--- a/reframe/utility/__init__.py
+++ b/reframe/utility/__init__.py
@@ -44,11 +44,11 @@ def import_module_from_file(filename):
     if os.path.isdir(filename):
         filename = os.path.join(filename, '__init__.py')
 
-    if filename.startswith('..') or os.path.isfile(filename):
+    if filename.startswith('..'):
         filename = os.path.abspath(filename)
 
     module_name = _get_module_name(filename)
-    if os.path.isabs(filename):
+    if os.path.isabs(filename) or os.path.isfile(filename):
         return _do_import_module_from_file(filename, module_name)
 
     return importlib.import_module(module_name)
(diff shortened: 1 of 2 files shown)


teojgo:
@rsarm can you running try with this branch?

reframe/utility/__init__.py
```diff
@@ -44,7 +44,7 @@ def import_module_from_file(filename):
     if os.path.isdir(filename):
         filename = os.path.join(filename, '__init__.py')
 
-    if filename.startswith('..'):
+    if filename.startswith('..') or os.path.isfile(filename):
```
vkarak: Wait a sec. Is this not always true??

teojgo: if the path is a directory no

reframe/utility/__init__.py
```diff
@@ -48,7 +48,7 @@ def import_module_from_file(filename):
         filename = os.path.abspath(filename)
 
     module_name = _get_module_name(filename)
-    if os.path.isabs(filename):
+    if os.path.isabs(filename) or os.path.basename(filename) == filename:
```
vkarak: This fix does not work. Try running reframe from a directory up:

```
./reframe/bin/reframe --system=generic -c reframe/tutorial/example1.py -l
```

The whole problem comes from how the `sys.path` is defined (check [this](https://docs.python.org/3.6/library/sys.html#sys.path)). Actually, for our case, the first element is the path where the `reframe.py` or `test_reframe.py` exists. That's why these relative imports work fine when we we invoke ReFrame from the same directory. I've done a PR into your branch with the correct implementation.

PR #325, opened 2018-06-12
Strip the reason details before testing for match

### LINKED ISSUE #303: 'ReqNodeNotAvail' is not always a reason of indefinite blocking of a job

teojgo:
Closes #303


teojgo:
@kraushm can you try now in kesch?

teojgo:
@jenkins-cscs retry all

kraushm:
First run on Kesch passed as well, will do some more runs.

kraushm:
Further tests are ok. Job on Kesch was queued for approx. 5min with `(ReqNodeNotAvail, UnavailableNodes:)` without being killed.

Issue #328, opened 2018-06-14
ddt_check_Cuda fails after maintenance


jgphpc:
FAILURE INFO for ddt_check_Cuda 

```
  * System partition: daint:gpu
  * Environment: PrgEnv-gnu
  * Stage directory: /scratch/snx3000/piccinal/reframe/stage/gpu/ddt_check_Cuda/PrgEnv-gnu
  * Job type: batch job (id=7993758)
  * Maintainers: ['MK', 'JG']
  * Failing phase: sanity
  * Reason: sanity error: pattern `MPI implementation' not found in `ddtreport.txt'
```

vkarak:
@teojgo As far as I understood you are also working on that, aren't you?

jgphpc:
>    [1]@teojgo As far as I understood you are also working on that, aren't
>    you?

I will do it:
self.maintainers = ['MK', 'JG']

vkarak:
Ok, marked as important and moved it to the current sprint.

teojgo:
@jg I managed to make it work by enforcing use of cudatoolkit 8.0

jgphpc:
During daint's maintenance, cuda driver was updated to support cuda/9 (was 375.74, is 384.11 now) but cudatoolkit/8.0.61_2.4.3-6.0.4.0_3.1__gb475d12
remains default => setting ALLINEA_FORCE_CUDA_VERSION=8.0 for daint checks.

PR #329, opened 2018-06-14
ddt_check_Cuda fails after maintenance

jgphpc:
fix for #328


teojgo:
@jgphpc why did  you change the `notool.py` and not the `ddt.py`?

jgphpc:
>    [1]@jgphpc why did you change the notool.py and not the ddt.py?
wrong push, vk is trying to fix it.

vkarak:
The problem, guys, is that this has been accidentally pushed to master. And perhaps the master is broken. It won't be an easy fix.. There are merges dispersed over time...

teojgo:
Maybe then reset to the `origin/master` and just make the change to `ddt.py`

vkarak:
@teojgo Not so simple. At commit bfbdb3742e06e420fd0d73b2d061bf0bb35f17eb, @jgphpc has pushed into his local master some changes in the settings, so every time he was pulling, new merge commits were inserted! I don't think we can get rid of them. I will just try to check that the repo is in a sane state.

vkarak:
I've forced pushed from my fork, which was up to date. I am closing this PR and you resubmit. 

@jgphpc Make sure to first `git reset` your master before pulling. Otherwise we will end up to the same situation.

PR #332, opened 2018-06-15
Fix paths for Boost Python check

teojgo:
* Update python3 test from python3.5 to python3.6.

* Use the PYTHONPATH environment variable to extract the python dir.

### DIFF
--- a/cscs-checks/libraries/boost/boost_python_check.py
+++ b/cscs-checks/libraries/boost/boost_python_check.py
@@ -1,20 +1,18 @@
 import os
 
+import reframe as rfm
 import reframe.utility.sanity as sn
-from reframe.core.pipeline import RegressionTest
 
 
-class BoostCrayGnuPythonTest(RegressionTest):
-
-    def __init__(self, boost_version, cray_gnu_version, python_version,
-                 **kwargs):
-
+@rfm.parameterized_test(['1.65.0', '17.08', '2.7'], ['1.65.0', '17.08', '3.6'])
+class BoostCrayGnuPythonTest(rfm.RegressionTest):
+    def __init__(self, boost_version, cray_gnu_version, python_version):
         def normalize(s):
             return s.replace('.', '_')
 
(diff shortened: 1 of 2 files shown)


cscs-checks/libraries/boost/boost_python_check.py
```diff
@@ -38,4 +38,4 @@ def normalize(s):
 
 def _get_checks(**kwargs):
     return [BoostCrayGnuPythonTest('1.65.0', '17.08', '2.7', **kwargs),
-            BoostCrayGnuPythonTest('1.65.0', '17.08', '3.5', **kwargs)]
+            BoostCrayGnuPythonTest('1.65.0', '17.08', '3.6', **kwargs)]
```
vkarak: I suggest converting the test to the new style.

cscs-checks/libraries/boost/boost_python_check.py
```diff
@@ -1,20 +1,18 @@
 import os
```
vkarak: This is not needed.

cscs-checks/libraries/boost/boost_python_check.py
```diff
@@ -1,20 +1,18 @@
...
+@rfm.parameterized_test(['1.65.0', '17.08', '2.7'], ['1.65.0', '17.08', '3.6'])
+class BoostCrayGnuPythonTest(rfm.RegressionTest):
+    def __init__(self, boost_version, cray_gnu_version, python_version):
         def normalize(s):
             return s.replace('.', '_')
 
         super().__init__('boost_%s_cray_gnu_%s_python_%s_check' % (
             normalize(boost_version), normalize(cray_gnu_version),
-            normalize(python_version)), os.path.dirname(__file__), **kwargs)
+            normalize(python_version)))
```
vkarak: Set the name explicitly, not by passing it to the base class constructor. It may be deprecated in the future.

vkarak: And btw I don't think that you need to set it at all. The default name should be fine.

PR #213, opened 2018-03-21
Redesign of performance logging and support for Graylog

victorusu:
**EDIT by @vkarak**

This PR, along with the support for Graylog performance logging, brings a redesign of the performance logging:

- New syntax for logging handler configuration.
  The new syntax is more like JSON.
  A warning is issued if the old syntax is used and it is automatically
  converted to the new.

- Logging fields for checks are enriched with performance information

- A special logging handler is introduced that dynamically creates a
  file based on live information of the first record to be logged.
  This handler is used for file performance logging, where each
  performance test must have its own performance file.

- No more properties in settings, because requires some programming
  skills to edit and adapt to future settings changes.

Still todo:

- ~~Update documentation~~

**The following description is now obsolete**

>This PR adds the support to log the performance checks that do use using
graylog.
>
> The implementation adds the following fields in graylog:
>
> * check_info which contains the check name the system:partition and PrgEnv
> * check_name which shows the check name
> * check_partition which displays the system partition used
> * check_perf_reference which shows the reference value of a given test
> * check_perf_upper_thres which displays the upper limit threshold of a test
> * check_perf_lower_thres which displays the lower limit threshold of a check
> * check_perf_value which holds the actual perf value recorded on that system
> * check_system which shows the system on which the test has ran
> * data-version which display the reframe version of the data
> * version which displays the reframe version

PR #334, opened 2018-06-18
[feat] Evaluate and log all performance values before asserting them

mand35:
One can create multiple Objectives for a singe test case (see example below). This commit improves the logging of the values. It add the tag to the log and it logs all values. (before it was all up to reference is not reached). Thus we can keep better track of all the values observed during the tests. 

Reference example:
       p_names = {'creation', 'stat', 'removal'}
        for p_name in p_names:
           self.perf_patterns[p_name] = sn.extractsingle(r'^\s+File '+p_name+
                    '\s+:\s+(?P<'+p_name+'>\S+) ', self.stdout, p_name, float)

        self.reference = {
            'kupe:compute' : {
                'creation' : (7747,  -(2*223.1)/7747, None),
                'stat' :     (16527, -(2*558.2)/16527,None),
                'removal' :  (7355, -(2*173.1)/7355, None)
            },
        }


vkarak:
@mand35 Thanks for your PR. The performance logging is completely redesigned in #213, which is planned to be merged this week. There you can specify exactly the format of the performance log in the configuration file and you can log the performance variables (tags) as well.

reframe/core/pipeline.py
```diff
@@ -1039,21 +1039,27 @@ def check_performance(self):
...
                     raise SanityError(
                         "tag `%s' not resolved in references for `%s'" %
                         (tag, self._current_partition.fullname)
                     )
-                evaluate(assert_reference(value, ref, low_thres, high_thres))
+            for tag, expr in self.perf_patterns.items():
+                key = '%s:%s' % (self._current_partition.fullname, tag)
+                ref, low_thres, high_thres = self.reference[key]
+                evaluate(assert_reference(value[tag], ref, low_thres,
+                         high_thres))
```
vkarak: I can't understand very well the purpose of this nested loop here.

mand35: It is not a nested loop. I needed to move the evaluation to an additional loop. The reason was that if one of the early reference comparison throw the assert the other values were not written to the log file. In this way first all values are written and then the comparison is performed.

vkarak:
@jenkins-cscs retry all

reframe/core/pipeline.py
vkarak: You are right, for some reason I overlooked the alignment! I see what you are trying to do.

reframe/core/pipeline.py
```diff
@@ -1039,21 +1039,27 @@ def check_performance(self):
             return
 
         with os_ext.change_dir(self._stagedir):
+            # first check and print all values with references
+            value = {}
```
vkarak: Better call this `perf_values`. I'd also change a bit the comment to "Evaluate all performance variables first, so as to log them, then assert them".

reframe/core/pipeline.py
```diff
@@ -1039,21 +1039,27 @@ def check_performance(self):
...
+                        (tag, value[tag], self.reference[key])
                     )
+                    ref, low_thres, high_thres = self.reference[key]
                 except KeyError:
                     raise SanityError(
                         "tag `%s' not resolved in references for `%s'" %
                         (tag, self._current_partition.fullname)
                     )
-                evaluate(assert_reference(value, ref, low_thres, high_thres))
+            for tag, expr in self.perf_patterns.items():
```
vkarak: There is no need to go over the `perf_patterns` again here. Just over the evaluated `perf_values`.

Also, leave a blank line after the previous `for` loop.

reframe/core/pipeline.py
```diff
@@ -1039,21 +1039,27 @@ def check_performance(self):
...
+                value[tag] = evaluate(expr)
                 key = '%s:%s' % (self._current_partition.fullname, tag)
                 try:
-                    ref, low_thres, high_thres = self.reference[key]
                     self._perf_logger.info(
-                        'value: %s, reference: %s' %
-                        (value, self.reference[key])
+                        '%s, value: %s, reference: %s' %
+                        (tag, value[tag], self.reference[key])
                     )
```
vkarak: For the logging, I suggest waiting for #213 to be merged (perhaps today).

vkarak:
Closing this. Replaced by #359.

PR #337, opened 2018-06-20
Add log hander attributes for retrieving the OS user and group

victorusu:
This PR attempts to add the username and group to perf logging.
This is especially good for grafana logging.

This should capture the name and primary group of the user that has
executed reframe.

REMARKS:
* capturing the username and group is not portable across OSes
* The fields were updated inside the `_update_check_extras`, but they
could be set during the initialization of the `LoggerAdapter`.

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -343,7 +343,6 @@ class LoggerAdapter(logging.LoggerAdapter):
                 'check_perf_ref': None,
                 'check_perf_lower_thres': None,
                 'check_perf_upper_thres': None,
-                'check_perf_upper_thres': None,
                 'username': None,
                 'group': None,
                 'check_tags': None,
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -12,6 +12,7 @@ from datetime import datetime
 
 import reframe
 import reframe.core.debug as debug
+import reframe.utility.os_ext as os_ext
 from reframe.core.exceptions import ConfigError, LoggingError
 
 
@@ -343,8 +344,8 @@ class LoggerAdapter(logging.LoggerAdapter):
                 'check_perf_ref': None,
                 'check_perf_lower_thres': None,
                 'check_perf_upper_thres': None,
-                'username': None,
-                'group': None,
+                'osuser':  os_ext.osuser()  or '<unknown>',
+                'osgroup': os_ext.osgroup() or '<unknown>',
                 'check_tags': None,
                 'version': reframe.VERSION,
             }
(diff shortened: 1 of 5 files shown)


reframe/core/logging.py
```diff
@@ -343,6 +343,9 @@ def __init__(self, logger=None, check=None):
                 'check_perf_ref': None,
                 'check_perf_lower_thres': None,
                 'check_perf_upper_thres': None,
+                'check_perf_upper_thres': None,
```
teojgo: Why is the `check_perf_upper_thres` two times here?

victorusu: copy and paste issues... sorry! 😯

reframe/core/logging.py
```diff
@@ -378,6 +380,24 @@ def _update_check_extras(self):
         if self.check.job:
             self.extra['check_jobid'] = self.check.job.jobid
 
+        try:
+            import pwd
+            self.extra['username'] = pwd.getpwuid(os.geteuid()).pw_name
```
teojgo: @vkarak is it better to get the `euid` or the `uid`?

reframe/core/logging.py
```diff
@@ -343,6 +343,8 @@ def __init__(self, logger=None, check=None):
                 'check_perf_ref': None,
                 'check_perf_lower_thres': None,
                 'check_perf_upper_thres': None,
+                'username': None,
+                'group': None,
```
vkarak: I suggest those being `osuser` and `osgroup`.

reframe/core/logging.py
```diff
@@ -378,6 +380,24 @@ def _update_check_extras(self):
...
             self.extra['check_jobid'] = self.check.job.jobid
 
+        try:
+            import pwd
+            self.extra['username'] = pwd.getpwuid(os.geteuid()).pw_name
+        except:
+            try:
+                import getpass
+                self.extra['username'] = getpass.getuser()
+            except:
```
vkarak: Bare except clauses are disallowed by PEP8. Either be more precise or catch `Exception` or `BaseException` (not this case).

reframe/core/logging.py
```diff
@@ -378,6 +380,24 @@ def _update_check_extras(self):
...
+                pass
+
+        try:
+            import grp
+            import pwd
+            gid = pwd.getpwnam(self.extra['username']).pw_gid
+            self.extra['group'] = grp.getgrgid(gid).gr_name
+        except:
+            pass
```
vkarak: The whole thing here is overly complicated. I suggest the following:

```python
try:
    user = getpass.getuser()
except Exception:
    user = '<unknown>'

try:
    group = grp.getgrgid(os.getgid()).gr_name
except KeyError:
    group = '<unknown>'

self.extra['osuser'] = user
self.extra['osgroup'] = group
```

The `getpass()` does already use `pwd` under the hood if necessary.

We would the extreme way proposed here only if we had actually to care about effective and real user ids. But (a) I don't think we have to and (b) if that was the case, this would have to go in `utility.os_ext`.

vkarak: Also @victorusu, I indeed think this should not go in the `_update_check_extras`. It should be in the `LoggerAdapter`, but the I suggest moving this query to the `os_ext.utility` as `osuser()` and `osgroup()` functions. These functions should return the user and group name or `None`, so that inside the `LoggerAdapter` you would do the following:

```python
                'osuser': os_ext.osuser() or '<unknown>',
                'osgroup': os_ext.osgroup() or '<unknown>',
```

vkarak: There are a couple of places in the framework that we are using `os.environ['USER']` to retrieve the current user. We should replace these with the new utility function. Sorry, you asked for it! 😛

reframe/core/logging.py
vkarak: Also we need to document those in the public documentation.

Issue #339, opened 2018-06-21
'TypedListField's accept also strings


vkarak:
This is because we are checking if the assigned value if a sequence and strings are a sequence. We should create a separate `TypedSequenceField` for accepting generic sequences.

vkarak:
Closed by #340.

Issue #297, opened 2018-05-28
Create a 'Make' build system


vkarak:
A build system that will use `make` for compiling the source code inside a directory. It will be used automatically by the framework if `sourcepath` refers to a directory and no build system is specified.

teojgo:
While preparing the cudamemtest check, I wanted to use `Make` with a nondefault target. It would be nice if the new build system had also an option to specify a target to the Makefile directly, without having to pass it to the `options` during the compilation.

vkarak:
It has an `options` attribute where you can pass anything to the make invocation. Check my PR. I don't think it makes sense to add a separate attribute just for specifying targets.

vkarak:
Closed by #340.

Issue #296, opened 2018-05-28
Create a single source file build system


vkarak:
A simple build system backend for compiling a single source file. It will be used by the framework when `sourcepath` refers to a file and no build system is specified.

vkarak:
Closed by #340.

Issue #176, opened 2018-03-05
Add inherent support of parallel 'make'


vkarak:
Currently, if somebody wants to make his build parallel, he has to override the `compile()` method as follows:

```python
    def compile(self):
        super().compile(options='-j')
```

We should somehow provide this automatically. What do you think?

vkarak:
Closed by #340.

PR #340, opened 2018-06-21
[feat] Add a new infrastructure for build systems

### LINKED ISSUE #295: Create class infrastructure to support multiple build systems
The current way of building source codes from within ReFrame is very restrictive, since it was designed for very simple codes. If ReFrame is to be used as a proper CI tool for HPC apps, we need to better support the classic build systems. Here's a proposed class hierarchy for the new design:

```
+-----------------+             +-------------------+
|   Environment   |             |    BuildSystem    |
+-----------------+       uses  +-------------------+
|  modules        |       +-----+  environ          |
|  variables      |       |     |  prebuild_cmd     |
|                 |       |     |  postbuild_cmd    |
+--------+--------+       |     |                   |
         ^                |     +-------------------+
         |inherits        |     |  build()          |
         |                |     +---+-----+-----+---+
+--------+--------+       |         ^     ^     ^
| ProgEnvironment <-------+         |     |     +-----------------------------------------------+
+-----------------+                 |     +--------------------------+                          |
|  cxxflags       |                 |                                |                          |
|  ...            |             +---+---------------+      +---------+---------+      +---------+--------+
|                 |             |       CMake       |      |   ConfigureMake   |      |       Make       |
+-----------------+             +-------------------+      +-------------------+      +------------------+
                                |  config_opts      |      |                   |      |                  |
                                |  ...              |      |                   |      |                  |
                                |                   |      |                   |      |                  |
                                +-------------------+      +-------------------+      +------------------+
```

The key change here is that the responsibility of building is moved from the programming environment to a new abstract class, called `BuildSystem`. This will provide a `build()` (perhaps also a `prebuild()` and `postbuild()`) that the framework will use to build the code of the test. Concrete subclasses will implement different build system by taking care of the different steps. Subclasses may provide additional fields for configuring their build process. The desired user interface is the following:

```python
self.sourcepath = 'mycode/'
self.build_system = 'CMake'
self.build_system.config_opts = '-DUSE_MYLIB'
```

Using Python descriptors we can do the magic to create a build system instance by directly assigning a string to it.

#### Backward compatibility

The framework internally will only use a build system to build a test's code. The current building process can be implemented with two different build systems:

1. A `SingleSource` build system that would simply compile a single source file.
2. A `Make` build system that would use make to compile a directory.

The `prebuild_cmd` and `postbuild_cmd` `RegressionTest` attributes can be used to set the corresponding attributes from the `BuildSystem`.

vkarak:
This PR provides two major contributions: 
1. A new build system infrastructure.
2. A major redesign of the internal infrastructure of shell script generation.

## New build system infrastructure

The compilation phase is now separated from the programming environment.
It is part of the new concept of build systems. A build system is
responsible for generating the required commands for compiling a
code. The framework then uses these commands to generate a build job
script which is then submitted. Currently, only local compilation is
supported.

The current behavior of the `sourcepath` and `sourcesdir` attributes is
maintained for convenience. Internally, they translate to concrete build
systems that are set up accordingly.

All build systems share some basic attributes and behavior:

- The compilers and compilation flags. By default, if not specified the
  corresponding values from the current programming environment will be
  used.
- The ability to ignore completely the current programming
  environment. A build system may be configured independently of the
  current programming environment by explicitly setting the compilers
  and compilation flags.

The above design allows the programming environment to become immutable
holding the global default values for each system.
Currently, for backward compatibility, it is not yet immutable, but
setting its attributes is now deprecated.

Two build systems are provided by this commit:

1. `SingleSource`: This build system is responsible for compiling a
    single source file in any of the recognized programming languages,
    i.e., C, C++, Fortran and CUDA.

2. `Make`: This build system is responsible for compiling a project
   using the `make` command.

## Redesign of the shell script generation infrastructure

Shell script generation is revised significantly. Here are the key
points of this revision:

- ReFrame generates only Bash. This has always been the case, but the
  previous design was such as to "enable" generation of other types of
  shell scripts. This pseudo-generic design was completely dropped and
  replaced by a Bash-only script generator. The rationale behind this is
  that in order to support fully portable script generation, it would
  require an intermediate language that would interface with the shell
  script generators and would require users to program in that language
  when setting the `pre_run`, `post_run` etc. attributes. This is way to
  much effort with zero or negative value.

  Another solution considered was to globally control the shell type
  used by ReFrame from the settings. The downside with this solution is
  that user tests are becoming inherently less portable, if a user wants
  to write his `pre_run`, `post_run` etc. commands in, say, fish.

  Finally, supporting multiple shell script generation backends would
  make the internal design more complicated and less consistent, since
  the different parts of the framework that need to emit shell commands
  would be required to do that through another API and not directly.

  For all these reasons, and given that Bash is the "standard" POSIX
  shell, I decided to drop completely the old pseudo-generic design in a
  favor of a simpler and more consistent behavior across the framework.

- The generated shell scripts are more sophisticated now. They can trap
  different events (signals, errors and exit) during their execution and
  act accordingly. These additions are crucial for two parts of the
  framework:

  1. For the generated build script. We want to trap errors in order to
     exit immediately if any of the commands fail without requiring the
     user to take extra precaution for that when setting the
     `prebuild_cmd` and `postbuild_cmd`.

  2. For getting reliable exit and signal code information for job
     scheduler backends that do not support it. By trapping
     signals (with a "terminate" or "core dump" default action) and the
     shell exit, we can always record the both the signal number and the
     exit code in the output. Then the scheduler backend can retrieve
     this information.

     NOTE: This feature is not yet implemented.

- A shell script inside ReFrame consists of the three parts:

  1. The shebang, i.e., the very first line of the script.
  2. The prolog
  3. The body
  4. The epilog

  Anyone wanting to generate a script may decide in which part (except
  the shebang) to emit the commands he likes. When the `finalize()`
  method is called the whole script will be generated. ReFrame emits its
  traps between the prolog and the body.

- The `Job` abstract class does not hold any more information not
  relating directly to the job creation and status, i.e., `pre_run`,
  `post_run`, `executable` etc. Instead, it provides a new, richer
  version of the `prepare` method for generating the job script. This
  new `prepare()` has the following signature:

  ```python
  def prepare(self, commands, environs, **gen_opts):
  ```

  The `commands` argument is a list of the actual shell commands to be
  emitted in the job script. The caller is responsible for filling up
  this list.

  The `environs` argument is a list of the environments to be set loaded
  before emitting the commands.

  The `gen_opts` arguments are passed through the bash script generator.

- This PR establishes also conventions for the functions emitting shell
  code. These should start with the `emit_` prefix and must return a
  list of shell commands. They must not accept a shell script generator
  object as argument. The standard way of generating a shell script is
  the following:

  ```python
  import reframe.core.shell as shell

  with shell.generate_script(filename) as gen:
      gen.write_prolog(emit_foo1())
      gen.write_body(emit_bar())
  ```

  The `generate_script` context manager takes care of the file creation
  and the finalization of the shell script.

Fixes #295, #297, #296, #339, #176.

Still todo:

- Update documentation (see issue #360)
- Adapt regression tests to avoid deprecation warnings (issues will be created inside the corresponding [project](https://github.com/eth-cscs/reframe/projects/8))


reframe/core/environments.py
```diff
@@ -266,133 +312,26 @@ def __init__(self,
...
+        self._cxx = cxx
+        self._ftn = ftn
+        self._nvcc = nvcc
+        self._cppflags = cppflags
+        self._cflags   = cflags
+        self._cxxflags = cxxflags
+        self._fflags   = fflags
+        self._ldflags  = ldflags
+        self._include_search_path = []
+        self._propagate = True
```
vkarak: I am not treating this variable in the new build systems. I should fix that for compatibility reasons.

reframe/core/exceptions.py
```diff
@@ -88,6 +88,10 @@ class StatisticsError(ReframeError):
     """Raised to denote an error in dealing with statistics."""
 
 
+class BuildSystemError(ReframeError):
+    """Raised when a build system is not configured properly."""
+
+
```
vkarak: I should also remove the `CompilationError`. Not valid any more.

reframe/core/pipeline.py
```diff
@@ -1114,6 +1162,13 @@ def compile(self, **compile_opts):
         This is a no-op for this type of test.
         """
 
+    def compile_wait(self):
+        """Wait for compilation phase to finish.
+
+        This is a no-op for this type of test.
+        .. versionadded:: 2.13
+        """
+
```
vkarak: I should remove this whole function from here.

unittests/test_environments.py
```diff
@@ -166,7 +166,7 @@ def test_swap(self):
         self.assertTrue(self.environ_other.is_loaded)
 
 
-class TestProgEnvironment(unittest.TestCase):
+class _TestProgEnvironment:
```
vkarak: Haha, a workaround I forgot to remove. I should fix this file.

reframe/core/pipeline.py
```diff
@@ -930,35 +947,66 @@ def compile(self, **compile_opts):
...
+        """Wait for compilation phase to finish.
+
+        .. versionadded:: 2.13
+        """
+        self._build_job.wait()
         self.logger.debug('compilation finished')
 
+        # FIXME: this check is not reliable for certain scheduler backends
+        if self._build_job.exitcode != 0:
+            raise PipelineError('compilation failed')
```
vkarak: I should perhaps raise a more specific error here, e.g., `BuildError`. What do you think?

vkarak:
@jenkins-cscs retry daint kesch monch

vkarak:
@teojgo @victorusu When reviewing this PR, I suggest playing a bit with your tests and try to adapt them to the build system syntax. This way you can have a better feeling of how it works and whether there are things that you don't like and need to be done differently.

reframe/core/exceptions.py
vkarak: Actually, I need to keep the `CompilationError` and change it. Before we used to print the output of the compilation output. We still need to treat such errors specially. Perhaps, we could print the path to the compilation output.

vkarak:
The regression tests affected by this PR are a lot. I don't think it makes sense to include them in this one, cos it will become huge. I see two options: either we merge this to the master and cope with the deprecation warnings for some time in production or I will do a separate PR on this one (to be reviewed separately) that will update the affected tests. What do you think?

teojgo:
I think that everyone can take some regression tests that is mostly familiar with and adapt them to the new syntax/new build system. That way we can also practice both of these new features and completed the transformation as quickly as possible.

vkarak:
Removed WIP.

PR #353, opened 2018-07-11
[ci] Check added/modified regression tests w.r.t master

teojgo:
* Find the added/modified regression tests based on diversion from
  the master branch.

### DIFF
--- a/ci-scripts/ci-runner.bash
+++ b/ci-scripts/ci-runner.bash
@@ -156,9 +156,8 @@ if [ $CI_GENERIC -eq 1 ]; then
 elif [ $CI_TUTORIAL -eq 1 ]; then
     # Run tutorial checks
     # Find modified or added tutorial checks
-    tutorialchecks=( $(git log --name-status --oneline --no-merges -1 | \
-                   awk '/^[AM]/ { print $2 } /^R0[0-9][0-9]/ { print $3 }' | \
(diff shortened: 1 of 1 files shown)


ci-scripts/ci-runner.bash
```diff
@@ -156,9 +156,8 @@ if [ $CI_GENERIC -eq 1 ]; then
 elif [ $CI_TUTORIAL -eq 1 ]; then
     # Run tutorial checks
     # Find modified or added tutorial checks
-    tutorialchecks=( $(git log --name-status --oneline --no-merges -1 | \
-                   awk '/^[AM]/ { print $2 } /^R0[0-9][0-9]/ { print $3 }' | \
-                   grep -e '^tutorial/(?!config/).*\.py') )
+    tutorialchecks=( $(git diff origin/master...HEAD --name-only --oneline --no-merges | \
```
vkarak: Why are you using three dots here instead of two. If there's no difference, I prefer `..`.

teojgo: The three dots are different from two. The `...` checks diversion from the latest common point while the `..` checks without taking the common point into account. Thus, `...` is better because there could be a change in the master in the meanwhile.

vkarak:
@jenkins-cscs retry dom

PR #356, opened 2018-07-17
[ci] Add testing for python3.7 in travis

teojgo:


teojgo:
It seems that python3.7 is not yet supported in Travis.

PR #359, opened 2018-07-17
[feat] Evaluate and log all performance values before asserting them

vkarak:
This is a useful feature for tests that check multiple performance values.

Replaces #334.


vkarak:
@jenkins-cscs retry daint dom

PR #400, opened 2018-07-18
Module purge each script

trolleyman:
This adds an option to allow us to add a `module purge` before the `module load`s of each script.


trolleyman:
Ah sorry, wrong repo.

Issue #59, opened 2018-02-14
Unnecessary 'module unload' commands in the generated job script


vkarak:
Here is an example output for the `PrgEnv-pgi` environment:

```bash
#!/bin/bash -l
#SBATCH --job-name="rrtmgp_check_daint_gpu_PrgEnv-pgi"
#SBATCH --time=0:10:0
#SBATCH --ntasks=1
#SBATCH --partition=debug
#SBATCH --exclusive
#SBATCH --output=/scratch/snx3000/karakasv/reframe-tests/stage/gpu/rrtmgp_check/PrgEnv-pgi/rrtmgp_check.out
#SBATCH --error=/scratch/snx3000/karakasv/reframe-tests/stage/gpu/rrtmgp_check/PrgEnv-pgi/rrtmgp_check.err
#SBATCH --constraint=gpu
module load daint-gpu
module unload PrgEnv-cray
module load PrgEnv-pgi
module unload PrgEnv-cray
module load craype-accel-nvidia60
module unload PrgEnv-cray
module load cray-netcdf
module unload PrgEnv-cray
module unload pgi
module load pgi/17.7.0
```

I think the problem comes from the way we record the conflicts.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/565

vkarak:
I raised this bug's priority, since it seems to affect the setting the `modules` attribute and how the module commands are generated.

PR #401, opened 2018-07-19
[test] Adapt Paraview check for PE 18.07

### LINKED ISSUE #381: Adapt Paraview test to PE 18.07

teojgo:
* Adapt to new regression test syntax.

* Make a change in `coloredSphere.py` to load different python modules
  depending on the ParaView version.

Closes #381.


teojgo:
@jenkins-cscs retry all

cscs-checks/apps/paraview/paraview_check.py
```diff
@@ -1,40 +1,32 @@
...
 
         # Uncomment and adjust to load the ParaView module
         self.modules = ['ParaView']
 
-        # Reset sources dir relative to the SCS apps prefix
         self.executable = 'pvbatch'
         self.executable_opts = ['coloredSphere.py']
 
+        # Needed in order to swap from the default version of gcc
+        self.pre_run = ['module swap gcc/6.2.0 gcc/7.1.0']
```
vkarak: This is not too portable. I suggest specifying `gcc/7.1.0` in `self.modules`. Then ReFrame will generate the necessary commands.

vkarak: It seems that sth is wrong with the module command generation. This looks like related to #59. Can you add a note before `pre_run` referring to this issue?

cscs-checks/apps/paraview/paraview_check.py
```diff
@@ -1,40 +1,34 @@
...
         self.modules = ['ParaView']
 
-        # Reset sources dir relative to the SCS apps prefix
         self.executable = 'pvbatch'
         self.executable_opts = ['coloredSphere.py']
 
+        # NOTE: This is needed in order to swap from the default
+        # version of gcc until issue #59 is fixed. Then it should
+        # be moved to the `self.modules` definition.
+        self.pre_run = ['module swap gcc/6.2.0 gcc/7.1.0']
```
vkarak: After merging with master, this workaround should not be needed.

teojgo: Yes I had it ready and tested with #405

vkarak:
@jenkins-cscs retry daint

PR #404, opened 2018-07-24
[test] Adapt the petsc test for PE 18.07

### LINKED ISSUE #380: Adapt 'petsc_helloworld.py' test to PE 18.07

### LINKED ISSUE #388: Adapt PETSc test to the new build systems feature

teojgo:
* Adapt to new regression test syntax.

* Change the sanity check so that the last residual norm of the KSP
  method is checked for convergence.

* Adapt the source code to be in sync with the newest version of the
  petsc library.

Closes #380.
Closes #388.


teojgo:
This should also be adapted to address issue #388 once #340 is merged

vkarak:
I will merge this after the build systems are merged, so as to adapt it to the syntax as well.

cscs-checks/libraries/petsc/petsc_helloworld.py
```diff
@@ -1,22 +1,13 @@
 import filecmp
 import os
```
vkarak: These two imports are not needed anymore.

cscs-checks/libraries/petsc/petsc_helloworld.py
```diff
@@ -1,22 +1,13 @@
...
-class PetscPoisson2DCheck(RegressionTest):
-    def __init__(self, variant, **kwargs):
-        super().__init__('petsc_2dpoisson_%s' % variant,
-                         os.path.dirname(__file__), **kwargs)
+@rfm.parameterized_test(['dynamic'], ['static'])
+class PetscPoisson2DCheck(rfm.RegressionTest):
+    def __init__(self, variant):
+        super().__init__()
         self.descr = ('Compile/run PETSc 2D Poisson example with cray-petsc '
                       '(%s linking case)') % variant
```
vkarak: Better remove the "case" here.

cscs-checks/libraries/petsc/petsc_helloworld.py
```diff
@@ -28,21 +19,18 @@ def __init__(self, variant, **kwargs):
...
 
-        self.maintainers = ['WS', 'AJ']
-        self.tags = {'production'}
+        norms = sn.extractall(r'\s+\d+\s+KSP Residual norm\s+(?P<norm>\S+)',
+                              self.stdout, 'norm', float)
 
-        self.sanity_patterns = sanity_filecmp(
-            'petsc_poisson2d.out', 'petsc_poisson2d.ref')
+        # Check the final residual norm for convergence
+        self.sanity_patterns = sn.assert_lt(norms[-1], 1.0e-5)
```
vkarak: Since you don't do anything with the rest of the norms that you extract, you could use `sn.extractsingle()` and pass `item=-1`, so to extract the last match. The advantage of this is that it will raise a `SanityError` instead of `IndexError` in case of an out-of-bounds error, which means that it will be printed nicely in the failure report. See [here](https://eth-cscs.github.io/reframe/sanity_functions_reference.html#reframe.utility.sanity.extractsingle) for the full spec of the `extractsingle()` function.

vkarak:
@jenkins-cscs retry daint

PR #405, opened 2018-07-24
[bugfix] Fix bug in unloading of conflicting modules

### LINKED ISSUE #59: Unnecessary 'module unload' commands in the generated job script

teojgo:
* Fix bug which led to accumulation of module unloading.

* Use raw string to avoid warnings.

Fixes #59


vkarak:
@teojgo Good, I get it now.

reframe/core/environments.py
```diff
@@ -79,8 +79,9 @@ def load(self):
             if rt.modules_system.is_module_loaded(m):
                 self._preloaded.add(m)
 
-            self._conflicted += rt.modules_system.load_module(m, force=True)
-            for conflict in self._conflicted:
+            module_conflicts = rt.modules_system.load_module(m, force=True)
+            self._conflicted += module_conflicts
+            for conflict in module_conflicts:
```
vkarak: The loop variable should be simply `m`.

reframe/core/environments.py
```diff
@@ -79,8 +79,9 @@ def load(self):
             if rt.modules_system.is_module_loaded(m):
                 self._preloaded.add(m)
 
-            self._conflicted += rt.modules_system.load_module(m, force=True)
-            for conflict in self._conflicted:
+            module_conflicts = rt.modules_system.load_module(m, force=True)
+            self._conflicted += module_conflicts
```
vkarak: Move this assignment after the loop. Placed here, it kind of lets you expect that `self._conflicted` will be used just right after.

reframe/core/environments.py
```diff
@@ -79,8 +79,9 @@ def load(self):
             if rt.modules_system.is_module_loaded(m):
                 self._preloaded.add(m)
 
-            self._conflicted += rt.modules_system.load_module(m, force=True)
-            for conflict in self._conflicted:
+            module_conflicts = rt.modules_system.load_module(m, force=True)
```
vkarak: I'd name this simply `conflicted`.

reframe/core/environments.py
teojgo: its a nested loop and the outer one already uses `m`

PR #406, opened 2018-07-24
[ci] Temporarily disable running of checks on Daint

vkarak:
For the duration of the PE upgrade.


vkarak:
@jenkins-cscs retry daint

PR #409, opened 2018-07-26
[test] Adapt Tensorflow tests for PE 18.07

### LINKED ISSUE #373: Adapt Tensorflow tests to PE 18.07

teojgo:
* Use the new regression test syntax.

* Modify the command line options to be in sync with tensorflow 1.7.

Closes #373

### DIFF
--- a/cscs-checks/apps/tensorflow/tensorflow_check.py
+++ b/cscs-checks/apps/tensorflow/tensorflow_check.py
@@ -1,15 +1,11 @@
-import os
-import itertools
+import reframe as rfm
 import reframe.utility.sanity as sn
 
-from reframe.core.pipeline import RunOnlyRegressionTest
 
-
-class TensorFlowBaseTest(RunOnlyRegressionTest):
-    def __init__(self, model_name, **kwargs):
-        super().__init__('tensorflow_%s_check' % model_name,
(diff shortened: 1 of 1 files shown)


cscs-checks/apps/tensorflow/tensorflow_check.py
```diff
@@ -18,38 +14,43 @@ def __init__(self, model_name, **kwargs):
...
+                                ' --data_dir', '.', '--train_epochs',
+                                str(train_epochs)]
 
         self.sanity_patterns = sn.all([
             sn.assert_found(r'INFO:tensorflow:Finished evaluation at',
                             self.stderr),
             sn.assert_gt(sn.extractsingle(
                 r"Evaluation results:\s+\{.*'accuracy':\s+(?P<accuracy>\S+)"
-                r"(?:,|\})", self.stdout, 'accuracy', float), 0.99)
+                r"(?:,|\})", self.stdout, 'accuracy', float, -1), 0.99)
```
vkarak: What was the difference between the first accuracy and the last accuracy value. I mean which one is more correct?

teojgo: Now I need to take the final value because there are many accuracy numbers printed throughout the calculation.

PR #410, opened 2018-07-26
[bugfix] Filter out empty strings before writing preamble

### LINKED ISSUE #407: Empty lines in built job scripts
For example for the petsc example the job script contains the following:

```
#!/bin/bash -l
#SBATCH --job-name="rfm_PetscPoisson2DCheck_dynamic_job"
#SBATCH --time=0:10:0
#SBATCH --ntasks=16
#SBATCH --ntasks-per-node=8








#SBATCH --output=/users/manitart/reframe_dev_public/stage/dom/gpu/PrgEnv-cray/PetscPoisson2DCheck_dynamic/rfm_PetscPoisson2DCheck_dynamic_job.out
#SBATCH --error=/users/manitart/reframe_dev_public/stage/dom/gpu/PrgEnv-cray/PetscPoisson2DCheck_dynamic/rfm_PetscPoisson2DCheck_dynamic_job.err

#SBATCH --constraint=gpu
module load daint-gpu
module load PrgEnv-cray
module load cray-petsc
srun ./PetscPoisson2DCheck_dynamic -da_grid_x 4 -da_grid_y 4 -ksp_monitor
```

This seems to have been introduced by #340 since in the `finalize` method of `ShellScriptGenerator` when joining `\n` with a list that contain empty strings, empty lines are introduced.

teojgo:
* Use `filter` to filter out empty strings before finalizing shell.

* User raw string to suppress invalid escape sequence warning.

Fixes #407

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -121,7 +121,7 @@ class SlurmJob(sched.Job):
             else:
                 preamble.append(opt)
 
-        # Filter out empty statements before returning"
+        # Filter out empty statements before returning
         return list(filter(None, preamble))
 
     def _run_command(self, cmd, timeout=None):
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -121,7 +121,8 @@ class SlurmJob(sched.Job):
             else:
                 preamble.append(opt)
 
-        return preamble
+        # Filter out empty statements before returning"
+        return list(filter(None, preamble))
 
     def _run_command(self, cmd, timeout=None):
         """Run command cmd and re-raise any exception as a JobError."""
--- a/reframe/core/shell.py
+++ b/reframe/core/shell.py
@@ -4,7 +4,7 @@
 import abc
 
 
-_RFM_TRAP_ERROR = '''
+_RFM_TRAP_ERROR = r'''
 _onerror()
 {
     exitcode=$?
@@ -98,8 +98,10 @@ class ShellScriptGenerator:
         self.write(s, 'body')
 
     def finalize(self):
-        ret = '\n'.join([self.shebang, *self._prolog,
-                         *self._body, *self._epilog])
+        # Here empty strings are filtered out before joining
+        stripped_shell = filter(None, [self.shebang, *self._prolog,
+                                       *self._body, *self._epilog])
+        ret = '\n'.join(stripped_shell)
 
         # end with a new line
         if ret:
--- a/reframe/core/shell.py
+++ b/reframe/core/shell.py
@@ -98,10 +98,8 @@ class ShellScriptGenerator:
         self.write(s, 'body')
 
     def finalize(self):
-        # Here empty strings are filtered out before joining
-        stripped_shell = filter(None, [self.shebang, *self._prolog,
-                                       *self._body, *self._epilog])
-        ret = '\n'.join(stripped_shell)
+        ret = '\n'.join([self.shebang, *self._prolog,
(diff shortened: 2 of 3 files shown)


reframe/core/shell.py
```diff
@@ -4,7 +4,7 @@
 import abc
 
 
-_RFM_TRAP_ERROR = '''
+_RFM_TRAP_ERROR = r'''
```
vkarak: Question: Which python version are you running and you're getting the string warnings?

reframe/core/shell.py
```diff
@@ -98,8 +98,10 @@ def write_body(self, s):
         self.write(s, 'body')
 
     def finalize(self):
-        ret = '\n'.join([self.shebang, *self._prolog,
-                         *self._body, *self._epilog])
+        # Here empty strings are filtered out before joining
```
vkarak: Write it as "Filter out empty statements before returning". And return directly, without using the `stripped_shell` variable.

vkarak: Nonetheless, have you found the place that actually generated these empty instructions? Perhaps, we could fix the problem there.

vkarak: I would also expand the shell script generator tests to check this stripping.

reframe/core/shell.py
teojgo: Pycodestyle shows me the warnings regarding invalid escape sequences. You must run with warnings enabled in order to take then from Python.

reframe/core/shell.py
vkarak: As discussed offline, I think it's better to fix that in the Slurm job script generator.

reframe/core/schedulers/slurm.py
```diff
@@ -121,7 +121,8 @@ def emit_preamble(self):
             else:
                 preamble.append(opt)
 
-        return preamble
+        # Filter out empty statements before returning"
```
vkarak: Remove the double quotes at the end.

unittests/test_schedulers.py
```diff
@@ -362,6 +362,16 @@ def test_prepare_without_smt(self):
         with open(self.testjob.script_filename) as fp:
             self.assertIsNotNone(re.search(r'--hint=nomultithread', fp.read()))
 
+    def test_no_empty_lines_in_preamble(self):
```
vkarak: This test is not specific to the Slurm backend. Instead it must be part of the base `_TestJob`. I am also not sure that you have to set the different attributes to `None` explicitly, since the default `setUp()` method sets up a minimal job.

unittests/test_schedulers.py
```diff
@@ -166,6 +166,11 @@ def test_poll_before_submit(self):
         self.testjob.prepare(self.commands, self.environs)
         self.assertRaises(JobNotStartedError, self.testjob.finished)
 
+    def test_no_empty_lines_in_preamble(self):
```
vkarak: This should be part of the `_TestJob` base class. Subclasses need not re-implement this, since it should be the same everywhere.

unittests/test_schedulers.py
```diff
@@ -166,6 +166,11 @@ def test_poll_before_submit(self):
...
+        self.setup_job()
```
vkarak: You should not call `setup_job()` here, because this sets everything and most probably the following test will have no meaning. I suggest omitting completely this call. In this case, `self.testjob` will be minimal.

teojgo:
@jenkins-cscs retry all

vkarak:
@jenkins-cscs retry dom

PR #420, opened 2018-08-15
[test] Adapt stream microbenchmark test to build systems

### LINKED ISSUE #391: Adapt STREAM test to the new build systems feature

teojgo:
* Adapt the stream microbenchmark regression test to
  use build systems.

* Use the new regression test syntax.

Closes #391


cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -91,16 +90,9 @@ def setup(self, partition, environ, **job_opts):
         self.reference = self.stream_bw_reference[self.current_environ.name]
         # On SLURM there is no need to set OMP_NUM_THREADS if one defines
         # num_cpus_per_task, but adding for completeness and portability
-        self.current_environ.variables['OMP_NUM_THREADS'] = \
-            str(self.num_cpus_per_task)
+        self.current_environ.variables['OMP_NUM_THREADS'] = str(
+            self.num_cpus_per_task)
```
vkarak: I don't like setting the `variables` attribute of the current environment directly. Besides, I don't think that they will be loaded. You should set `self.variables` instead and call `super().setup(...)` later.

vkarak: Same applies for the `OMP_PROC_BIND`.

vkarak: Everything fits in a single line here.

vkarak:
@teojgo The failure on Kesch is expected. I will merge this one.

vkarak:
@jenkins-cscs retry daint dom

PR #425, opened 2018-08-21
[test] Adapt ddt regression test to PE 18.07

### LINKED ISSUE #368: Adapt 'ddt.py' to PE 18.07

teojgo:
* Adapt to build systems.

* Adapt to new syntax.

* Remove unused Makefiles

* Add build dependencies in the F90 Makefile in order to
  make the Fortran modules available.

Closes #368 
Addresses #397


cscs-checks/tools/profiling_and_debugging/ddt.py
```diff
@@ -120,12 +117,11 @@ def __init__(self, lang, extension):
             '--trace-at _jacobi-cuda-kernel.cu:111,residue'
         ]
 
-        self.flags += ['-DUSE_MPI']
-        self.flags += ['-D_CSCS_ITMAX=5']
+        self.build_system.cflags = ['-DUSE_MPI', '-D_CSCS_ITMAX=5']
```
vkarak: These are `cppflags` (i.e., preprocessor flags). Here, you will only set the `CFLAGS` in the `Makefile` invocation.

vkarak:
@teojgo There seems to be a bug in the test for Kesch.

PR #436, opened 2018-09-03
[ci] Remove monch from the machinesList

teojgo:


vkarak:
@jenkins-cscs retry dom

PR #439, opened 2018-09-04
[test] Adapt perftools to use build systems

### LINKED ISSUE #397: Adapt profiling and debugger tests to the new build systems feature

teojgo:
* Adapt the perftools test to use build systems.

* Use new regression test syntax.

Solves the test failure due to the following error:
`attempt to set an aggregate field of different type. Required typespec: list[str]`

This is the last regression test that addresses issue #397.

Closes #397

### DIFF
--- a/cscs-checks/tools/profiling_and_debugging/perftools.py
+++ b/cscs-checks/tools/profiling_and_debugging/perftools.py
@@ -1,13 +1,13 @@
 import os
+
+import reframe as rfm
 import reframe.utility.sanity as sn
 
-from reframe.core.pipeline import RegressionTest
 
-
-class PerftoolsCheck(RegressionTest):
-    def __init__(self, lang, **kwargs):
-        super().__init__('perftools_check_' + lang.replace('+', 'p'),
-                         os.path.dirname(__file__), **kwargs)
+@rfm.parameterized_test(*([lang] for lang in ['C', 'Cpp', 'F90', 'Cuda']))
+class PerftoolsCheck(rfm.RegressionTest):
+    def __init__(self, lang):
+        super().__init__()
         if lang == 'Cuda':
             self.valid_systems = ['daint:gpu', 'dom:gpu']
         else:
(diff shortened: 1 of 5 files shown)


cscs-checks/tools/profiling_and_debugging/perftools.py
```diff
@@ -22,48 +22,51 @@ def __init__(self, lang, **kwargs):
             # FIXME: PGI Fortran is hanging after completion
             self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu']
 
+        # NOTE: Reduce time limit because for PrgEnv-pgi even if the output
+        # is correct, the batch job uses all the time.
```
vkarak: Should we open a ticket to ops for that?

kraushm: JG also suggested to open a ticket earlier today. (yet, I could imagine he meant the tools-rt and not immediately ops)

cscs-checks/tools/profiling_and_debugging/perftools.py
```diff
@@ -22,48 +22,51 @@ def __init__(self, lang, **kwargs):
...
-    for lang in ['C', 'C++', 'F90', 'Cuda']:
-        ret.append(PerftoolsCheck(lang, **kwargs))
+        self.maintainers = ['MK', 'JG']
+        self.tags = {'production'}
 
-    return ret
+    def setup(self, environ, partition, **job_opts):
+        super().setup(environ, partition, **job_opts)
+        flags = self.prgenv_flags[self.current_environ.name]
+        flags += ['-g', '-D_CSCS_ITMAX=1', '-DUSE_MPI']
```
vkarak: I would use the `build_system.cppflags` for the `-D` definitions and as for the `-g`, I would put it in each entry in the `prgenv_flags` dictionary. This way, you will have all the compilation flags in a single place in the `__init__()`.

PR #442, opened 2018-09-06
[test] Update OSU microbenchmark test to new syntax and build system

### LINKED ISSUE #327: Update G2G latency tests
The current latency test does not really measure latency, but it measures bandwidth.
Instead of using `osu_bw` we should be using `osu_latency`. Especially, because the latter uses a different algorithm than the former.

### LINKED ISSUE #390: Adapt OSU tests to the new build systems feature

rsarm:
Closes #327 
Closes #390


cscs-checks/microbenchmarks/osu/osu_tests.py
```diff
@@ -114,23 +111,21 @@ def setup(self, partition, environ, **job_opts):
```
vkarak: The `setup()` method here obfuscates a bit the code. It does a setup for GPU benchmarks practically and the base class is supposed to be agnostic. I suggest moving this code in the constructor of the corresponding GPU benchmark subclasses, where you will only check against the `self.current_system` and not against the partition. Also for Daint/Dom you should load `craype-accel-nvidia60` instead of `cudatoolkit`.

cscs-checks/microbenchmarks/osu/osu_tests.py
```diff
@@ -1,17 +1,16 @@
...
-from reframe.core.pipeline import RegressionTest
 
-
-class AlltoallBaseTest(RegressionTest):
-    def __init__(self, name, **kwargs):
-        super().__init__(name,
-                         os.path.dirname(__file__), **kwargs)
+@rfm.parameterized_test([['production']])
+class AlltoallTest(rfm.RegressionTest):
+    def __init__(self, tags):
```
vkarak: I didn't mean it like that. The tests should take a single *string* argument of whether it is production or not. Then you set the tags as usual:

```python
@rfm.parameterized_test(['production'])
class AlltoallTest(rfm.RegressionTest):
    def __init__(self, variant):
        self.tags = {variant}
```

cscs-checks/microbenchmarks/osu/src/Makefile_p2p
```diff
@@ -0,0 +1,25 @@
...
+OBJS_BW = osu_util.o osu_bw.o
+OBJS_LT = osu_util.o osu_latency.o
+
+OBJS := $(SRCS:.c=.o)
+
+$(OBJS):
+	$(CC) $(CPPFLAGS) -o $(@) -c $(@:.o=.c)
+
+p2p_osu_bw: $(OBJS_BW)
+	$(CC) $(CPPFLAGS) $(LDFLAGS) $(OBJS_BW) -o $(@)
```
vkarak: You should still pass the `CXXFLAGS` if you want to be completely correct. `CPPFLAGS` are the *preprocessor* flags, i.e., the `-D` and `-I` options mostly.

cscs-checks/microbenchmarks/osu/osu_tests.py
```diff
@@ -226,17 +199,24 @@ def __init__(self, **kwargs):
...
-        self.valid_systems = ['daint:gpu', 'dom:gpu', 'kesch:cn']
         if self.current_system.name == 'kesch':
-            self.valid_prog_environs = ['PrgEnv-cray']
+            self.variables = {'MV2_USE_CUDA': '1'}
+
+        self.build_system.cppflags = ['-D_ENABLE_CUDA_']
 
+
+@rfm.simple_test
+class GPGLatencyTest(P2PBaseTest):
```
vkarak: `GPG` -> `G2G`.

vkarak:
@jenkins-cscs retry dom kesch

vkarak:
@jenkins-cscs retry dom

PR #448, opened 2018-09-12
[bugfix] Fix redundant export statements

### LINKED ISSUE #423: The environment variables are exported twice in the job script

kraushm:
In some cases (e. g. GPUDirectACC test), an export statement was repeated
in the job submission script. The modifications check upfront, whether
the same environment variable has been already defined and omit its definition,
if necessary.

Fixes #423.


kraushm:
The list `ret` used to point to load statements in the function def `emit_load_commands(self):` in core/environments.py:
```
...
         ret = self._load_stmts
...

     for k, v in self._variables.items():
         ret.append('export %s=%s' % (k, v))

     return ret
```
Changing the code to `ret = list(self._load_stmts)`, and thus creating a local copy, prevented the occurrence of the redundant export statements found, e. g., in the test gpu_direct_acc.py job script.


#423 led to a revision of the environments interface (see #456) and any further modification will be addressed in a new PR.

reframe/core/environments.py
```diff
@@ -121,7 +121,8 @@ def unload(self):
     def emit_load_commands(self):
         rt = runtime()
         if self.is_loaded:
-            ret = self._load_stmts
+        # FIXME: This is a workaround for issue #423; revisit environment interface
```
vkarak: Please align the comment correctly. Also replace the `revisit [...]` with `the environment interface must be revisited (see issue #456)`

PR #459, opened 2018-09-14
[test] Increase GROMACS sanity check tolerance to 0.1%

victorusu:


vkarak:
@victorusu Is the PR title correct? I had the impression the old threshold was at 0.05%.

PR #461, opened 2018-09-17
[test] Adapt cuda nvml check to build systems

teojgo:


teojgo:
@jenkins-cscs retry all

PR #464, opened 2018-09-18
[feat] Use different stage and output directories for different retries of the same test

### LINKED ISSUE #403: When a test succeeds in retries, the original failure information is lost
We should perhaps prefix the stage directories differently for each retry. It needs some investigation on how this could be done nicely.

omlins:
Includes the migration of the `current_run` variable to the class RuntimeContext and related changes.
Closes #403.

### DIFF
--- a/reframe/core/runtime.py
+++ b/reframe/core/runtime.py
@@ -128,8 +128,9 @@ class HostResources:
             return os.path.join(self.prefix, 'output' + self._run_suffix(),
                                 self.timestamp)
         else:
-            return os.path.join(os.path.normpath(self.outputdir) +
-                                self._run_suffix(), self.timestamp)
+            return os.path.join(self.outputdir + self._run_suffix(),
+                                self.timestamp)
+
 
     @property
     def stage_prefix(self):
@@ -138,8 +139,8 @@ class HostResources:
             return os.path.join(self.prefix, 'stage' + self._run_suffix(),
                                 self.timestamp)
         else:
-            return os.path.join(os.path.normpath(self.stagedir) +
-                                self._run_suffix(), self.timestamp)
+            return os.path.join(self.stagedir + self._run_suffix(),
+                                self.timestamp)
 
     @property
     def perflog_prefix(self):
@@ -218,10 +219,7 @@ class RuntimeContext:
 
     @property
     def current_run(self):
-        """The current run.
-
-        :type: `integer`
-        """
+        # Not publicly documented
         return self._current_run
 
     @property
--- a/reframe/core/runtime.py
+++ b/reframe/core/runtime.py
@@ -219,7 +219,6 @@ class RuntimeContext:
 
     @property
     def current_run(self):
-        # Not publicly documented
         return self._current_run
 
     @property
--- a/reframe/core/runtime.py
+++ b/reframe/core/runtime.py
@@ -113,6 +113,10 @@ class HostResources:
         os.makedirs(ret, exist_ok=True)
         return ret
 
+    def _run_suffix(self):
+        current_run = runtime().current_run
+        return '_%s' % current_run if current_run > 0 else ''
+
     @property
     def timestamp(self):
         return self._timestamp.strftime(self.timefmt) if self.timefmt else ''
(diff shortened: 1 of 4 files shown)


reframe/core/runtime.py
```diff
@@ -204,6 +213,17 @@ def mode(self, name):
...
             raise ConfigError('unknown execution mode: %s' % name) from None
 
+    def next_run(self):
+        self._current_run += 1
+
+    @property
+    def current_run(self):
+        """The current run.
+
+        :type: `integer`
```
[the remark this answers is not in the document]
vkarak: The documentation here must be expanded a bit. Whatever you put here is exposed to the user API. So what is a run, how it can be useful for a regression test?

reframe/core/runtime.py
```diff
@@ -204,6 +213,17 @@ def mode(self, name):
         except KeyError:
             raise ConfigError('unknown execution mode: %s' % name) from None
 
+    def next_run(self):
+        self._current_run += 1
+
+    @property
+    def current_run(self):
```
vkarak: What about calling this property `runid`? I think it is better from the user's point of view. It will also help explaining it in the documentation.

reframe/core/runtime.py
```diff
@@ -204,6 +213,17 @@ def mode(self, name):
...
+        """The current run.
+
+        :type: `integer`
+        """
+        return self._current_run
```
vkarak: Rename this to `_runid` as well.

reframe/frontend/statistics.py
```diff
@@ -9,21 +9,15 @@ class TestStats:
...
-    def current_run(self):
-        return self._current_run
-
-    def next_run(self):
-        self._current_run += 1
-        self._tasks.append([])
-
     def add_task(self, task):
-        self._tasks[self._current_run].append(task)
+        current_run = rt.runtime().current_run
```
vkarak: I think `runid` fits much better as name here.

reframe/frontend/statistics.py
```diff
@@ -9,21 +9,15 @@ class TestStats:
...
+        if current_run == len(self._tasks):
+            self._tasks.append([])
```
vkarak: Add a blank line here.

unittests/test_policies.py
```diff
@@ -132,6 +133,9 @@ def test_system_exit_within_test(self):
         stats = self.runner.stats
         self.assertEqual(1, stats.num_failures())
 
+    # Retries tests are executed in a different runtime as they modify
+    # the global rt.runtime().current_run what makes subsequent tests fail
```
vkarak: This comment is not very accurate. You should say that the policy tests need a freshly initialized runtime, that's why we need to switch before every test.

reframe/core/runtime.py
vkarak: Having looked the whole PR, I think we can do the following:

```python
class RuntimeContext:
    def __init__(...):
        self._runid = 0

    @property
    def runid(self):
        """Fully documented, since it is meant to be used also be regression tests"""
        return self._runid

    @property
    def run_count(self):
        # Not publicly documented
        return self._runid
```

Then, in places where conceptually you need the count, you call `run_count`, whereas in places where the `id` is more meaningful you call `runid`.

vkarak:
Another thing is to augment the unit tests to check that the directory suffixes are created as expected.

reframe/core/runtime.py
```diff
@@ -121,17 +125,21 @@ def timestamp(self):
     def output_prefix(self):
         """The output prefix directory of ReFrame."""
         if self.outputdir is None:
-            return os.path.join(self.prefix, 'output', self.timestamp)
+            return os.path.join(self.prefix, 'output' + self._run_suffix(),
+                                self.timestamp)
         else:
-            return os.path.join(self.outputdir, self.timestamp)
+            return os.path.join(os.path.normpath(self.outputdir) +
```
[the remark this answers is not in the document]
omlins: I did not realize that is already guaranteed that `self.outputdir` and `self.stagedir` are normalized. I will remove it then.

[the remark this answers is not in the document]
omlins: Done.

reframe/core/runtime.py
```diff
@@ -218,10 +219,7 @@ def next_run(self):
 
     @property
     def current_run(self):
-        """The current run.
-
-        :type: `integer`
-        """
+        # Not publicly documented
```
vkarak: This comment is not actually needed.

PR #466, opened 2018-09-19
[test] Adapt trilinos regression test to PE18.08 and build systems

### LINKED ISSUE #363: Adapt 'trilinos_compile_run.py' to PE 18.07

### LINKED ISSUE #385: Adapt Trilinos test to the new build systems feature

teojgo:
The test fails for `PrgEnv-cray` since the executable gibes a segmentation fault for cray-trilinos/12.12.1.1

Fixes #385 
Fixes #363

### DIFF
--- a/cscs-checks/libraries/math/trilinos_compile_run.py
+++ b/cscs-checks/libraries/math/trilinos_compile_run.py
@@ -1,28 +1,13 @@
 import os
+
+import reframe as rfm
 import reframe.utility.sanity as sn
 
-from reframe.core.pipeline import RegressionTest
 
-
-class TrilinosTest(RegressionTest):
-    def __init__(self, linkage, **kwargs):
-        super().__init__('trilinos_compile_run_%s' % linkage,
-                         os.path.dirname(__file__), **kwargs)
-
-        self.flags = ' -DHAVE_MPI -DEPETRA_MPI -lparmetis -%s ' % linkage
-        self.prgenv_flags = {
-            'PrgEnv-cray': ' -homp -hstd=c++11 -hmsglevel_4 ',
-            'PrgEnv-gnu': ' -fopenmp -std=c++11 -w -fpermissive ',
-            'PrgEnv-intel': ' -qopenmp -w -std=c++11 ',
-            'PrgEnv-pgi': ' -mp -w '
-        }
-        self.descr = 'Trilinos ' + linkage.capitalize()
-        self.sourcepath = 'example_AmesosFactory_HB.cpp'
-        #self.sourcepath = 'trilinos_compile_run.cpp'
-        self.input_file = os.path.join(self.current_system.resourcesdir,
-                                       'Trilinos', 'trilinos_compile_run.rua')
-
-        self.executable_opts = self.input_file.split()
+@rfm.parameterized_test(['static'], ['dynamic'])
+class TrilinosTest(rfm.RegressionTest):
+    def __init__(self, linkage):
+        super().__init__()
         self.valid_systems = ['daint:gpu', 'daint:mc',
                               'dom:gpu', 'dom:mc']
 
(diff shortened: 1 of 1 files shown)


cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -1,40 +1,42 @@
...
+class TrilinosTest(rfm.RegressionTest):
+    def __init__(self, linkage):
+        super().__init__()
         self.valid_systems = ['daint:gpu', 'daint:mc',
                               'dom:gpu', 'dom:mc']
 
-        # Removed CRAY env in dynamic because of CrayBug/809265
-        if linkage == 'dynamic':
+        # NOTE: PrgEnv-cray in dynamic does not work because of CrayBug/809265
+        # PrgEnv-cray seg faults for cray-trilinos/12.12.1.1
```
vkarak: I don't think this note has to be in this test, since you are not doing anything special for it, do you? If the bug is fixed, the test will pass automatically. I suggest putting this note in the PR description instead.

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -1,40 +1,42 @@
...
             self.valid_prog_environs = ['PrgEnv-gnu', 'PrgEnv-intel']
-        elif linkage == 'static':
+        else:
             self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu',
                                         'PrgEnv-intel']
 
+        self.build_system = 'SingleSource'
+        self.build_system.ldflags = ['-%s' % linkage]
+        self.build_system.cppflags = ['-DHAVE_MPI', '-DEPETRA_MPI']
+        self.build_system.cxxflags = ['-lparmetis']
```
vkarak: This belongs to `ldflags`.

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -1,40 +1,42 @@
...
+        self.prgenv_flags = {
+            'PrgEnv-cray': ['-homp', '-hstd=c++11', '-hmsglevel_4'],
+            'PrgEnv-gnu': ['-fopenmp', '-std=c++11', '-w', '-fpermissive'],
+            'PrgEnv-intel': ['-qopenmp', '-w', '-std=c++11'],
+            'PrgEnv-pgi': ['-mp', '-w']
+        }
+        self.sourcepath = 'example_AmesosFactory_HB.cpp'
+        input_file = os.path.join(self.current_system.resourcesdir,
+                                  'Trilinos', 'trilinos_compile_run.rua')
+        self.executable_opts = input_file.split()
```
vkarak: This split here is kind of strange, isn't it? I don't think it's needed: `[input_file]` should be fine.

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -1,40 +1,42 @@
...
+
+        # NOTE: default cray-trilinos module in PE/18.08 does not work
         self.modules = ['cray-mpich', 'cray-hdf5-parallel',
-                        'cray-tpsl', 'cray-trilinos']
+                        'cray-tpsl', 'cray-trilinos/12.12.1.1']
```
vkarak: Why are we not using the default?

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -45,18 +47,10 @@ def __init__(self, linkage, **kwargs):
         self.tags = {'production'}
 
     def setup(self, partition, environ, **job_opts):
+        prgenv_flags = self.prgenv_flags[environ.name]
+        self.build_system.cxxflags += prgenv_flags
         if environ.name == 'PrgEnv-intel':
             # CrayBug/836679
```
vkarak: Is this still valid?

teojgo: Yes, gcc is needed

cscs-checks/libraries/math/trilinos_compile_run.py
teojgo: Because the default is 12.12.1.0 and it does not work with any PrgEnv

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -12,17 +12,15 @@ def __init__(self, linkage):
                               'dom:gpu', 'dom:mc']
 
         # NOTE: PrgEnv-cray in dynamic does not work because of CrayBug/809265
```
vkarak: This note is a bit confusing. You are excluding `PrgEnv-cray` from the `static` variant, but you include it in the `dynamic`, where it fails. I get your point, but I think this note does not belong here. It should be part of the PR description only. As soon as the Cray bug is fixed, this test (without any modification) should pass, shouldn't it? The note will become then obsolete and we would have to fix the test again..

teojgo: This is an error that I introduced in this PR. The dynamic one was not working in the previous PE versions as well.

PR #474, opened 2018-09-20
[bugfix] Remove time limit from build jobs

### LINKED ISSUE #465: Build jobs are killed after 10 minutes
It would be nice to have an option to set the time limit for the build job.

vkarak:
This PR changes basically the `Job` class by changing the default time limit.

- Time limit now defaults to `None`, meaning that the backends should
  not impose any time limit on the running job.
- Build jobs do not have any time limit.

Fixes #465.


teojgo:
@jenkins-cscs retry all

PR #480, opened 2018-09-27
[bugfix] Remove the validation of the name of the test's base classes

### LINKED ISSUE #451: ReFrame silently ignores wrong test files
The following wrong test file is simply ignored by ReFrame:

```python
import reframe as rfm


@rfm.simple_test
class SedTest(rfm.CompileOnlyTest):
    def __init__(self):
        super().__init__()
        self.valid_systems = ['*']
        self.sourcesdir = 'https://github.com/mirror/sed'
        self.prebuild_cmd = ['autoreconf -vfi']
        self.build_system = 'Autotools'
        self.build_system.builddir = 'build'
        self.sanity_patterns = sn.assert_found('.*', self.stdout)
```

```
Command line: ./bin/reframe -C config/tresa.py -c ../misc/rfm-checks/sed.py -l
Reframe version: 2.14-dev2
Launched by user: karakasv
Launched on host: tresa.local
Reframe paths
=============
    Check prefix      :
    Check search path : '../misc/rfm-checks/sed.py'
    Stage dir prefix     : /Users/karakasv/Repositories/reframe/stage/
    Output dir prefix    : /Users/karakasv/Repositories/reframe/output/
    Perf. logging prefix : /Users/karakasv/Repositories/reframe/perflogs
List of matched checks
======================
Found 0 check(s).
```

The problem in this file is that the base class for compile-only tests is `CompileOnlyRegressionTest` not `CompileOnlyTest`. I think it has something to do with the initial parsing of the source files.

rsarm:
Closes #451.

### DIFF
--- a/reframe/frontend/loader.py
+++ b/reframe/frontend/loader.py
@@ -21,7 +21,7 @@ class RegressionCheckValidator(ast.NodeVisitor):
 
     @property
     def valid(self):
-        return self._has_import and self._has_regression_test
+        return self._has_import
 
     def visit_Import(self, node):
         for m in node.names:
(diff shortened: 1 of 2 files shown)


reframe/frontend/loader.py
```diff
@@ -29,7 +29,9 @@ def visit_Import(self, node):
                 self._has_import = True
 
     def visit_ImportFrom(self, node):
-        if node.module.startswith('reframe'):
+        if not node.module:
+            self._has_import = False
+        elif node.module.startswith('reframe'):
```
vkarak: I would simply test as follows:

```python
if node.module is not None and node.module.startswith('reframe'):
    self._has_import = True
```

unittests/resources/checks/bad/relative_imports.py
```diff
@@ -0,0 +1,4 @@
+from . import invalid_iterable
```
vkarak: Can you put a small comment at the beginning of this fail explaining its purpose?

PR #485, opened 2018-09-28
[refactor] Pass `sched_*` options to `options` list in slurm

teojgo:
* Prepend the 'sched_*' options to `options` list for the slurm
  scheduler.

* Add method `guess_num_tasks` to `Job` abstract class.


teojgo:
@jenkins-cscs retry dom kesch

unittests/test_schedulers.py
```diff
@@ -472,6 +477,11 @@ def test_submit_timelimit(self):
         # Skip this test for PBS, since we the minimum time limit is 1min
         self.skipTest("PBS minimum time limit is 60s")
 
+    def test_guess_num_tasks(self):
+        self.testjob._num_tasks = 0
+        with self.assertRaises(JobError):
+            self.testjob.guess_num_tasks()
+
```
vkarak: I think this test should go at the base class for two reasons:
1. It belongs to `Job`'s public interface.
2. All implementations must throw `JobError` by default. If a backend treats this thing differently, it should simply override it and specialize it accordingly.

vkarak: You still need to remove this `test_guess_num_tasks()` from here. It's only relevant at the base class.

PR #487, opened 2018-09-28
[feat] List only the checks that are supported on the current system

### LINKED ISSUE #472: When listing tests, print only those that are actually supported for the current system
Or the system specified with the `--system` option.

rsarm:
Closes #472

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -19,12 +19,6 @@ from reframe.frontend.loader import RegressionCheckLoader
 from reframe.frontend.printer import PrettyPrinter
 
 
-def list_supported_systems(systems, printer):
-    printer.info('List of supported systems:')
-    for s in systems:
-        printer.info('    %s' % s)
-
-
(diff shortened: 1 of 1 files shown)


reframe/frontend/cli.py
```diff
@@ -390,6 +384,15 @@ def filter_prgenv(c):
         if not options.skip_prgenv_check:
             checks_matched = filter(filter_prgenv, checks_matched)
 
+        # Filter checks by system
+        def filter_system(c):
+            current_systems = ['%s:%s' % (rt.system.name, p)
+                               for p in rt.system.partitions]
+            return any([c.supports_system(s) for s in current_systems])
```
vkarak: This function can be simplified:

```python
def filter_system(c):
    return any([c.supports_system(s) for s.fullname in rt.system.partitions])
```

vkarak:
@rsarm I will merge this PR, but we need to update the documentation. I have created a new issue for that: #490.

PR #493, opened 2018-10-04
[WIP] Investigate CI failure on Dom

vkarak:


vkarak:
@jenkins-cscs retry dom

vkarak:
There is a problem with the system configuration. Follow up ticket [here](https://webrt.cscs.ch/Ticket/Display.html?id=33309).

PR #494, opened 2018-10-04
[refactor] Better support for dynamic type checking and field validation

### LINKED ISSUE #47: Improve attribute validation
Currently, we can only define aggregate types of builtin types. For example, it is not possible to create a "list of alphanumeric fields". The solution would be to separate the type information from the field. Here's a proof-of-concept:

```python
import re

# Type generators
class Type:
    def __init__(self, typespec):
        self.typespec = typespec


    def check(self, value):
        if not isinstance(value, self.typespec):
            raise TypeError


class SequenceType(Type):
    def __init__(self, container_type, elem_type):
        self.container_type = container_type
        self.elem_type = elem_type

    def check(self, container):
        if not isinstance(container, self.container_type):
            raise TypeError

        for e in container:
            print(e, self.elem_type)
            self.elem_type.check(e)

# class MapType(Type):
#    pass


# Concrete types
class AlphanumericType:
    base_type = Type(str)

    def check(value):
        AlphanumericType.base_type.check(value)
        if not re.fullmatch('\w*', value, re.ASCII):
            raise TypeError



class Field(object):
    def __init__(self, fieldname, fieldtype):
        self.name = fieldname
        self.type = fieldtype


    def __get__(self, obj, objtype):
        return obj.__dict__[self.name]


    def __set__(self, obj, value):
        self.type.check(value)
        obj.__dict__[self.name] = value


class C:
    v = Field('v', SequenceType(list, AlphanumericType))


if __name__ == '__main__':
    c = C()
    c.v = [ 'foo', 'bar', 'foo/bar' ]

```

The above program will raise a `TypeError`. The whole logic of the type system is separate from the fields and we only need a generic field that carries on its name and its type. The type verification happens in the type system. Notice that concrete types can be simple static classes.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/244

vkarak:
This PR offers a type checking facility for aggregate types based on [PEP484](https://www.python.org/dev/peps/pep-0484/) type hints. It also adjusts and simplifies the data validation descriptors (`reframe.core.fields`) by using the new facility.

More specifically, the `reframe.utility.typecheck` module defines types for aggregate data structures, such as lists, dictionaries etc. that you can use with the [`isinstance()`](https://docs.python.org/3.6/library/functions.html#isinstance) builtin function to recursively check the type of all the elements of an aggregate data structure. Suppose you have a list of integers, such as `[1, 2, 3]`, the following checks should hold:

```python
from reframe.utility.typecheck import List

l = [1, 2, 3]
assert isinstance(l, List[int]) == True
assert isinstance(l, List[float]) == False
```

Aggregate types can be combined in an arbitrary depth, so that can type check any complex data structure:

```python
from reframe.utility.typecheck import Dict, List

d = {'a': [1, 2], 'b': [3, 4]}
assert isisntance(d, Dict) == True
assert isisntance(d, Dict[str, List[int]]) == True
```

This module offers the following aggregate types:
- `List[T]`: This corresponds to a list with elements of type `T`.
- `Set[T]`: This corresponds to a set with elements of type `T`.
- `Dict[K,V]`: This corresponds to a dictionary with keys of type `K` and values of type ``V``.
- `Tuple[T]`: This corresponds to a tuple with elements of type `T`.
- `Tuple[T1,T2,...,Tn]`: This corresponds to a tuple with `n` elements, whose types are exactly `T1`, `T2`, ..., `Tn` and in that order.
- `Str[patt]`: This corresponds to a string type, whose members are all the strings matching the regular expression `patt`.

This modules internally leverages metaclasses and the [`__isinstancecheck__()`](https://docs.python.org/3.6/reference/datamodel.html#class.__instancecheck__) method to customize the behaviour of the `isinstance()` builtin.
By implementing also the [`__getitem__`](https://docs.python.org/3.6/reference/datamodel.html#object.__getitem__) accessor method, it follows the look-and-feel of the type hints proposed in PEP484.
This method returns a new type that is a subtype of the base container type.
Using the facilities of [`abc.ABCMeta`](https://docs.python.org/3.6/library/abc.html#abc.ABCMeta), builtin types, such as `list`, `str` etc. are registered as subtypes of the base container types offered by this module.
The type hierarchy of the types defined in this module is the following (example shown for `List`, but it is analogous for the rest of the types):

```
          List
        /   |
       /    |
      /     |
    list  List[T]
```

In the above example `T` may refer to any type, so that `List[List[int]]` is an instance of `List`, but not an instance of `List[int]`.

Fixes #47.

### DIFF
--- a/reframe/utility/typecheck.py
+++ b/reframe/utility/typecheck.py
@@ -9,7 +9,7 @@ checks should hold:
 ::
     l = [1, 2, 3]
     assert isinstance(l, List[int]) == True
-    assert isinstance(l, List[false]) == False
+    assert isinstance(l, List[float]) == False
 
 
 Aggregate types can be combined in an arbitrary depth, so that can type check
@@ -197,9 +197,6 @@ class _MappingType(_TypeFactory):
 
 class StrType(_ContainerType):
     """A metaclass for type checking string types."""
-
-    def __instancecheck__(cls, inst):
-        return False
 
     def __instancecheck__(cls, inst):
         if not issubclass(type(inst), cls):
--- a/reframe/utility/typecheck.py
+++ b/reframe/utility/typecheck.py
@@ -1,0 +1,171 @@
+import abc
+import collections
+import re
+
+
+class _TypeFactory(abc.ABCMeta):
+    def register_subtypes(cls):
+        for t in cls._subtypes:
+            cls.register(t)
+
+
+class ContainerType(_TypeFactory):
+    def __init__(cls, name, bases, namespace):
+        super().__init__(name, bases, namespace)
+        cls._elem_type = None
+        cls._bases = bases
+        cls._namespace = namespace
+        cls.register_subtypes()
+
+    def __instancecheck__(cls, inst):
+        if not issubclass(type(inst), cls):
+            return False
+
+        if cls._elem_type is None:
+            return True
+
+        return all(isinstance(c, cls._elem_type) for c in inst)
+
+    def __getitem__(cls, elem_type):
+        if not isinstance(elem_type, type):
+            raise TypeError('{0} is not a valid type'.format(elem_type))
+
+        if isinstance(elem_type, tuple):
+            raise TypeError('invalid type specification for container type: '
+                            'expected ContainerType[elem_type]')
+
+        ret = ContainerType('%s[%s]' % (cls.__name__, elem_type.__name__),
+                            cls._bases, cls._namespace)
+        ret._elem_type = elem_type
+        ret.register_subtypes()
+        cls.register(ret)
+        return ret
+
+
+class TupleType(ContainerType):
+    def __instancecheck__(cls, inst):
+        if not issubclass(type(inst), cls):
+            return False
+
+        if cls._elem_type is None:
+            return True
+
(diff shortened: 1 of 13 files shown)


reframe/utility/typecheck.py
```diff
@@ -0,0 +1,244 @@
...
+This module defines types for aggregate data structures, such as lists,
+dictionaries etc. that you can use with the ``isinstance`` builtin function to
+recursively check of all the elements of an aggregate data structure.
+Suppose you have a list of integers, suchs as ``[1, 2, 3]``, the following
+checks should hold:
+
+::
+    l = [1, 2, 3]
+    assert isinstance(l, List[int]) == True
+    assert isinstance(l, List[false]) == False
```
teojgo: Do you mean `assert isinstance(l, List[bool]) == False`

vkarak: Yes!

vkarak:
@teojgo Do you think it would make sense to move the `fields` module under `reframe.utility` instead of `reframe.core`?

teojgo:
@vkarak I think that it better belongs to the `core` package.

vkarak:
Ok, I will leave it there then.

PR #509, opened 2018-10-11
[test] Add MCH halo exchange benchmark

### LINKED ISSUE #252: Add performance test for halo exchange (CPU + GPU)
MCH will provide the test case.

vkarak:
Also adjusted the collectives test, so as to combine them in a single test file, since the halo exchange test is almost identical to it.

Fixes #252.


vkarak:
@jenkins-cscs retry all

vkarak:
@jenkins-cscs retry daint

PR #512, opened 2018-10-12
[ci] Add C2SM environments on Kesch

### LINKED ISSUE #430: Create COSMO environments on Kesch/Daint
We should also verify that the existing MCH tests pass with these environments.

ajocksch:
Fixes #430.


config/cscs.py
```diff
@@ -288,7 +330,8 @@ class ReframeSettings:
                     'cc':  'gcc',
                     'cxx': 'g++',
                     'ftn': 'gfortran',
-                }
+                },
+
```
vkarak: This should be reverted to the old syntax.

vkarak:
@jenkins-cscs retry dom kesch

PR #514, opened 2018-10-12
[test] Add MCH CUDA stress test

### LINKED ISSUE #261: Add MCH CUDA stress test
MCH would like to have a computational kernel that could run on all node to detect GPU slow down.

vkarak:
Fixes #261.

### DIFF
--- a/cscs-checks/mch/src/cuda_stencil_test.cu
+++ b/cscs-checks/mch/src/cuda_stencil_test.cu
@@ -1,0 +1,216 @@
+#include "cuda.h"
+#include <cassert>
+#include <chrono>
+#include <iostream>
+#include <stdio.h>
+
+#define N 800
+#define TPB 16
+#define RED_NB 50
+
+#define BigN 4000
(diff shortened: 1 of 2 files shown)


vkarak:
@jenkins-cscs retry dom

cscs-checks/mch/src/cuda_stencil_test.cu
```diff
@@ -0,0 +1,218 @@
...
+}
+__global__ void bigstencil(int* in, int* out) {
+  int i = blockIdx.x * blockDim.x + threadIdx.x;
+  out[i] = in[i] + 2;
+}
+
+int main() {
+
+  // CSCS: Initialize CUDA runtime outside measurement region
+  cudaFree(0);
```
vkarak: @lxavier I have added this line to your benchmark, because the performance measurements were not reliable, since you were also counting the CUDA runtime initialization time. The initialization overhead is much more significant on CUDA 9 than in CUDA 8.

PR #529, opened 2018-10-18
[test] Remove read-only files attribute from the CP2K check

lucamar:
Removing read-only files (no symbolic lik)


vkarak:
@jenkins-cscs retry dom

PR #535, opened 2018-10-22
[bugfix] Fix retry report when test fails before the setup phase

### LINKED ISSUE #531: Something is broken in the retries code
Have a look [here](https://jenkins.cscs.ch/blue/organizations/jenkins/reframe-kesch-production-daily/detail/reframe-kesch-production-daily/144/pipeline) in today's regression run on Kesch.

omlins:
Fixes #531.

### DIFF
--- a/unittests/test_policies.py
+++ b/unittests/test_policies.py
@@ -11,8 +11,8 @@ from reframe.frontend.loader import RegressionCheckLoader
 import unittests.fixtures as fixtures
 from unittests.resources.checks.hellocheck import HelloTest
 from unittests.resources.checks.frontend_checks import (
-    KeyboardInterruptCheck, SleepCheck, SleepCheckPollFail,
-    SleepCheckPollFailLate, BadSetupCheck, RetriesCheck, SystemExitCheck)
+    KeyboardInterruptCheck, SystemExitCheck, SleepCheck, SleepCheckPollFail,
+    SleepCheckPollFailLate, BadSetupCheck, BadSetupCheckEarly, RetriesCheck)
 
 
 class TestSerialExecutionPolicy(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


omlins:
I could trigger the bug with the `BadSetupCheckEarly` and made a unit test.

vkarak:
@jenkins-cscs retry daint

unittests/test_policies.py
```diff
@@ -147,6 +147,22 @@ def test_retries_bad_check(self):
...
+
+        # Ensure that the test was retried #max_retries times and failed.
+        self.assertEqual(1, self.runner.stats.num_cases())
+        self.assertEqual(max_retries, rt.runtime().current_run)
+        self.assertEqual(1, self.runner.stats.num_failures())
+
+        # Ensure that the report succeeds and that the output is as expected.
+        report = self.runner.stats.retry_report()
+        self.assertIn("Test BadSetupCheckEarly was retried 2 time(s) and "
+                      "failed.", report)
```
vkarak: This check is quite restrictive. We will have to update the unit test every time we change either the automatic test naming scheme or message. I think we were getting a stack trace. You could then check that no `Traceback` is found in the output as we do with other similar cases.

vkarak: I think the check is even easier here. I think `retry_report()` was raising an exception. So you may simply call it. If it raises an exception, the unit test will fail.

unittests/test_policies.py
```diff
@@ -147,6 +147,22 @@ def test_retries_bad_check(self):
         self.assertEqual(max_retries, rt.runtime().current_run)
         self.assertEqual(1, self.runner.stats.num_failures())
 
+    def test_retries_bad_check_early(self):
```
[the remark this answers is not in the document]
omlins: Good idea! I merged the two tests. Done.

unittests/test_policies.py
vkarak: Yes, you simply call. If an unexpected exception is thrown, the unit tests will fail.

PR #537, opened 2018-10-24
[test] Remove Monch from supported systems

### LINKED ISSUE #489: Remove Monch from supported systems
Since Monch has been taken out of service, its references in the test cases should be removed. The ***MonchAcceptance tests***, on the other hand, should be kept, since they can be reused (latest example being Fulen).

vkarak:
- Monch-specific tests are commented as obsolete.

Fixes #489.

### DIFF
--- a/cscs-checks/apps/gromacs/gromacs_check.py
+++ b/cscs-checks/apps/gromacs/gromacs_check.py
@@ -129,6 +129,7 @@ class GromacsCPUProdCheck(GromacsCPUCheck):
         }
 
 
+# FIXME: This test is obsolete; it is kept only for reference.
 @rfm.parameterized_test([1], [2], [4], [6], [8])
 class GromacsCPUMonchAcceptance(GromacsBaseCheck):
     def __init__(self, num_nodes):
--- a/cscs-checks/libraries/math/scalapack_compile_run.py
+++ b/cscs-checks/libraries/math/scalapack_compile_run.py
@@ -67,16 +67,14 @@ class ScaLAPACKSanity(ScaLAPACKTest):
         ])
 
 
+# FIXME: This test is obsolete; it is kept only for reference.
+# NOTE:  The test case is very small, but larger cases did not succeed!
 @rfm.required_version('>=2.14')
 @rfm.parameterized_test(['dynamic'])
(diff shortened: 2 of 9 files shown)


vkarak:
@jenkins-cscs retry dom kesch

cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -129,6 +129,7 @@ def __init__(self):
         }
 
 
+# FIXME: This test is obsolete; it is kept only for reference.
```
victorusu: should we comment the valid systems in this file?

cscs-checks/libraries/math/scalapack_compile_run.py
```diff
@@ -67,16 +67,14 @@ def scalapack_sanity(number1, number2, expected_value):
         ])
 
 
+# FIXME: This test is obsolete; it is kept only for reference.
```
victorusu: should we comment the valid systems in this file?

cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -11,8 +11,7 @@ def __init__(self):
         self.exclusive_access = True
         # All available systems are supported
         self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc',
-                              'kesch:cn', 'kesch:pn', 'leone:normal',
-                              'monch:compute']
+                              'kesch:cn', 'kesch:pn', 'leone:normal']
```
victorusu: we are removing the system, but not the performance reference number.
Is there any reason why? Do we want to keep for future reference?

cscs-checks/apps/gromacs/gromacs_check.py
vkarak: I was thinking that these tests will be replaced by your versions, so, I've just put a `FIXME` tag.

cscs-checks/microbenchmarks/stream/stream.py
vkarak: I thought we wanted them for reference, don't we?

PR #576, opened 2018-11-11
[feat] Add support for units in performance references

### LINKED ISSUE #321: Extend reference tuple to include units and ordering
The idea is to allow some optional elements in the reference tuple of performance values, namely the units of the performance variable and the ordering, that could also be logged, so that plotting of the performance data will be straightforward by just reading the performance logs.

vkarak:
- Reference tuple may optionally accept a fourth element specifying the unit of measurement of the performance value.
- The possibility of logging the units is also supplied.
- Unit tests are adjusted.
- CSCS performance logging configuration adjusted to log also the units.
- Documentation is added.

Fixes #321.

### DIFF
--- a/config/cscs.py
+++ b/config/cscs.py
@@ -414,7 +414,8 @@ class ReframeSettings:
                     '%(check_perf_var)s=%(check_perf_value)s|'
                     'ref=%(check_perf_ref)s '
                     '(l=%(check_perf_lower_thres)s, '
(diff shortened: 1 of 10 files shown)


vkarak:
@jenkins-cscs retry daint

vkarak:
@victorusu Can you have a quick look in this PR. Does it affect the data collection? If not, we can merge it in this sprint. I believe it's an interesting feature.

PR #580, opened 2018-11-13
[feat] Update retries stage/output directory naming convention

### LINKED ISSUE #536: Revisit the naming convention of the stage directories of retries
After some real situations of production runs, I think that the current naming convention is confusing and unnatural:

```
/path/to/stage/sys/part/env/testname
/path/to/stage_1/sys/part/env/testname
/path/to/stage_2/sys/part/env/testname
```

This convention also increases the i-node count without a reason. I suggest changing to the following, which more natural and more intuitive:

```
/path/to/stage/sys/part/env/testname_0
/path/to/stage/sys/part/env/testname_1
/path/to/stage/sys/part/env/testname_2
```

We can agree on the suffix, but any change other in the last component of the path will be confusing.

omlins:
Modifies the behaviour of ReFrame to create stage and output folders as in the following example:
```
/path/to/stage/sys/part/env/testname
/path/to/stage/sys/part/env/testname_retry1
/path/to/stage/sys/part/env/testname_retry2
```

Closes  #536.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -807,10 +807,10 @@ class RegressionTest:
             resources = rt.runtime().resources
             self._stagedir = resources.make_stagedir(
                 self.current_system.name, self._current_partition.name,
-                self._current_environ.name, self.name + resources.run_suffix())
+                self._current_environ.name, self.name)
             self._outputdir = resources.make_outputdir(
                 self.current_system.name, self._current_partition.name,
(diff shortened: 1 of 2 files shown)


vkarak:
@jenkins-cscs retry daint

omlins:
Thanks @vkarak for the improvement. I changed the test-directory suffix back to the form `_retry1` as agreed in the stand-up meeting. I believe the MR is ready for merging.

PR #581, opened 2018-11-13
[test] Run libsci_acc and libsci_resolve tests on compute nodes as well

### LINKED ISSUE #538: Run libsci_acc resolution tests on the compute nodes as well
Currently we are only checking the login nodes, but the images might differ. These tests must also be added to the maintenance suite. They would have caught problems such those with COSMO OPCODE.

* `cscs-checks/compile/libsci_resolve.py`
* `cscs-checks/compile/libsci_acc_symlink.py`

teojgo:
* In addition, adapt the tests to the new syntax.

Closes #538

### DIFF
--- a/cscs-checks/compile/libsci_acc_symlink.py
+++ b/cscs-checks/compile/libsci_acc_symlink.py
@@ -21,8 +21,7 @@ class LibSciAccSymLinkTest(rfm.RunOnlyRegressionTest):
         self.valid_prog_environs = ['PrgEnv-cray']
         self.executable = 'ls'
         self.executable_opts = ['-al', '/opt/cray/pe/lib64/libsci_a*']
-        lib_so = lib_name + '.so'
-        self.sanity_patterns = sn.assert_found(lib_so, self.stdout)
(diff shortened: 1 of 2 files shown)


cscs-checks/compile/libsci_acc_symlink.py
```diff
@@ -1,36 +1,28 @@
...
 
         # The prgenv is irrelevant for this case, so just chose one
         self.valid_prog_environs = ['PrgEnv-cray']
-
         self.executable = 'ls'
-        self.executable_opts = '-al /opt/cray/pe/lib64/libsci_a*'.split()
-        self.sanity_patterns = sn.assert_found(lib_name, self.stdout)
+        self.executable_opts = ['-al', '/opt/cray/pe/lib64/libsci_a*']
+        lib_so = lib_name + '.so'
+        self.sanity_patterns = sn.assert_found(lib_so, self.stdout)
```
vkarak: Remove `lib_so`.

```suggestion
        self.sanity_patterns = sn.assert_found(lib_name + '.so', self.stdout)
```

Issue #510, opened 2018-10-12
Add regression test for testing the environment of the login nodes when accessed remotely over SSH


vkarak:
This has been a recurring problem with the new upgrade. The test should run on the login nodes and do the following:

```
ssh LOGIN echo '$SCRATCH'
```

The returned value must not be empty. Feel free to add other tests.

This test must be part of the maintenance suite.

Issue #524, opened 2018-10-18
Add regression test to check the environment


lucamar:
We have recently experienced several changes in the environment on the TDS system (e.g.: new NVIDIA driver, new Linux kernel, new Programming Environment...) and the list of modules loaded at the time we build and run applications unfortunately is not enough to have a full picture.

It would be therefore useful to have a test `environment_check` that does at least the following:
```
rpm -qa
env
module -t avail 2>&1
```
The data can be compared against a reference list of `rpm`, `env` and `module` commands (filtering default ones in the last case, for instance) and raise a `WARNING` or `FAIL` in case of a mismatch.

victorusu:
We should also check the compute node images. The proper command for that is
`rpm --root=$(findmnt -t squashfs /dev/loop3 -o target -f -n) -qa | sort`

PR #583, opened 2018-11-15
[test] Testing the environment of the login nodes when accessed remotely over SSH

rsarm:
Closes https://github.com/eth-cscs/reframe/issues/510

### DIFF
--- a/cscs-checks/prgenv/ssh_environ_check.py
+++ b/cscs-checks/prgenv/ssh_environ_check.py
@@ -2,17 +2,31 @@ import reframe as rfm
 import reframe.utility.sanity as sn
 
 
-@rfm.simple_test
+@rfm.parameterized_test(['CRAY_CPU_TARGET'], ['CRAYPE_NETWORK_TARGET'],
+                        ['MODULEPATH'], ['MODULESHOME'], ['PE_PRODUCT_LIST'],
+                        ['SCRATCH'], ['XDG_RUNTIME_DIR'])
 class SSHLoginEnvCheck(rfm.RunOnlyRegressionTest):
-    def __init__(self):
+    def __init__(self, envvar):
         super().__init__()
-        self.descr = 'Test the environment when accessed remotely over SSH'
+        self.descr = ('Check if $%s is defined when accessing '
+                      'remotely over SSH' % envvar)
         self.valid_systems = ['daint:login', 'dom:login']
         self.valid_prog_environs = ['PrgEnv-cray']
-        self.executable = "ssh"
-        self.executable_opts = ["%s" % self.current_system.name,
-                                "echo", "'$SCRATCH'"]
-        self.sanity_patterns = sn.assert_found(
-            r'/scratch/[\S+]', self.stdout)
-        self.maintainers = ['RS', 'VK']
+        self.executable = 'ssh'
+        self.executable_opts = [self.current_system.name,
+                                'echo', "'$%s'" % envvar]
+        values = {
+            'CRAY_CPU_TARGET': 'haswell',
+            'CRAYPE_NETWORK_TARGET': 'aries',
+            'MODULEPATH': r'[\S+]',
+            'MODULESHOME': r'/opt/cray/pe/modules/[\d+\.+]',
+            'PE_PRODUCT_LIST': 'CRAYPE_HASWELL:CRAY_RCA:CRAY_ALPS:DVS:'
+                               'CRAY_XPMEM:CRAY_DMAPP:CRAY_PMI:CRAY_UGNI:'
+                               'CRAY_UDREG:CRAY_LIBSCI:CRAYPE:CRAY:'
+                               'PERFTOOLS:CRAYPAT',
+            'SCRATCH': r'/scratch/[\S+]',
+            'XDG_RUNTIME_DIR': r'/run/user/[\d+]'
+        }
+        self.sanity_patterns = sn.assert_found(values[envvar], self.stdout)
+        self.maintainers = ['RS', 'LM']
         self.tags = {'maintenance'}
--- a/cscs-checks/prgenv/ssh_environ_check.py
+++ b/cscs-checks/prgenv/ssh_environ_check.py
@@ -2,20 +2,15 @@ import reframe as rfm
 import reframe.utility.sanity as sn
 
 
-@rfm.parameterized_test(['CRAY_CPU_TARGET'], ['CRAYPE_NETWORK_TARGET'],
-                        ['MODULEPATH'], ['MODULESHOME'], ['PE_PRODUCT_LIST'],
-                        ['SCRATCH'], ['XDG_RUNTIME_DIR'])
+@rfm.simple_test
 class SSHLoginEnvCheck(rfm.RunOnlyRegressionTest):
-    def __init__(self, envvar):
+    def __init__(self):
         super().__init__()
-        self.descr = ('Check if $%s is defined when accessing '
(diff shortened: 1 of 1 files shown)


cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,18 @@
...
+    def __init__(self):
+        super().__init__()
+        self.descr = 'Test the environment when accessed remotely over SSH'
+        self.valid_systems = ['daint:login', 'dom:login']
+        self.valid_prog_environs = ['PrgEnv-cray']
+        self.executable = "ssh"
+        self.executable_opts = ["%s" % self.current_system.name,
+                                "echo", "'$SCRATCH'"]
+        self.sanity_patterns = sn.assert_found(
+            r'/scratch/[\S+]', self.stdout)
```
vkarak: This should fit in a single line.

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@rfm.simple_test
+class SSHLoginEnvCheck(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        super().__init__()
+        self.descr = 'Test the environment when accessed remotely over SSH'
+        self.valid_systems = ['daint:login', 'dom:login']
+        self.valid_prog_environs = ['PrgEnv-cray']
+        self.executable = "ssh"
```
vkarak: Please use single quotes.

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,18 @@
...
+        self.executable_opts = ["%s" % self.current_system.name,
+                                "echo", "'$SCRATCH'"]
```
vkarak: Single quotes.

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,18 @@
...
+
+@rfm.simple_test
+class SSHLoginEnvCheck(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        super().__init__()
+        self.descr = 'Test the environment when accessed remotely over SSH'
+        self.valid_systems = ['daint:login', 'dom:login']
+        self.valid_prog_environs = ['PrgEnv-cray']
+        self.executable = "ssh"
+        self.executable_opts = ["%s" % self.current_system.name,
```
vkarak:
```suggestion
        self.executable_opts = [self.current_system.name,
```

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,18 @@
...
+                                "echo", "'$SCRATCH'"]
+        self.sanity_patterns = sn.assert_found(
+            r'/scratch/[\S+]', self.stdout)
+        self.maintainers = ['RS', 'VK']
```
vkarak:
```suggestion
        self.maintainers = ['RS', 'LM']
```

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,18 @@
...
+        self.tags = {'maintenance'}
```
teojgo: Add also production so that we can catch if there is something problematic with the profiles.

lucamar:
Thanks @rsarm: is it possible to check also the items proposed in #524 ?

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,18 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+@rfm.simple_test
+class SSHLoginEnvCheck(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        super().__init__()
+        self.descr = 'Test the environment when accessed remotely over SSH'
```
victorusu: The test name and description are not very informative. I would vote to have this test parameterised, with the individual env variables that we want to check. I would extend the variables to `$SCRATCH`, `$XDG_DATA_DIRS`, `$XDG_RUNTIME_DIR`, `$MODULEPATH`, `$MODULESHOME` and finally on Cray we should have `CRAY_CPU_TARGET`, `CRAYPE_NETWORK_TARGET` and `PE_PRODUCT_LIST`.

This way, I would vote to have the test description to **check if %(env_var)s is defined when accessed remotely over SSH**

vkarak:
@rsarm The problem you are seeing is due to the fact that a regression test is deep copied for each combination of system partition and programming environment, but the closure of the lambda function still refers to the original object, because functions are not deep copied. More specifically, when you create the lambda function for the `map()`, its closure contains `self`. At the time the lambda is defined `self` refers to the object created during loading of the test (by the `@simple_test` decorator). Before executing the test, ReFrame clones it, but the `self` of its closure still refers to the original object, which is not executed. As a result, `self._job` is `None`. Here is a small reproducer that showcases the problem:

```python
import copy


class C:
    def __init__(self):
        self.fn = lambda: self


if __name__ == '__main__':
    c = C()
    d = copy.deepcopy(c)
    print('c is bound to', c)
    print('d is bound to', d)
    print('c.fn() is bound to', c.fn())
    print('d.fn() is bound to', d.fn())
```

This will print the following, which demonstrates exactly the problem:

```
c is bound to <__main__.C object at 0x2b073e23e780>
d is bound to <__main__.C object at 0x2b073e267470>
c.fn() is bound to <__main__.C object at 0x2b073e23e780>
d.fn() is bound to <__main__.C object at 0x2b073e23e780>
```

A workaround to this bug is to avoid functions with closures. So you could rewrite your sanity function as follows:

```python
    def __init__(self):
        ...
        self.sanity_patterns = sn.all(
            sn.map(self.assert_envvar, ['CRAY_CPU_TARGET=haswell']))

    def assert_envvar(self, v):
        return sn.assert_found(v, self.stdout)
```

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,36 @@
...
+            'SCRATCH': r'/scratch/[\S+]',
+            'XDG_RUNTIME_DIR': r'/run/user/[\d+]'
+        }
+        self.executable = 'ssh'
+        echo_args = ' '.join('%s=$%s' % (i, i)  for i in reference.keys())
+        self.executable_opts = [self.current_system.name,
+                                'echo', "'%s'" % echo_args]
+        self.sanity_patterns = sn.all(sn.map(self.assert_envvar,
+                                      list(reference.keys()),
+                                      list(reference.values())))
```
vkarak: You can rewrite this in a simpler way:

```python
        self.sanity_patterns = sn.all(
            sn.map(self.assert_envvar, list(reference.items())))
```

See below how the `assert_envvar` is also simplified.

cscs-checks/prgenv/ssh_environ_check.py
```diff
@@ -0,0 +1,36 @@
...
+        self.maintainers = ['RS', 'LM']
+        self.tags = {'maintenance'}
+
+    def assert_envvar(self, var, value):
+        return sn.assert_found('='.join([var, value]), self.stdout)
```
vkarak:
```python
    def assert_envvar(self, v):
        return sn.assert_found('='.join(v), self.stdout)
```

vkarak:
@rsarm Can you revert your latest changes? I have fixed the PR myself yesterday, but forgot to merge it.

PR #584, opened 2018-11-15
[feat] Add a new `count_uniq()` sanity function

### LINKED ISSUE #556: Add a `count_uniq` sanity function
I think it is useful for flexible tests since it avoids caveats like the following:

```python
import reframe as rfm


@rfm.simple_test
class HostnameCheck(rfm.RunOnlyRegressionTest):
    def __init__(self):
        super().__init__()
        self.valid_systems = ['daint', 'dom']
        self.executable = 'hostname'
        self.sourcesdir = None
        self.num_tasks = 0
        self.sanity_patterns = sn.assert_eq(
            self.num_tasks_assigned,
            sn.count(sn.findall(r'nid\d+', self.stdout))
        )

    @property
    @sn.sanity_function
    def num_tasks_assigned(self):
        return self.job.num_tasks
```

The test passes fine, but it's missing the `num_tasks_per_node = 1`, which would have been caught easily if I could replace `count` with `count_uniq`:
```
nid00000
nid00000
nid00000
nid00000
nid00000
nid00000
nid00000
nid00000
nid00001
nid00001
nid00001
nid00001
nid00001
nid00001
nid00001
```

teojgo:
* Add the corresponding unit tests.

Closes #556

### DIFF
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -710,9 +710,9 @@ def count(iterable):
 
 
 @deferrable
-def count_unique(iterable):
+def count_uniq(iterable):
     """Return the unique element count of ``iterable``."""
-    return builtins.len(set(iterable))
+    return builtins.len(builtins.set(iterable))
 
 
 @deferrable
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -710,6 +710,12 @@ def count(iterable):
 
 
 @deferrable
+def count_unique(iterable):
+    """Return the unique element count of ``iterable``."""
+    return builtins.len(set(iterable))
+
+
+@deferrable
 def glob(pathname, *, recursive=False):
     """Replacement for the :func:`glob.glob() <python:glob.glob>` function."""
     return pyglob.glob(pathname, recursive=recursive)
--- a/unittests/test_sanity_functions.py
+++ b/unittests/test_sanity_functions.py
@@ -505,11 +505,7 @@ class TestUtilityFunctions(unittest.TestCase):
 
         # Test deferred expressions
         d = [1, 2, 2, 1]
-        dl = make_deferrable(d)
-        self.assertEqual(2, sn.count_uniq(dl))
-
-        d.extend([3, 4, 3])
-        self.assertEqual(4, sn.count_uniq(dl))
+        self.assertEqual(2, sn.count_uniq(make_deferrable(d)))
 
     def test_glob(self):
         filepatt = os.path.join(TEST_RESOURCES_CHECKS, '*.py')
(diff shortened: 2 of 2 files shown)


reframe/utility/sanity.py
```diff
@@ -709,6 +709,12 @@ def count(iterable):
         return ret
 
 
+@deferrable
+def count_unique(iterable):
```
vkarak: I prefer `count_uniq` as a name.

reframe/utility/sanity.py
```diff
@@ -709,6 +709,12 @@ def count(iterable):
...
+    """Return the unique element count of ``iterable``."""
+    return builtins.len(set(iterable))
```
vkarak:
```suggestion
    return builtins.len(builtins.set(iterable))
```

unittests/test_sanity_functions.py
```diff
@@ -483,6 +483,26 @@ def myrange(n):
         self.assertEqual(0, sn.count(range(0)))
         self.assertEqual(0, sn.count(myrange(0)))
 
+    def test_count_unique(self):
```
vkarak:
```suggestion
    def test_count_uniq(self):
```

unittests/test_sanity_functions.py
```diff
@@ -483,6 +483,26 @@ def myrange(n):
...
+        self.assertEqual(2, sn.count_unique(my_mod_range(10)))
+        self.assertEqual(3, sn.count_unique(my_mod_range(10, 3)))
+
+        # Test empty sequences
+        self.assertEqual(0, sn.count_unique([]))
+        self.assertEqual(0, sn.count_unique({}))
+        self.assertEqual(0, sn.count_unique(set()))
+        self.assertEqual(0, sn.count_unique(my_mod_range(0)))
+        self.assertEqual(0, sn.count_unique(range(0)))
+
```
vkarak: You should also add a couple of assertions, where the argument passed to the tested function is a deferred expression. Check other unit tests, e.g., `test_len` etc. You may simply convert one of the arguments passed above to a deferrable.

unittests/test_sanity_functions.py
```diff
@@ -483,6 +483,34 @@ def myrange(n):
...
+        self.assertEqual(0, sn.count_uniq([]))
+        self.assertEqual(0, sn.count_uniq({}))
+        self.assertEqual(0, sn.count_uniq(set()))
+        self.assertEqual(0, sn.count_uniq(my_mod_range(0)))
+        self.assertEqual(0, sn.count_uniq(range(0)))
+
+        # Test deferred expressions
+        d = [1, 2, 2, 1]
+        dl = make_deferrable(d)
+        self.assertEqual(2, sn.count_uniq(dl))
```
vkarak: Better write this as a single liner:

```suggestion
        self.assertEqual(2, sn.count_uniq(make_deferrable([1, 2, 2, 1])))
```

unittests/test_sanity_functions.py
```diff
@@ -483,6 +483,34 @@ def myrange(n):
...
+
+        d.extend([3, 4, 3])
+        self.assertEqual(4, sn.count_uniq(dl))
```
vkarak: This test is superfluous.

PR #585, opened 2018-11-16
[test] Remove `ifdef USE_MPI` from the MCH OpenACC/CUDA/MPI/C++ test

lxavier:
This reduce the complexity, and forces to compile with MPI

If refremce does not need a version that can run without MPI, I suggest to integrate this change, otherwise just close the PR

### DIFF
--- a/cscs-checks/mch/openacc_cuda_mpi_cppstd.py
+++ b/cscs-checks/mch/openacc_cuda_mpi_cppstd.py
@@ -2,12 +2,11 @@ import reframe as rfm
 import reframe.utility.sanity as sn
 
 
-@rfm.parameterized_test([True], [False])
+@rfm.simple_test
 class OpenaccCudaCpp(rfm.RegressionTest):
-    def __init__(self, withmpi):
+    def __init__(self):
         super().__init__()
-        name_suffix = 'WithMPI' if withmpi else 'WithoutMPI'
-        self.name = 'OpenaccCudaCPP' + name_suffix
+        self.name = 'OpenaccCudaCPP'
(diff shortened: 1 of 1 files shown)


teojgo:
@jenkins-cscs retry all

lxavier:
yes we are only interested to run on at least 2 MPI threads you can remove the non mpi test

vkarak:
@lxavier Thanks for your input.
@ajocksch Could you update the tests as well in this PR, so that we can merge?

ajocksch:
I suggest to merge pull request #513 first. Then we should work on this one.

vkarak:
@ajocksch If we merge, it will cause the current ReFrame test to fail. So this PR should fix the test as well.

cscs-checks/mch/openacc_cuda_mpi_cppstd.py
```diff
@@ -2,12 +2,11 @@
...
 
-@rfm.parameterized_test([True], [False])
+@rfm.simple_test
 class OpenaccCudaCpp(rfm.RegressionTest):
-    def __init__(self, withmpi):
+    def __init__(self):
         super().__init__()
-        name_suffix = 'WithMPI' if withmpi else 'WithoutMPI'
-        self.name = 'OpenaccCudaCPP' + name_suffix
+        self.name = 'OpenaccCudaCPP'
```
vkarak: You don't need this anymore.

ajocksch: done

vkarak:
@jenkins-cscs retry all

vkarak:
@jenkins-cscs retry daint

Issue #589, opened 2018-11-20
Set `exclusive_access` to all tests for Kesch


vkarak:

teojgo:
The issue can be closed. The only remaining tests are on the private repository. Comments were made at the issues for the individual tests.

vkarak:
@teojgo Could we have PR in the private repo that sets this flags and fixes the issue completely?

teojgo:
@vkarak There are individual issues for each of the tests that need to be adapted anyway.

vkarak:
Ok, we can close this then.

PR #591, opened 2018-11-21
[test] Set `exclusive_access` for Kesch tests

teojgo:
The private tests have to be changed accordingly.

Partially addresses #589


teojgo:
@vkarak There are some private tests that have to be adapted, that's why.

vkarak:
@jenkins-cscs retry daint dom

PR #597, opened 2018-11-23
WIP [test] Add Allocation Speed Test

victorusu:
The test is inconclusive. I am still making the PR in order to discuss
possible solutions.

The allocation times fluctuate from 1.90735e-06 to ~2.5 sec.
I am not sure if this is because of the OS memory load by the time of
the execution of if this is related to the small memory allocation of
the test.


cscs-checks/prgenv/src/alloc_speed.cpp
```diff
@@ -0,0 +1,34 @@
...
+    return (t1 - t0) - (t2 - t1);
+}
+
+int main(int argn, char** argv)
+{
+    for (int i = 20; i < 33; i++) {
+        size_t sz = size_t(1) << i;
+        std::cout << sz / 1024.0 / 1024.0 << " MB, allocation time " << test_alloc(sz) << " sec. \n";
+    }
+}
```
jgphpc: you should add a newline

victorusu:
@toxa81, can you help me with your thoughts on the test? I am not sure why I am getting a huge fluctuation in both checks. 

The worst results of huge pages are consistently 10 times smaller than the worst results of no hugepages.

But both fluctuate by 6 orders of magnitude.

cscs-checks/prgenv/src/alloc_speed.cpp
```diff
@@ -0,0 +1,34 @@
...
+double test_alloc(size_t n)
+{
+    double t0 = wtime();
+    /* time to allocate + fill */
+    char* ptr = (char*)std::malloc(n);
+    std::fill(ptr, ptr + n, 0);
+    double t1 = wtime();
+    /* time fo fill */
+    std::fill(ptr, ptr + n, 0);
+    double t2 = wtime();
```
toxa81: can you please add this lines after line 22:
/* harmless: add zero to t0 to prevent full optimization of the code with GCC */
    t0 += ptr[0];

cscs-checks/prgenv/src/alloc_speed.cpp
toxa81: The test should accumulate the total timer to allocate blocks of different size. This single number should be parsed by the Reframe

victorusu: The time required to allocate the 4096 MB is bigger than the sum of all the others.

cscs-checks/prgenv/src/alloc_speed.cpp
victorusu: This solved the problem. Thanks @toxa81.

cscs-checks/prgenv/alloc_speed.py
```diff
@@ -0,0 +1,79 @@
...
+                'daint:gpu': {
+                    'perf': (0.16, None, 0.10)
+                },
+                'daint:mc': {
+                    'perf': (0.50, None, 0.10)
+                },
+            },
+        }
+        self.reference = self.sys_reference[hugepages]
+
```
teojgo: Remove one of the empty lines

cscs-checks/prgenv/alloc_speed.py
```diff
@@ -0,0 +1,78 @@
+from datetime import datetime
```
teojgo: Remove this unused import

cscs-checks/prgenv/src/alloc_speed.cpp
```diff
@@ -0,0 +1,27 @@
+#include <chrono>
+#include <iostream>
+#include <algorithm>
+
+double test_alloc(size_t n)
+{
+    auto t0 = std::chrono::system_clock::now();
+    /* time to allocate + fill */
+    char* ptr = (char*)std::malloc(n);
```
vkarak: Why not use `new` and wrap it in an `std::unique_ptr`. You'd avoid the `free()` at the end.

cscs-checks/prgenv/src/alloc_speed.cpp
```diff
@@ -0,0 +1,27 @@
...
+    std::fill(ptr, ptr + n, 0);
+    auto t1 = std::chrono::system_clock::now();
+    /* time fo fill */
+    std::fill(ptr, ptr + n, 0);
+    auto t2 = std::chrono::system_clock::now();
+    t0 += static_cast<std::chrono::seconds>(ptr[0]);
```
vkarak: This is *VERY* cryptic. I understand you are doing it for fooling the compiler in order not to optimize out the `std::fill`, but you should explain. It is also quite error-prone. What if sb accidentally changes the array fill element to non-zero? Also, are you sure that none of the `std::fill()` is optimized out? I'd suggest declaring `ptr` as `volatile char *ptr` to avoid such optimizations from the compiler.

teojgo: `volatile` does not seem to work well with `unique_ptr`.

cscs-checks/prgenv/src/alloc_speed.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+    return std::chrono::duration_cast<std::chrono::duration<double>>((t1 - t0) - (t2 - t1)).count();
+}
+
+int main(int argn, char** argv)
+{
+    for (int i = 20; i < 33; i++) {
+        size_t sz = size_t(1) << i;
+        std::cout << sz / 1024.0 / 1024.0 << " MB, allocation time " << test_alloc(sz) << " sec. \n";
+    }
```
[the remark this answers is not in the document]
victorusu: Not really. The C++ [Standard states](https://github.com/cplusplus/draft/blob/5aaec13ef8f29527e762f8e4bf8e2362c8b1419c/source/basic.tex#L5774):

```
A \tcode{return} statement\iref{stmt.return} in \tcode{main} has the effect of leaving the main
function (destroying any objects with automatic storage duration) and
calling \tcode{std::exit} with the return value as the argument.
If control flows off the end of
the \grammarterm{compound-statement} of \tcode{main},
the effect is equivalent to a \tcode{return} with operand \tcode{0}
```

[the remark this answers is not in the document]
vkarak: Ah, ok! The classic C++ inconsistencies: "`main()` it's a function that must return an `int`, but if you don't return anything, nevermind, I'll do it for you." My suggestion is always to return to avoid such an inconsistency. I won't insist, though.

vkarak:
@jenkins-cscs retry kesch

PR #598, opened 2018-11-26
[test] mch_openacc_cuda_cppstd only with mpi

ajocksch:


vkarak:
@ajocksch Please push directly to #585.

PR #599, opened 2018-11-26
[bugfix] Check if 'UnavailableNodes' are down to cancel job

### LINKED ISSUE #303: 'ReqNodeNotAvail' is not always a reason of indefinite blocking of a job

teojgo:
* Use scontrol to retrieve the state of the nodes reported
  as unavailable and cancel the job if they are down.

* Add methods `in_state` and `is_down` to `SlurmNode` class.

* Add unittests for the above methods.

Fixes #303

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -293,11 +293,24 @@ class SlurmJob(sched.Job):
             reason, reason_details = reason_descr, None
 
         if reason in self._cancel_reasons:
-            # Here we handle the case were the UnavailableNodes list is empty,
-            # which actually means that the job is pending
             if reason == 'ReqNodeNotAvail' and reason_details:
-                if re.match(r'UnavailableNodes:$', reason_details.strip()):
-                    return
+                nodes_match = re.match(
+                    r'UnavailableNodes:(?P<node_names>\S+)?',
+                    reason_details.strip())
+                if nodes_match:
+                    node_names = node_match.group('node_names')
+                    if nodes_names:
+                        # Here we retrieve the info of the unavailable nodes
+                        # and check if they are indeed down
+                        nodes = self._get_nodes_by_name(node_names)
+                        down_nodes = [n for n in nodes if n.is_down()]
+                        if not down_nodes:
+                            return
+                    else:
+                        # Here we handle the case where the UnavailableNodes
+                        # list is empty which actually means that the job
+                        # is pending
+                        return
 
             self.cancel()
             reason_msg = ('job cancelled because it was blocked due to '
@@ -420,6 +433,13 @@ class SlurmNode:
 
     def is_available(self):
         return self._state == 'IDLE'
+
+    def is_down(self):
+        return any(
+            self.in_state(s) for s in ['DOWN', 'DRAIN', 'MAINT', 'NO_RESPOND'])
+
+    def in_state(self, state):
+        return state in self._state
 
     @property
     def active_features(self):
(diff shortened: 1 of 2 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -293,11 +293,24 @@ def _check_and_cancel(self, reason_descr):
...
-            # Here we handle the case were the UnavailableNodes list is empty,
-            # which actually means that the job is pending
             if reason == 'ReqNodeNotAvail' and reason_details:
-                if re.match(r'UnavailableNodes:$', reason_details.strip()):
-                    return
+                nodes_match = re.match(
+                    r'UnavailableNodes:(?P<node_names>\S+)?',
+                    reason_details.strip())
+                if nodes_match:
+                    node_names = node_match.group('node_names')
```
vkarak: I think the following will work:

```suggestion
                    node_names = node_match['node_names']
```

reframe/core/schedulers/slurm.py
```diff
@@ -293,11 +293,24 @@ def _check_and_cancel(self, reason_descr):
...
+                    if nodes_names:
+                        # Here we retrieve the info of the unavailable nodes
```
```suggestion
                        # Retrieve the info of the unavailable nodes
```

reframe/core/schedulers/slurm.py
```diff
@@ -293,11 +293,24 @@ def _check_and_cancel(self, reason_descr):
...
+                        # and check if they are indeed down
+                        nodes = self._get_nodes_by_name(node_names)
+                        down_nodes = [n for n in nodes if n.is_down()]
+                        if not down_nodes:
+                            return
+                    else:
+                        # Here we handle the case where the UnavailableNodes
+                        # list is empty which actually means that the job
+                        # is pending
```
```python
# List of unavailable nodes is empty; assume job is pending
```

reframe/core/schedulers/slurm.py
```diff
@@ -293,11 +293,24 @@ def _check_and_cancel(self, reason_descr):
...
+                    reason_details.strip())
+                if nodes_match:
+                    node_names = node_match.group('node_names')
+                    if nodes_names:
+                        # Here we retrieve the info of the unavailable nodes
+                        # and check if they are indeed down
+                        nodes = self._get_nodes_by_name(node_names)
+                        down_nodes = [n for n in nodes if n.is_down()]
+                        if not down_nodes:
+                            return
```
```python
if not any(n.is_down for n in nodes):
    return
```

reframe/core/schedulers/slurm.py
```diff
@@ -421,6 +434,13 @@ def __hash__(self):
     def is_available(self):
         return self._state == 'IDLE'
 
+    def is_down(self):
+        return any(
+            self.in_state(s) for s in ['DOWN', 'DRAIN', 'MAINT', 'NO_RESPOND'])
+
+    def in_state(self, state):
+        return state in self._state
+
```
vkarak: I think it would be better if `self._state` was actually a set holding all the states a Slurm node is in. Then `is_down` could be rewritten more efficiently as follows:

```python
def is_down(self):
    return bool({'DOWN', 'DRAIN', 'MAINT', 'NO_RESPOND'} & self._states)
```

reframe/core/schedulers/slurm.py
```diff
@@ -432,14 +431,13 @@ def __hash__(self):
         return hash(self.name)
 
     def is_available(self):
-        return self._state == 'IDLE'
+        return self._states == {'IDLE'}
```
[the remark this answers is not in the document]
teojgo: The problem here is that the states may be both 'IDLE' and 'DRAIN' which would mean that the node is not available.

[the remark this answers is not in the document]
vkarak: I thought there was a case such as `IDLE+ALLOCATED`, but this is represented by `MIXED` AFAICS. So I'm fine.

PR #600, opened 2018-11-26
[ci] Use integer comparison for ReFrame versions in deployment script

### LINKED ISSUE #593: Deployment script treats ReFrame versions as integers during comparison
The operator `-ne` is used for comparison, while the `!=` should be used.

teojgo:
Fixes #593


vkarak:
@jenkins-cscs retry daint

PR #604, opened 2018-11-27
[test] Add flexible HPCG test

### LINKED ISSUE #550: Add a flexible HPCG test

sebkelle1:
Added checks for two HPCG implementations.

Fixes #550.


cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+class HPCGCheckRef(rfm.RegressionTest):
+    def __init__(self, **kwargs):
+        super().__init__()
+
+        self.descr = 'HPCG reference benchmark'
+        self.valid_systems = ['daint:mc', 'daint:gpu', 'dom:gpu', 'dom:mc']
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.modules = ['craype-hugepages8M']
+        self.build_system = 'Make'
+        self.build_system.options = ['arch=MPI_GCC_OMP']
```
victorusu: Does it make sense to add the compilation flags here, per system?
```
self.build_system.cxxflags = '-xCORE-AVX2 -qopenmp -std=c++11 $(HPCG_DEFS)'
```

Because not all systems are AVX2. Again it only makes sense if we extend the test systems.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+    @property
+    @sn.sanity_function
+    def num_tasks_assigned(self):
+        return self.job.num_tasks
+
+    @property
+    @sn.sanity_function
+    def outfile_lazy(self):
+        pattern = 'n%d-%dp-%dt-*.yaml' % (self.problem_size,
+                                       self.job.num_tasks,
```
teojgo: Align the tuple.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        output_file = sn.getitem(sn.glob('HPCG*.txt'), 0)
+        self.sanity_patterns = sn.assert_eq(4, sn.count(
+            sn.findall(r'PASSED', output_file)))
+        self.num_tasks = 12
+        self.num_tasks_per_node = 12
+        self.num_cpus_per_task = 1
+
+        self.reference = {
+            'daint:gpu': {
+                'perf': (7.6, -0.1, 0.1)
```
victorusu: The compilation is MPI_OpenMP. So, does it make sense to have some sort of ramp, where the performance is dependent on the number of tasks used?

And also, there is an upper limit. I would not consider a failure to run faster than expected.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        self.build_system.options = ['arch=MPI_GCC_OMP']
+        self.sourcesdir = None
+        self.prebuild_cmd = ['git clone https://github.com/hpcg-benchmark/hpcg.git', 'cd hpcg']
+
+        self.executable = 'hpcg/bin/xhpcg'
+        self.executable_opts = ['--nx=104', '--ny=104', '--nz=104', '-t2']
+        output_file = sn.getitem(sn.glob('HPCG*.txt'), 0)
+        self.sanity_patterns = sn.assert_eq(4, sn.count(
+            sn.findall(r'PASSED', output_file)))
+        self.num_tasks = 12
```
victorusu: Should the number of tasks depend on the system:partition combo? I think we should be using a different number for daint/dom:mc.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        super().__init__()
+
+        self.descr = 'HPCG reference benchmark'
+        self.valid_systems = ['daint:mc', 'daint:gpu', 'dom:gpu', 'dom:mc']
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.modules = ['craype-hugepages8M']
+        self.build_system = 'Make'
+        self.build_system.options = ['arch=MPI_GCC_OMP']
+        self.sourcesdir = None
+        self.prebuild_cmd = ['git clone https://github.com/hpcg-benchmark/hpcg.git', 'cd hpcg']
```
teojgo: Better use the `sourcesdir` alternative to clone the repository, as explained in https://eth-cscs.github.io/reframe/advanced.html?highlight=git#retrieving-the-source-code-from-a-git-repository

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        }
+        self.maintainers = ['SK']
+
+@rfm.simple_test
+class HPCGCheckMKL(rfm.RegressionTest):
+    def __init__(self, **kwargs):
+        super().__init__()
+
+        self.descr = 'HPCG benchmark Intel MKL implementation'
+        self.valid_systems = ['dom:mc']
```
victorusu: Did you consider running this on daint also? and on the gpu partitions?

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        self.perf_patterns = {
+            'perf': sn.extractsingle(
+                r'HPCG result is VALID with a GFLOP\/s rating of=\s*'
+                r'(?P<perf>\S+)', output_file, 'perf',  float)
+        }
+        self.maintainers = ['SK']
+
+@rfm.simple_test
+class HPCGCheckMKL(rfm.RegressionTest):
+    def __init__(self, **kwargs):
```
teojgo: Remove `**kwargs`

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        self.reference = {
+            'dom:mc': {
+                'perf': (22, -0.1, 0.1)
+            },
+        }
+
+        self.perf_patterns = {
+            'perf': sn.extractsingle(
+                r'HPCG result is VALID with a GFLOP\/s rating of:\s*'
+                r'(?P<perf>\S+)', self.outfile_lazy, 'perf',  float) / (self.num_tasks_assigned/self.num_tasks_per_node)
```
teojgo: Use `self.num_tasks_assigned // self.num_tasks_per_node` to get an integer result.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+                             'mv Make.CrayXC setup',
+                             './configure CrayXC']
+
+        self.num_tasks = 0
+        self.num_tasks_per_core = 2
+        self.num_tasks_per_node = 4
+        self.num_cpus_per_task = 18
+        self.problem_size = 104
+
+        self.variables  = {
```
victorusu: Does it make sense to have similar env flags on the other check too?

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+class HPCGCheckMKL(rfm.RegressionTest):
+    def __init__(self, **kwargs):
+        super().__init__()
+
+        self.descr = 'HPCG benchmark Intel MKL implementation'
+        self.valid_systems = ['dom:mc']
+        self.valid_prog_environs = ['PrgEnv-intel']
+        self.modules = ['craype-hugepages8M']
+        #self.sourcesdir needed for "CrayXC" config file
+        self.build_system = 'Make'
```
teojgo: Use the `Autotools` `build_system` that handles the configuration step.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+@rfm.simple_test
+class HPCGCheckRef(rfm.RegressionTest):
+    def __init__(self, **kwargs):
```
teojgo: Remove `**kwargs`

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: daint has the issue with hanging flexible tests. dom:gpu would be possible if you think it makes sense to test there

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: I don't want an integer result. If `num_tasks` is a multiple of `num_tasks_per_node`, the outcome
is the same. If not, the floating point result will be a better estimate.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: Performance already is dependent on number of tasks, because the extracted perf gets divided by number of nodes. I removed the upper limit.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: No, because different implementation

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: ok on daint:mc
the benchmark obviously runs on the GPU partitions as well. If it makes sense to test
cpu-only conjugate gradient there is another question and I'll leave that to you. Yes or No on daint/dom:gpu?

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: ok

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: If I use `Autotools` the configure script doesnt work, unless I also specify `self.build_system.flags_from_environ=False`, which is not really what I want.
Also, I need `self.build_system.config_opts = ['CrayXC']` in addition. In summary, I think
the Autotools build_system makes it more complicated in this case

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        super().__init__()
+
+        self.descr = 'HPCG reference benchmark'
+        self.valid_systems = ['daint:mc', 'daint:gpu', 'dom:gpu', 'dom:mc']
```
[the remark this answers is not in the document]
victorusu: @vkarak, what do you think? Should we extend the list of systems or it should be fine.?

cscs-checks/analytics/hpcg/hpcg_benchmark.py
victorusu: @sebkelle1, I would vote for yes. Since this test can also be used for node diagnostic.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
sebkelle1: not that it makes a difference for performance in this case, but for completeness I added this

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.modules = ['craype-hugepages8M']
```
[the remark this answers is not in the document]
vkarak: As I said previously, this is not needed for the moment. Whoever extends the test to a different system should take care of this.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
vkarak: Let's focus on Daint/Dom right now. You could open an issue for extending this test on Fulen.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
vkarak: @sebkelle1 Yes, it would make sense, cos they are different microarchitectures: Haswell @ gpu, Broadwell @ mc.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,138 @@
...
+        self.build_system = 'Make'
+        self.build_system.options = ['arch=MPI_GCC_OMP']
+        self.sourcesdir = 'https://github.com/hpcg-benchmark/hpcg.git'
+
+        self.executable = 'bin/xhpcg'
+        self.executable_opts = ['--nx=104', '--ny=104', '--nz=104', '-t2']
+        output_file = sn.getitem(sn.glob('HPCG*.txt'), 0)
```
vkarak: Can you put a comment before this line explaining why this is necessary. It's not so straightforward for a first-time reader to understand why you are using filename globbing and you are picking only the first element. AFAIR, there is just a single file generated, but it's name is not known.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,138 @@
...
+        self.system_num_tasks = {
+            'daint:mc':  36,
+            'daint:gpu': 12,
+            'dom:mc':  36,
+            'dom:gpu': 12,
+        }
+
+        self.reference = {
+            'daint:gpu': {
+                'perf': (7.6, -0.1, None)
```
vkarak: Don't use `perf`, but rather a more descriptive name for the performance variable. `perf` was just a generic placeholder I've put it in some very early examples, and then survived and multiplied thanks to copy-paste...

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,138 @@
...
+            'perf': sn.extractsingle(
+                r'HPCG result is VALID with a GFLOP\/s rating of=\s*'
+                r'(?P<perf>\S+)', output_file, 'perf',  float)
+        }
+        self.maintainers = ['SK']
+        self.tags = {'diagnostic'}
+
+    def setup(self, partition, environ, **job_opts):
+        self.num_tasks = self.system_num_tasks[self.current_system.name
+                                               + ":" + partition.name]
```
vkarak: We split lines after the binary operator, but you can rewrite this much easier as follows:

```python
self.num_tasks = self.system_num_tasks[partition.fullname]
```

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,138 @@
...
+    @property
+    @sn.sanity_function
+    def outfile_lazy(self):
+        pattern = 'n%d-%dp-%dt-*.yaml' % (self.problem_size,
+                                          self.job.num_tasks,
+                                          self.num_cpus_per_task)
+        return sn.getitem(sn.glob(pattern), 0)
+
+    def setup(self, partition, environ, **job_opts):
+        if partition.name == 'gpu':
```
vkarak: I'd rewrite this as follows in order to be more future-proof:

```suggestion
        if partition.fullname in ['daint:gpu', 'dom:gpu']:
```

cscs-checks/analytics/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,138 @@
...
+            self.num_tasks_per_node = 2
+            self.num_cpus_per_task = 12
+        else:
+            self.num_tasks_per_node = 4
+            self.num_cpus_per_task = 18
+
+        self.perf_patterns = {
+            'perf': sn.extractsingle(
+                r'HPCG result is VALID with a GFLOP\/s rating of:\s*'
+                r'(?P<perf>\S+)', self.outfile_lazy, 'perf',  float) / (self.num_tasks_assigned/self.num_tasks_per_node)
```
vkarak: Why you need to divide the extracted value? Please put a short comment explaining why this is needed for the MKL version of the test.

cscs-checks/analytics/hpcg/hpcg_benchmark.py
vkarak: I suggest adding another element in the tuple that denotes the units. This is feasible from the current version only, so you should put `@rfm.required_version('>=2.16-dev0') on the top of the test.

vkarak:
@sebkelle1 Can you address the review comments so that we can merge this?

cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,148 @@
...
+@rfm.simple_test
+class HPCGCheckMKL(rfm.RegressionTest):
+    def __init__(self):
+        super().__init__()
+
+        self.descr = 'HPCG benchmark Intel MKL implementation'
+        self.valid_systems = ['daint:mc', 'dom:mc', 'daint:gpu', 'dom:gpu']
+        self.valid_prog_environs = ['PrgEnv-intel']
+        self.modules = ['craype-hugepages8M']
+        #self.sourcesdir needed for "CrayXC" config file
```
vkarak: Is this comment needed?

cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
```diff
@@ -0,0 +1,148 @@
...
+        if partition.fullname in ['daint:gpu', 'dom:gpu']:
+            self.num_tasks_per_node = 2
+            self.num_cpus_per_task = 12
+        else:
+            self.num_tasks_per_node = 4
+            self.num_cpus_per_task = 18
+
+        # since this is a flexible test, we divide the extracted
+        # performance by the number of nodes and compare
+        # against a single reference
```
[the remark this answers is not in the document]
sebkelle1: The number of tasks_per_node is different between the tests.
While `--flex-alloc-tasks=8` would run on 4 nodes on daint:gpu and 2 nodes daint:mc for the MKL case, the reference implementation has to have at least 36 tasks.
If there was a `--flex-alloc-nodes`, it would work. Or I split the test in two files.
In fact, there should be a check in the test that throws an error if the allocated tasks
is not a multiple of `num_tasks_per_node`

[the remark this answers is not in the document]
vkarak: This I understand. My question is basically, why the base test is not flexible as well?

vkarak:
@teojgo @victorusu Could you please have a look and approve this PR?

vkarak:

PR #615, opened 2018-11-29
[test] Fail if test is not compiled with OpenACC

lxavier:
We probably should add this to all the OpenACC tests, they have no purpose on CPU and we want to make sure there is no mistake in the flags/compiler


lxavier:
could we rename the test automatic_arrays.F90 to automatic_arrays_OpenACC.F90 ?

vkarak:
@jenkins-cscs retry dom kesch

PR #617, opened 2018-11-30
[test] Add HPX 1d stencil tests

victorusu:
The stencil 4 check is a single-local test
The stencil 8 check is a multi-local test

### DIFF
--- a/cscs-checks/libraries/hpx/hpx_stencil.py
+++ b/cscs-checks/libraries/hpx/hpx_stencil.py
@@ -1,0 +1,189 @@
+import itertools
+import os
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+@rfm.simple_test
+class Stencil4HPXCheck(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        super().__init__()
+
+        self.descr = 'HPX 1d_stencil_4 check'
+        self.valid_systems = ['daint:gpu, daint:mc', 'dom:gpu', 'dom:mc']
+        self.valid_prog_environs = ['PrgEnv-gnu']
+
+        self.modules = ['HPX']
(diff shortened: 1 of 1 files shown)


cscs-checks/libraries/hpx/hpx_stencil.py
```diff
@@ -0,0 +1,183 @@
...
+            self.num_cpus_per_task = 36
+        elif partition.fullname == 'dom:gpu':
+            self.num_tasks = 1
+            self.num_tasks_per_node = 1
+            self.num_cpus_per_task = 12
+        elif partition.fullname == 'dom:mc':
+            self.num_tasks = 1
+            self.num_tasks_per_node = 1
+            self.num_cpus_per_task = 36
+        self.executable_opts += ['--hpx:threads=%s' % self.num_cpus_per_task]
```
teojgo: Add an empty line before this.

cscs-checks/libraries/hpx/hpx_stencil.py
```diff
@@ -0,0 +1,186 @@
+import itertools
+import os
```
vkarak: These imports are not redundant.

victorusu:
@vkarak, we can combine both tests, but the combination is a bit ugly since both tests have different regexes for the sanity and performance. If you want I can make a base class that encompasses what is common between both tests and have different specialisations for both tests. What do you think?

vkarak:
@victorusu I've now seen more carefully the test. You are right. They do seem quite similar, but in fact differ in several minor aspects. So leave them as is. Just fix the redundant imports and we can merge it.

vkarak:
@teojgo Can you approve as well?

vkarak:
@jenkins-cscs retry none

vkarak:
@jenkins-cscs retry none

PR #621, opened 2018-12-03
[bugfix] Do not crash if Slurm node does not contain partition information

### LINKED ISSUE #605: Slurm node description may not contain partition info
The node information returned by `scontrol show` may not contain information regarding partitions. This should be correctly handled by ReFrame, by setting the corresponding `SlurmNode` field to an empty set.

teojgo:
* The `SlurmNode` attibutes: partitions, active_features and `state` can
  now be `None`.

* Throw a `JobError` when the `nodename` cannot be extracted.

* Adjust the unit tests to reflect the above changes.

Fixes #605

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -418,10 +418,10 @@ class SlurmNode:
             raise JobError('could not extract NodeName from node description')
 
         self._partitions = self._extract_attribute(
-            'Partitions', node_descr, delim=',')
+            'Partitions', node_descr, sep=',')
         self._active_features = self._extract_attribute(
-            'ActiveFeatures', node_descr, delim=',')
-        self._states = self._extract_attribute('State', node_descr, delim='+')
+            'ActiveFeatures', node_descr, sep=',')
+        self._states = self._extract_attribute('State', node_descr, sep='+')
 
     def __eq__(self, other):
         if not isinstance(other, type(self)):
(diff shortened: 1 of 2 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -452,13 +455,15 @@ def partitions(self):
...
         attr_match = re.search(r'%s=(\S+)' % attr_name, node_descr)
         if attr_match:
-            return attr_match.group(1)
-        else:
-            raise JobError("could not extract attribute '%s' from "
-                           "node description" % attr_name)
+            if delim:
+                return set(attr_match.group(1).split(delim))
+            else:
+                return attr_match.group(1)
```
vkarak: You can rewrite this in a more compact way:

```python
if attr_match:
    attr = attr_match.group(1)
    return attr.split(delim) if delim else attr
```

reframe/core/schedulers/slurm.py
```diff
@@ -431,7 +433,8 @@ def __hash__(self):
         return hash(self.name)
 
     def is_available(self):
-        return self._states == {'IDLE'}
+        return (self._states == {'IDLE'} and
+                all([self._partitions, self._active_features, self._states]))
```
vkarak: You may rewrite this as follows:

```python
return all([self._states == {'IDLE'}, self._partitions, self._active_features, self._states])
```

PR #625, opened 2018-12-04
[feat] Omit printing "start/finished ..." for completely skipped tests

### LINKED ISSUE #411: Do no print "start/finished processing test" for tests that are skipped
Because the output is confusing. See this:

```
[==========] Running 3 check(s)
[==========] Started on Thu Jul 26 18:02:05 2018

[----------] started processing amber_gpu_prod_check (Amber parallel GPU production check)
[----------] finished processing amber_gpu_prod_check (Amber parallel GPU production check)

[----------] started processing amber_gpu_maint_check (Amber parallel GPU maintenance check)
[----------] finished processing amber_gpu_maint_check (Amber parallel GPU maintenance check)

[----------] started processing amber_cpu_check (Amber parallel CPU check)
[----------] finished processing amber_cpu_check (Amber parallel CPU check)

[  PASSED  ] Ran 0 test case(s) from 3 check(s) (0 failure(s))
[==========] Finished on Thu Jul 26 18:02:05 2018
```

I think we should also count the skipped tests and/or test cases and print that, too, because now somebody sees "Running 3 check(s)" at the beginning of the output and then finally "Run 0 test case(s)". We should just enhance the output.

omlins:
There are the following two possible solutions to improve the output of ReFrame:
1) filter first all test cases (printing filter results only in verbose mode), then run only the filtered test cases (print also in non-verbose mode). Seems to be a straightforward and clean solution; however, it is a considerable change to the implementation of ReFrame for filtering/launching tests (which could turn out actually also better for other issues).
2) filter the tests simply for one additional criteria: whether the test has test cases to be launched, given the partition and environments. This is the simplest solution; it leads though to a little bit of duplication of code currently; once #488 is solved, most of it will be removable.
3) keep the current implementation of ReFrame for filtering/launching tests cases (go through each test case, and run or skip it immediately) and improve the non-verbose output by delaying the "start processing/running <test>" until it is sure that at least one test case of <test> is run (this way the output is omitted if no test case is run). Problem: requires independent printing calls for verbose and non-verbose and needs in general quite some changes to the printing/logging mechanisms.

The solution 2 was chosen as it is simple and has several advantages over solution 3 (solution 1 would require many fundamental changes that we do not want - at least not now).

closes #411

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -45,6 +45,7 @@ def list_checks(checks, printer, detailed=False):
     printer.info('Found %d check(s).' % len(checks))
 
 
+# FIXME: This function is a duplicate of the corresponding function in the class Runner.
 def partition_supported(check, partition, options):
     if options.skip_system_check:
         return True
@@ -52,6 +53,7 @@ def partition_supported(check, partition, options):
     return check.supports_system(partition.name)
 
 
+# FIXME: This function is a duplicate of the corresponding function in the class Runner.
 def environ_supported(check, environ, options):
     ret = True
     if options.prgenv:
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -45,6 +45,40 @@ def list_checks(checks, printer, detailed=False):
     printer.info('Found %d check(s).' % len(checks))
 
 
+def partition_supported(check, partition, options):
+    if options.skip_system_check:
+        return True
+
+    return check.supports_system(partition.name)
+
+
+def environ_supported(check, environ, options):
+    ret = True
+    if options.prgenv:
+        ret = environ.name in options.prgenv
+
+    if options.skip_prgenv_check:
+        return ret
+    else:
+        return ret and check.supports_environ(environ.name)
+
+
+def has_testcase(check, options):
+    system = runtime.runtime().system
+    has_testcase = False
+    for partition in system.partitions:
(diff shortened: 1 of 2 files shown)


reframe/frontend/cli.py
```diff
@@ -45,6 +45,40 @@ def list_checks(checks, printer, detailed=False):
     printer.info('Found %d check(s).' % len(checks))
 
 
+def partition_supported(check, partition, options):
```
omlins: this function `partition_supported` and the below function `environ_supported` are duplicates of the corresponding functions in the class `Runner`. Once Christopher has finished the issue #488 where he is abstracting the test selection phase of the frontend in a separate class, we may move this two functions to this new class and remove them from here and from the class `Runner`

omlins:
I believe it can be merged as is even if it is not beautiful; most of it will change anyway again once #488 is solved...

omlins:
> I can't see a case where this output can appear anymore, can you? Perhaps, a case where a test defines valid systems, but not any programming environment for that system.
- Yes, exactly: the case where system is valid, but no program environment is not covered by #487.

vkarak:
The case where a system is defined but not any programming environments is rather corner case not causing any problems and it is essentially a logic mistake when writing a test (I can't see any reason behind it), but we add considerable code duplication to address it. Thus, I don't think it makes sense to merge this. We can close the issue with a comment that this has been addressed by #487.

PR #626, opened 2018-12-04
[ci] Create the yaml file for pep8speaks

### LINKED ISSUE #167: Add commit hook to check the coding style
I suggest adding a commit hook that would disable committing if the changed files do not follow the coding style. The downside is that this requires us to have `pycodestyle` installed on the machine that we commit.

@eth-cscs/reframe-devteam What do you think?

teojgo:
* Automatically check for ReFrame Coding Style violations.

Fixes #167

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,17 @@
+message:  # Customize the comment made by the bot
+    opened:  # Messages when a new PR is submitted
+        header: "Hello @{name}, Thank you for submitting the Pull Request!"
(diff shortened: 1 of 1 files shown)


vkarak:
@teojgo Why is this WIP?

teojgo:
@vkarak I need to recheck what it prints when there are no errors since now it should be updated after I fixed the coding style.

teojgo:
@vkarak it's ok to merge and we can make any changes in the future.

PR #629, opened 2018-12-12
[refactor] Refactor test selection phase

### LINKED ISSUE #488: Abstract test selection phase of the frontend in a separate class, so that it can be unit tested efficiently

ChristopherBignamini:
Closes #488


vkarak:
@ChristopherBignamini Check [here](https://github.com/eth-cscs/reframe/wiki/Coding-Style-Guide) for the exact error codes we ignore. I don't know why pep8speaks reports a couple of them.

reframe/frontend/cli.py
```diff
@@ -375,46 +375,27 @@ def main():
         # Locate and load checks
         try:
             checks_found = loader.load_all()
-        except OSError as e:
+        except OSError as oe:
```
teojgo: Why did you change to `oe`. If the change is valid the next line should raise from `oe`.

vkarak: The `e` should remain, unless its used elsewhere. This is a convention throughout the code. All exception objects are named `e`.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,41 @@
+import unittest
+
+import reframe.frontend.check_filters as filters
+from reframe.core.pipeline import RegressionTest
+
+class TestCheckFilters(unittest.TestCase):
+    def setUp(self):
+        #TODO: use correct initialization and add checks
+        self.check = RegressionTest()
+        self.check.name = "check1"
```
teojgo: Use single quotes

reframe/frontend/check_filters.py
```diff
@@ -0,0 +1,48 @@
...
+def have_prgenv(prgenv):
+    def _fn(c):
+        if prgenv:
+            return util.allx(c.supports_environ(e) for e in prgenv)
+        else:
+            return bool(c.valid_prog_environs)
+
+    return _fn
+
+def have_system():
```
teojgo: Since `have_system` does not take any arguments, why return a function and not directly the result?

reframe/frontend/check_filters.py
```diff
@@ -0,0 +1,48 @@
...
+    def _fn(c):
+        return any([c.supports_system(s.fullname)
+                    for s in rt.runtime().system.partitions])
+
+    return _fn
+
+def is_gpu_only():
```
teojgo: Same as above.

reframe/frontend/check_filters.py
```diff
@@ -0,0 +1,48 @@
...
+    def _fn(c):
+        return c.num_gpus_per_node > 0
+
+    return _fn
+
+def is_cpu_only():
```
teojgo: Same as above.

reframe/frontend/check_filters.py
vkarak: I would simply change the name of this function to `have_gpu_only()`.

reframe/frontend/check_filters.py
vkarak: I would simply change the name of this function to `have_cpu_only()`.

reframe/frontend/cli.py
```diff
@@ -375,46 +375,27 @@ def main():
...
 
         if options.names:
-            checks_matched = filter(
-                lambda c: c if c.name in options.names else None,
-                checks_matched
-            )
+            checks_matched = filter(filters.have_name(options.names), checks_matched)
 
         # Filter checks by tags
         user_tags = set(options.tags)
```
vkarak: There is no need for this one any more. The `have_tags` converts its argument to a set already.

reframe/frontend/cli.py
```diff
@@ -375,46 +375,27 @@ def main():
...
-        checks_matched = filter(
-            lambda c: c if user_tags.issubset(c.tags) else None,
-            checks_matched
-        )
+        checks_matched = filter(filters.have_tag(options.tags), checks_matched)
 
+        # TODO: should we move this in utils?
```
vkarak: Which one?

unittests/test_check_filters.py
```diff
@@ -0,0 +1,41 @@
...
+        self.check.tags = {'a', 'b', 'c', 'd'} 
+        self.check.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu']
```
vkarak: I would put more generic names, such as `env1`, `env2` etc.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,41 @@
...
+        self.check.valid_systems = ['daint:gpu', 'daint:mc',
+            'dom:gpu', 'dom:mc', 'kesch:pn', 'generic:login']
```
vkarak: Same here.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,41 @@
...
+        self.check.num_gpus_per_node = 1
+
+    def test_have_name(self):
+        self.assertTrue((filters.have_name('check1'))(self.check))
```
vkarak: I see too many parentheses here.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,41 @@
+import unittest
+
+import reframe.frontend.check_filters as filters
+from reframe.core.pipeline import RegressionTest
+
+class TestCheckFilters(unittest.TestCase):
+    def setUp(self):
+        #TODO: use correct initialization and add checks
+        self.check = RegressionTest()
```
vkarak: You should use more checks here.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertTrue((filters.have_tag(['a', 'c']))(self.check)) 
+        self.assertFalse((filters.have_tag(['a', 'b', 'c', 'd', 'e']))(self.check)) 
+        self.assertFalse((filters.have_tag(['x', 'y', 'z']))(self.check)) 
+
+    def test_have_prgenv(self):
+        self.assertTrue((filters.have_prgenv(['PrgEnv-cray', 'PrgEnv-gnu']))(self.check))
+        self.assertFalse((filters.have_prgenv(['PrgEnv-pgi']))(self.check))
+
+    def test_system(self):
+        self.assertTrue((filters.have_system())(self.check))
```
vkarak: If `have_system()` does not take an argument you rely here on the actual system that you are running the unit tests. That's why `have_system()` must accept the system name as an argument.

reframe/frontend/cli.py
ChristopherBignamini: Sorry, that is just a typo.

reframe/frontend/cli.py
ChristopherBignamini: Removed.

reframe/frontend/check_filters.py
ChristopherBignamini: Done

reframe/frontend/check_filters.py
ChristopherBignamini: Done

reframe/frontend/cli.py
ChristopherBignamini: Comment removed, it referred to the old filtering version

unittests/test_check_filters.py
ChristopherBignamini: done

unittests/test_check_filters.py
ChristopherBignamini: done

reframe/frontend/check_filters.py
```diff
@@ -0,0 +1,48 @@
...
+            return util.allx(c.supports_environ(e) for e in prgenv)
+        else:
+            return bool(c.valid_prog_environs)
+
+    return _fn
+
+def have_system(partitions):
+    def _fn(c):
+        return any([c.supports_system(s.fullname)
+                    for s in partitions])
```
vkarak: Put everything in a single line here. It should fit.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,65 @@
+import unittest
+
+import reframe.core.runtime as rt
+import reframe.frontend.check_filters as filters
+import unittests.fixtures as fixtures
+from reframe.core.systems import SystemPartition
+from reframe.core.pipeline import RegressionTest
+from reframe.utility.typecheck import Str
```
vkarak: This import is not needed.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,65 @@
+import unittest
+
+import reframe.core.runtime as rt
+import reframe.frontend.check_filters as filters
+import unittests.fixtures as fixtures
+from reframe.core.systems import SystemPartition
```
vkarak: This import is not needed.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,65 @@
...
+        self.check1.valid_prog_environs = ['env1', 'env2']
+        self.check1.valid_systems = ['testsys:gpu', 'testsys2:mc']
+        self.check1.num_gpus_per_node = 1
+        #check 2
+        self.check2 = RegressionTest()
+        self.check2.name = 'check2'
+        self.check2.tags = {'x', 'y', 'z'}
+        self.check2.valid_prog_environs = ['env3']
+        self.check2.valid_systems = ['testsys:mc', 'testsys2:mc']
+        self.check2.num_gpus_per_node = 0
```
vkarak: You should not individual tests here. You should have a list of tests instead:

```python
def create_check(attrs):
    ret = RegressionTest()
    for k, v in attrs.items():
        setattr(ret, k, v)

    return ret

self.checks = [create_check({'name': 'check1', ...}), create_check({'name': 'check2', ...})]
```

All subsequent unit tests should act on the list of checks `self.checks`. See example in my comments below.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_have_name(self):
+        self.assertTrue(filters.have_name('check1')(self.check1))
```
vkarak: This is the correct way to test this, because it reflects the way we are using the filter functions.

```python

self.assertEqual(1, sn.count(filter(filters.have_name('check1'), self.checks)))
```

The reason that you cannot use `len()` directly on the result of the `filter()` function is because `filter()` returns an generator, so you would have to create a concrete data structure (e.g., `list`) out of it first. `sn.count` handles this case automatically, but it consumes the generator (not a problem in our case). It also returns a deferrable expression, but since it's included in boolean check, it will be automatically evaluated.

unittests/test_check_filters.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_have_prgenv(self):
+        self.assertTrue(filters.have_prgenv(['env1', 'env2'])(self.check1))
+        self.assertFalse(filters.have_prgenv(['env3'])(self.check1))
+        self.assertTrue(filters.have_prgenv(['env3'])(self.check2))
+        self.assertFalse(filters.have_prgenv(['env1', 'env2'])(self.check2))
+
+    @rt.switch_runtime(fixtures.TEST_SITE_CONFIG, 'testsys')
+    def test_system(self):
+        self.assertTrue(filters.have_system(rt.runtime().system.partitions)(self.check1))
```
vkarak: Here you should explicitly look for a system partition by name. For example,

```python
have_system(['testsys'])
have_system(['testsys:gpu', 'testsys:mc'])
```

reframe/frontend/check_filters.py
ChristopherBignamini: Done.

unittests/test_check_filters.py
ChristopherBignamini: Done.

unittests/test_check_filters.py
ChristopherBignamini: Done.

ChristopherBignamini:
@jenkins-cscs retry dom

vkarak:
@teojgo Can you approve this PR?

unittests/test_check_filters.py
```diff
@@ -0,0 +1,93 @@
...
+                                            self.checks)))
+        self.assertEqual(1, sn.count(filter(filters.have_name('check3'),
+                                            self.checks)))
+        self.assertEqual(0, sn.count(filter(filters.have_name('check4'),
+                                            self.checks)))
+
+    def test_have_not_name(self):
+        self.assertEqual(1, sn.count(filter(filters.have_not_name('check1'),
+                                            self.checks)))
+        self.assertEqual(1, sn.count(filter(filters.have_not_name('check2'),
```
teojgo: Shouldn't this and the next one be equal to `2`?

vkarak: Yes, I haven't run the unit tests again. I am fixing it now and will re-push.

vkarak:
@teojgo It should be fine now.

PR #646, opened 2019-01-10
[test] Add benchmark measuring CUDA kernel launch latency

teojgo:


cscs-checks/microbenchmarks/kernel_latency/kernel_latency.py
```diff
@@ -0,0 +1,68 @@
...
+
+        self.build_system.cxxflags += ['-arch=compute_%s' % gpu_arch,
+                                       '-code=sm_%s' % gpu_arch]
+
+        if kernel_version == 'sync':
+            self.build_system.cppflags = ['-D SYNCKERNEL=1']
+        else:
+            self.build_system.cppflags = ['-D SYNCKERNEL=0']
+
+        self.sanity_patterns = sn.assert_found(r'Found \d+ gpu\(s\)',
```
vkarak: What if no GPUs are found?

cscs-checks/microbenchmarks/kernel_latency/kernel_latency.py
```diff
@@ -0,0 +1,68 @@
...
+                                               self.stdout)
+        self.perf_patterns = {
+            'latency': sn.extractsingle(
+                r'Kernel launch latency: (?P<latency>\S+) seconds',
+                self.stdout, 'latency', float)
+        }
+        self.sys_reference = {
+            'sync': {
+                'dom:gpu': {
+                    'latency': (6.6e-06, None, 0.10, 's')
```
vkarak: I think it would be better to report in `us`. Also from inside the code.

cscs-checks/microbenchmarks/kernel_latency/kernel_latency.py
```diff
@@ -2,14 +2,12 @@
 import reframe.utility.sanity as sn
 
 
-@rfm.required_version('>=2.16-dev0')
```
vkarak: Why did you remove this? It's still valid, since you're using units in the reference tuples.

cscs-checks/microbenchmarks/kernel_latency/src/kernel_latency.cu
```diff
@@ -8,16 +9,20 @@ __global__ void null_kernel() {
...
     cudaError_t error;
-    int gpu_count;
+    int gpu_count = 0;
 
     error = cudaGetDeviceCount(&gpu_count);
 
     if (error == cudaSuccess) {
-        std::cout << "Found " << gpu_count << " gpu(s)" << std::endl;
+        if (gpu_count <= 0) {
+            std::cout << "Could not found any gpu\n";
```
vkarak: `s/found/find`.

cscs-checks/microbenchmarks/kernel_latency/kernel_latency.py
```diff
@@ -0,0 +1,66 @@
...
+                'kesch:cn': {
+                    'latency': (5.7, None, 0.10, 'us')
+                },
+            },
+        }
+
+        self.reference = self.sys_reference[kernel_version]
+
+        self.maintainers = ['TM']
+        self.tags = {'production'}
```
vkarak: This test should not be production. It should be "benchmark" and "diagnostic" and it should flexible as well.

PR #650, opened 2019-01-17
Add option to increase output verbosity level

victorusu:
This PR adds the option to increase the verbosity level of the
PrettyPrinter. However, there is only one call to printer.debug and
there is on call to printer.verbose in the entire code. The former is
related to a module not being properly loaded, which is difficult to
reproduce, since we do have correct checks on ReFrame.

I am not sure if it is part of the issue to also add the printer.debug
and printer.verbose calls into the code.

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -86,6 +86,9 @@ def main():
         '--save-log-files', action='store_true', default=False,
         help='Copy the log file from the work dir to the output dir at the '
              'end of the program')
+    output_options.add_argument(
+        '--verbose', '-v', action='count',
+        help='Increase output verbosity level')
 
     # Check discovery options
     locate_options.add_argument(
@@ -258,6 +261,7 @@ def main():
     # Setup printer
     printer = PrettyPrinter()
     printer.colorize = options.colorize
+    printer.set_verbose_level(options.verbose)
 
     try:
         runtime.init_runtime(settings.site_configuration, options.system)
@@ -452,7 +456,7 @@ def main():
                 rt.modules_system.load_module(m, force=True)
                 raise EnvironError("test")
             except EnvironError as e:
-                printer.warning("could not load module '%s' correctly: " 
+                printer.warning("could not load module '%s' correctly: "
                                 "Skipping..." % m)
                 printer.debug(str(e))
 
--- a/reframe/frontend/printer.py
+++ b/reframe/frontend/printer.py
@@ -61,6 +61,14 @@ class PrettyPrinter:
     def __repr__(self):
         return debug.repr(self)
 
+    def set_verbose_level(self, number):
+        if not number:
+            self._logger.setLevel(logging.INFO)
+        elif number == 1:
+            self._logger.setLevel(logging.VERBOSE)
+        elif number >= 2:
+            self._logger.setLevel(logging.DEBUG)
+
     def separator(self, linestyle, msg=''):
         if linestyle == 'short double line':
             line = self.status_width * '='


reframe/core/logging.py
```diff
@@ -284,13 +298,22 @@ def __init__(self, name, level=logging.NOTSET):
...
 
     def __repr__(self):
         return debug.repr(self)
 
     def setLevel(self, level):
         self.level = _check_level(level)
 
+    def addHandler(self, hdlr):
+        if isinstance(hdlr, StreamHandler):
+            self._stream_handlers.append(hdlr)
```
vkarak: Even if the `Logger.handlers` attribute is not documented, I think we can make use of it, since it's public. We do this already. This would make the whole implementation much more straightforward.

reframe/frontend/printer.py
```diff
@@ -61,6 +61,19 @@ def __init__(self):
     def __repr__(self):
         return debug.repr(self)
 
+    def set_verbose_level(self, number):
+        if number:
+            for hdlr in self._logger.stream_handlers():
```
vkarak: Using whatever Python's logging provides us, here you would go over all the handlers and pick only those that are instance of `logging.StreamHandler` and their `stream` is `sys.stdout` (should we really bother for `sys.stderr` here?) and set their level.

reframe/core/logging.py
```diff
@@ -362,6 +362,15 @@ def setLevel(self, level):
...
             super().setLevel(level)
 
+    def stream_handlers(self):
+        if self.logger:
+            _stream_handlers = []
+            for hdlr in self.logger.handlers:
+                 if isinstance(hdlr, logging.StreamHandler) and not \
+                    isinstance(hdlr, logging.FileHandler):
+                    _stream_handlers.append(hdlr)
+            return _stream_handlers
```
vkarak: You should do sth if `self.logger` is `None`. Perhaps, return an empty list.

vkarak: You may also rewrite this return statement in a simpler way:

```python
            return [h for h in self.logger.handlers
                    if (isinstance(h, logging.StreamHandler) and
                        not isinstance(h, logging.FileHandler))]
```

reframe/frontend/cli.py
```diff
@@ -87,6 +87,9 @@ def main():
         '--save-log-files', action='store_true', default=False,
         help='Copy the log file from the work dir to the output dir at the '
              'end of the program')
+    output_options.add_argument(
+        '--verbose', '-v', action='count', default=0,
+        help='Increase output verbosity level')
```
```suggestion
        help='Increase verbosity level of output')
```

reframe/frontend/printer.py
```diff
@@ -20,6 +20,19 @@ def __init__(self):
         self.line_width = 78
         self.status_width = 10
 
+    def set_verbosity(self, number):
+        if number:
+            for hdlr in self.stream_handlers():
```
vkarak: I would make `stream_handlers` a property.

reframe/frontend/printer.py
```diff
@@ -20,6 +20,19 @@ def __init__(self):
...
+                for l in range(0,number):
+                    loglevel = hdlr.level
+                    if loglevel == logging.INFO:
```
vkarak: I think you can avoid this set of `if`s if you do the following:

```python
def inc_verbosity(self, num_steps):
    log_levels = sorted(_log_level_names.keys())[1:] # [1:] to exclude level NOTSET
    for h in self.stream_handlers:
        level_idx = log_levels.index(h.level)
        try:
            new_level = log_levels[level_idx - num_steps]
        except IndexError:
            new_level = 0

        h.setLevel(new_level)
```

vkarak: I also believe `inc_verbosity()` is better than `set_verbosity()`, because you can only increase the verbosity.

reframe/frontend/printer.py
```diff
@@ -20,18 +20,16 @@ def __init__(self):
...
-                        new_level = max(loglevel - 10, logging.DEBUG)
-                    hdlr.setLevel(new_level)
+    def inc_verbosity(self, num_steps):
+        log_levels = sorted(logging._log_level_names.keys())[1:] # [1:] to exclude level NOTSET
+        for h in self.stream_handlers:
+            level_idx = log_levels.index(h.level)
+            try:
+                new_level = log_levels[level_idx - num_steps]
+            except IndexError:
+                new_level = 0
```
vkarak: I think this is not correct. It should be `new_level = log_levels[0]`. Otherwise it will set the level to `NOTSET`.

vkarak: I verified it. Passing `-vvv` produces no output.

PR #652, opened 2019-01-18
[feat] Add support for setting a minimum number of required tasks in tests with flexible task allocation

### LINKED ISSUE #640: Allow flexible tests to specify the minimum amount of nodes that they need
Some flexible tests may need a minimum amount of nodes to run. Currently, you there is no way to restrict this. Perhaps we will need to add a new field in the `RegressionTest`, e.g., `num_nodes_min`?

teojgo:
Fixes #640

### DIFF
--- a/reframe/core/schedulers/__init__.py
+++ b/reframe/core/schedulers/__init__.py
@@ -222,12 +222,22 @@ class Job(abc.ABC):
     def prepare(self, commands, environs=None, **gen_opts):
         environs = environs or []
         if self.num_tasks <= 0:
+            num_tasks_per_node = (self.num_tasks_per_node if
+                                  self.num_tasks_per_node else 1)
+            min_num_tasks = (abs(self.num_tasks) if self.num_tasks < 0 else
+                             num_tasks_per_node)
             try:
-                self._num_tasks = (self.guess_num_tasks(abs(self.num_tasks)) if
-                                   self.num_tasks != 0 else
-                                   self.guess_num_tasks())
-                getlogger().debug('flex_alloc_tasks: setting num_tasks to %s' %
-                                  self._num_tasks)
+                guessed_num_tasks = self.guess_num_tasks()
+                if guessed_num_tasks >= min_num_tasks:
+                    self._num_tasks = guessed_num_tasks
+                    getlogger().debug('flex_alloc_tasks: setting num_tasks '
+                                      'to %s' % self._num_tasks)
+                else:
+                    raise JobError(
+                        'not enough nodes satisfying the minimum '
+                        'number of tasks required: %s < %s' %
+                        (guessed_num_tasks, min_num_tasks))
+
             except NotImplementedError as e:
                 raise JobError('guessing number of tasks is not implemented '
                                'by the backend') from e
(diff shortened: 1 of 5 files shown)


reframe/core/schedulers/__init__.py
```diff
@@ -243,11 +245,15 @@ def prepare(self, commands, environs=None, **gen_opts):
...
-    def guess_num_tasks(self):
+    def guess_num_tasks(self, min_num_tasks=1):
         if isinstance(self.sched_flex_alloc_tasks, int):
             if self.sched_flex_alloc_tasks <= 0:
                 raise JobError('invalid number of flex_alloc_tasks: %s' %
                                self.sched_flex_alloc_tasks)
+            elif self.sched_flex_alloc_tasks < min_num_tasks:
+                raise JobError('invalid number of flex_alloc_tasks: %s > '
+                               '%s (min number of tasks)' %
+                               self.sched_flex_alloc_tasks, min_num_tasks)
```
vkarak: The logic of the minimum tasks should not go inside the `guess_num_tasks()`. `guess_num_tasks()` should return the number of tasks that it has found; that's it. As before. Inside the prepare, you compare against the minimum and either return or raise a `JobError`. Going even further, the only reason I see raising a `JobError` from the `guess_num_tasks()` currently is just for having a more precise error message. You could even replace the raising with `return 0` and just log the message. In this case, the implementation will become more consistent.

reframe/core/schedulers/__init__.py
```diff
@@ -222,12 +222,22 @@ def sched_exclusive_access(self):
     def prepare(self, commands, environs=None, **gen_opts):
         environs = environs or []
         if self.num_tasks <= 0:
+            num_tasks_per_node = (self.num_tasks_per_node if
+                                  self.num_tasks_per_node else 1)
```
vkarak: Rewrite as follows:

```python
num_tasks_per_node = self.num_tasks_per_node or 1
```

reframe/core/schedulers/__init__.py
```diff
@@ -222,12 +222,22 @@ def sched_exclusive_access(self):
...
+            min_num_tasks = (abs(self.num_tasks) if self.num_tasks < 0 else
```
vkarak: You know it is negative, so you can simply use `-self.num_tasks`. You might be able to fit this in a single line if you rewrite it as follows:

```python
min_num_tasks = -self.num_tasks if self.num_tasks else num_tasks_per_node
```

reframe/core/schedulers/__init__.py
```diff
@@ -222,12 +222,22 @@ def sched_exclusive_access(self):
...
+                             num_tasks_per_node)
             try:
-                self._num_tasks = (self.guess_num_tasks(abs(self.num_tasks)) if
-                                   self.num_tasks != 0 else
-                                   self.guess_num_tasks())
-                getlogger().debug('flex_alloc_tasks: setting num_tasks to %s' %
-                                  self._num_tasks)
+                guessed_num_tasks = self.guess_num_tasks()
+                if guessed_num_tasks >= min_num_tasks:
```
vkarak: I would invert the check here so as to first raise the exception and gain an identation level afterwards.

reframe/core/schedulers/__init__.py
```diff
@@ -222,12 +222,22 @@ def sched_exclusive_access(self):
...
+                    self._num_tasks = guessed_num_tasks
+                    getlogger().debug('flex_alloc_tasks: setting num_tasks '
+                                      'to %s' % self._num_tasks)
+                else:
+                    raise JobError(
+                        'not enough nodes satisfying the minimum '
+                        'number of tasks required: %s < %s' %
```
vkarak: I would simply say:

`could not find enough nodes: required X, found Y`

But you should better convert the num tasks you get to node counts.

reframe/core/schedulers/__init__.py
```diff
@@ -222,12 +222,22 @@ def sched_exclusive_access(self):
...
+                        (guessed_num_tasks, min_num_tasks))
+
             except NotImplementedError as e:
                 raise JobError('guessing number of tasks is not implemented '
```
vkarak: `flexible task allocation is not supported by this backend`

PR #623, opened 2018-12-03
[bugfix] Improve handling of bad tests to avoid frontend crashing

rsarm:
Closes #554

PR #654, opened 2019-01-19
[refactor] Refactor `PrettyPrinter` and enable colors in the general logger

vkarak:
- `PrettyPrinter` is now a proxy to the ReFrame's logger and takes particular care in formatting specially the progress output.
- The general framework logger is now also capable of colorizing its output. This allows errors and warnings to be colorized when logged from anywhere in the framework. It also prefixes errors and warnings with the program name.

This is useful for #623.


reframe/utility/__init__.py
```diff
@@ -310,3 +311,47 @@ def __eq__(self, *args, **kwargs):
...
+
+
+class Colorizer(abc.ABC):
+    """Base class for colorizing strings on the terminal."""
+
+    def __repr__(self):
+        return debug.repr(self)
+
+    @abc.abstractmethod
+    def colorize(string, foreground, background):
```
teojgo: Should this be also `@staticmethod`?

vkarak: You mean `@classmethod`? Yes, you have a point.

teojgo: No I mean the `@staticmethod` since you don't need access to the class itself.

teojgo: Actually yes, the `@classmethod` would be even better to access the corresponding escape sequences, etc.

vkarak: I'll take a look and fix it.

vkarak: I refactored completely the coloring code.

victorusu:
@vkarak and @teojgo, Can I merge?

vkarak:
@victorusu Yeah, sure! I wanted to do that yesterday.

PR #662, opened 2019-01-28
[test] Extend DGEMM test for using MKL

lucamar:
I have added the build and run of `DGEMM` using `PrgEnv-intel` with `MKL`. 
The performance values are still the same, I have tested `Dom` for the time being and the performance fits the current range. 
I have inserted a new tag `benchmark`, therefore the performance thresholds might change with respect to `diagnostic`.


lucamar:
Note that you need to run `reframe -P normal` on `Dom` to get the node assignment correctly.

cscs-checks/libraries/math/dgemm.py
```diff
@@ -33,9 +33,26 @@ def __init__(self):
         }
 
         self.maintainers = ['AJ', 'VH', 'VK']
-        self.tags = {'diagnostic'}
+        self.tags = {'benchmark', 'diagnostic'}
 
     def setup(self, partition, environ, **job_opts):
+
+        if environ.name.startswith('PrgEnv-gnu'):
+            envname = 'PrgEnv-gnu'
```
victorusu: As far as I can see, the `envname` variable is defined but never used.

cscs-checks/libraries/math/dgemm.py
```diff
@@ -33,9 +33,26 @@ def __init__(self):
...
+            envname = 'PrgEnv-intel'
+            self.build_system.cflags += [
+                '-qopenmp', '-DMKL_ILP64', '-I${MKLROOT}/include',
+                '-Wl,--start-group',
+                '${MKLROOT}/lib/intel64/libmkl_intel_ilp64.a',
+                '${MKLROOT}/lib/intel64/libmkl_intel_thread.a',
+                '${MKLROOT}/lib/intel64/libmkl_core.a',
+                '-Wl,--end-group',
+                '-liomp5', '-lpthread', '-lm', '-ldl']
+        else:
```
victorusu: I think you can remove completely this `else:` branch.

cscs-checks/libraries/math/dgemm.py
lucamar: Is it used if I select on the command line `-p PrgEnv-gnu` when I invoke `reframe`?

cscs-checks/libraries/math/dgemm.py
```diff
@@ -33,9 +33,22 @@ def __init__(self):
...
+            self.build_system.cflags += ['-fopenmp']
+        elif environ.name.startswith('PrgEnv-intel'):
+            self.build_system.cflags += [
+                '-qopenmp', '-DMKL_ILP64', '-I${MKLROOT}/include',
+                '-Wl,--start-group',
+                '${MKLROOT}/lib/intel64/libmkl_intel_ilp64.a',
+                '${MKLROOT}/lib/intel64/libmkl_intel_thread.a',
+                '${MKLROOT}/lib/intel64/libmkl_core.a',
+                '-Wl,--end-group',
+                '-liomp5', '-lpthread', '-lm', '-ldl']
```
vkarak: To be typically correct the options should be as follows:

```python
self.build_system.cppflags = ['-DMKL_ILP64', '-I{MKLROOT}/include']
self.build_system.cflags = ['-qopenmp']
self.build_system.ldflags = ['--the-rest']
```

lucamar: Thanks @vkarak, I have modified the code as you suggested

vkarak:
@victorusu Can you recheck and approve?

vkarak:
I will move the test to microbenchmarks

victorusu:
I will merge, as soon as the tests pass.

PR #683, opened 2019-02-18
[feat] Support command substitution in variable expansion

### LINKED ISSUE #682: PR #635 has broken Kesch environments
The error is visible also in the failure of #681 [here](https://jenkins.cscs.ch/blue/organizations/jenkins/ReframeCI/detail/ReframeCI/1528/pipeline). It appears that loading the environment during `setup()` crashes if an environment variable contains inline commands.

vkarak:
- Extend `expandvars()` to support command substitutions.
- Replace almost all usages of `os.path.expandvars()` with our version.

Fixes #682.

I still need to update a bit the documentation. I believe this is a feature rather than a bugfix.

### DIFF
--- a/reframe/utility/os_ext.py
+++ b/reframe/utility/os_ext.py
@@ -314,12 +314,12 @@ def expandvars(path):
     This function is the same as ``os.path.expandvars()``, except that it
     understands also the syntax: $(cmd)`` or `cmd`.
     """
-    cmd_subst = re.compile(r'(`|\$\()(.*)(`|\))')
+    cmd_subst = re.compile(r'`(.*)`|\$\((.*)\)')
     cmd_subst_m = cmd_subst.search(path)
     if not cmd_subst_m:
         return os.path.expandvars(path)
 
-    cmd = cmd_subst_m.group(2)
+    cmd = cmd_subst_m.groups()[0] or cmd_subst_m.groups()[1]
 
     # We need shell=True to support nested expansion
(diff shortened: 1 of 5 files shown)


reframe/utility/os_ext.py
```diff
@@ -305,3 +305,25 @@ def git_repo_exists(url, timeout=5):
...
+
+
+def expandvars(path):
+    """Expand environment variables in ``path`` and
+        perform any command substitution
+
+    This function is the same as ``os.path.expandvars()``, except that it
+    understands also the syntax: $(cmd)`` or `cmd`.
+    """
+    cmd_subst = re.compile(r'(`|\$\()(.*)(`|\))')
```
teojgo: Replace regex with 
```
`.*`|\$\(.*\)
```
Since now matches like:
``` 
`my_command) are allowed
```

vkarak: The problem with your suggestion as the proper regex is that you cannot perform nicely the grouping. You will need to have two groups and sometimes you need to extract group 1, some others group2. I will continue using mine and check explicitly the prefix and suffix.

teojgo: Yes, you can do:
```
cmd_subst = re.compile(r'`(.*)`|\$\((.*)\)')
```
And then use:
```
cmd = cmd_subst_m.groups()[0] or cmd_subst_m.groups()[1]
```

PR #684, opened 2019-02-19
[test] Add Fulen as a supported system in CSCS config

### LINKED ISSUE #609: Enable Fulen
The virtual partitions should be
* fulen:normal
* fulen:fat
* fulen:gpu

rsarm:
Fixes #609


ci-scripts/ci-runner.bash
```diff
@@ -5,6 +5,7 @@ CI_FOLDER=""
 CI_GENERIC=0
 CI_TUTORIAL=0
 CI_EXITCODE=0
+CI_LOCAL=0
```
vkarak: This is equivalent to `CI_GENERIC`, isn't it?

config/cscs.py
```diff
@@ -99,6 +99,42 @@ class ReframeSettings:
...
+                        'scheduler': 'nativeslurm',
+                        'environs': ['PrgEnv-gnu'],
+                        'descr': ('Compute nodes - '
+                                  'default partition'),
+                    },
+
+                    'fat': {
+                        'scheduler': 'nativeslurm',
+                        'environs': ['PrgEnv-gnu'],
+                        'access':  ['--partition fat'],
```
vkarak: There is a redundant space here. Either align it to the above value or remove the space.

vkarak: Same in other places.

config/cscs.py
```diff
@@ -99,6 +99,42 @@ class ReframeSettings:
...
+                        'environs': ['PrgEnv-gnu'],
+                        'descr': 'Login nodes',
+                        'max_jobs': 1
+                    },
+
+                    'normal': {
+                        'scheduler': 'nativeslurm',
+                        'environs': ['PrgEnv-gnu'],
+                        'descr': ('Compute nodes - '
+                                  'default partition'),
```
vkarak: Put everything in a single line.

config/cscs.py
```diff
@@ -99,6 +99,42 @@ class ReframeSettings:
...
+                        'environs': ['PrgEnv-gnu'],
+                        'access':  ['--partition fat'],
+                        'descr': ('High-memory compute nodes'),
+                    },
+
+                    'gpu': {
+                        'scheduler': 'nativeslurm',
+                        'environs': ['PrgEnv-gnu'],
+                        'access':  ['--partition gpu'],
+                        'descr': ('Hybrid compute nodes'),
```
vkarak: Remove parentheses.

ci-scripts/ci-runner.bash
teojgo: Not exactly because in the `CI_GENERIC` case no added regression tests are actually executed. Only the public unit tests are run.

vkarak: So you run the unit tests with the generic config and the regression tests with the CSCS config?

teojgo: Yes, since regression tests including `fulen` should be tested.

vkarak: That looks like a workaround just for Fulen. I don't agree. Either we enable it fully as for all other systems or run with only the public settings overall. We might need to increase the priority of the Jenkins jobs? A special account? @victorusu What do you think?

vkarak: Guys, @victorusu @teojgo, what do we do with this?

vkarak:
Adding Fulen to CI has no added value. We can achieve the same with VMs on Pollux. We should only keep the configuration entry in ReFrame's entry.

vkarak:
@victorusu Can you approve this PR?

PR #687, opened 2019-02-22
[doc] Remove link to prior to 2.6.1 documentation

kraushm:
fixes the Jira issue: https://jira.cscs.ch/browse/UES-219


kraushm:
I guess it makes sense to then also remove the `old` folder inside the docs?

PR #689, opened 2019-02-22
[test] Enlarge buffer sizes for H2D and D2D multi GPU tests

sebkelle1:
increased from 32MB up to 1024MB


cscs-checks/cuda/multi_gpu.py
```diff
@@ -30,11 +30,14 @@ def __init__(self):
         self.sourcepath = 'bandwidthtestflex.cu'
         self.executable = 'gpu_bandwidth_check.x'
 
-        # NOTE: Perform a range of bandwidth tests from 2MB to 32MB
-        # with 2MB increments to avoid initialization overhead in bandwidth
+        # NOTE: Perform a range of bandwidth tests from 64MB to 1024MB
+        # with 64MB increments
```
vkarak: I would go for a single buffer size. If we ever need smaller buffer sizes, we will lower the min buffer size.

PR #690, opened 2019-02-25
[test] Add description to the collective halo tests

ajocksch:


cscs-checks/mch/collectives_halo.py
```diff
@@ -2,7 +2,7 @@
 import reframe.utility.sanity as sn
 
 
-class CollectivesBaseTest(rfm.RegressionTest):
+class CommunicationBaseTest(rfm.RegressionTest):
```
vkarak: Since it's not a test, but rather a base class, write it as follows:

```suggestion
class CommunicationTestBase(rfm.RegressionTest):
```

cscs-checks/mch/collectives_halo.py
```diff
@@ -102,13 +105,14 @@ def __init__(self, variant):
                                  'default': 0.0138493
                              }
                          })
+        self.descr = 'Alltoall communication; Meteoswiss check'
```
vkarak: I don't see the reason for adding the `Meteoswiss check` annotation. We do have the `mch` tag already.

cscs-checks/mch/collectives_halo.py
```diff
@@ -123,5 +127,6 @@ def __init__(self, variant):
                                  'default': 2.53509
                              }
                          })
+        self.descr = 'Halo-cell exchange; Meteoswiss check'
```
vkarak:
```suggestion
        self.descr = 'Halo-cell exchange test'
```

vkarak:
@jenkins-cscs retry daint

PR #691, opened 2019-02-25
[test] Extend DGEMM test to build with OpenBLAS

lucamar:
I have inserted the `PrgEnv-gnu-nompi` on `Kesch` in order to build `DGEMM` using `OpenBLAS`: 
 - we don't need to link to the MPI library as we only use a single task to run
 - only this programming environment loads the module `gmvolf` (which loads the module `openblas`), while `PrgEnv-gnu` only loads `gmvapich2/17.02_cuda_8.0_gdr` (see `config/cscs.py`)


lucamar:
Note that `Dom` is currently offline, therefore the CI will not manage to run.

cscs-checks/microbenchmarks/dgemm/dgemm.py
```diff
@@ -14,8 +14,17 @@ def __init__(self):
...
 
-        self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc']
-        self.valid_prog_environs = ['PrgEnv-gnu', 'PrgEnv-intel']
+        self.valid_systems = [
+            'daint:gpu', 'daint:mc',
+            'dom:gpu', 'dom:mc',
+            'kesch:cn', 'kesch:pn'
+        ]
+
+        if self.current_system.name == ['daint', 'dom']:
```
vkarak: This will never be true! You should write it as follows:

```suggestion
        if self.current_system.name in ['daint', 'dom']:
```

lucamar:
Thanks for checking @vkarak! I have updated the if test as you suggested.

vkarak:
@jenkins-cscs retry daint

lucamar:
It looks like the error was the following in `reframeCI`:
```
E           sbatch: error: Batch job submission failed: Socket timed out on send/recv operation
```

vkarak:
@lucamar No problem, we will retry it. This is a Slurm problem.

PR #695, opened 2019-02-28
[test] Change resources dir for GROMACS test

victorusu:
This is the first effort in the series of extending GROMACS test cases


vkarak:
@jenkins-cscs retry none

PR #700, opened 2019-03-01
[test] Hardcode LD_PRELOAD path for Kesch

ajocksch:
This should fix the performance problem with the collectives_halo.py benchmark. It has been tested for 2 nodes only.


vkarak:
@ajocksch This PR is redundant. The fact that you need to hardcode the `LD_PRELOAD` here is that `PrgEnv-gnu` on Kesch loads the old mvapich. I will close this PR.

PR #704, opened 2019-03-06
[bugfix] Fix error message in case of performance errors

### LINKED ISSUE #677: Performance failure report prints 'sanity error' which is confusing
This is known "bug" and is due to the fact that mechanism of sanity and performance patterns is the same; the same sanity functions are used in both. The sanity functions always raise a `SanityError` in case of errors, which is reported as "sanity error". We can fix that by reraising sanity errors as a `PerformanceError`, which will be reported differently.

teojgo:
* Create the `PerformanceError` exception class when used
  the performance of a test is not within the reference values.
* Modify the corresponding unittests to except a `PerformanceError`.
* Refactor how exception messages are handled.

Closes #677


reframe/core/pipeline.py
```diff
@@ -1120,7 +1121,10 @@ def check_performance(self):
 
             for val, reference in perf_values:
                 ref, low_thres, high_thres, *_ = reference
-                evaluate(assert_reference(val, ref, low_thres, high_thres))
+                try:
+                    evaluate(assert_reference(val, ref, low_thres, high_thres))
+                except SanityError as e:
+                    raise PerformanceError(e)
```
```suggestion
                    raise PerformanceError() from e
```

reframe/core/pipeline.py
```diff
@@ -1075,7 +1076,7 @@ def sanity(self):
     def performance(self):
         try:
             self.check_performance()
-        except SanityError:
+        except (PerformanceError, SanityError):
```
vkarak: If `SanityError` is thrown here, it means that a performance value could not be extracted. This is still a sanity error that happens during the performance stage. Since now we have the distinction, it's better to catch only `PerformanceError`.

reframe/core/pipeline.py
teojgo: If I do it like that then the formatted message contains also the `:` from the `SanityError`.

vkarak: Can you copy here the output you get?

teojgo: `Reason: performance error: : 10 is beyond reference value 20 (l=18.0, u=22.0)`

vkarak: I see, but `PerformanceError(e)` is not the way to go. We don't do it anywhere else in the framework. You may improve the way that exception messages are chained, by not printing the `:` the current exception's message is empty.

reframe/core/pipeline.py
teojgo: Then `SanityError` will be raised even if the self is not strict

teojgo:
@jenkins-cscs retry all

reframe/core/exceptions.py
```diff
@@ -243,7 +243,8 @@ def format_user_frame(frame):
         return 'sanity error: %s' % exc_value
 
     if isinstance(exc_value, PerformanceError):
-        return 'performance error: %s' % exc_value
+        exc = exc_value.__cause__ or exc_value
```
vkarak: The problem with `: :` is more general and should be treated inside the `ReFrameError`.

unittests/test_policies.py
```diff
@@ -53,43 +53,43 @@ def test_runall(self):
 
         stats = self.runner.stats
         self.assertEqual(7, stats.num_cases())
-        self.assertEqual(4, stats.num_failures())
+        self.assertEqual(5, stats.num_failures())
```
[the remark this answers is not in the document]
teojgo: Because the tests fail even if they are not strict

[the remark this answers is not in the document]
vkarak: No, the unit tests are right. The problem is that the `CustomPerformanceFailureCheck` raises a `SanityError` instead of a `PerformanceError`. You should fix it there.

vkarak:
@teojgo I am taking over the changes for fixing how the error message prefixes are printed.

teojgo:
@vkarak good, I have pushed the reverted unittests so you can continue from there.

vkarak:
@teojgo Yep, I've took the changes already.

vkarak:
@jenkins-cscs retry daint

PR #716, opened 2019-03-15
[bugfix] Define the set of run environments correctly when `-p` option is used with regular expressions

vkarak:
This PR fixes the behaviour of the `-p` option when it is used with regular expressions. The `-p` option has also the effect of restricting the programming environments to run with. Although this still works fine in old-style cases, e.g., `-p PrgEnv-gnu`, it fails to do so when passing regular expressions, for example, `-p PrgEnv-(gnu|pgi)`. Although this will select all the tests that support either `PrgEnv-gnu` or `PrgEnv-pgi`, it will not run them. This PR fixes this.

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -1,6 +1,7 @@
 import inspect
 import json
 import os
+import re
 import socket
 import sys
 import traceback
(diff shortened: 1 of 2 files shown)


vkarak:
@jenkins-cscs retry kesch

PR #718, opened 2019-03-21
[bugfix] Fix string error in header file of Cuda memtest

teojgo:
* Apply a patch to fix the compilation error in the header
  file of cuda memtest.


cscs-checks/cuda/cuda_memtest_check.py
```diff
@@ -20,7 +20,8 @@ def __init__(self):
             'wget %s' % src_url,
             'tar -xzf cuda_memtest-1.2.3.tar.gz',
             'cd cuda_memtest-1.2.3',
-            'patch -p1 < ../cuda_memtest-1.2.3.patch'
+            'patch -p1 < ../cuda_memtest-1.2.3.patch',
+            'patch -p1 < ../cuda_memtest-1.2.3.patch2'
```
vkarak: Why don't you combine the two patches in one?

teojgo: I tried doing it but I could not make it work!

vkarak: That's strange. You still have the "vanilla" version in the test's resources, so you should be able to produce a single patch by comparing the latest version vs. the original one.

teojgo: Let me retry.

PR #724, opened 2019-04-01
[refactor] Refactor frontend to execute directly test cases

vkarak:
- All the test cases from a test are generated in advance, before calling the `runall()` method of `Runner`. A test case consists of (a clone of) the check, the partition and the environment on which it is running. `Runner` sees now only test cases.  Adaptations were made to the statistics and the retries feature.
- `SystemPartition` is now immutable. As a result, only the check needs to be copied in the test case.
- The redesign improves the retries feature; now only the failing test cases are retried not the whole test.

This step is a prerequisite for supporting dependencies of tests.

Closes https://jira.cscs.ch/browse/UES-281

### DIFF
--- a/reframe/frontend/executors/__init__.py
+++ b/reframe/frontend/executors/__init__.py
@@ -1,4 +1,5 @@
 import abc
+import copy
 import sys
 
 import reframe.core.debug as debug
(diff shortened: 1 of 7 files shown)


reframe/frontend/executors/__init__.py
```diff
@@ -29,9 +92,13 @@ def __init__(self, check, listeners=[]):
...
         self.zombie = False
 
+    @property
+    def testcase(self):
+        return self._case
+
     @property
     def check(self):
-        return self._check
+        return self._case.check
```
victorusu: @vkarak, I did not understand the concept...
Is the `case` part of the `check` or the `check` part of the `case`?

reframe/frontend/executors/__init__.py
```diff
@@ -53,7 +120,7 @@ def _notify_listeners(self, callback_name):
     def _safe_call(self, fn, *args, **kwargs):
         self._current_stage = fn.__name__
         try:
-            with logging.logging_context(self._check) as logger:
+            with logging.logging_context(self.check) as logger:
```
victorusu: I am a bit confused... Shouldn't we pass tests cases instead of checks here?

vkarak:
@victorusu Here is what happening:

- A test case is essentially a tuple of the check (or test), the partition and the environment. This particular check instantiation will be modified along its way down the pipeline. That's why we also store its original version (the one instantiated after loading), in order to be able to repeat (cleanly) a test case in case of retries.
- A check holds references to the current partition and environment. These references are updated for each test as soon as it enters the pipeline, i.e., in the `setup()` method. The check implements the pipeline.
- A regression task holds a test case, pipeline information and the result of the test. It also holds all the necessary information for deferring and restoring the execution of a check and is responsible for sending pipeline events to listeners (currently execution policy).

Now that things are becoming more clear, we can have a drastically different design of the internals, but I leave this as a future work. We could do the following:

- Have a very thin `RegressionTest` that does not incorporate the logic of the pipeline.
- The logic of the pipeline currently in `RegressionTest` moves completely inside the test case.

There are still things to consider in such redesign, but it could be possible and could make things clearer at the code level as well.

Hope this explains a bit the design.

reframe/frontend/executors/__init__.py
vkarak: You should not be confused with this. It's exactly as before. The only difference is that the `RegressionTask` now stores a `TestCase` rather than a check. The reason for that is that for supporting retries, we need to generate a new `TestCase` from a failed task. For that, we need a handle to the original check, which now we don't have any more, cos `runall()` acts on test cases and the cloning happens outside. So this information is stored in the `__check_orig` field of the `TestCase`.

reframe/frontend/executors/__init__.py
vkarak: The check is part of the test case. Here you are looking into the `RegressionTask`, which is a different thing. See my explanations at the general comments.

vkarak:
@teojgo What do you mean? It will appear at the "Other" section in any case. Unless we consider this a feature.

teojgo:
@vkarak, ok, I thought that we would like to support also the `refactor` tag.

PR #729, opened 2019-04-03
[bugfix] Fix crash in flexible task allocation due to refactoring of `SystemPartition`

### LINKED ISSUE #728: Flexible task allocation is broken on master
It has broken by the changes introduced in #724. More specifically, `SystemPartition` became immutable and returns read-only views of its internals. `guess_num_tasks()` tries to add some options and the `+` operator is not supported by the view:

Example backtrace:

```
FAILURE INFO for GpuBurnTest
  * System partition: dom:gpu
  * Environment: PrgEnv-gnu
  * Stage directory: /users/karakasv/Devel/reframe/stage/dom/gpu/PrgEnv-gnu/GpuBurnTest
  * Maintainers: ['AJ', 'VK', 'TM']
  * Failing phase: run
  * Reason: unexpected error: unsupported operand type(s) for +: 'SequenceView' and 'list'
  File "/users/karakasv/Devel/reframe/reframe/frontend/executors/__init__.py", line 125, in _safe_call
    return fn(*args, **kwargs)
  File "/users/karakasv/Devel/reframe/reframe/core/pipeline.py", line 1044, in run
    self._job.prepare(commands, environs, login=True)
  File "/users/karakasv/Devel/reframe/reframe/core/schedulers/slurm.py", line 100, in prepare
    super().prepare(commands, environs, **gen_opts)
  File "/users/karakasv/Devel/reframe/reframe/core/schedulers/__init__.py", line 231, in prepare
    guessed_num_tasks = self.guess_num_tasks()
  File "/users/karakasv/Devel/reframe/reframe/core/schedulers/__init__.py", line 274, in guess_num_tasks
    self.sched_access + self.options)
TypeError: unsupported operand type(s) for +: 'SequenceView' and 'list'

```

teojgo:
* Create a list with the contents of `sched_access` and append
  the `options` to it.

Fixes #728

### DIFF
--- a/reframe/core/schedulers/__init__.py
+++ b/reframe/core/schedulers/__init__.py
@@ -270,8 +270,10 @@ class Job(abc.ABC):
                           len(available_nodes))
 
         # Try to guess the number of tasks now
-        available_nodes = self.filter_nodes(available_nodes,
-                                            self.sched_access + self.options)
+        # Append `options` to a list containing the elements of `sched_access`
+        # whis is immutable.
+        available_nodes = self.filter_nodes(
+            available_nodes, list(self.sched_access) + self.options)
 
         if self.sched_flex_alloc_tasks == 'idle':
             available_nodes = {n for n in available_nodes
--- a/reframe/core/schedulers/__init__.py
+++ b/reframe/core/schedulers/__init__.py
@@ -270,10 +270,8 @@ class Job(abc.ABC):
                           len(available_nodes))
 
         # Try to guess the number of tasks now
-        # Append `options` to a list containing the elements of `sched_access`
-        # whis is immutable.
-        available_nodes = self.filter_nodes(
-            available_nodes, list(self.sched_access) + self.options)
+        available_nodes = self.filter_nodes(available_nodes,
+                                            self.sched_access + self.options)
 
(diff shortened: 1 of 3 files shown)


reframe/core/schedulers/__init__.py
```diff
@@ -270,8 +270,10 @@ def guess_num_tasks(self):
                           len(available_nodes))
 
         # Try to guess the number of tasks now
-        available_nodes = self.filter_nodes(available_nodes,
-                                            self.sched_access + self.options)
+        # Append `options` to a list containing the elements of `sched_access`
+        # whis is immutable.
+        available_nodes = self.filter_nodes(
+            available_nodes, list(self.sched_access) + self.options)
```
vkarak: It's better if the `SequenceView` supports the `+` operator and returns a new list. The introduction of `SequenceView` must be transparent to the rest of the code.

vkarak:
Perhaps, we can have a unit test the exercises the flexible allocation code, because this is not covered and we often get bugs.

reframe/core/schedulers/__init__.py
teojgo: Should addition of two `SequenceView` instances be supported?

PR #738, opened 2019-04-05
[test] Enable small scale application tests also on Daint

vkarak:
- Several tests were also refactored.

Fixes: https://jira.cscs.ch/browse/UES-314


vkarak:

PR #739, opened 2019-04-05
[test] Sort cores list in affinity test before comparing

sebkelle1:
This would be my suggestion to fix the test. I sort the list of cores, because arguably
for performance we don't care which OpenMP thread is executed on which core.
It should be enough to check that the list of cores contains no duplicates and the total
number of distinct cores is the same relative to the reference.

### DIFF
--- a/cscs-checks/prgenv/affinity_check.py
+++ b/cscs-checks/prgenv/affinity_check.py
@@ -9,12 +9,12 @@ class AffinityTestBase(rfm.RegressionTest):
                               'dom:gpu', 'dom:mc']
         self.valid_prog_environs = ['PrgEnv-gnu']
         self.build_system = 'Make'
-        self.build_system.options = ['--directory=affinity', 'MPI=1']
+        self.build_system.options = ['-C affinity', 'MPI=1']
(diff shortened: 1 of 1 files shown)


vkarak:
@sebkelle1 Please have a look at the PR title guide: https://github.com/eth-cscs/reframe/wiki/Contributing-to-ReFrame#pull-request-policy

cscs-checks/prgenv/affinity_check.py
```diff
@@ -50,10 +50,14 @@ def parse_cpus(x):
...
             sn.assert_eq(self.aff_thrds, self.ref_thrds),
             sn.assert_eq(self.aff_ranks, self.ref_ranks),
-            sn.assert_eq(self.aff_cores, self.ref_cores)])
+            sn.assert_eq(self.sn_sorted(self.aff_cores),
+                         self.sn_sorted(self.ref_cores))])
 
         super().setup(partition, environ, **job_opts)
 
+    @sn.sanity_function
+    def sn_sorted(self, lst):
```
vkarak: This is already provided by ReFrame. `sn.sorted()` should be fine.

PR #742, opened 2019-04-09
[doc] Fix reference to `exclusive_access` in tutorial

mand35:
You now named the variable exclusive_access, and not just exclusive anymore. This needs to be reflected in the documentation


mand35:
Sure. I verified, that self.exclusive does not providing the mentioned feature but self.exclusive_access does.

vkarak:
@mand35 Thanks for spotting this! It has always been `exclusive_access` the variable, so this bug has been existing forever :-)

PR #744, opened 2019-04-09
[feat] Enable environment variable substitution in shell code emitted by build systems

### LINKED ISSUE #702: Custom make flags in ReFrame
It could be useful to customize `make` flags that `ReFrame` uses when building. For instance, if a modulefile sets `mpif90=$MPIF90`, then `ReFrame` creates  the following `make` line:
```
make -j CC='mpicc' CXX='mpicxx' FC='$MPIF90' NVCC='nvcc' FCFLAGS='-O2 -acc -ta=tesla,cc35' LDFLAGS='-acc -ta:tesla:cc35 -lstdc++' NVCC_FLAGS="-arch=compute_37"
```
The command would fail in this case:
```
PIF90 compute_cuda.o openacc_cuda_mpi_cppstd.o std_cpp_call.o -o openacc_cuda_mpi_cppstd -acc -ta:tesla:cc35 -lstdc++
make: PIF90: Command not found
```

vkarak:
This was achieved by simply using double quotes instead of single ones in the emitted variables, e.g., `CXXFLAGS="-Wall -std=c++11 $MOREFLAGS". In your test you may write things like the following:

```python
self.build_system.cxxflags = '-Wall -std=c++11 $MOREFLAGS'
```

Fixes #702.

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -172,7 +172,7 @@ class Make(BuildSystem):
 
     .. code::
 
-      make -j [N] [-f MAKEFILE] [-C SRCDIR] CC='X' CXX='X' FC='X' NVCC='X' CPPFLAGS='X' CFLAGS='X' CXXFLAGS='X' FCFLAGS='X' LDFLAGS='X' OPTIONS
+      make -j [N] [-f MAKEFILE] [-C SRCDIR] CC="X" CXX="X" FC="X" NVCC="X" CPPFLAGS="X" CFLAGS="X" CXXFLAGS="X" FCFLAGS="X" LDFLAGS="X" OPTIONS
 
     The compiler and compiler flags variables will only be passed if they are
     not :class:`None`.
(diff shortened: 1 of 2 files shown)


vkarak:
@pep8speaks This is an exception, sorry!

reframe/core/buildsystems.py
```diff
@@ -246,31 +246,31 @@ def emit_build_commands(self, environ):
         fflags   = self._fflags(environ)
         ldflags  = self._ldflags(environ)
         if cc is not None:
```
teojgo: Maybe you can change all the `if <var> is not None:` with `if <var>:`

vkarak: No, it has a different meaning if it is `None`. Check [here](https://reframe-hpc.readthedocs.io/en/stable/reference.html#reframe.core.buildsystems.BuildSystem). In short if it is not `None` but empty, it will be emitted the following: `cc=''`. Same thing for the different flags.

vkarak:
@jenkins-cscs retry all

PR #748, opened 2019-04-10
[test] Update boost python test to use the latest module

teojgo:


vkarak:
@jenkins-cscs retry none

PR #749, opened 2019-04-10
[ci] Disable running modified tests on Daint

vkarak:
Also

- Temporarily remove `PrgEnv-pgi` from Dom.

### DIFF
--- a/ci-scripts/ci-runner.bash
+++ b/ci-scripts/ci-runner.bash
@@ -197,8 +197,8 @@ else
     userchecks=( $(git diff origin/master...HEAD --name-only --oneline --no-merges | \
                    grep -e '^cscs-checks/.*\.py') )
     if [[ $(hostname) =~ daint ]]; then
-        # Temporarily disable running modified checks on Daint
-        userchecks=""
+        # FIXME: temporarily disable running modified checks on Daint
+        userchecks=()
     fi
 
     if [ ${#userchecks[@]} -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


ci-scripts/ci-runner.bash
```diff
@@ -194,9 +194,13 @@ else
     fi
 
     # Find modified or added user checks
+    if [[ $(hostname) =~ daint ]]; then
+        # Temporarily disable running modified checks on Daint
+        userchecks=""
+    fi
+
     userchecks=( $(git diff origin/master...HEAD --name-only --oneline --no-merges | \
```
teojgo: This needs to go to an else block

vkarak: I will move the assignment after the git commands, in order to avoid re-indentation.

ci-scripts/ci-runner.bash
```diff
@@ -196,6 +196,10 @@ else
     # Find modified or added user checks
     userchecks=( $(git diff origin/master...HEAD --name-only --oneline --no-merges | \
                    grep -e '^cscs-checks/.*\.py') )
+    if [[ $(hostname) =~ daint ]]; then
+        # Temporarily disable running modified checks on Daint
+        userchecks=""
```
teojgo: This should be declared as an empty array, otherwise `${userchecks[@]}` equals 1.

vkarak: Sth like `userchecks=( )` ?? Thought it was easier... S**y Bash.

teojgo: Yes, `userchecks=()` should work.

PR #754, opened 2019-04-11
[test] Update magma tests

vkarak:
This PR fixes several issues with the MAGMA tests:

1. Performance thresholds were completely wrong! They were allowing indefinitely lower performances. This was fixed.
2. Units are added to the reference values.
3. Remove all unnecessary specializations.
4. Ported them to the new s/w stack (tests only for MAGMA 2.4)


vkarak:
@jenkins-cscs retry none

PR #755, opened 2019-04-12
[ci] Update deployment script

vkarak:
- Relax requirement for Python version. Since we are not building the documentation anymore, any Python >= 3.5 is ok.


vkarak:
@jenkins-cscs retry none

PR #758, opened 2019-04-12
[test] Adapt Paraview GPU check to CLE7

vkarak:


vkarak:
@jenkins-cscs retry dom

jfavre:
I have finally found the lines of ParaView Python code which causes the job to hang for ever. I will inquire with the ParaView development team. In teh meantime, it is OK to user the lower limit time as suggested by @vkarak

PR #759, opened 2019-04-12
[bugfix] Force units in gperftools report

jgphpc:
Fixes https://jira.cscs.ch/browse/MAINT-54


cscs-checks/tools/profiling_and_debugging/gperftools_mpi_omp.py
```diff
@@ -62,10 +62,10 @@ def __init__(self, lang):
             'chmod u+x %s' % (self.split_file),
         ]
         self.post_run = [
-            'pprof --text --lines %s %s &> %s' %
+            'pprof --unit=ms --text --lines %s %s &> %s' %
```
teojgo: It seems that there is no `--unit` option.

PR #760, opened 2019-04-15
[test] Add GridTools mini app benchmark

ChristopherBignamini:
Fixes UES-77.


cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,180 @@
...
+import reframe.utility.sanity as sn
+
+
+class GridToolsCheckBase(rfm.RegressionTest):
+    def __init__(self):
+        super().__init__()
+        self.descr = 'GridTools test base'
+
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.prebuild_cmd = ['module switch gcc/6.2.0 gcc/5.3.0']# TODO: fix this
```
vkarak: Why do you do this here? Normally, the following should be enough:

```python
self.modules = ['gcc/5.3.0']
```

Also why do you need to switch back to such an old GCC version?

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,180 @@
...
+        self.modules = ['/scratch/snx3000/bignamic/easybuild_install/modules/all/CMake/3.12.4', 'Boost', 'cudatoolkit']
```
vkarak: We do have CMake 3.12 installed on Daint/Dom already, so you should simply do

```python
self.modules = ['CMake', 'cudatoolkit']
```

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,180 @@
...
+        self.sourcesdir = '/scratch/snx3000/bignamic/gridtools/' #'git@github.com:eth-cscs/gridtools'
```
vkarak: That's the whole gridtools code?

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,180 @@
...
+                                         '-DBOOST_ROOT=$BOOST_ROOT',
+                                         '-DGT_ENABLE_PYUTILS=OFF',
+                                         '-DGT_TESTS_REQUIRE_FORTRAN_COMPILER=ON',
+                                         '-DGT_TESTS_REQUIRE_C_COMPILER=ON',
+                                         '-DCMAKE_EXPORT_NO_PACKAGE_REGISTRY=ON']
+
+
+        self.sanity_patterns = sn.assert_found(r'PASSED', self.stdout)
+        self.perf_patterns = {
+            'perf': sn.extractsingle(r'(?P<timer>\w+) ms total',
```
vkarak: Do not name the performance variable `perf`. What is the thing your are timing.

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,180 @@
...
+        self.num_gpus_per_node = 0
+        self.num_tasks = 1
+
+        if test_name == 'VerticalAdvectionNaive':
+            self.build_system.make_opts = ['vertical_advection_dycore_naive']
+            self.executable = 'regression/vertical_advection_dycore_naive'
+            self.executable_opts = ['150', '150', '150']
+            self.reference = {
+                'daint:mc': {
+                    'perf': (3400, None, 0.1)
```
vkarak: Put units in your performance variables.

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,180 @@
...
+class GridToolsCheckCPU(GridToolsCheckBase):
+    def __init__(self, test_name):
+        super().__init__()
+        self.descr = 'GridTools host test'
+
+        self.valid_systems = ['daint:mc', 'daint:gpu']
+        self.num_gpus_per_node = 0
+        self.num_tasks = 1
+
+        if test_name == 'VerticalAdvectionNaive':
```
vkarak: Why don't you use as test names the underscore variables so that you can easily construct your variables later on?

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,180 @@
...
+                'daint:mc': {
+                    'perf': (3300, None, 0.1)
+                },
+                'daint:gpu': {
+                    'perf': (3700, None, 0.1)
+                }
+        }
+
+@rfm.parameterized_test(['VerticalAdvectionCuda'], ['SimpleHoriDiffCuda'])
+class GridToolsCheckGPU(GridToolsCheckBase):
```
vkarak: I think the CPU/GPU can be also a parameter. The differences are really small and you can write your test in a nice way in any case with not so many `if`s, especially if you follow my suggestion about not using camel notation in the test type parameters.

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: The minimum CMake version for GT is currently 3.12.4

vkarak: I suggest installing this version through Easybuild and providing it as a standard module.

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: There is a combination of supported gcc/nvcc versions in the GT wiki (https://github.com/GridTools/gridtools, "Supported compilers") and I took the oldest supported gcc version.

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: Updated, I can clone the repo now.

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: It is wall time, variable name updated.

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: Done.

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: PR created https://github.com/eth-cscs/production/pull/1137

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: Done.

cscs-checks/libraries/gridtools/gridtools_base.py
ChristopherBignamini: Done.

vkarak:
@ChristopherBignamini Is this still a WIP?

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -1,180 +1,170 @@
...
-            self.reference = {
-                'daint:mc': {
-                    'perf': (3400, None, 0.1)
-                },
-                'daint:gpu': {
-                    'perf': (3800, None, 0.1)
+        self.variant_data = {
+            'vertical_advection_dycore_naive': {
+                'executable_opts':
+                    ['150', '150', '150'],
```
vkarak: This can be put in the previous line.

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,164 @@
...
+                }
+            }
+        }
+
+        self.build_system.make_opts = [variant]
+        self.executable = os.path.join('regression', variant)
+        self.executable_opts = self.variant_data[variant]['executable_opts']
+        self.reference = self.variant_data[variant]['reference']
+
+        self.tags = {'production'}
```
vkarak: I don't think we should run these tests daily in production, since the purpose is not to be a CI for GridTools. The tags should be just `scs` and `benchmarking`.

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,164 @@
...
+                '-DCUDA_ARCH:STRING=sm_60',
+                '-DCMAKE_CUDA_HOST_COMPILER:STRING=CC'))
+        else:
+            self.build_system.config_opts.extend((
+                '-DGT_ENABLE_BACKEND_X86:BOOL=ON',
+                '-DGT_ENABLE_BACKEND_NAIVE:BOOL=ON',
+                '-DGT_ENABLE_BACKEND_MC=ON',
+                '-DGT_ENABLE_BACKEND_CUDA:BOOL=OFF'))
+
+        self.valid_systems = ['daint:gpu']
```
vkarak: Why not Dom as well?

ChristopherBignamini: Done.

cscs-checks/libraries/gridtools/gridtools_base.py
```diff
@@ -0,0 +1,194 @@
...
+                '-DGT_ENABLE_BACKEND_NAIVE:BOOL=ON',
+                '-DGT_ENABLE_BACKEND_MC=ON',
+                '-DGT_ENABLE_BACKEND_CUDA:BOOL=OFF'))
+
+        self.valid_systems = ['daint:gpu', 'dom:gpu']
+        if is_cuda_test:
+            self.num_gpus_per_node = 1
+            self.num_tasks = 1
+        else:
+            self.valid_systems.append('daint:mc', 'dom:mc')
```
teojgo: change to `self.valid_systems.extend(['daint:mc', 'dom:mc'])`

vkarak: `+=` is preferred, cos it's more readable.

cscs-checks/libraries/gridtools/gridtools_base.py
vkarak: It fails on Dom:

```
gcc(3):ERROR:105: Unable to locate a modulefile for 'gcc/5.3.0'
```

ChristopherBignamini: I fixed this with a conditional module loading. According to gridtools wiki cuda 10 is not officially supported at the moment but it seems to work anyway (with gcc 7.3).

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -0,0 +1,203 @@
...
+        super().__init__()
+
+        # Check if this is a device check
+        is_cuda_test = "cuda" in variant
+
+        self.descr = 'GridTools test base'
+
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.modules = ['CMake', 'Boost']
+        if self.current_system.name is 'daint':
```
teojgo: Use `==` instead of `is`

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -0,0 +1,203 @@
...
+            self.modules.append('gcc/5.3.0')
+        else:
+            self.modules.append('gcc')
+        if is_cuda_test:
+            if self.current_system.name is 'daint':
```
teojgo: You don't need to make the system distinction if you use `craype-accel-nvidia60` instead of the `cudatoolkit` module.

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -0,0 +1,203 @@
...
+                        ['vertical_advection_dycore_mc'],
+                        ['simple_hori_diff_naive'], ['simple_hori_diff_mc'],
+                        ['vertical_advection_dycore_cuda'],
+                        ['simple_hori_diff_cuda'])
+class GridToolsCheck(rfm.RegressionTest):
+    def __init__(self, variant):
+        super().__init__()
+
+        # Check if this is a device check
+        is_cuda_test = "cuda" in variant
```
teojgo: Use single quotes for string.

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self.descr = 'GridTools test base'
+
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        self.modules = ['CMake', 'Boost']
+        if self.current_system.name is 'daint':
+            self.modules.append('gcc/5.3.0')
+        else:
+            self.modules.append('gcc')
```
teojgo: For me I don't need to load gcc for dom, it is already loaded by `PrgEnv-gnu`. Only if you wanted a particular version it would make sense to append it.

vkarak: Also the gcc/5.3 will fail as soon as we upgrade Daint and we will have to change the test. The test should target Dom only, but Daint should still be on its valid systems so as to be ready.

cscs-checks/libraries/gridtools/gridtools_tests.py
ChristopherBignamini: Done.

cscs-checks/libraries/gridtools/gridtools_tests.py
ChristopherBignamini: Done.

cscs-checks/libraries/gridtools/gridtools_tests.py
ChristopherBignamini: Done.

cscs-checks/libraries/gridtools/gridtools_tests.py
ChristopherBignamini: Yes sorry, this is useless.

ChristopherBignamini: @vkarak what should I do for Daint+gcc 5.3.0? What if we leave that part as it is and then we delete the gcc/5.3.0 loading command as soon as that version will not be available anymore on Daint? Otherwise, is it ok if I remove Daint from the valid system list and I leave the reference numbers?

vkarak: You remove `gcc/5.3.0` completely and you still keep Daint in valid systems. The tests in master target CLE7, so we don't care if it doesn't run on current Daint.

ChristopherBignamini: Ok, done.

vkarak:
@jenkins-cscs retry daint

vkarak:
@jenkins-cscs retry none

PR #761, opened 2019-04-15
[doc] Document the `--performance-report` command line option

### LINKED ISSUE #756: Document the `--performance-report` option

rsarm:
Fixes #756.


docs/running.rst
```diff
@@ -515,6 +515,105 @@ They are summarized below:
...
 * ``--max-retries NUM``: Specify the maximum number of times a failed regression test may be retried (default: 0).
 
+
+
+Generating a Performance Report
+------------------------------
+
+A report of the performance values extracted during the regression tests can be generated by running ReFrame with the command line option `--performance-report`.
+
+The performance report is printed after the output of the regression and has the following format.
```
teojgo: The performance report is printed after the output of the regression tests and has the following format:

docs/running.rst
```diff
@@ -515,6 +515,105 @@ They are summarized below:
...
+            * ...
+        - PrgEnv2
+            * perf_variable1: <value> <units>
+            * perf_variable2: <value> <units>
+            * ...
+    ------------------------------------------------------------------------------
+
+where the performance variables are listed by programming environment and partition for each of the tests that are run.
+
+For instance
```
teojgo: For instance:

docs/running.rst
```diff
@@ -515,6 +515,105 @@ They are summarized below:
...
+
+  .. code-block:: bash
+  reframe -C tutorial/config/settings.py -c tutorial/example7.py -r --performance-report
+
+prints something like this
```
teojgo: is going to produce the following output:

docs/running.rst
```diff
@@ -515,6 +515,105 @@ They are summarized below:
...
+  - daint:gpu
+     - PrgEnv-cray
+        * perf: 49.994311 Gflop/s
+     - PrgEnv-gnu
+        * perf: 50.748701 Gflop/s
+     - PrgEnv-pgi
+        * perf: 49.844147 Gflop/s
+  ------------------------------------------------------------------------------
+
+If a test needs to be run on a situation where there are no references, the `self.reference` dictionary may be written as
```
teojgo: If a tests needs to be run without any specific reference values, the `self.reference` dictionary should be defined as follows:

docs/running.rst
```diff
@@ -515,6 +515,105 @@ They are summarized below:
...
+   
+  self.reference = {
+      '*': {
+          'perf_variable1': (0, None, None),
+          'perf_variable2': (0, None, None)
+          ...
+      }
+  }
+
+This will pass the performance checks and the performance report will be printed.
```
teojgo: According to the above, the test is going to pass the performance test and the performance report is going to be printed as expected.

docs/running.rst
```diff
@@ -515,6 +515,105 @@ They are summarized below:
 
 * ``--max-retries NUM``: Specify the maximum number of times a failed regression test may be retried (default: 0).
 
+
+
+Generating a Performance Report
+------------------------------
```
teojgo: Extend the `---` to the end of the header name

PR #762, opened 2019-04-15
[test] Add flexible HPCG GPU benchmark test

sebkelle1:
Adapted from the original test for the full Piz Daint.


cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
```diff
@@ -163,3 +163,65 @@ def setup(self, partition, environ, **job_opts):
...
+                'gflops': (94.7, -0.1, None, 'Gflop/s')
+            },
+        }
+
+    @property
+    @sn.sanity_function
+    def num_tasks_assigned(self):
+        return self.job.num_tasks
+
+    def setup(self, partition, environ, **job_opts):
```
teojgo: The is no need for a setup methods since every information is available in the `__init__`.

vkarak: I agree with @teojgo. You can move everything to `__init__`.

vkarak:
@teojgo Can you also approve this PR so that we can merge it?

PR #766, opened 2019-04-17
[test] Update IPM check for PE/19.03

jgphpc:
Will fix https://jira.cscs.ch/browse/MAINT-70

### DIFF
--- a/cscs-checks/tools/profiling_and_debugging/ipm_mpi_omp.py
+++ b/cscs-checks/tools/profiling_and_debugging/ipm_mpi_omp.py
@@ -5,7 +5,7 @@ import reframe.utility.sanity as sn
 
 
 @rfm.required_version('>=2.14')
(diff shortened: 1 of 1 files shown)


cscs-checks/tools/profiling_and_debugging/ipm_mpi_omp.py
```diff
@@ -5,7 +5,7 @@
 
 
 @rfm.required_version('>=2.14')
-@rfm.parameterized_test(['C'], ['C++'], ['F90'])
+@rfm.parameterized_test(['C++'], ['F90'])
```
vkarak: Don't you need C anymore?

jgphpc: No, if it works in C++, it will work in C (+ it will take less time to run)

PR #769, opened 2019-04-18
[feat] Implement test dependency API

vkarak:
- Extend RegressionTest API to allow defining and retrieving dependencies between tests
- Build dependency graph of test cases
- Corresponding unit tests


reframe/frontend/dependency.py
```diff
@@ -0,0 +1,80 @@
...
+
+import reframe as rfm
+from reframe.core.exceptions import DependencyError
+
+
+def build_deps(cases):
+    """Build dependency graph from raw test cases.
+
+    The dependency graph is represented in adjacency list format and is stored
+    as a Python dictionary.
```
vkarak: This is inaccurate.

reframe/core/pipeline.py
```diff
@@ -1187,6 +1198,35 @@ def cleanup(self, remove_files=False, unload_env=True):
...
+        if (subdeps is not None and
+            not isinstance(subdeps, typ.Dict[str, typ.List[str]])):
+            raise TypeError("subdeps argument must be of type "
+                            "`Dict[str, List[str]]' or `None'")
+
+        self._userdeps.append((target, how, subdeps))
+
+    def getdep(self, target, environ):
+        if self._case() is None:
+            raise DependencyError('no test case is associated with this test')
```
vkarak: This is not tested.

reframe/frontend/executors/__init__.py
```diff
@@ -22,14 +23,33 @@ class TestCase:
...
         return iter([self.__check, self.__partition, self.__environ])
 
+    def __hash__(self):
+        return (hash(self.check.name) ^
+                hash(self.partition.fullname) ^
+                hash(self.environ.name))
+
+    def __eq__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
```
vkarak: `__hash__` and `__eq__` are not tested.

unittests/test_dependencies.py
```diff
@@ -0,0 +1,8 @@
+import unittest
+
+
```
vkarak: Too many blank lines here.

unittests/test_policies.py
```diff
@@ -383,3 +393,174 @@ def test_poll_fails_busy_loop(self):
...
+        has_edge = TestDependencies.has_edge
+        num_deps = TestDependencies.num_deps
+        find_check = TestDependencies.find_check
+        find_case = TestDependencies.find_case
+
+        cases = executors.generate_testcases(self.loader.load_all())
+        deps = dependency.build_deps(cases)
+
+        # Check DEPEND_FULLY dependencies
+        assert num_deps(deps, 'Test1_fully') == 8
```
vkarak: I switched intentionally to pytest notation. I find it less verbose and more integrated to the language. I suggest new unit tests to follow this notation and gradually migrate the older tests.

vkarak:
@jenkins-cscs retry dom

reframe/frontend/executors/__init__.py
```diff
@@ -22,14 +23,33 @@ class TestCase:
...
+        self.__check._case = weakref.ref(self)
+        self.__deps = []
 
     def __iter__(self):
         # Allow unpacking a test case with a single liner:
         #       c, p, e = case
         return iter([self.__check, self.__partition, self.__environ])
 
+    def __hash__(self):
+        return (hash(self.check.name) ^
```
teojgo: Maybe just return the hash of a tuple consisting of the corresponding strings.

reframe/frontend/dependency.py
```diff
@@ -0,0 +1,85 @@
...
+    # NOTE on variable names
+    #
+    # c stands for check or case depending on the context
+    # p stands for partition
+    # e stands for environment
+    # t stands for target
+
+    graph = {}
+    for c in cases:
+        graph[c] = c
```
teojgo: Why don't you put directly the dependencies here, like:
`graph[c] = c.deps`
and after just append the dependencies on `graph[c]`?

reframe/core/pipeline.py
```diff
@@ -1187,6 +1198,35 @@ def cleanup(self, remove_files=False, unload_env=True):
             self._current_environ.unload()
             self._current_partition.local_env.unload()
 
+    # Dependency API
+    def user_deps(self):
+        return util.SequenceView(self._userdeps)
+
+    def depends_on(self, target, how=DEPEND_BY_ENV, subdeps=None):
```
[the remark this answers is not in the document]
vkarak: No, this the public API and `depends_on` is a much better when writing a test. It also matches other frameworks when defining dependencies.

unittests/resources/settings.py
```diff
@@ -90,7 +105,15 @@ class ReframeSettings:
                     'cc':  'gcc',
                     'cxx': 'g++',
                     'ftn': 'gfortran',
-                }
+                },
+                'e0': {
+                    'type': 'ProgEnvironment',
+                    'modules': ['e0'],
```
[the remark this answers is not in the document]
vkarak: It doesn't make any difference. I have put `e0`, as the module of the `e0` programming environment. I could change it, though.

reframe/frontend/executors/__init__.py
vkarak: I think this one is more efficient.

reframe/frontend/dependency.py
vkarak: I'll check that. I don't remember why I have done so.

reframe/frontend/executors/__init__.py
vkarak: I won't change this; I don't see the advantage of what you propose.

vkarak:
@jenkins-cscs retry none

PR #772, opened 2019-04-26
[test] Fix ddt+cuda/10.0 check 

jgphpc:
Will fix https://jira.cscs.ch/browse/MAINT-19


jgphpc:
@jenkins-cscs retry all

PR #777, opened 2019-05-11
[ci] Add Ault TDS to the config file

jgphpc:
Will fix https://jira.cscs.ch/browse/UES-288

### DIFF
--- a/config/cscs.py
+++ b/config/cscs.py
@@ -281,7 +281,7 @@ class ReframeSettings:
                 'PrgEnv-gnu': {
                     'type': 'ProgEnvironment',
                     # defaults were gcc/8.3.0, cuda/10.1, openmpi/4.0.0
-                    'modules': ['gcc/8.3.0', 'cuda', 'openmpi'],
+                    'modules': ['gcc', 'cuda', 'openmpi'],
                     'cc':  'mpicc',
                     'cxx': 'mpicxx',
                     'ftn': 'mpif90',
--- a/config/cscs.py
+++ b/config/cscs.py
@@ -288,7 +288,8 @@ class ReframeSettings:
             'ault': {
                 'PrgEnv-gnu': {
                     'type': 'ProgEnvironment',
-                    'modules': ['gcc/8.3.0', 'cuda/10.1', 'openmpi/4.0.1'],
+                    # defaults were gcc/8.3.0, cuda/10.1, openmpi/4.0.0
+                    'modules': ['gcc/8.3.0', 'cuda', 'openmpi'],
                     'cc':  'mpicc',
                     'cxx': 'mpicxx',
                     'ftn': 'mpif90',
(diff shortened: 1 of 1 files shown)


config/cscs.py
```diff
@@ -244,6 +284,29 @@ class ReframeSettings:
         },
 
         'environments': {
+
+            'ault': {
+                'PrgEnv-gnu': {
+                    'type': 'ProgEnvironment',
+                    'modules': ['gcc/8.3.0', 'cuda/10.1', 'openmpi/4.0.1'],
```
vkarak: I would just use the defaults so that one could easily do `-M gcc:gcc/10` when gcc 10 comes out and is installed.

config/cscs.py
```diff
@@ -11,6 +11,46 @@ class ReframeSettings:
...
+                    },
+
+                    'v100': {
+                        'scheduler': 'nativeslurm',
+                        'access':  ['-pintelv100'],
+                        'environs': ['PrgEnv-gnu'],
+                        'descr': 'Hybrid Nvidia nodes (Skylake36c/4*V100)',
+                        'max_jobs': 100,
+                        'resources': {
+                            'switches': ['--switches={num_switches}']
```
vkarak: No need for switches.

config/cscs.py
```diff
@@ -11,6 +11,46 @@ class ReframeSettings:
...
+                    },
+
+                    'skl': {
+                        'scheduler': 'nativeslurm',
+                        'access':  ['-pintel'],
+                        'environs': ['PrgEnv-gnu'],
+                        'descr': 'Multicore nodes (Skylake36c)',
+                        'max_jobs': 100,
+                        'resources': {
+                            'switches': ['--switches={num_switches}']
```
vkarak: No need for switches.

config/cscs.py
```diff
@@ -11,6 +11,46 @@ class ReframeSettings:
...
+            'ault': {
+                'descr': 'Ault TDS',
+                'hostnames': ['ault'],
+                'modules_system': 'lmod',
+                'resourcesdir': '/apps/common/UES/reframe/resources',
+                'partitions': {
+                    'login': {
+                        'scheduler': 'local',
+                        'modules': [],
+                        'access':  [],
```
vkarak: You can remove these two attributes. Default is `[]` in any case.

config/cscs.py
```diff
@@ -288,7 +280,8 @@ class ReframeSettings:
             'ault': {
                 'PrgEnv-gnu': {
                     'type': 'ProgEnvironment',
-                    'modules': ['gcc/8.3.0', 'cuda/10.1', 'openmpi/4.0.1'],
+                    # defaults were gcc/8.3.0, cuda/10.1, openmpi/4.0.0
+                    'modules': ['gcc/8.3.0', 'cuda', 'openmpi'],
```
vkarak: Why not just `gcc` here?

vkarak:
@jenkins-cscs retry none

PR #782, opened 2019-05-16
[ci] Expand Ault configuration

vkarak:


vkarak:
@jenkins-cscs retry none

PR #783, opened 2019-05-16
[test] Revise MAGMA library benchmark

ajocksch:
Fixes MAINT-61 and UES-344.

### DIFF
--- a/cscs-checks/libraries/magma/magma_checks.py
+++ b/cscs-checks/libraries/magma/magma_checks.py
@@ -40,6 +40,7 @@ class MagmaCheck(rfm.RegressionTest):
                 },
             }
         elif subtest == 'zgemm':
+            self.executable_opts = ['--range 1088:3136:1024']
             self.perf_patterns = {
                 'magma': sn.extractsingle(
                              r'MAGMA GFlops: (?P<magma_gflops>\S+)',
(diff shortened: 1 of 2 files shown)


cscs-checks/libraries/magma/src/patch.txt
```diff
@@ -73,7 +73,7 @@ diff -Naur src/testing_zgemm.cpp src_patched/testing_zgemm.cpp
      printf("%%========================================================================================================\n");
 -    for( int itest = 0; itest < opts.ntest; ++itest ) {
 +//ajajajaj    for( int itest = 0; itest < opts.ntest; ++itest ) {
-+    for( int itest = 0; itest < opts.ntest; itest+=100 ) {
++    for( int itest = 0; itest < 3; itest++ ) {
```
vkarak: It seems that what you are trying to do here, you could accomplish by a command line option: `opts.ntest` seems like an option.

cscs-checks/libraries/magma/magma_checks.py
```diff
@@ -41,79 +41,67 @@ def __init__(self, subtest):
...
-                                        self.stdout, 'cpu_gflops', float)
+                    'cublas_gflops', float, 2)
             }
             self.reference = {
                 'daint:gpu': {
-                    'magma':  (3344, -0.05, None, 'Gflop/s'),
-                    'cublas': (3709, -0.05, None, 'Gflop/s'),
-                    'cpu':    (42.8, -0.27, None, 'Gflop/s'),
+                    'magma':  (3749, -0.05, None, 'Gflop/s'),
+                    'cublas': (4050, -0.05, None, 'Gflop/s'),
```
vkarak: Are you sure about your reference values? I have updated them in Apr. using some statistical data.

ajocksch: The old values are outdated unfortunately, since the case is executed with 3 different matrix sizes, and the 3rd is used. In the past it was the 1st. The alternative would be to let the loop count down:
for( int itest = 3-1; itest >= 0; itest-- )

vkarak: That's fine. In any case we would have to update them.

vkarak:

vkarak:
@jenkins-cscs retry daint

PR #785, opened 2019-05-17
[test] Hardcode version in Intel Vtune test to prevent failure when default changes

jgphpc:
Will fix https://jira.cscs.ch/browse/MAINT-75

### DIFF
--- a/cscs-checks/tools/profiling_and_debugging/intel_vtune.py
+++ b/cscs-checks/tools/profiling_and_debugging/intel_vtune.py
@@ -45,10 +45,6 @@ class IntelVTuneAmplifierTest(rfm.RegressionTest):
         self.version_rpt = 'Intel_VTuneAmplifier_version.rpt'
         self.summary_rpt = 'Intel_VTuneAmplifier_summary.rpt'
         self.paranoid_rpt = 'Intel_VTuneAmplifier_paranoid.rpt'
-        self.pre_run = [
-            'source $INTEL_PATH/../vtune_amplifier/amplxe-vars.sh',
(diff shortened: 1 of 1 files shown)


cscs-checks/tools/profiling_and_debugging/intel_vtune.py
```diff
@@ -73,10 +69,17 @@ def setup(self, partition, environ, **job_opts):
             self.job.options = ['--constraint="mc&perf"']
 
         if self.current_system.name == 'dom':
+            vtuneversion = '2019'
             toolsversion = '579888'
         elif self.current_system.name == 'daint':
```
vkarak: I am wondering whether we need this `if`, since the test on Daint are frozen.

cscs-checks/tools/profiling_and_debugging/intel_vtune.py
```diff
@@ -73,10 +69,17 @@ def setup(self, partition, environ, **job_opts):
...
+            vtuneversion = '2018'
             toolsversion = '551022'
 
+        self.pre_run = [
+            'source $INTEL_PATH/../vtune_amplifier_%s/amplxe-vars.sh' %
+            vtuneversion,
+            'amplxe-cl -help collect |tail -20',
```
teojgo: This fits in the previous line. Also put a space before `tail`.

cscs-checks/tools/profiling_and_debugging/intel_vtune.py
teojgo: The if is not needed since the daint runs the frozen version which passes fine.

PR #786, opened 2019-05-17
[feat] Print task allocation type (standard or flexible) for tests in detailed listing

### LINKED ISSUE #746: Print the number of nodes required by a test in the detailed listing

teojgo:
Fixes #746

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -27,7 +27,7 @@ from reframe.frontend.printer import PrettyPrinter
 def format_check(check, detailed):
     lines = ['  * %s (found in %s)' % (check.name,
                                        inspect.getfile(type(check)))]
-    flex = 'flexible' if check.num_tasks <= 0 else 'inflexible'
+    flex = 'flexible' if check.num_tasks <= 0 else 'standard'
 
     if detailed:
         lines += [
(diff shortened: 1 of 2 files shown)


reframe/frontend/cli.py
```diff
@@ -32,7 +32,18 @@ def format_check(check, detailed):
...
+            '      - modules: %s' % ', '.join(check.modules)
+        ]
+
+        if check.num_tasks <= 0:
+            num_tasks_per_node = check.num_tasks_per_node or 1
+            min_num_tasks = (-check.num_tasks if check.num_tasks else
+                             num_tasks_per_node)
+            num_nodes_required = min_num_tasks // num_tasks_per_node
+            lines.append(
+                '      - required nodes: %s' % num_nodes_required)
```
vkarak: Here we duplicating the same logic we have internally. I don't think it makes sense to be so precise, because for example, you don't print the required nodes for a non-flexible test. I suggest being simple and showing only if the test is flexible or not. To check the number of nodes, the user has to look into the check as is the case for every other check.

reframe/frontend/cli.py
```diff
@@ -27,23 +27,15 @@
 def format_check(check, detailed):
     lines = ['  * %s (found in %s)' % (check.name,
                                        inspect.getfile(type(check)))]
+    flex = 'flexible' if check.num_tasks <= 0 else 'inflexible'
```
```python
    flex = 'flexible' if check.num_tasks <= 0 else 'standard'
```

vkarak: Or should we just add sth to the output if only the test is flexible?

teojgo: I think it's better to print something in any case.

vkarak:
@jenkins-cscs retry none

PR #787, opened 2019-05-17
[feat] Add high-level API for container platforms

rsarm:
Fixes https://jira.cscs.ch/browse/UES-186


vkarak:
@rsarm Rename also the `containerplatform.py` to `containers.py`.

reframe/core/containers.py
```diff
@@ -0,0 +1,108 @@
...
+            raise ContainerError('no command specified')
+
+
+class Docker(ContainerPlatform):
+    """An implementation of ContainerPlatform to run containers with
+    Docker.
+    """
+    def emit_prepare_cmds(self):
+        pass
+
```
vkarak: You should also define `validate()` in subclasses as well.

reframe/core/containers.py
```diff
@@ -0,0 +1,108 @@
...
+            raise ContainerError('no image specified.')
+
+        if not self.commands:
+            raise ContainerError('no command specified')
+
+
+class Docker(ContainerPlatform):
+    """An implementation of ContainerPlatform to run containers with
+    Docker.
+    """
```
vkarak: Everything fits in a single line.

unittests/test_containers.py
```diff
@@ -0,0 +1,68 @@
+import abc
+import unittest
+
+import pytest
+import reframe.core.containers as containers
+from reframe.core.exceptions import ContainerError
+
+
+class _ContainerPlatformTest:
```
vkarak:
```suggestion
class _ContainerPlatformTest(abc.ABC):
```

vkarak:
@rsarm Is this still a WIP?

reframe/core/containers.py
```diff
@@ -0,0 +1,109 @@
...
+
+class Docker(ContainerPlatform):
+    """An implementation of ContainerPlatform to run containers with Docker."""
+
+    def emit_prepare_cmds(self):
+        pass
+
+    def emit_launch_cmds(self):
+        super().emit_launch_cmds()
+        docker_opts = ['-v "%s":"%s"' % mp for mp in self.mount_points]
```
teojgo: You should use `--mount` as described here https://docs.docker.com/storage/bind-mounts/ since the stage dir is already created on the host machine.

teojgo: As discussed the `-v` option is ok for our case.

PR #791, opened 2019-05-20
[feat] Print performance variable name in case of performance errors

### LINKED ISSUE #753: Message about performance errors is ambiguous if multiple performance variables are present
The message does not specify the performance variable that has not met the criteria, so the use has to look into the test or use `--performance-report` to figure out which one it was.

```
  * Reason: performance error: 2448.454 is beyond reference value 3330.0 (l=2997.0, u=inf)
```

teojgo:
* Allow `assert_reference` to accept a `tag` parameter.

* Use the performance tag in `assert_reference` at the
  `check_performance` stage.

Fixes #753

### DIFF
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -514,7 +514,7 @@ def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
     try:
         evaluate(assert_bounded(val, lower, upper))
     except SanityError:
-        error_msg = '%s ' % msg or ''
+        error_msg = '%s ' % msg if msg else ''
         error_msg += '{0} is beyond reference value {1} (l={2}, u={3})'
         raise SanityError(_format(error_msg, val, ref, lower, upper))
     else:
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -466,7 +466,7 @@ def assert_bounded(val, lower=None, upper=None, msg=None):
 
 
 @deferrable
-def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
+def assert_reference(val, ref, lower_thres=None, upper_thres=None, tag=None):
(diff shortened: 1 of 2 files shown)


reframe/utility/sanity.py
```diff
@@ -466,7 +466,7 @@ def assert_bounded(val, lower=None, upper=None, msg=None):
 
 
 @deferrable
-def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
+def assert_reference(val, ref, lower_thres=None, upper_thres=None, tag=None):
```
vkarak: I see two problems here:

1. The `msg` is a valid parameter used in every sanity functions and it is an omission that it is ignored here. I will open an issue for that.
2. I don't like the idea about tag as an argument. What does it represent in the context of `assert_reference()`?

vkarak: A solution would be to implement the logic for the `msg` as for other sanity functions and have the caller, i.e., the `check_performance()` set the message using the tag. But again I'm not 100% sure.

reframe/utility/sanity.py
```diff
@@ -514,7 +514,8 @@ def calc_bound(thres):
     try:
         evaluate(assert_bounded(val, lower, upper))
     except SanityError:
-        error_msg = '{0} is beyond reference value {1} (l={2}, u={3})'
+        error_msg = '%s ' % msg if msg else ''
+        error_msg += '{0} is beyond reference value {1} (l={2}, u={3})'
```
vkarak: This violates the purpose of `msg`. `msg` allows the user to override completely the default message. This is not the case here.

vkarak: I fixed it.

teojgo:
@vkarak lgtm

vkarak:
@jenkins-cscs retry dom

PR #793, opened 2019-05-20
[bugfix] Fix exception message formatting in reservation node extraction

teojgo:


reframe/core/schedulers/slurm.py
```diff
@@ -233,7 +233,7 @@ def _get_reservation_nodes(self, reservation):
             reservation_nodes = node_match[1]
         else:
             raise JobError("could not extract the nodes names for "
```
vkarak: Can you change the message as well?

```suggestion
            raise JobError("could not extract the node names for "
```

vkarak:
@jenkins-cscs retry none

PR #794, opened 2019-05-21
[bugfix] Add version checks for TMod and TMod4

### LINKED ISSUE #780: ReFrame fails with Tcl versions that use modulecmd.tcl as executable
ReFrame checks for the module system by running the `modulecmd` command.

Some versions of Tcl do not have the `modulecmd` binary. Instead they use the script `~/libexec/modulecmd.tcl` which is used by the `module` function in a similar way to `modulecmd`. They can not be replaced between each other, doing a symlink from `modulecmd.tcl` to `modulecmd` doesn't work as ReFrame gives `configuration error: could not find a sane Tmod installation`.

The details of the `module` function can be found with `type -a _moduleraw`. In the cases where the `modulecmd.tcl` is used, `module` depends on another function: `_moduleraw`. With `type -a _moduleraw` it can be seen that it used `libexec/modulecmd.tcl`.

### LINKED ISSUE #781: Tmod4 backend crashes when an unexpected installation is found
How to reproduce on Arolla:

```
ln -s /cm/local/apps/environment-modules/4.0.0/libexec/modulecmd.tcl modulecmd
PATH=.:$PATH ./bin/reframe -C config/cscs.py --system=generic -l
 PATH=.:$PATH ./bin/reframe -C config/cscs.py --system=generic -l
Command line: ./bin/reframe -C config/cscs.py --system=generic -l
Reframe version: 2.18-dev2
Launched by user: karakasv
Launched on host: arolla-ln003
Reframe paths
=============
    Check prefix      : /users/karakasv/Devel/reframe
(R) Check search path : 'checks/'
    Stage dir prefix     : /users/karakasv/Devel/reframe/stage/
    Output dir prefix    : /users/karakasv/Devel/reframe/output/
    Perf. logging prefix : /users/karakasv/Devel/reframe/perflogs
./bin/reframe: unexpected error: '_mlstatus'
  File "/users/karakasv/Devel/reframe/reframe/frontend/cli.py", line 464, in main
    rt.modules_system.unload_module(settings.reframe_module)
  File "/users/karakasv/Devel/reframe/reframe/core/modules.py", line 210, in unload_module
    self._unload_module(m)
  File "/users/karakasv/Devel/reframe/reframe/core/modules.py", line 213, in _unload_module
    self._backend.unload_module(Module(name))
  File "/users/karakasv/Devel/reframe/reframe/core/modules.py", line 463, in unload_module
    msg="could not unload module '%s' correctly" % module)
  File "/users/karakasv/Devel/reframe/reframe/core/modules.py", line 512, in _exec_module_command
    if not namespace['_mlstatus']:
KeyError: '_mlstatus'
```

It shouldn't crashed, but give an error instead.

That's for an older Tmod4 installation:
```
Modules Release 4.0.0 (2017-10-16)
```

teojgo:
* Throw `ConfigError` for TMod4 versions older than 4.1.0.
* Throw `ConfigError` for TMod versions older than 3.2.
* Change the corresponding docs.

Fixes #780  
Fixes #781

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -97,8 +97,8 @@ The valid attributes of a system are the following:
 * ``modules_system``: The modules system that should be used for loading environment modules on this system (default :class:`None`).
   Three types of modules systems are currently supported:
 
-  - ``tmod``: The classic Tcl implementation of the `environment modules <https://sourceforge.net/projects/modules/files/Modules/modules-3.2.10/>`__.
-  - ``tmod4``: The version 4 of the Tcl implementation of the `environment modules <http://modules.sourceforge.net/>`__ (versions less than 4.1.0 are not supported).
+  - ``tmod``: The classic Tcl implementation of the `environment modules <https://sourceforge.net/projects/modules/files/Modules/modules-3.2.10/>`__ (versions less than 3.2 are not supported).
+  - ``tmod4``: The version 4 of the Tcl implementation of the `environment modules <http://modules.sourceforge.net/>`__ (versions less than 4.1 are not supported).
   - ``lmod``: The Lua implementation of the `environment modules <https://lmod.readthedocs.io/en/latest/>`__.
 * ``prefix``: Default regression prefix for this system (default ``.``).
 * ``stagedir``: Default stage directory for this system (default :class:`None`).
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -97,8 +97,8 @@ The valid attributes of a system are the following:
 * ``modules_system``: The modules system that should be used for loading environment modules on this system (default :class:`None`).
   Three types of modules systems are currently supported:
 
-  - ``tmod``: The classic Tcl implementation of the `environment modules <https://sourceforge.net/projects/modules/files/Modules/modules-3.2.10/>`__ (versions less than 3.2 are not supported).
-  - ``tmod4``: The version 4 of the Tcl implementation of the `environment modules <http://modules.sourceforge.net/>`__ (versions less than 4.1 are not supported).
+  - ``tmod``: The classic Tcl implementation of the `environment modules <https://sourceforge.net/projects/modules/files/Modules/modules-3.2.10/>`__ (versions older than 3.2 are not supported).
+  - ``tmod4``: The version 4 of the Tcl implementation of the `environment modules <http://modules.sourceforge.net/>`__ (versions older than 4.1 are not supported).
   - ``lmod``: The Lua implementation of the `environment modules <https://lmod.readthedocs.io/en/latest/>`__.
 * ``prefix``: Default regression prefix for this system (default ``.``).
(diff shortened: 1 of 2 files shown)


reframe/core/modules.py
```diff
@@ -484,22 +485,30 @@ def emit_unload_instr(self, module):
...
+        version_match = re.match(r'^Modules Release (\S+)\s+',
+                                 completed.stderr)
         if not version_match:
             raise ConfigError('could not retrieve the TMod4 version')
 
-        self._version = version_match.group(1)
+        version = version_match.group(1)
+        if tuple(version.split('.')) < self.MIN_VERSION:
+            raise ConfigError(
+                'version %s of TMod4 is not supported' % version)
```
vkarak: We should say what is the minimum version required:

```
unsupported TMod4 version: {found} (required >= {min_version})
```

vkarak:
@teojgo Can you also do the same for standard Tmod. There the minimum version is `3.2`.

docs/configure.rst
```diff
@@ -97,8 +97,8 @@ The valid attributes of a system are the following:
 * ``modules_system``: The modules system that should be used for loading environment modules on this system (default :class:`None`).
   Three types of modules systems are currently supported:
 
-  - ``tmod``: The classic Tcl implementation of the `environment modules <https://sourceforge.net/projects/modules/files/Modules/modules-3.2.10/>`__.
-  - ``tmod4``: The version 4 of the Tcl implementation of the `environment modules <http://modules.sourceforge.net/>`__ (versions less than 4.1.0 are not supported).
+  - ``tmod``: The classic Tcl implementation of the `environment modules <https://sourceforge.net/projects/modules/files/Modules/modules-3.2.10/>`__ (versions less than 3.2 are not supported).
```
vkarak: `less` -> `older`.

docs/configure.rst
```diff
@@ -97,8 +97,8 @@ The valid attributes of a system are the following:
...
+  - ``tmod4``: The version 4 of the Tcl implementation of the `environment modules <http://modules.sourceforge.net/>`__ (versions less than 4.1 are not supported).
```
vkarak: `less` -> `older`.

reframe/core/modules.py
```diff
@@ -391,7 +394,14 @@ def __init__(self):
         if version_match is None or tcl_version_match is None:
             raise ConfigError('could not find a sane Tmod installation')
 
-        self._version = version_match.group(1)
+        version = version_match.group(1)
+        ver_major, ver_minor, *_ = [int(v) for v in version.split('.')]
```
vkarak: You should also guard here against `ValueError` thrown by the conversion to `int`.
In such case, you should raise an error that you couldn't parse the version information.

vkarak:
@jenkins-cscs retry none

PR #796, opened 2019-05-22
[test] Make CUDA kernel latency tests generic

vkarak:
The idea of making a test generic is to make no assumption as of what is `self.current_system`. This essentially means the following:

1. If you check against `self.current_system` or `self.current_partition` or `self.current_environ`, make sure you handle the case of an unknown system or environment.
2. If you are running on an unknown system, make sure to include `*` in `self.valid_systems` and `self.valid_prog_environs`.
3. If you are on a performance test, make sure that a `*` entry exists inside the `references` dictionary.   Its value should be `(0, None, None, 'unit')`.

Addresses [UES-339](https://jira.cscs.ch/browse/UES-339)

### DIFF
--- a/cscs-checks/microbenchmarks/kernel_latency/kernel_latency.py
+++ b/cscs-checks/microbenchmarks/kernel_latency/kernel_latency.py
@@ -21,7 +21,7 @@ class KernelLatencyTest(rfm.RegressionTest):
             self.modules = ['craype-accel-nvidia60']
             self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-pgi',
                                         'PrgEnv-gnu']
-        elif self.current_system == 'kesch':
+        elif self.current_system.name == 'kesch':
             self.num_gpus_per_node = 16
             self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-pgi']
             self.modules = ['craype-accel-nvidia35']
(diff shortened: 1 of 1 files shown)


jgphpc:
Modyfing  https://github.com/eth-cscs/reframe/blob/master/config/cscs.py#L295
to have `cuda/10.1` instead of `cuda` will fix https://jira.cscs.ch/browse/UES-339

It would help me test this PR.

jgphpc:
I get:

```
Found 4 gpu(s)
[gpu 0] Kernel launch latency: 8.67954 us
[gpu 1] Kernel launch latency: 9.11042 us
[gpu 2] Kernel launch latency: 8.99908 us
[gpu 3] Kernel launch latency: 8.63197 us
```

```
  * Failing phase: sanity
  * Reason: sanity error: 8 != 0
```

where ` 'latency': (0.0, None, None, 'us')`

vkarak:
@jgphpc There two things here:

1. There is a bug in my implementation, because I should have set `self.num_gpus_per_node = 1`.
2. This wouldn't prevent the test from failing though on an unknown system, but this is not a problem. You can't have a really generic test without any sort of hardware detection. The important things is that the test is loaded correctly and runs through on an unknown system. Then you inspect possible failures.

teojgo:
@jgphpc Also it is a flexible test therefore it will run on all idle nodes unless instructed with `--flex-alloc-tasks`

jgphpc:
Adding `self.num_gpus_per_node = 4` makes the sanity test to pass, thanks @teojgo

vkarak:
@jgphpc Yes, but you don't know that if running on an unknown system. Most probably the test will fail. At least it's already ready to run. That's the point.

vkarak:
> Modyfing [config](https://github.com/eth-cscs/reframe/blob/master/config/cscs.py#L295)
> to have `cuda/10.1` instead of `cuda` will fix [UES-339](https://jira.cscs.ch/browse/UES-339)
> 
> It would help me test this PR.

Can you do a separate PR for this? Or open an issue?

vkarak:

vkarak:
@jenkins-cscs retry none

PR #797, opened 2019-05-23
[feat] Validate test case graph

vkarak:
This PR adds the validation routine for the test graph. Dependencies are validated at the test level only, so pseudo-dependencies like the following are not allowed:

```
(t0, e1) -> (t1, e1)
(t1, e0) -> (t0, e0)
```

Fixes [UES-283](https://jira.cscs.ch/browse/UES-283).

### DIFF
--- a/reframe/frontend/dependency.py
+++ b/reframe/frontend/dependency.py
@@ -100,24 +100,29 @@ def validate_deps(graph):
 
     # Check for cyclic dependencies in the test name graph
     visited = set()
-    unvisited = list(
-        itertools.zip_longest(test_graph.keys(), [], fillvalue=None)
-    )
+    sources = set(test_graph.keys())
     path = []
-    while unvisited:
-        node, parent = unvisited.pop()
-        while path and path[-1] != parent:
-            path.pop()
 
-        adjacent = reversed(test_graph[node])
-        path.append(node)
-        for n in adjacent:
-            if n in path:
-                cycle_str = '->'.join(path + [n])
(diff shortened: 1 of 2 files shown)


vkarak:
There seems to be a problem with Python 3.5 in this PR.

teojgo:
> There seems to be a problem with Python 3.5 in this PR.

@vkarak the cycles found are correct but are not the expected ones. It might have to do with the fact that Python3.5 does not respect the insertion order of the dictionary keys. Maybe alter the test cases so that only one cycle is possible or test for either one.

reframe/frontend/dependency.py
```diff
@@ -79,3 +80,44 @@ def print_deps(graph):
...
+        for n in adjacent:
+            if n in path:
+                cycle_str = '->'.join(path + [n])
+                raise DependencyError(
+                    'found cyclic dependency between tests: ' + cycle_str)
+
+            if n not in visited:
+                unvisited.append((n, node))
+
+        visited.add(node)
```
teojgo: Here you either have to move also `node` from the unvisited, or add an if block on line 108 to continue if the node is already in the visited ones.

vkarak: `node` is popped out from the `unvisited` at the beginning of the loop (`pop()` removes the element):

```python
        node, parent = unvisited.pop()
```

If I had to check if node inside `unvisited` is already visited, then there is a semantic problem.

vkarak:
@jenkins-cscs retry none

PR #799, opened 2019-05-23
[test] Make the STREAM test generic

jgphpc:
Will fix [339](https://jira.cscs.ch/browse/UES-339)

### DIFF
--- a/cscs-checks/microbenchmarks/stream/stream.py
+++ b/cscs-checks/microbenchmarks/stream/stream.py
@@ -112,4 +112,5 @@ class StreamTest(rfm.RegressionTest):
             self.reference = self.stream_bw_reference[envname]
         except KeyError:
             self.reference = {'*': {'triad': (0.0, None, None, 'MB/s')}}
+
         super().setup(partition, environ, **job_opts)
--- a/cscs-checks/microbenchmarks/stream/stream.py
+++ b/cscs-checks/microbenchmarks/stream/stream.py
@@ -15,12 +15,8 @@ class StreamTest(rfm.RegressionTest):
         self.exclusive_access = True
         self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc',
                               'kesch:cn', 'kesch:pn', 'leone:normal']
-        if self.current_system.name in {'dom', 'daint', 'kesch', 'leone'}:
-            self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu',
-                                        'PrgEnv-intel', 'PrgEnv-pgi']
-        else:
-            self.valid_prog_environs = ['*']
-
+        self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu',
+                                    'PrgEnv-intel', 'PrgEnv-pgi']
         if self.current_system.name == 'kesch':
             self.exclusive_access = True
             self.valid_prog_environs += ['PrgEnv-cray-nompi',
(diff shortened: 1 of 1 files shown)


cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -5,15 +5,22 @@
...
-        # All available systems are supported
         self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc',
                               'kesch:cn', 'kesch:pn', 'leone:normal']
-        self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu',
-                                    'PrgEnv-intel', 'PrgEnv-pgi']
+        if self.current_system.name in {'dom', 'daint', 'kesch', 'leone'}:
+            self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu',
+                                        'PrgEnv-intel', 'PrgEnv-pgi']
+        else:
+            self.valid_prog_environs = ['*']
```
vkarak: As discussed with @rsarm this is not needed. When running on an unknown system when running as follows should be enough:

```
reframe --system=mysys -p myenv --skip-system-check --skip-prgenv-check ...
```

cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -28,8 +35,7 @@ def __init__(self):
         }
         self.sourcepath = 'stream.c'
         self.build_system = 'SingleSource'
-        self.sanity_patterns = sn.assert_found(
-            r'Solution Validates: avg error less than', self.stdout)
+        self.prebuild_cmd = ['module list']
```
vkarak: This is not portable, since you assume a module system.

cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -42,55 +48,64 @@ def __init__(self):
...
         self.tags = {'production', 'monch_acceptance'}
         self.maintainers = ['RS', 'VK']
 
     def setup(self, partition, environ, **job_opts):
-        self.num_cpus_per_task = self.stream_cpus_per_task[partition.fullname]
+        if self.current_system.name in {'dom', 'daint', 'kesch', 'leone'}:
+            self.num_cpus_per_task = self.stream_cpus_per_task[
+                partition.fullname]
+        else:
+            self.num_cpus_per_task = 1
```
vkarak: I would replace all this with the following:

```python
self.num_cpus_per_task = self.stream_cpus_per_task.get(partition.fullname, 1)
```

vkarak: And you should the same a bit later with the `cflags`.

cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -24,12 +28,11 @@ def __init__(self):
             'PrgEnv-cray': ['-homp'],
             'PrgEnv-gnu': ['-fopenmp', '-O3'],
             'PrgEnv-intel': ['-qopenmp', '-O3'],
-            'PrgEnv-pgi': ['-mp', '-O3']
+            'PrgEnv-pgi': ['-mp', '-O3'],
+            '*': ['-O3']
```
vkarak: This has no effect. This is a normal dictionary. You should do handle this case in the setup method:

```python
        self.build_system.cflags = self.prgenv_flags.get(envname, ['-O3'])
```

vkarak:
@jenkins-cscs retry none

PR #801, opened 2019-05-24
[test] Make OSU tests check generic

rsarm:
Addresses [UES-339](https://jira.cscs.ch/browse/UES-339)

### DIFF
--- a/cscs-checks/microbenchmarks/osu/osu_tests.py
+++ b/cscs-checks/microbenchmarks/osu/osu_tests.py
@@ -32,14 +32,18 @@ class AlltoallTest(rfm.RegressionTest):
             'daint:gpu': {
                 'latency': (20.73, None, 2.0, 'us')
             },
+            '*': {
+                'latency': (0, None, None, 'us')
+            },
         }
         self.num_tasks_per_node = 1
         self.num_gpus_per_node  = 1
         if self.current_system.name == 'dom':
             self.num_tasks = 6
-
-        if self.current_system.name == 'daint':
+        elif self.current_system.name == 'daint':
             self.num_tasks = 16
+        else:
+            self.num_tasks = 2
 
         self.extra_resources = {
             'switches': {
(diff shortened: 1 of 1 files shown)


jgphpc:
Minor comments:

add comment [L18](https://github.com/eth-cscs/reframe/pull/801/files#diff-30f12f0f9fdeb330c8a8ca02b6b6bfe9L18)
```
# -x sets the minimum message size
```

cscs-checks/microbenchmarks/osu/osu_tests.py
```diff
@@ -32,14 +32,18 @@ def __init__(self, variant):
...
+                'latency': (0, None, None, 'us')
+            },
         }
         self.num_tasks_per_node = 1
         self.num_gpus_per_node  = 1
         if self.current_system.name == 'dom':
             self.num_tasks = 6
-
-        if self.current_system.name == 'daint':
+        elif self.current_system.name == 'daint':
```
jgphpc: not an issue but why 6 and not 12 on dom:gpu ? and 16 on daint:gpu ? A comment would help

jgphpc:
I get:

jgphpc:
`AlltoallMonchAcceptanceTest` is obsolete but will run in generic mode.

jgphpc:
P2PBaseTest derived test will not be on the list of checks to run by default, is it because they are `self.tags |= {'monch_acceptance'}`

cscs-checks/microbenchmarks/osu/osu_tests.py
```diff
@@ -32,14 +32,18 @@ def __init__(self, variant):
...
             self.num_tasks = 16
+        else:
+            self.num_tasks = 2
```
vkarak: After second thoughts, I believe that we should not do that. The test is meant for X nodes; we should not assume less for an unknown system. The `sbatch` will fail if the request cannot be satisfied. This test has another problem though; it runs different configurations for Daint/Dom; we should do it as for the AllReduce with two variants, `small` and `large`. This is different issue.

cscs-checks/microbenchmarks/osu/osu_tests.py
vkarak: @jgphpc I think that it is just because it's easier to get an allocation and also all the application tests on Dom are with 6 nodes.

vkarak:
@jgphpc We should not take care of the Monch test at all.

PR #805, opened 2019-05-24
[refactor] Fix abstract class definitions

vkarak:
Abstract classes must derive from `abc.ABC`.


vkarak:
@jenkins-cscs retry dom

vkarak:
@jenkins-cscs retry none

PR #819, opened 2019-06-05
[test] Make the Jacobi check generic

jgphpc:
Will fix UES-384

### DIFF
--- a/cscs-checks/tools/profiling_and_debugging/notool.py
+++ b/cscs-checks/tools/profiling_and_debugging/notool.py
@@ -62,28 +62,28 @@ class JacobiNoToolHybrid(rfm.RegressionTest):
         }
         self.reference = {
             '*': {
-                'elapsed_time': (0, None, None, 'seconds')
+                'elapsed_time': (0, None, None, 's')
             }
         }
         if lang == 'C++':
             self.reference_prgenv = {
-                'PrgEnv-cray': (0.38, -0.6, None, 'seconds'),
-                'PrgEnv-gnu': (0.38, -0.6, None, 'seconds'),
-                'PrgEnv-intel': (0.38, -0.6, None, 'seconds'),
-                'PrgEnv-pgi': (18.0, -0.6, None, 'seconds'),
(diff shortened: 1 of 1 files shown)


cscs-checks/tools/profiling_and_debugging/notool.py
```diff
@@ -23,66 +21,79 @@ def __init__(self, lang):
...
+                'PrgEnv-intel': (0.17, -0.6, None, 'seconds'),
+                'PrgEnv-pgi': (18.0, -0.6, None, 'seconds'),
+            }
 
     def setup(self, partition, environ, **job_opts):
         super().setup(partition, environ, **job_opts)
-        environ_name = self.current_environ.name
-        prgenv_flags = self.prgenv_flags[environ_name]
+        envname = self.current_environ.name
+        prgenv_flags = self.prgenv_flags[envname]
```
vkarak: This is not generic. What if sb tries it on a system with a different prgenv? This will make the test crash. You should write it as follows:

```python
prgenv_flags = self.prgenv_flags.get(envname, ['put', 'some', 'default', 'flags', 'if', 'any'])
```

cscs-checks/tools/profiling_and_debugging/notool.py
```diff
@@ -23,66 +21,79 @@ def __init__(self, lang):
...
+                'elapsed_time': (0, None, None, 'seconds')
             }
         }
-        self.post_run = ['module list -t']
-
-        self.maintainers = ['JG', 'MK']
-        self.tags = {'production'}
+        if lang == 'C++':
+            self.reference_prgenv = {
+                'PrgEnv-cray': (0.38, -0.6, None, 'seconds'),
```
vkarak: Use SI units, so `s` is better.

jgphpc: https://github.com/eth-cscs/reframe/pull/819/commits/7924290da6343608a367f6192468344aa004528b

vkarak:
@jenkins-cscs retry dom

PR #820, opened 2019-06-05
[test] Generic fft revised

ajocksch:


vkarak:
This test is already generic. I've run it on an unknown system without  a problem.

PR #826, opened 2019-06-07
[test] Make MCH halo cell exchange generic

ajocksch:


cscs-checks/mch/collectives_halo.py
```diff
@@ -63,7 +63,11 @@ def __init__(self, variant, bench_reference):
         else:
             sysname = self.current_system.name
 
-        ref = bench_reference[sysname][variant]
+        try:
+            ref = bench_reference[sysname][variant]
+        except KeyError:
+            ref = {}
```
vkarak:
```suggestion
            ref = 0.0
```

cscs-checks/mch/collectives_halo.py
```diff
@@ -73,6 +77,9 @@ def __init__(self, variant, bench_reference):
             },
             'dom': {
                 'elapsed_time': (ref, None, 0.15)
+            },
+            '*': {
+                'elapsed_time': (0, None, None)
```
vkarak:
```suggestion
                'elapsed_time': (ref, None, None)
```

ajocksch:
I hope it is generic enough; the new else branch is a bit guesswork

cscs-checks/mch/collectives_halo.py
```diff
@@ -40,6 +40,16 @@ def __init__(self, variant, bench_reference):
                 '-DCUDA_COMPUTE_CAPABILITY="sm_60"'
             ]
             self.build_system.max_concurrency = 8
+        else:
+            self.num_tasks = 4
+            self.num_gpus_per_node = 1
+            self.num_tasks_per_node = 1
+            self.variables['MPICH_RDMA_ENABLED_CUDA'] = '1'
+            self.variables['MV2_USE_CUDA'] = '1'
+            self.build_system.config_opts += [
```
teojgo: Here, don't use the `sm` flag. Since it is generic it does not have to be there.

cscs-checks/mch/collectives_halo.py
```diff
@@ -40,6 +40,16 @@ def __init__(self, variant, bench_reference):
                 '-DCUDA_COMPUTE_CAPABILITY="sm_60"'
             ]
             self.build_system.max_concurrency = 8
+        else:
+            self.num_tasks = 4
+            self.num_gpus_per_node = 1
+            self.num_tasks_per_node = 1
+            self.variables['MPICH_RDMA_ENABLED_CUDA'] = '1'
```
teojgo: If this and the next environment variables are optional don't use them for the generic case.

vkarak: Exactly. The generic version should be bare bones. You don't know anything about the system and the will need adaptations in any case. The important thing is for the test not to crash.

vkarak:
@jenkins-cscs retry none

PR #830, opened 2019-06-09
[test] Make HPCG basic test generic

vkarak:
Only the basic test is converted to a generic one. The other two are already quite specific.

Fixes UES-404.


vkarak:
@jenkins-cscs retry none

PR #831, opened 2019-06-13
[test] Set `num_cpus_per_task` to the TensorFlow Horovod test

rsarm:
@jgphpc found that in this test, doing `export OMP_NUM_THREADS=$SLURM_CPUS_PER_TASK` produce the message `libgomp: Invalid value for environment variable OMP_NUM_THREADS` because as `--cpus-per-task` is not defined, `SLURM_CPUS_PER_TASK` has no value. This PR adds `self.num_cpus_per_task = 12` to the test.


vkarak:
@jenkins-cscs retry kesch

vkarak:
@jenkins-cscs retry none

PR #845, opened 2019-06-28
[bugfix] Fix how Slurm states are mapped internally by the Slurm backend

### LINKED ISSUE #844: Squeue backend unit tests fail on Dom
```
____________________________________________________________________________ TestSqueueJob.test_poll _____________________________________________________________________________

self = <unittests.test_schedulers.TestSqueueJob testMethod=test_poll>

    @fixtures.switch_to_user_runtime
    def test_poll(self):
        self.setup_user()
        self.parallel_cmd = 'sleep 2'
        self.testjob.prepare(self.commands, self.environs)
        self.testjob.submit()
>       self.assertFalse(self.testjob.finished())
E       AssertionError: True is not false

unittests/test_schedulers.py:161: AssertionError
```

teojgo:
Fixes #844

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -19,8 +19,6 @@ def slurm_state_completed(state):
         'BOOT_FAIL',
         'CANCELLED',
         'COMPLETED',
-        'CONFIGURING',
-        'COMPLETING',
         'DEADLINE',
         'FAILED',
         'NODE_FAIL',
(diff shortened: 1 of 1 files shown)


vkarak:
@jenkins-cscs retry none

vkarak:
So it appears that the job was entering the `CONFIGURING` state and the Slurm backend was erroneously reporting as finished. Then when we were checking in the unit test if the job has actually finished (an operation that triggers a status update), the job was in (most probably) the `RUNNING` phase, and `finished()` returned `False`.

vkarak:
This regression was introduced in #835.

PR #850, opened 2019-07-01
[Test] Stream Benchmark update

sebkelle1:
-Cray cc OpenMP flag change to -fopenmp
-need to set the Slurm flag --hint=nomultithread to get good performance on 2-socket systems

### DIFF
--- a/cscs-checks/microbenchmarks/stream/stream.py
+++ b/cscs-checks/microbenchmarks/stream/stream.py
@@ -24,8 +24,10 @@ class StreamTest(rfm.RegressionTest):
                                          'PrgEnv-gnu-nompi',
                                          'PrgEnv-pgi-nompi']
 
+        self.use_multithreading = False
+
         self.prgenv_flags = {
-            'PrgEnv-cray': ['-homp'],
+            'PrgEnv-cray': ['-fopenmp'],
             'PrgEnv-gnu': ['-fopenmp', '-O3'],
             'PrgEnv-intel': ['-qopenmp', '-O3'],
             'PrgEnv-pgi': ['-mp', '-O3']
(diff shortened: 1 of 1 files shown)


sebkelle1:
`FAILURE INFO for StreamTest 

  * System partition: kesch:cn

  * Environment: PrgEnv-gnu

  * Stage directory: None

  * Job type: batch job (id=-1)

  * Maintainers: ['RS', 'VK']

  * Failing phase: setup

  * Reason: environ error: could not show module 'gmvapich2/17.02_cuda_8.0_gdr': command 'modulecmd python show gmvapich2/17.02_cuda_8.0_gdr' failed with exit code 0:

=== STDOUT ===

=== STDERR ===

ModuleCmd_Display.c(151):ERROR:105: Unable to locate a modulefile for 'gmvapich2/17.02_cuda_8.0_gdr'`

seems unrelated to this PR

vkarak:
@sebkelle1 Config for Kesch is broken.

cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -56,30 +58,30 @@ def __init__(self):
         }
         self.stream_bw_reference = {
             'PrgEnv-cray': {
-                'daint:gpu': {'triad': (50223.0, -0.15, None, 'MB/s')},
-                'daint:mc': {'triad': (56643.0, -0.25, None, 'MB/s')},
-                'dom:gpu': {'triad': (50440.0, -0.15, None, 'MB/s')},
-                'dom:mc': {'triad': (56711.0, -0.25, None, 'MB/s')},
+                'daint:gpu': {'triad': (42000, -0.15, None, 'MB/s')},
```
vkarak: Should the thresholds be this high? I suggest doing ten runs, remove the slowest and fastest measurements, put as reference the median of the rest and base the threshold on the variance (but its minimum value should be 5%).

cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -24,8 +24,10 @@ def __init__(self):
                                          'PrgEnv-gnu-nompi',
                                          'PrgEnv-pgi-nompi']
 
+        self.use_multithreading = False
+
         self.prgenv_flags = {
-            'PrgEnv-cray': ['-homp'],
+            'PrgEnv-cray': ['-fopenmp'],
```
vkarak: Why not also here `-O3`. It doesn't make a big difference in this case, but it's better if you put it.

cscs-checks/microbenchmarks/stream/stream.py
```diff
@@ -56,30 +58,30 @@ def __init__(self):
         }
         self.stream_bw_reference = {
             'PrgEnv-cray': {
```
vkarak: You should merge with master to take the `PrgEnv-cray_classic` prgenv and this to the supported systems as well.

cscs-checks/microbenchmarks/stream/stream.py
vkarak: I think we have the `-O3` set in a different place in the test. I believe it's better to set all the flags here.

vkarak:

vkarak:
@sebkelle1 Test fails on Kesch. Can you have a look please?

PR #852, opened 2019-07-02
[ci] Define machines to run the CI with Jenkins params

teojgo:


teojgo:
@jenkins-cscs retry all

Jenkinsfile
```diff
@@ -3,7 +3,7 @@
 def dirPrefix = 'reframe-ci'
 def loginBash = '#!/bin/bash -l'
 def bashScript = 'ci-scripts/ci-runner.bash'
-def machinesList = ['daint', 'dom', 'kesch']
+def machinesList = params.machines.split()
```
vkarak: Where these `params` are gotten from? From the Jenkins project? Could you then remove `Kesch` from there?

teojgo: The `params` is injected by Jenkins and if the build is not triggered manually the parameters have the default values.

teojgo:
@jenkins-cscs retry all

Jenkinsfile
vkarak: So the defaults are set by the Jenkins project?

teojgo: You add a parameter and give it default values yes.

vkarak:
@jenkins-cscs retry daint

PR #853, opened 2019-07-03
[feat] Enable running containers inside the test pipeline

### LINKED ISSUE #644: Add support for containers
ReFrame should be able to launch tests making use of various containers. It should abstract away the way images are pulled or built. A similar abstraction to the build systems may be used here.

rsarm:
Fixes #644.

This is a simple test to try this feature.
```python
import reframe as rfm


@rfm.simple_test
class UbuntuContainerCheck(rfm.RunOnlyRegressionTest):
    def __init__(self, **kwargs):
        super().__init__()
        self.descr = ('Run the some commands inside a container')
        self.valid_systems = ['*']

        self.container_platform = 'Docker'
        self.container_platform.image = 'ubuntu:18.04'
        self.container_platform.workdir = '/workdir'
        self.container_platform.commands = [
            'pwd', 'ls', 'cat /etc/os-release']

            sn.assert_found(r'^' + self.container_platform.workdir,
                self.stdout),
            sn.assert_found(r'^dummy_src.txt', self.stdout),
            sn.assert_found(r'18.04.2 LTS \(Bionic Beaver\)', self.stdout),
            ])
```

I also added a `--rm` to the `docker run` command after discussing with @teojgo. Without this the container stays on status `stopped` after the test ends. `--rm` takes cares of completely clean the container after the command exits.


config/cscs.py
```diff
@@ -95,6 +100,14 @@ class ReframeSettings:
 
                     'gpu': {
                         'scheduler': 'nativeslurm',
+                        'container_platforms': {
+                            'ShifterNG': {
+                                'modules': ['shifter-ng']
+                            },
+                            'Singularity': {
+                                'modules': ['singularity']
```
vkarak: @rsarm @teojgo Do you think we will need other config parameters, e.g., default mounts, environment variables, etc.?

reframe/core/pipeline.py
```diff
@@ -640,8 +632,6 @@ def __init__(self, name=None, prefix=None):
         self._build_job = None
         self._compile_proc = None
         self.build_system = None
-
-        # Container platform
         self.container_platform = None
```
vkarak: This should be inside the `RunOnlyRegressionTest`.

reframe/core/pipeline.py
```diff
@@ -1260,6 +1247,17 @@ class RunOnlyRegressionTest(RegressionTest):
     module.
     """
 
+    #: The container platform to be used for this test.
+    #:
+    #: If the `self.container_platform.name` is defined on the test, both
```
vkarak: `self.container_platform.name` -> `self.container_platform`

reframe/core/pipeline.py
```diff
@@ -1288,12 +1286,14 @@ def run(self):
         if self.container_platform:
             self.container_platform.validate()
             self.container_platform.mount_points = [
-                (self._stagedir, self.container_platform.workdir)]
+                (self._stagedir, self.container_platform.workdir)
+            ]
+            # We replace executable and executable_opts in the case of containers.
```
vkarak: Leave a blank line before comment.

reframe/core/pipeline.py
```diff
@@ -1288,12 +1286,14 @@ def run(self):
...
             self.executable = self.container_platform.emit_launch_cmds()
-            if self.container_platform.emit_prepare_cmds():
-                self.pre_run += [self.container_platform.emit_prepare_cmds()]
-
             self.executable_opts = []
+            emit_prepare_cmds = self.container_platform.emit_prepare_cmds()
```
vkarak: `emit_prepare_cmds -> pre_run_container`

reframe/core/pipeline.py
```diff
@@ -1288,12 +1286,14 @@ def run(self):
...
+            if emit_prepare_cmds:
+                self.pre_run += [emit_prepare_cmds]
```
vkarak: I think, the `emit_prepare_cmds()` already returns a list.

vkarak:
@rsarm Can you also merge with master to get the latest changes and resolve the conflicts?

config/cscs.py
teojgo: For the moment I think we are fine. If we plan to support image building in the future we can add more.

vkarak: So Singularity works fine out of the box for MPI and GPU programs, as soon as you load the module?

teojgo: We have a module wrapper for this. In the singularity case you can add also environment variables to the test or add the module wrapper to the modules.

vkarak: Ok, my question then is would it make sense for this wrapper to be a partition configuration parameter, i.e., available by default to all checks that are using Singularity?

teojgo: The wrapper module we make available on both gpu/mc partitions (for dom it is missing at the moment but for daint it is there). I don't know what would make more sense.

reframe/core/pipeline.py
```diff
@@ -1259,6 +1258,10 @@ class RunOnlyRegressionTest(RegressionTest):
     #: .. versionadded:: 2.19
     container_platform = ContainerPlatformField('container_platform', type(None))
 
+    def __init__(self, name=None, prefix=None):
+        self.container_platform = None
```
vkarak: I would set that after calling `super().__init__()`.

reframe/core/pipeline.py
```diff
@@ -1057,6 +1060,9 @@ def run(self):
         commands = [*self.pre_run, ' '.join(exec_cmd), *self.post_run]
         environs = [self._current_partition.local_env,
                     self._current_environ, self._user_environ]
+        if self._extra_env:
+            environs.append(self._extra_env)
```
vkarak: Remove completely this part. Also the order of the environments should be the following: local_env, current_environ, container_env, user_environ.

reframe/core/pipeline.py
```diff
@@ -1288,6 +1294,11 @@ def run(self):
                                                     self.sourcesdir))
 
         if self.container_platform:
+            try:
+                self._extra_env = self._current_partition.container_environs[self.container_platform.__class__.__name__]
```
vkarak: Simply set the `environs` and remove completely the `_extra_env`.

vkarak: Also split this statement in two:

```python
cp_name = self.container_platform.__class__.__name__
cp_env = self._current_partition.container_environs[cp_name]
```

vkarak:
@rsarm Unit tests are failing as well.

reframe/core/pipeline.py
```diff
@@ -1085,14 +1098,40 @@ def run(self):
         This call is non-blocking.
         It simply submits the job associated with this test and returns.
         """
+
         if not self.current_system or not self._current_partition:
             raise PipelineError('no system or system partition is set')
 
+        if self.container_platform:
+            try:
+                cp_name = self.container_platform.__class__.__name__
```
vkarak: Rewrite this as follows:

```suggestion
                cp_name = type(self.container_platform).__name__
```

reframe/core/pipeline.py
```diff
@@ -1085,14 +1098,40 @@ def run(self):
...
+                cp_env = self._current_partition.container_environs[cp_name]
+            except KeyError as e:
+                cp_env = None
+                self.logger.debug('no configuration found for container '
+                                  'platform: %s' % e)
+
+            self.container_platform.validate()
+            self.container_platform.mount_points = [
```
vkarak: This should by  `+=`. The user may add his own mount points as well.

reframe/core/pipeline.py
```diff
@@ -1085,14 +1098,40 @@ def run(self):
...
+                (self._stagedir, self.container_platform.workdir)
+            ]
+
+            # We replace executable and executable_opts in the case of containers.
+            self.executable = self.container_platform.emit_launch_cmds()
+            self.executable_opts = []
+            pre_run_container = self.container_platform.emit_prepare_cmds()
```
```suggestion
            prepare_container = self.container_platform.emit_prepare_cmds()
```

unittests/test_containers.py
```diff
@@ -59,10 +59,48 @@ def create_container_platform(self):
 
     @property
     def exp_cmd_mount_points(self):
```
vkarak: Change all `exp_` to `expected_`.

unittests/test_containers.py
```diff
@@ -59,10 +59,48 @@ def create_container_platform(self):
...
+    @property
+    def exp_cmd_mount_points(self):
+        return ('shifter run '
+                '--mount=type=bind,source="/path/one",destination="/one" '
+                '--mount=type=bind,source="/path/two",destination="/two" '
+                "name:tag bash -c 'cd /stagedir; cmd1; cmd2'")
+
+    @property
+    def exp_cmd_custom_registry(self):
+        self.container_platform.with_mpi = True
```
vkarak: You should test also when this is false.

unittests/test_containers.py
```diff
@@ -59,10 +59,48 @@ def create_container_platform(self):
...
+        return containers.Singularity()
+
+    @property
+    def exp_cmd_mount_points(self):
+        return ('singularity exec -B"/path/one","/one" -B"/path/two","/two" '
+                "name:tag bash -c 'cd /stagedir; cmd1; cmd2'")
+
+    @property
+    def exp_cmd_custom_registry(self):
+        self.container_platform.with_cuda = True
```
vkarak: You should test also when this is false.

reframe/core/pipeline.py
```diff
@@ -1319,7 +1358,6 @@ class RunOnlyRegressionTest(RegressionTest):
     This class is also directly available under the top-level :mod:`reframe`
     module.
     """
-
```
vkarak: Leave this line.

unittests/test_containers.py
```diff
@@ -58,49 +67,81 @@ def create_container_platform(self):
...
+                '--mount=type=bind,source="/path/two",destination="/two" '
+                "name:tag bash -c 'cd /stagedir; cmd1; cmd2'")
+
+    @property
+    def expected_cmd_prepare(self):
+        return ['shifter pull name:tag']
+
+    @property
+    def expected_cmd_with_run_opts(self):
+        self.container_platform.requires_mpi = True
```
vkarak: You still don't test the case without MPI. My suggestion was the following:

```python
class TestShifterNG(_ContainerPlatformTest, unittest.TestCase):
    # as of now, but without setting the MPI flag.

class TestShifterNGWithMPI(TestShifterNG):
    @property
    def expected_cmd_with_run_opts(self):
        self.container_platform.requires_mpi = True
        return ...
```

You should do the same for the CUDA option of Singularity.

unittests/test_containers.py
```diff
@@ -58,49 +67,81 @@ def create_container_platform(self):
...
     @property
-    def exp_cmd_custom_registry(self):
-        self.container_platform.with_mpi = True
-        return ('shifter run --mpi '
+    def expected_cmd_prepare(self):
+        return ['sarus pull name:tag']
+
+    @property
+    def expected_cmd_with_run_opts(self):
+        self.container_platform.requires_mpi = True
```
vkarak: Same as my suggestion for ShifterNG.

unittests/test_pipeline.py
```diff
@@ -791,3 +791,51 @@ def extract_perf(patt, tag):
         self.assertIn('v1', log_output)
         self.assertIn('v2', log_output)
         self.assertIn('v3', log_output)
+
+
+class TestContainerPlatformsTest(TestRegressionTest):
```
vkarak: By inheriting from `TestRegressionTest`, all its tests will run, which you don't want it.

unittests/test_pipeline.py
```diff
@@ -791,3 +791,51 @@ def extract_perf(patt, tag):
...
+            sn.assert_found(r'18.04.3 LTS \(Bionic Beaver\)', test.stdout),
+        ])
+        return test
+
+    def run_test(self, test):
+        try:
+            self._run_test(test)
+        except PipelineError:
+            self.skipTest('no configuration found for container platform: %s' %
+                          type(test.container_platform).__name__)
```
vkarak: No, you can't skip the test like that here. The fact that a `PipelineError` was thrown does not mean that it is valid to skip the test. It might be a bug. You should check individually here that a test may not be run and only then skip the test.

unittests/test_pipeline.py
vkarak: I see that you are using some functions of the `TestRegressionTest`, but we can figure it out how to do it nicely.

config/cscs.py
```diff
@@ -140,6 +161,14 @@ class ReframeSettings:
 
                     'gpu': {
                         'scheduler': 'nativeslurm',
+                        'container_platforms': {
+                            'Singularity': {
+                                'modules': ['singularity']
+                            },
+                            'Sarus': {
+                                'modules': ['sarus/cle7up01-test']
```
vkarak: Where is this module? I am getting the following error:

```]
E           ModuleCmd_Display.c(157):ERROR:105: Unable to locate a modulefile for 'sarus/cle7up01-test'
```

vkarak:
This PR is now ready. Thanks @rsarm.

PR #855, opened 2019-07-03
[feat] Allow self loops in module mappings

### LINKED ISSUE #843: Allow self referring module mappings
Running ReFrame with the following will issue an error of cyclic dependency:

```
-M 'A: A B'
```
This scenario is useful in testing when you want to load an additional module without touching the test, and actually it is not a cyclic dependency.

@lucamar has also came across this problem in the past if I'm not mistaken.

teojgo:
Fixes #843

### DIFF
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -76,7 +76,6 @@ class ModulesSystem:
     """A modules system abstraction inside ReFrame.
 
     This class interfaces between the framework internals and the actual
-re
     modules systems implementation.
     """
 
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -117,7 +117,7 @@ class ModulesSystem:
                 path.pop()
 
             # Handle modules mappings with self loops
-            if node == parent and node not in ret:
+            if node == parent:
                 ret.add(node)
                 continue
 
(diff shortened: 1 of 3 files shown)


reframe/core/modules.py
```diff
@@ -109,13 +109,19 @@ def resolve_module(self, name):
...
+        loop_nodes = set()
         while unvisited:
             node, parent = unvisited.pop()
-
             # Adjust the path
             while path and path[-1] != parent:
                 path.pop()
 
+            # Handle modules mappings with self loops
+            if node == parent and node not in loop_nodes:
```
vkarak: Why do you need the `loop_nodes` here?

teojgo: I need it to avoid duplicate modules.

vkarak: Can't you just use an `OrderedSet` for the `ret` value and convert it to a list at return?

reframe/core/modules.py
```diff
@@ -105,21 +106,19 @@ def resolve_module(self, name):
...
             node, parent = unvisited.pop()
             # Adjust the path
             while path and path[-1] != parent:
                 path.pop()
 
             # Handle modules mappings with self loops
-            if node == parent and node not in loop_nodes:
-                loop_nodes.add(node)
-                ret.append(node)
+            if node == parent and node not in ret:
```
vkarak: You don't need the `not in ret` part. It's handled the `OrderedSet`.

teojgo: A ok, I thought that it would update the order if it was already in.

PR #861, opened 2019-07-05
[ci] Update system config for Kesch

kraushm:
Updating the environments to account for the new meta-modules, which became necessary to support two different toolchain versions. The config file targets the production PE/17.06.

### DIFF
--- a/config/cscs.py
+++ b/config/cscs.py
@@ -337,29 +337,34 @@ class ReframeSettings:
             'kesch': {
                 'PrgEnv-pgi-nompi': {
                     'type': 'ProgEnvironment',
-                    'modules': ['PrgEnv-pgi/17.10'],
+                    'modules': ['PE/17.06',
+                                'PrgEnv-pgi/18.05'],
                     'cc': 'pgcc',
                     'cxx': 'pgc++',
                     'ftn': 'pgf90',
                 },
                 'PrgEnv-pgi': {
                     'type': 'ProgEnvironment',
-                    'modules': ['PrgEnv-pgi/17.10_gdr'],
+                    'modules': ['PE/17.06',
+                                'PrgEnv-pgi/18.05'],
                     'cc': 'mpicc',
                     'cxx': 'mpicxx',
                     'ftn': 'mpif90',
                 },
                 'PrgEnv-cray': {
                     'type': 'ProgEnvironment',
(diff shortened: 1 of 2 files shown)


kraushm:
@jenkins-cscs retry all

vkarak:
@kraushm Why is this a WIP?

kraushm:
@jenkins-cscs retry all

kraushm:
@jenkins-cscs retry all

kraushm:
@jenkins-cscs retry all

config/cscs.py
```diff
@@ -340,29 +340,34 @@ class ReframeSettings:
...
+                                'PrgEnv-pgi/18.05'],
                     'cc': 'pgcc',
                     'cxx': 'pgc++',
                     'ftn': 'pgf90',
                 },
                 'PrgEnv-pgi': {
                     'type': 'ProgEnvironment',
-                    'modules': ['PrgEnv-pgi/17.10_gdr'],
+                    'modules': ['PE/17.06',
+                                'PrgEnv-pgi/18.05'],
```
vkarak: What's the difference between this and the previous one?

kraushm: Now, there is only a single (with respect to mpi/nompi) PGI programming environment available on the system. So, the only difference remaining in this config file would be the compilers used, i. e. PGI vs. MPI. 
The question then is, whether we want to keep both or not?

config/cscs.py
```diff
@@ -340,29 +340,34 @@ class ReframeSettings:
...
+                                'PrgEnv-pgi/18.5'],
                     'cc': 'pgcc',
                     'cxx': 'pgc++',
                     'ftn': 'pgf90',
                 },
                 'PrgEnv-pgi': {
                     'type': 'ProgEnvironment',
-                    'modules': ['PrgEnv-pgi/17.10_gdr'],
+                    'modules': ['PE/17.06',
+                                'PrgEnv-pgi/18.5'],
```
vkarak: My comment was that we don't need both `PrgEnv-pgi` and `PrgEnv-pgi-nompi`, since they load completely the same modules.

config/cscs.py
vkarak: Ok, got it. Let's keep both then.

config/cscs.py
vkarak: Ignore this.

PR #866, opened 2019-07-09
[feat] Add automatically a default `*` reference for performance checks

### LINKED ISSUE #811: Add always a default `*` reference if it does not exist
This is one of the guidelines for writing generic performance tests, and I think it's a good idea that the framework provides that automatically.

ChristopherBignamini:
Fixes #811


vkarak:
@ChristopherBignamini Why this is a WIP?

vkarak:
@ChristopherBignamini Please take a look at the unit tests failures.

reframe/core/pipeline.py
```diff
@@ -1124,6 +1124,24 @@ def check_performance(self):
             return
 
         with os_ext.change_dir(self._stagedir):
+
+            # Check if default reference perf values are provided and
+            # store all the variables  tested in the performance check
+            is_default_present = False
+            variables = set()
+            for system, perf_data in self.reference.items():
+                variables.add((system.split(":")[-1], perf_data[-1]))
```
teojgo: You should handle also the case that there are no units defined.

ChristopherBignamini: Is it possible to provide a reference value without unit?

teojgo: Yes the unit is optional. Am I right @vkarak ?

ChristopherBignamini: I may be wrong but I remember a discussion about rejecting performance tests without units in order to avoid mistakes

reframe/core/pipeline.py
```diff
@@ -1124,6 +1124,24 @@ def check_performance(self):
...
+                if system.split(":")[0] == '*':
+                    is_default_present = True
+                    break
+
+            if is_default_present is False:
```
teojgo: Change this to `if not is_default_present` or change variable to something that when it is `True` then you need to handle the missing default case.

unittests/test_pipeline.py
```diff
@@ -573,7 +573,24 @@ def test_unknown_system(self):
                 'value3': (3.1, -0.1, 0.1),
             }
         }
-        self.assertRaises(SanityError, self.test.check_performance)
+        self.test.check_performance()
+
+    def test_unknown_multiple_systems(self):
```
vkarak: I don't think this unit test offers sth more compared to the `test_unknown_system` one.

In one of the reference numbers you should add a unit as well and we are still missing sth to test that the units are emitted correctly.

reframe/core/pipeline.py
```diff
@@ -1124,6 +1124,24 @@ def check_performance(self):
...
 
         with os_ext.change_dir(self._stagedir):
+
+            # Check if default reference perf values are provided and
+            # store all the variables  tested in the performance check
+            is_default_not_present = True
+            variables = set()
+            for system, perf_data in self.reference.items():
+                variables.add((system.split(":")[-1], perf_data[-1]))
+                if system.split(":")[0] == '*':
```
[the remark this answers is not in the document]
ChristopherBignamini: Done

reframe/core/pipeline.py
```diff
@@ -1124,6 +1124,24 @@ def check_performance(self):
             return
 
         with os_ext.change_dir(self._stagedir):
+
+            # Check if default reference perf values are provided and
+            # store all the variables  tested in the performance check
+            is_default_not_present = True
```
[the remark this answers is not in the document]
ChristopherBignamini: Done.

reframe/core/pipeline.py
```diff
@@ -1124,6 +1124,24 @@ def check_performance(self):
...
+            variables = set()
+            for system, perf_data in self.reference.items():
```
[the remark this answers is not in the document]
ChristopherBignamini: Done.

reframe/core/pipeline.py
```diff
@@ -1124,6 +1124,24 @@ def check_performance(self):
...
+                if system.split(":")[0] == '*':
+                    is_default_not_present = False
+                    break
+
+            if is_default_not_present:
+                # If default value is not provided add one for all the
+                # tested variables
+                for variable in variables:
+                    self.reference.update(
+                        {'*': {variable[0]: (0, None, None, variable[1])}})
```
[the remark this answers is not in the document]
ChristopherBignamini: Done.

unittests/test_pipeline.py
ChristopherBignamini: The idea behind this unit test was checking the ability of the code to collect different variables from different systems and include all of them in the added default case.

reframe/core/pipeline.py
```diff
@@ -1124,6 +1124,24 @@ def check_performance(self):
             return
 
         with os_ext.change_dir(self._stagedir):
+
+            # Check if default reference perf values are provided and
+            # store all the variables  tested in the performance check
+            is_default_not_present = True
+            variables = set()
+            for system, perf_data in self.reference.items():
+                variables.add((system.split(":")[-1], perf_data[-1]))
```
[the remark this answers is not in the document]
ChristopherBignamini: Is it correct to assume that the unit is always a string? In that case I could put the varname variable in the reference tuple only if it is of string type.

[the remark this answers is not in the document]
vkarak: No, it's better if you do sth like that:

```python
try:
    unit = keyparts[3]
except IndexError:
    unit = None
```

And then include the unit only if it's not `None`.

unittests/test_pipeline.py
vkarak: I guess we could merge then this and the previous one. Just add another unknown system to the previous one.

PR #867, opened 2019-07-09
[feat] Fall back to a generic system configuration if system auto-detection fails

### LINKED ISSUE #832: out-of-the-box user experience "could not auto-detect system"
Just took `reframe` for a quick spin (without checking the documentation, I must admit), and I think there's some low hanging fruit to create a better initial user experience.

After checking `reframe --help`, I wanted to check which tests are available out-of-the-box (if any):

```
$ reframe -l
/software/ReFrame-HPC/2.18/bin/reframe: could not auto-detect system; please use the `--system' option to specify one explicitly
```

Same output for `reframe --show-config`.

One minor annoyance here is the full `reframe` part being printed, I don't think that's really necessary to be helpful?

I couldn't find a `--list-systems` option based on the `--help` output, and a naive attempt to use `--system localhost` failed miserably...

Maybe a pointer to the documentation should be spit out when `reframe` noticed it's "starting from scratch" (no systems defined yet), or simply default to using `localhost` (if that's an option, for example by auto-detecting # available cores)?

vkarak:
We now issue a warning instead of an error and we fall back to a default system configuration, that is generic enough to allow ReFrame to run anywhere. User is pointed to the documentation for configuring ReFrame for her system.

Fixes #832.

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -314,17 +314,57 @@ The possible attributes of an environment are the following:
 System Auto-Detection
 ---------------------
 
-When the ReFrame is launched, it tries to auto-detect the current system based on its site configuration. The auto-detection process is as follows:
+When the ReFrame is launched, it tries to detect the current system and select the correct site configuration entry. The auto-detection process is as follows:
 
 ReFrame first tries to obtain the hostname from ``/etc/xthostname``, which provides the unqualified *machine name* in Cray systems.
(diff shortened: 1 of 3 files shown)


docs/configure.rst
```diff
@@ -314,17 +314,57 @@ The possible attributes of an environment are the following:
 System Auto-Detection
 ---------------------
 
-When the ReFrame is launched, it tries to auto-detect the current system based on its site configuration. The auto-detection process is as follows:
+When the ReFrame is launched, it tries to detect the current system and select the correct site configuration entry. The auto-detection process is as follows:
```
teojgo: When ReFrame ...

PR #870, opened 2019-07-12
[test] Workaround for netcdf test with PrgEnv-pgi

teojgo:


vkarak:
@omlins I don't see a problem merging the version with the workaround so far that this is clearly stated in the test. We don't know when Cray will be fixing this and we want somehow to "document" this workaround inside the test. We can stall this PR, h however, until we get a fix, but if it takes too long, we will merge it.

cscs-checks/libraries/io/netcdf_compile_run.py
```diff
@@ -66,11 +66,8 @@ def setup(self, partition, environ, **job_opts):
...
         else:
             self.build_system.ldflags = ['-%s' % self.linkage]
-            if (environ.name == 'PrgEnv-pgi' and
-                # NOTE: Workaround to fix static linking for C++ with
-                # PrgEnv-pgi
-                self.lang == 'cpp' and
-                self.linkage == 'static'):
-                self.build_system.ldflags += ['-lstdc++']
+            if environ.name == 'PrgEnv-pgi' and self.lang == 'cpp':
+                # NOTE: Workaround to fix compiling for PrgEnv-pgi
```
vkarak: - `NOTE` -> `FIXME`
- `compiling` -> `compilation`

Please put also the the Cray case number in the comment.

vkarak:
I am merging despite the failure on Kesch, because it fails also on master.

PR #874, opened 2019-07-17
[test] Add CPU memory latency test

sebkelle1:


vkarak:
@jenkins-cscs retry daint

jgphpc:
Should lower reference be smaller ?

```
  * Stage directory: /scratch/snx3000tds/piccinal/reframe/stage/dom/gpu/PrgEnv-cray/CPULatencyTest
  * Failing phase: performance
  * Reason: performance error: failed to meet reference: latencyL2=1.15884, expected 3.44 (l=3.4055999999999997, u=4.3344)
```

I need to retest this one:

```
FAILURE INFO for CPULatencyTest
  * System partition: dom:gpu
  * Environment: PrgEnv-gnu
  * Stage directory: /scratch/snx3000tds/piccinal/reframe/stage/dom/gpu/PrgEnv-gnu/CPULatencyTest
  * Maintainers: ['SK']
  * Failing phase: run
  * Reason: job error: [jobid=None] could not find enough nodes: required 1, found 0

**Comment by jgphpc** [comment]
There seems to be an issue with `--exec-policy=async` and `self.num_tasks = 0` (independent of this PR). Running without `--exec-policy=async` gives:

```
FAILURE INFO for CPULatencyTest
  * Environment: PrgEnv-cray
  * Reason: performance error: failed to meet reference: latencyL2=1.21557, expected 3.65 (l=3.6134999999999997, u=4.599)
  * Environment: PrgEnv-gnu
  * Reason: performance error: failed to meet reference: latencyL2=1.21533, expected 3.65 (l=3.6134999999999997, u=4.599)
```

**Comment by jgphpc** [comment]
'ault:intel', 'ault:amdvega' and 'tave:compute' are not in the ci

File: cscs-checks/microbenchmarks/latency/src/latency.cpp [hunk 1]
```diff
@@ -0,0 +1,230 @@
...
+void convert_to_pointers(CacheLine* start, std::vector<unsigned> const& path)
+{
+    for (unsigned i = 0; i < path.size(); ++i)
+        start[i].payload = start + path[i];
+}
+
+
+size_t chainload(CacheLine* next, size_t reps)
+{
+    // compiler doesn't unroll
```
vkarak: Is there a particular reason that you've chosen to unroll four times? Empirically set or not, I'd mention that in the comment.

cscs-checks/microbenchmarks/latency/src/latency.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include <pthread.h>
+
+#define PAGESIZE 4096
+#define CACHELINESIZE 64
+
+// only used for runtime estimation
+#define L1 32768
+#define L2 262144
+#define L3 25600*1024
+#define GHZ 3.3
```
vkarak: Why not guarding these with `ifndef`s so as to allow them to be defined during compilation per processor? Same for `PAGESIZE` and `CACHELINESIZE`.

vkarak: Since you're printing out the cycles based on the `GHZ`, and `PAGESIZE` and `CACHELINESIZE` are essential to your algorithm, I'd guard at least those with `ifndef`.

cscs-checks/microbenchmarks/latency/src/latency.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+
+// convert any permutation visit_sequence to a path
+// --
+// A path is equivalent to a permutation of [0, length-1]
+// that contains no valid permutation in any subset
+// [0,n] for n < length-1
+// --
+// e.g for length=5, 12034 would be invalid, because 120 is
+// a valid permutation of [0,2], leading to a closed loop
```
vkarak: This is correct, but I can't see how do you check for that? How do you guarantee that you can't have such a permutation?

cscs-checks/microbenchmarks/latency/src/latency.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+        double ns_per_load = duration / reps * 1e9;
+        double clocks = GHZ * ns_per_load;
+
+        std::cout << "latency (ns): " << ns_per_load
+                  << " clocks: " << clocks << " size " << buffer_size
+                  << " ngroups: " << buffer_size/group_size << std::endl;
+
+        std::free(chain);
+    }
```
vkarak: I would print an error here, e.g., "progname: too few arguments", if the user does not pass any arguments and also print a small usage message.

cscs-checks/microbenchmarks/latency/latency.py
```diff
@@ -0,0 +1,96 @@
...
+        if self.current_system.name in {'tave'}:
+            self.modules = ['craype-hugepages512M']
+
+        self.sanity_patterns = sn.assert_eq(
+            sn.count(sn.findall(r'latency', self.stdout)),
+            self.num_tasks_assigned * len(self.executable_opts))
+
+        self.perf_patterns = {
+            'latencyL1': sn.extractall(
+                r'latency \(ns\): (?P<bw>\S+) clocks',
```
vkarak: Since the pattern is common, I wouldn't repeat in all perf patterns here.

cscs-checks/microbenchmarks/latency/src/latency.cpp
sebkelle1: The way I construct the path from the permutation guarantees that all elements are visited.
To see this, one needs to realize that a random permutation \sigma interpreted as a sequence of elements to be visited is not the same as the array of indices (the path) that traverses the elements by dependency chaining in the order specified by \sigma.

The function make_path creates the dependency chain for the input permutation \sigma.
Since \sigma is a valid permutation, it contains all elements in the range, and therefore the path traverses all elements.

e.g. the example given: 12034 is a valid permutation (but not a valid path)
the corresponding path (visit [1], then [2], [0], [3], [4] and then again [1]) is 32041.
This will produce the order 12034 in the for loop

`int next = sigma[0];`
`for (int i = 0; i < sigma.size(); ++i)`
`    next = path[next];`

vkarak: I got it, you are right. It's a bit confusing the comment as it is now. The mathematical definition is correct, but I would like to see in the comment how you achieve that. I guess that the simply stating clearly the difference between the `visit_sequence` and `path` would help + the example of how the `12034` permutation gets into `32041` (not step-by-step of course, just to mention it). So sth like the following would be enough:

- `visit_sequence` is the order that we want to visit the nodes
- Each element of the `path` contains a "pointer" (index) to the next element to visit according to the visit sequence. For example, a `visit_sequence` of 12034 will produce as path `32041`.

vkarak:
@sebkelle1 Can you address my PR comments, so that we can merge this one?

cscs-checks/microbenchmarks/latency/src/latency.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+    return make_path(grouped_seq);
+}
+
+
+size_t estimate_reps(unsigned sz, unsigned l1, unsigned l2, unsigned l3)
+{
+    const double l1_lat_ns = 1.2;
+    const double l2_lat_ns = 3;
+    const double l3_lat_ns = 12;
```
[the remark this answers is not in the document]
sebkelle1: as you prefer, this is irrelevant for the code/benchmark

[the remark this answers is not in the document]
vkarak: I know, that's why I think it makes sense to have "rounded" numbers.

cscs-checks/microbenchmarks/latency/src/latency.cpp
```diff
@@ -0,0 +1,230 @@
...
+            group_size = pages_per_group * PAGESIZE;
+            buffer_size = group_size * ngroups;
+        }
+        else
+            group_size = buffer_size;
+
+        unsigned num_elements = buffer_size / sizeof(CacheLine);
+        unsigned group_elements = group_size / sizeof(CacheLine);
+
+        size_t reps = estimate_reps(buffer_size, L1, L2, L3);
```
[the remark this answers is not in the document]
sebkelle1: This would require a separate definition of the `reps` variable and require an additional statement / line.

cscs-checks/microbenchmarks/latency/src/latency.cpp
sebkelle1: It really seems the unrolling is not needed and I removed it.

cscs-checks/microbenchmarks/latency/src/latency.cpp
sebkelle1: ok

cscs-checks/microbenchmarks/latency/latency.py
sebkelle1: I'll need to fix this tomorrow morning, the pattern needs to be size specific, otherwise
it doesn't work when run on multiple nodes simultaneously.

vkarak:
@jenkins-cscs retry dom

PR #876, opened 2019-07-18
[feat] Remove support for the very old test syntax using `_get_checks()`

vkarak:
I have also adapted the OpenFOAM tests to the new syntax, since they were the only ones remaining to be ported. @kraushm Can you have a look into them?

This PR renders unnecessary the call to `super().__init__()` inside the user test's constructor.


reframe/frontend/loader.py
```diff
@@ -86,25 +81,14 @@ def recurse(self):
     def load_from_module(self, module):
         """Load user checks from module.
 
-        This method tries to call the `_get_checks()` method of the user check
-        and validates its return value."""
+        This method tries to call the `_rfm_gettests()` method of the user
+        check and validates its return value."""
         from reframe.core.pipeline import RegressionTest
 
-        old_syntax = hasattr(module, '_get_checks')
```
vkarak: Instead of removing it completely, I could perhaps issue a warning here that this syntax is now dropped and point the users to use the decorators.

teojgo: I think that we can remove completely the old syntax. We should just have a note in the documentation that support for old style checks was removed.

vkarak: I have put a simple check in place that issues a warning, but I have not pushed yet.

vkarak:
@victorusu @teojgo This PR is a requirement for the stuff I'm doing with the delayed initialization of tests.

vkarak:
I am converting it to wip, because I still need to remove the ability to pass arguments to the `RegressionTest`'s `__init__()`, which was only meaningful with the old syntax.

vkarak:
With this PR it will no more be necessary to call `super().__init__()` inside the user tests. Should I change all the tutorial example?

teojgo:
@vkarak yes change them and remove `super().__init__()` from the OpenFoam tests included here.

unittests/resources/checks/bad/invalid_check.py
```diff
@@ -1,17 +1,12 @@
-import os
+import reframe as rfm
 
-from reframe.core.pipeline import RegressionTest
 
+@rfm.simple_test
+class SomeTest(rfm.RegressionTest):
+    def __init__(self):
+        super().__init__()
```
teojgo: Is this needed?

vkarak: Not any more.

cscs-checks/apps/openfoam/check_openfoam_extend.py
```diff
@@ -0,0 +1,174 @@
+import os
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+class OpenfoamExtendBaseTest(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        super().__init__()
```
teojgo: Also remove the `super().__init__()` from these tests

PR #878, opened 2019-07-23
[ci] Add setuptools as requirement to reframe

victorusu:


vkarak:
@jenkins-cscs retry daint

teojgo:
for me it works when I create a virtual environment and use the requirements without setuptools

vkarak:
@teojgo Yes, that's the norm, but there might be situation that `setuptools` are required. We had gone through this when preparing the official EB recipe for reframe.

PR #883, opened 2019-07-25
fix broken documentation link

jgphpc:


jgphpc:
Is this the expected page ?

eff

jgphpc:
link works fine here: https://reframe-hpc.readthedocs.io/en/latest/index.html
so maybe not an issue after all

docs/index.rst
```diff
@@ -20,7 +20,7 @@ Use Cases
 
 The ReFrame framework has been in production at `CSCS <http://www.cscs.ch>`__ since the upgrade of the `Piz Daint <http://www.cscs.ch/computers/piz_daint/index.html>`__ system in early December 2016.
 
-`Read the full story <usecases.html>`__...
+`Read the full story <usecases.rst>`__...
```
vkarak: I don't think this is going to work. RST files are not visible. There are used to generate HTML.

PR #886, opened 2019-07-25
[bugfix] Fix wrong error message emitted when a user module is loaded with the `-m` option

### LINKED ISSUE #885: ReFrame complains that it failed to load a module with the `-m` command line option, but it actually does load it
This was originally reported by @bcfriesen in the chat:

> Somewhat confusingly, ReFrame complains that the module load failed, but in fact it succeeds :slightly_smiling_face:

I just came across it today, while working on #806. It's just stale test code inserted here for quite some time: https://github.com/eth-cscs/reframe/blame/master/reframe/frontend/cli.py#L500

I'm going to submit a PR asap.

vkarak:
- This line was producing a fake error message when trying to load user modules
  with the `-m` option.

Fixes #885.

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -497,7 +497,6 @@ def main():
         for m in options.user_modules:
             try:
                 rt.modules_system.load_module(m, force=True)
-                raise EnvironError("test")
             except EnvironError as e:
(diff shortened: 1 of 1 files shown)


vkarak:
@jenkins-cscs retry dom

PR #887, opened 2019-07-25
[feat] Enable modifications to ReFrame's environment per system basis

### LINKED ISSUE #806: ReFrame fails to load modules before job submission
Issue reported by @bcfriesen in the Slack channel: 

> I have a question about how ReFarme handles Tcl modules. A few tests I am writing for NERSC require certain modules to be loaded before the job is submitted (the Slurm submit filter checks that the modules are loaded and rejects the job submission if they are not). However, when I add the necessary modules to the `setup()` phase, the logs indicate that ReFrame is indeed attempting to load them via `modulecmd python module load <module>`, but that seems to fail because when it submits the job to Slurm, the job is rejected because the module is not loaded. Any ideas what is going on?
I notice that `modulecmd python load <module>` takes no action other than printing the environment changes that it _would_ make to the terminal. I wonder if ReFrame is not actually evaluating these outputs for some reason?

This looks like a regression in ReFrame. Some time ago (when we introduced the build systems) we changed slightly the behaviour of the `setup()`, and its only purpose now is to record the necessary load/unload sequence of modules. In the past, we kept the environment (in Python) until the end of the test, that’s why this scenario worked then but not now. I will open an issue about that. For the moment, you can run ReFrame with `-m MODULE`, which will load `MODULE` before starting running any test.

vkarak:
This PR introduces the `modules` and `variables` parameters at the system level of the site configuration. This is useful in cases were a module is always needed to be loaded for launching jobs for example:

```python
        'systems': {
            'foosys': {
                'modules': ['slurm']
                'hostnames': ['localhost'],
                'partitions': {
                    'compute': {
                        'scheduler': 'nativeslurm',
                        'environs': ['builtin-gcc'],
                    },
                }
            },
```

Notice that this modifies the ReFrame environment, whereas the same parameters defined for a partition do not.

Fixes #806.


vkarak:
@bcfriesen This fixes the issue you had when trying to load a module providing Slurm. You may now specify this module as parameter at the system level (check the description of this PR). Can you tell me if this works for you? Passing `-m` option should not be necessary any more.

docs/configure.rst
```diff
@@ -100,6 +100,11 @@ The valid attributes of a system are the following:
   - ``tmod``: The classic Tcl implementation of the `environment modules <https://sourceforge.net/projects/modules/files/Modules/modules-3.2.10/>`__ (versions older than 3.2 are not supported).
   - ``tmod4``: The version 4 of the Tcl implementation of the `environment modules <http://modules.sourceforge.net/>`__ (versions older than 4.1 are not supported).
   - ``lmod``: The Lua implementation of the `environment modules <https://lmod.readthedocs.io/en/latest/>`__.
+
+* ``modules``: Modules to be loaded always when running on this system.
```
victorusu: Would it be better to say...
>  Modules to load when running on this system.
?

reframe/frontend/cli.py
```diff
@@ -494,6 +494,14 @@ def main():
         if options.purge_env:
             rt.modules_system.unload_all()
 
+        # Load the environment for the current system
+        try:
+            rt.system.preload_environ.load()
+        except EnvironError as e:
+            printer.error("failed to load current's system environment; "
+                          "please check your configuration")
+            raise
```
victorusu: Are we raising empty?

vkarak: No, this means raise the current exception. But I should remove `e`, which I don't use.

docs/configure.rst
vkarak: I don't see such a big difference.. Unless you want me to remove the "always".

bcfriesen:
Hi @vkarak,

> @bcfriesen This fixes the issue you had when trying to load a module providing Slurm. You may now specify this module as parameter at the system level (check the description of this PR). Can you tell me if this works for you? Passing `-m` option should not be necessary any more.

Great! This is one of the last outstanding bugs that we have to work around at NERSC. We're returning from a system-wide maintenance today, so I will test this as soon as the systems come back (today or tomorrow).

bcfriesen:
Hi @vkarak, sorry for the delay. This patch indeed works for me, thanks!

PR #893, opened 2019-08-05
[ci] Temporarily disable the params variable

victorusu:


teojgo:
@victorusu why do you need this change?

gppezzi:
@teojgo the params API was not working properly for a couple of days, CI was not running properly:
```
No machines were found. Aborting...
```
See for example [build #2247](https://jenkins.cscs.ch/blue/organizations/jenkins/ReframeCI/detail/ReframeCI/2247/pipeline)

teojgo:
@gppezzi this is strange since the `params` was defined for that build. In any case I update all the related jenkins plugins. Let's see how it goes.

gppezzi:
I also updated the plugins last week (twice actually) and it didn't help. Now it seems to be working again (CI build 2251 worked without merging this PR), so I assume that was a temporary glitch.

I think we can close this PR for the moment.

teojgo:
I am closing this since everything seems to be working again.

PR #894, opened 2019-08-07
[test] Update IPM tests for PE19.06

teojgo:


teojgo:
@jenkins-cscs retry all

PR #897, opened 2019-08-09
Fix openacc_cuda_mpi_cppstd.F90

lucamar:
After merging the pull request #888, we are getting the following error:
```
 MPI test on GPU with OpenACC using  8 tasks
 Result : FAIL
 Expected value sum(f3):  55. actual value: 110.
``` 
I think that there is a mistake in the fix submitted by @cosunae: the array that should be checked for the `EXPECTED_CPP_STD_SUM` is `f2` and not `f3`, since only the first one is called as argument in `call_cpp_std`. What do you think @lxavier?

P.S.: I have replaced `cudaThreadSynchronize()` with `cudaDeviceSynchronize()` in `compute_cuda.cu` since the first one is deprecated.

### DIFF
--- a/cscs-checks/mch/src/openacc_cuda_mpi_cppstd.F90
+++ b/cscs-checks/mch/src/openacc_cuda_mpi_cppstd.F90
@@ -74,9 +74,9 @@ program openacc_cuda_mpi_cppstd
     if (sum(f1) /= EXPECTED_CUDA_SUM) then
       write (*,*) "Result : FAIL"
       write (*,*) "Expected value sum(f1): ", EXPECTED_CUDA_SUM, "actual value:", sum(f1)
-    else if (sum(f3) /= EXPECTED_CPP_STD_SUM) then
+    else if (sum(f2) /= EXPECTED_CPP_STD_SUM) then
       write (*,*) "Result : FAIL"
-      write (*,*) "Expected value sum(f3): ", EXPECTED_CPP_STD_SUM, "actual value:", sum(f3)
+      write (*,*) "Expected value sum(f2): ", EXPECTED_CPP_STD_SUM, "actual value:", sum(f2)
     else if (data_sum(1) /= ref_val) then
       write (*,*) "Result : FAIL"
       write (*,*) "Expected value data_sum: ", ref_val, "actual value:", data_sum(1)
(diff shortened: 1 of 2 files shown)


lxavier:
it was in fact already failing before the fix, and we try to correct but obviously we did it wrong

cscs-checks/mch/src/openacc_cuda_mpi_cppstd.F90
```diff
@@ -74,9 +74,9 @@ program openacc_cuda_mpi_cppstd
     if (sum(f1) /= EXPECTED_CUDA_SUM) then
       write (*,*) "Result : FAIL"
       write (*,*) "Expected value sum(f1): ", EXPECTED_CUDA_SUM, "actual value:", sum(f1)
-    else if (sum(f3) /= EXPECTED_CPP_STD_SUM) then
+    else if (sum(f2) /= EXPECTED_CPP_STD_SUM) then
```
lxavier: no it should be f3, but the error is on the right side

ajocksch: like this:

else if (sum(f3) /= EXPECTED_CUDA_SUM) then

?

lxavier: Yes, basicall revertting the change from PR #888 only the file openacc_cuda_mpi_cppstd.F90  . I did that with Carlos, because the code was failling. But probably because we did too quickly and did not compile with the correct OpenACC flag. Note that you also need to adapt the 
```
 write (*,*) "Expected value sum(f3): ", EXPECTED_CUDA_SUM, "actual value:", sum(f3)
```
below

lxavier: sorry for messing this up

ajocksch:
```suggestion
    else if (sum(f3) /= EXPECTED_CUDA_SUM) then
```

cscs-checks/mch/src/openacc_cuda_mpi_cppstd.F90
```diff
@@ -74,9 +74,9 @@ program openacc_cuda_mpi_cppstd
...
       write (*,*) "Result : FAIL"
-      write (*,*) "Expected value sum(f3): ", EXPECTED_CPP_STD_SUM, "actual value:", sum(f3)
+      write (*,*) "Expected value sum(f2): ", EXPECTED_CPP_STD_SUM, "actual value:", sum(f2)
```
ajocksch:
```suggestion
      write (*,*) "Expected value sum(f3): ", EXPECTED_CUDA_SUM, "actual value:", sum(f3)
```

PR #910, opened 2019-08-27
[test] Increase iterations for async kernel latency test

teojgo:
* Add PrgEnv-cray_classic in programming environments


cscs-checks/microbenchmarks/kernel_latency/kernel_latency.py
```diff
@@ -75,10 +74,10 @@ def __init__(self, kernel_version):
             },
             'async': {
                 'dom:gpu': {
-                    'latency': (2.2, None, 0.10, 'us')
+                    'latency': (2.2, None, 0.15, 'us')
```
vkarak: From my experiments, I think that the reference should change and we might need to investigate.

teojgo:
@vkarak can you merge #912 first to see if it runs correctly the benchmark test?

vkarak:
@teojgo I've check it manually. Everything's green.

PR #911, opened 2019-08-28
[bugfix] Use login shell when building executables

### LINKED ISSUE #882: Add option to use login shell to build executables
Currently `ReFrame` invokes `/bin/bash` without `-l` in the build scripts. 
It could be useful to add an option to select a login shell as well if needed, i.e. using the option `-l` of the `bash` command.

teojgo:
Fixes #882


vkarak:
@lucamar Can you check if this PR fixes the problem you had in #882?

vkarak:
@jenkins-cscs retry none

PR #914, opened 2019-08-29
[ci] Fix tags of modified regression tests to run in ci

teojgo:


ci-scripts/ci-runner.bash
```diff
@@ -50,7 +50,7 @@ run_tutorial_checks()
 run_user_checks()
 {
     cmd="./bin/reframe -C config/cscs.py --exec-policy=async --save-log-files \
--r --flex-alloc-tasks=2 -t 'production|benchmark' $@"
+-r --flex-alloc-tasks=2 -t production|benchmark $@"
```
vkarak: This is strange! Why is this happening?

teojgo: I don't know. It seems that the quotes are passed to the regex parsed by Python

teojgo:
@jenkins-cscs retry all

PR #917, opened 2019-08-30
[test] Add performance report to Empirical Roofline Tool test

jgphpc:
Will fix UES-511


jgphpc:
Failed because of unittest: `E   AssertionError: 3.131918 not less than 3`

PR #918, opened 2019-09-02
[feat] Print compiler commands and flags when showing the environment configuration

### LINKED ISSUE #884: Show the compiler commands and flags in the --show-config-env option
Currently we are only showing the modules and the variables:

```
$ ./bin/reframe -C config/racklette.py --show-config-env intel
intel:
    modules: intel/compiler/64
    variables:
```

teojgo:
Fixes #884

### DIFF
--- a/reframe/core/environments.py
+++ b/reframe/core/environments.py
@@ -334,3 +334,19 @@ class ProgEnvironment(Environment):
     @property
     def nvcc(self):
         return self._nvcc
+
+    def details(self):
+        base_details = super().details()
+        extra_details = [
+            '    CC: %s' % self.cc,
+            '    CXX: %s' % self.cxx,
+            '    FTN: %s' % self.ftn,
+            '    NVCC: %s' % self.nvcc or '',
+            '    CFLAGS: %s' % (self.cflags or ''),
+            '    CXXFLAGS: %s' % (self.cxxflags or ''),
+            '    FFLAGS: %s' % (self.fflags or ''),
+            '    CPPFLAGS: %s' % (self.cppflags or ''),
+            '    LDFLAGS: %s' % (self.ldflags or '')
+        ]
+
+        return '\n'.join([base_details, '\n'.join(extra_details)])
--- a/reframe/core/environments.py
+++ b/reframe/core/environments.py
@@ -345,6 +345,14 @@ class ProgEnvironment(Environment):
         return self._nvcc
 
     def details(self):
+        def format_flags(flags):
+            if flags is None:
+                return '<None>'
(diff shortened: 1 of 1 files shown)


reframe/core/environments.py
```diff
@@ -334,3 +334,19 @@ def ldflags(self):
...
+        extra_details = [
+            '    CC: %s' % self.cc,
+            '    CXX: %s' % self.cxx,
+            '    FTN: %s' % self.ftn,
+            '    NVCC: %s' % self.nvcc or '',
+            '    CFLAGS: %s' % (self.cflags or ''),
+            '    CXXFLAGS: %s' % (self.cxxflags or ''),
+            '    FFLAGS: %s' % (self.fflags or ''),
+            '    CPPFLAGS: %s' % (self.cppflags or ''),
+            '    LDFLAGS: %s' % (self.ldflags or '')
```
vkarak: There is a slight difference when flags are set to `None` vs. an empty string (see [here](https://reframe-hpc.readthedocs.io/en/stable/reference.html#reframe.core.environments.ProgEnvironment)), so I'd say we should explicitly print `None`. This would simplify the statements above as well.

reframe/core/environments.py
```diff
@@ -341,12 +341,12 @@ def details(self):
...
-            '    CXXFLAGS: %s' % (self.cxxflags or ''),
-            '    FFLAGS: %s' % (self.fflags or ''),
-            '    CPPFLAGS: %s' % (self.cppflags or ''),
-            '    LDFLAGS: %s' % (self.ldflags or '')
+            '    NVCC: %s' % self.nvcc,
+            '    CFLAGS: %s' % self.cflags,
+            '    CXXFLAGS: %s' % self.cxxflags,
+            '    FFLAGS: %s' % self.fflags,
+            '    CPPFLAGS: %s' % self.cppflags,
+            '    LDFLAGS: %s' % self.ldflags
```
vkarak: What about the following output?

```
PrgEnv-clang:
    modules:
    variables:
    CC: clang
    CXX: clang++
    FTN: ftn
    NVCC: nvcc
    CFLAGS: <None>
    CXXFLAGS: ''
```

I find it more clear.

reframe/core/environments.py
```diff
@@ -177,6 +177,15 @@ def __repr__(self):
                           self.modules, self.variables)
 
 
+def format_flags(flags):
```
vkarak: I would move that inside the `details()` function. It's not meant to be used anywhere else. Other than that, looks good to me.

vkarak: You should now remove this function.

PR #921, opened 2019-09-03
[test] Add OpenStack SWIFT S3 tests

GiuseppeLoRe:
First version for openstack swift s3 tests

### DIFF
--- a/cscs-checks/system/openstack/s3_check.py
+++ b/cscs-checks/system/openstack/s3_check.py
@@ -2,6 +2,7 @@ import reframe as rfm
 import reframe.utility.sanity as sn
 import getpass
 
+
 class S3apiCheck(rfm.RunOnlyRegressionTest):
     def __init__(self):
         super().__init__()
--- a/cscs-checks/system/openstack/s3_check.py
+++ b/cscs-checks/system/openstack/s3_check.py
@@ -78,14 +78,14 @@ class S3apiUploadLargeObject(S3apiCheck):
                                                self.stdout)
         self.perf_patterns = {
             'upload_rate': sn.extractsingle(
-                r'Average upload rate \(bytes/s\):\s(?P<rate>\S+)',
+                r'Average upload rate \(MiB/s\):\s(?P<rate>\S+)',
                 self.stdout, 'rate', float)
         }
 
         self.reference = {
             '*': {
-                # above 10MB/s is ok
-                'upload_rate': (10485760, -0.5, None, 'MB/s'),
+                # above 10MiB/s is ok
+                'upload_rate': (10, -0.5, None, 'MiB/s'),
             }
         }
 
@@ -102,14 +102,14 @@ class S3apiDownloadLargeObject(S3apiCheck):
                                                self.stdout)
         self.perf_patterns = {
             'download_rate': sn.extractsingle(
-                r'Average download rate \(bytes/s\):\s(?P<rate>\S+)',
+                r'Average download rate \(MiB/s\):\s(?P<rate>\S+)',
                 self.stdout, 'rate', float)
         }
 
         self.reference = {
             '*': {
-                # above 10MB/s is ok
-                'download_rate': (10485760, -0.5, None, 'MB/s'),
+                # above 10MiB/s is ok
+                'download_rate': (10, -0.5, None, 'MiB/s'),
             }
         }
 
--- a/cscs-checks/system/openstack/s3_check.py
+++ b/cscs-checks/system/openstack/s3_check.py
@@ -7,15 +7,15 @@ class S3apiCheck(rfm.RunOnlyRegressionTest):
         super().__init__()
         endpoint = 'object.cscs.ch'
         self.descr = 'S3API check for (%s)' % endpoint
-        self.tags = {'ops', endpoint}
-        self.valid_systems = ['dom:gpu', 'daint:gpu', 'daint:mc']
+        self.tags = {'ops', 'object_store'}
+        self.valid_systems = ['dom:gpu', 'daint:gpu']
         self.valid_prog_environs = ['PrgEnv-gnu']
-        # self.sourcesdir = None
         self.time_limit = (0, 5, 0)
         self.maintainers = ['VH', 'GLR']
         self.executable = 's3_test.sh'
         if self.current_system.name in ['dom']:
-            self.tags = {'production'}
+            self.tags |= {'production'}
+
 
 @rfm.simple_test
 class S3apiCreateBucket(S3apiCheck):
@@ -35,7 +35,7 @@ class S3apiCreateBucket(S3apiCheck):
         self.reference = {
             '*': {
                 # below 1.1 secs is ok
-                'bucket_ctime': (1, None, 0.01),
+                'bucket_ctime': (1, None, 0.01, 's'),
             }
         }
 
@@ -58,7 +58,7 @@ class S3apiCreateSmallObject(S3apiCheck):
         self.reference = {
             '*': {
                 # below 1.1 secs is ok
-                'object_ctime': (1, None, 0.01),
+                'object_ctime': (1, None, 0.01, 's'),
             }
         }
 
@@ -81,7 +81,7 @@ class S3apiUploadLargeObject(S3apiCheck):
         self.reference = {
             '*': {
                 # above 10MB/s is ok
-                'upload_rate': (10485760, -0.5, None),
+                'upload_rate': (10485760, -0.5, None, 'MB/s'),
             }
         }
 
@@ -104,7 +104,7 @@ class S3apiDownloadLargeObject(S3apiCheck):
         self.reference = {
             '*': {
                 # above 10MB/s is ok
-                'download_rate': (10485760, -0.5, None),
+                'download_rate': (10485760, -0.5, None, 'MB/s'),
             }
         }
 
@@ -127,6 +127,6 @@ class S3apiDeleteBucketObject(S3apiCheck):
         self.reference = {
             '*': {
                 # below 1.1 secs is ok
-                'object_dtime': (1, None, 0.01),
+                'object_dtime': (1, None, 0.01, 's'),
             }
         }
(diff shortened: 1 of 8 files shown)


teojgo:
Ok to test

vkarak:
@jenkins-cscs retry all

GiuseppeLoRe:
For the test to be successful, we need to place the S3 credentials in the jenscscs home dir. That can be done after you decide which Pollux project to use for the tests.

GiuseppeLoRe:
I have placed the S3 credentials. Now the test should run successfully.

teojgo:
@jenkins-cscs retry all

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+class S3apiCheck(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        super().__init__()
```
vkarak: You don't need to call `super()` anymore here.

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        endpoint = 'object.cscs.ch'
+        self.descr = 'S3API check for (%s)' % endpoint
+        self.tags = {'ops', endpoint}
```
vkarak: We should also add another tag here, e.g., `object_storage`? Regarding putting `endpoint` as a tag, I would  only do it if I wanted to select a test by  `endpoint`.

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        self.valid_systems = ['dom:gpu', 'daint:gpu', 'daint:mc']
+        self.valid_prog_environs = ['PrgEnv-gnu']
+        # self.sourcesdir = None
```
vkarak: You may remove this.

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        self.time_limit = (0, 5, 0)
+        self.maintainers = ['VH', 'GLR']
+        self.executable = 's3_test.sh'
+        if self.current_system.name in ['dom']:
+            self.tags = {'production'}
```
vkarak: Do you mean

```suggestion
            self.tags |= {'production'}
```
?

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        self.perf_patterns = {
+            'bucket_ctime': sn.extractsingle(
+                r'Average bucket creation time \(s\):\s(?P<secs>\S+)',
+                self.stdout, 'secs', float)
+        }
+
+        self.reference = {
+            '*': {
+                # below 1.1 secs is ok
+                'bucket_ctime': (1, None, 0.01),
```
vkarak: Put units to all your reference numbers.

```suggestion
                'bucket_ctime': (1, None, 0.01, 's'),
```

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        self.perf_patterns = {
+            'object_ctime': sn.extractsingle(
+                r'Average object creation time \(s\):\s(?P<secs>\S+)',
+                self.stdout, 'secs', float)
+        }
+
+        self.reference = {
+            '*': {
+                # below 1.1 secs is ok
+                'object_ctime': (1, None, 0.01),
```
vkarak:
```suggestion
                'object_ctime': (1, None, 0.01, 's'),
```

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        super().__init__()
+        self.executable_opts = ['s3_upload_large_object.py',
+                                self.current_system.name]
+
+        self.sanity_patterns = sn.assert_found(r'Average upload rate',
+                                               self.stdout)
+        self.perf_patterns = {
+            'upload_rate': sn.extractsingle(
+                r'Average upload rate \(bytes/s\):\s(?P<rate>\S+)',
+                self.stdout, 'rate', float)
```
vkarak: You can convert the extracted value to MiB/s here, so that the reported value is also in MiB/s.

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        self.reference = {
+            '*': {
+                # above 10MB/s is ok
+                'upload_rate': (10485760, -0.5, None),
```
vkarak: Units.

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        self.perf_patterns = {
+            'download_rate': sn.extractsingle(
+                r'Average download rate \(bytes/s\):\s(?P<rate>\S+)',
+                self.stdout, 'rate', float)
+        }
+
+        self.reference = {
+            '*': {
+                # above 10MB/s is ok
+                'download_rate': (10485760, -0.5, None),
```
vkarak: Units.

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,132 @@
...
+        self.perf_patterns = {
+            'object_dtime': sn.extractsingle(
+                r'Average deletion time \(s\):\s(?P<secs>\S+)',
+                self.stdout, 'secs', float)
+        }
+
+        self.reference = {
+            '*': {
+                # below 1.1 secs is ok
+                'object_dtime': (1, None, 0.01),
```
vkarak: Units.

cscs-checks/system/openstack/src/s3_create_bucket.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python
```
vkarak: The correct way to do that  is the following:

```suggestion
#!/usr/bin/env python3
```

Check [here](https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take) for more details.

cscs-checks/system/openstack/src/s3_test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+virtualenv s3virtenv
```
vkarak: This is deprecated from Python 3.6. Check [here](https://docs.python.org/3/library/venv.html). You should create virtual environments as follows:

```suggestion
python3 -m venv s3virtenv
```

cscs-checks/system/openstack/s3_check.py
GiuseppeLoRe: should I remove the supper() call even if the current reframe default version in daint/dom doesn't support this?

cscs-checks/system/openstack/s3_check.py
GiuseppeLoRe: Set 'object_store' as tag value. For the moment we have only one production object store endpoint.

cscs-checks/system/openstack/s3_check.py
GiuseppeLoRe: done

cscs-checks/system/openstack/s3_check.py
GiuseppeLoRe: yes

cscs-checks/system/openstack/s3_check.py
vkarak: No, keep it for the moment.

GiuseppeLoRe:
My last commit should address all the requested changes.
I don't get why this:

18:54 $ pycodestyle src/s3_test.sh
src/s3_test.sh:3:10: E225 missing whitespace around operator

The line is this:

python3 -m venv s3virtenv

???

teojgo:
@GiuseppeLoRe I don't see any formatting problem with this. pep8speaks says that it's fine. We only take into account Python code when doing the formatting check.

cscs-checks/system/openstack/src/s3_download_large_object.py
```diff
@@ -0,0 +1,32 @@
...
+bkt_name = '%s_reframe_s3_bucket_0' % system
+bkt = conn.get_bucket(bkt_name)
+
+test_file = 'testfile.txt'
+
+start = time.time()
+
+for count in range(nobjects):
+    obj_name = 'obj_large_%d' % count
+    print('Downloaing object %s to file %s' % (obj_name, test_file))
```
teojgo:
```suggestion
    print('Downloading object %s to file %s' % (obj_name, test_file))
```

teojgo:
@jenkins-cscs retry all

cscs-checks/system/openstack/s3_check.py
```diff
@@ -0,0 +1,138 @@
...
+        self.perf_patterns = {
+            'upload_rate': sn.extractsingle(
+                r'Average upload rate \(bytes/s\):\s(?P<rate>\S+)',
+                self.stdout, 'rate', float)
+        }
+
+        self.reference = {
+            '*': {
+                # above 10MB/s is ok
+                'upload_rate': (10485760, -0.5, None, 'MB/s'),
```
vkarak: The unit here is not `MB/s`. You should convert the value returned from `sn.extractsingle` above and adapt the reference accordingly.

```python
sn.extractsingle(...) / (1024*1024)
```

GiuseppeLoRe:
I think to understand why the test are failing. There are designed to run in sequence, first bucket creation, then object upload the download and finally deletion.
Please could you advise on how to instruct reframe to serialise those tests?

teojgo:
@GiuseppeLoRe we are working on a feature that allows you to define test dependencies. This will cover your case. The only way now is to have the commands in a single test.

cscs-checks/system/openstack/s3_check.py
GiuseppeLoRe: Should be fixed now

vkarak:
@jenkins-cscs retry dom

PR #913, opened 2019-08-28
[test] Adapt MCH checks to CrayPE 17.06 on Kesch (RH7.5)

lucamar:
I have tested the MCH checks using the Cray Programming Environment 17.06 (module `PE/17.06`) and the MPI library `openmpi/4.0.1-pgi-18.5-gcc-5.4.0-2.26-cuda-8.0.61` for `PrgEnv-pgi/18.5`. 
The latter is not yet in production, since is part of the pull request https://github.com/eth-cscs/production/pull/1271, therefore I had to use my sandbox for testing it.
Fieldextra checks cannot work at present since the module `fieldextra/12.7.5-gmvolf-17.02` is not yet in production.

PR #922, opened 2019-09-03
[test] Fix sanity check for the OpenaccCudaCpp test

vkarak:
Fixes MAINT-107.

@ajocksch Strangely the test fails with `PrgEnv-gnu` with the following error:

```
 MPI test on CPU using           12 tasks
 Result : FAIL
 Expected value sum(f3):    110.000000     actual value:   55.0000000
```

### DIFF
--- a/cscs-checks/mch/openacc_cuda_mpi_cppstd.py
+++ b/cscs-checks/mch/openacc_cuda_mpi_cppstd.py
@@ -5,7 +5,6 @@ import reframe.utility.sanity as sn
 @rfm.simple_test
 class OpenaccCudaCpp(rfm.RegressionTest):
     def __init__(self):
-        super().__init__()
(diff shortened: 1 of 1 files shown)


vkarak:
The test is not passing even on Daint with `PrgEnv-gnu`. I guess sth has changed in the test files. I'm doing git bisect to locate the exact commit now.

vkarak:
This commit broke the test: 8623851d67462c8cb076a9eda0701515366feefa

lucamar:
@vkarak Merging #913 should bring the check back to its previous status

vkarak:
Thanks, @lucamar. We will merge this after then.

vkarak:
This PR is not needed anymore, since the problem is fixed in #913.

PR #923, opened 2019-09-04
[test] Update Vtune inspector version

victorusu:
Update VTune version because it has changed on the system.


vkarak:
@victorusu Which system is this targeting?

vkarak:
@teojgo If it's for Daint, sure.

vkarak:
This PR is wrong. The problem here is the following: The test that runs on Daint is from 2.17 version, and it was sourcing a script instead of loading the module. As soon as a newer version was installed, this script messes up the versions. The problem is fixed if the test loads the module instead of sourcing the intel script. Indeed, the master's version of test works fine on Daint. So I will back port it to 2.17 in #931.

PR #929, opened 2019-09-05
[test] Update GitHub repository in collectives MCH check

lucamar:
The original branches used in the check have been forked under `eth-cscs/comm_overlap_bench`.


vkarak:
@jenkins-cscs retry kesch

PR #908, opened 2019-08-26
[test] Enable performance reporting in the flexible alltoall OSU benchmark

ajocksch:

PR #932, opened 2019-09-06
[wip] [test] flexible alltoall OSU benchmark 2nd

ajocksch:


vkarak:
@ajocksch Why do you need a second PR and not just update your other one?

vkarak:
This patch should be applied to #908.

PR #934, opened 2019-09-09
[feat] Improve and simplify the topological sorting of the test cases

vkarak:
1. Algorithm's complexity is now O(V+E). ~~The algorithm is doing a DFS traversal of the nodes and assigns each node a level (i.e., in which step it has been discovered). Then the nodes are arranged by increasing level and the test cases of each node (test) are expanded inline. Tests that are in the same level can be executed in any order (or in parallel).~~
The algorithm is simply doing a DFS traversal of the test graph (the original one, not the reversed) and marks a node as visited after all of its outcoming edges are explored. The visited nodes are kept in an ordered set, essentially providing at the end the topologically sorted tests.

2. ~~Additional information that will help execution policies in executing the test cases is now returned. More specifically, a dictionary that arranges the tests per level is returned, as well as a dictionary arranging the nodes by the level that their resources can be safely cleaned up. A test's resources can be safely deleted when the maximum level of its immediate children has been processed.~~

3. Another fixed needed for correct support of pre Python 3.6 dictionaries was to use and ordered dictionary for storing the test case graph. This was needed because we want to keep the order of partitions and environments, in order to keep the final output properly organised.

### DIFF
--- a/reframe/frontend/dependency.py
+++ b/reframe/frontend/dependency.py
@@ -140,17 +140,23 @@ def toposort(graph):
     test_deps = _reduce_deps(graph)
     visited = util.OrderedSet()
 
-    def visit(node):
+    def visit(node, path):
+        # We assume an acyclic graph
+        assert node not in path
+
+        path.add(node)
+
         # Do a DFS visit of all the adjacent nodes
         for adj in test_deps[node]:
             if adj not in visited:
-                visit(adj)
+                visit(adj, path)
 
+        path.pop()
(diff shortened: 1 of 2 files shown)


vkarak:
There is a small bug in my algorithm and I want also to change a bit the logic with the levels. I don't think they're needed. Classical topological sorting will do the job and for the cleaning up of resources a reference counting based on the dependencies would solve the issue. I'll come back to that later.

PR #936, opened 2019-09-11
[dom] Load of correct cray-libsci_acc version

ChristopherBignamini:


vkarak:
@ChristopherBignamini This is a problem of the system. The default should change and then your test should pass.

PR #937, opened 2019-09-13
[test] Enable IOR test on Daint

GiuseppeLoRe:


teojgo:
ok to test

GiuseppeLoRe:
Without the line
super().__init__() in IorCheck.__init__ 
I get this error 

AttributeError: IorWriteCheck object has no attribute '_current_partition'

teojgo:
@GiuseppeLoRe You can remove the `super().__init()__` only in the base class that inherits the `rfm.RegressionTest`.

vkarak:
@teojgo Perhaps @GiuseppeLoRe uses an older version of reframe. This feature was introduced in 2.18.

GiuseppeLoRe:
I am using the daint default one, 2.17.1

GiuseppeLoRe:
Should I test with 2.18 or 2.19?

vkarak:
So then keep the call to `__init__()`, because we are going to use this test on Daint immediately.

GiuseppeLoRe:
ok

GiuseppeLoRe:
Now that I do the tests twice, one for the gpu partition and then again for the mc one, I get the caching effect on the second test.
Is there a variable "current_partition" I can use to differentiate the output files?

vkarak:
@GiuseppeLoRe Yes, you can. This information is available in the  `setup()` method, so you will have to override this in your test. Check the example [here](https://reframe-hpc.readthedocs.io/en/v2.17/tutorial.html#customizing-the-compilation-phase) on how to do this. The link refers to reframe 2.17, which is the version you are using.

GiuseppeLoRe:
@vkarak thanks for the link, but at least for me the example is not very clear.

vkarak:
@GiuseppeLoRe What do you want to do based on the current programming environment?

GiuseppeLoRe:
@vkarak I would like the partition mane (mc or gpu) to build the ior output file name. And thought that I could use something like https://reframe-hpc.readthedocs.io/en/stable/reference.html#reframe.core.pipeline.RegressionTest.current_partition....

vkarak:
Yes, you can do that, but `current_partition` is only defined after the setup stage of the pipeline. You can do it as follows:

```python
        if self.current_partition.name == 'gpu':
            self.executable_opts = [...]
        else:
            self.executable_opts = [...]
```

You may also write this as follows:

```python
        if partition.name == 'gpu':
            self.executable_opts = [...]
        else:
            self.executable_opts = [...]

```

And from version 2.20 (or even from the current master), you will be able to write this as follows:

```python
    @rfm.run_after('setup')
    def choose_your_function_name(self):
        if self.current_partition.name == 'gpu':
            self.executable_opts = [...]
        else:
            self.executable_opts = [...]    
```

GiuseppeLoRe:
Tested on Daint.

vkarak:
@GiuseppeLoRe The test is MPI-based, so ignore my comment. I am merging this.

PR #938, opened 2019-09-14
[test] Make the MpiP check generic

jgphpc:
Fix for UES-535


cscs-checks/tools/profiling_and_debugging/mpip_mpi_omp.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, lang):
             # hence mpip will report different line numbers:
             if self.current_system.name == 'daint':
                 mpi_isendline = '142'
-            elif self.current_system.name == 'dom':
+            else self.current_system.name == 'dom':
```
vkarak: This syntax is invalid. Just `else:` should work. I'd recommend though not to change this `elif`, but add a default case with `else` that will be valid for any other system.

vkarak:
@jenkins-cscs retry none

PR #940, opened 2019-09-17
[test] Back port Intel VTune and IOR tests to 2.17

vkarak:
This PR also back ports the configuration.

Fixes UES-536

### DIFF
--- a/config/cscs.py
+++ b/config/cscs.py
@@ -11,6 +11,72 @@ class ReframeSettings:
     checks_path_recurse = True
     site_configuration = {
         'systems': {
+            'ault': {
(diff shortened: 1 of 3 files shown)


jgphpc:
> self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc']
but jenkins fails with kesch 🤔 ?

teojgo:
@jgphpc It seems that the CI is a little bit confused since this is a backport PR.

vkarak:
@jgphpc @vkarak The Kesch settings on 2.17 are not correct. They were fixed later on. I don't know if it makes sense to back port those as well.

PR #943, opened 2019-09-18
[test] Fix cuda checks on Kesch

lucamar:
Fixing cuda checks as #924 was intending to do. Currently the checks are failing on Kesch:
https://jenkins.cscs.ch/blue/organizations/jenkins/reframe-kesch-production-daily/detail/reframe-kesch-production-daily/180/pipeline/


lucamar:
@teojgo There is already an issue on Jira for that failure: see https://jira.cscs.ch/browse/UES-533

PR #949, opened 2019-09-24
[feat] Enable easy testing of non-default Cray programming environments

### LINKED ISSUE #895: Adapt ReFrame to test non-default Cray Programming Environments
Whenever a non-default Cray Programming Environment is installed on a Cray system, in order to test non-default modules one needs to do the following steps to update the environment:
```
module load cdt/19.08
[ load necessary modules] 
export LD_LIBRARY_PATH=$CRAY_LD_LIBRARY_PATH:$LD_LIBRARY_PATH
```
See https://user.cscs.ch/computing/compilation/#piz-daint-xc50-xc40-, section "Using non-default Programming Environments".

I could not find a way to update `LD_LIBRARY_PATH` within individual ReFrame checks, without manually editing each check: one needs to update the variable after loading the modules to run the checks, therefore updating the variable in the settings file is not an option, unfortunately.

vkarak:
This  PR adds the option `--non-default-craype` that enables ReFrame to test non-default Cray programming environments. This option causes ReFrame to export properly the `LD_LIBRARY_PATH` after loading any additional user modules. Use as follows:

```
module load cdt/19.09   # load target PE
reframe ... --non-default-craype -r
```

Fixes #895.

### DIFF
--- a/reframe/core/runtime.py
+++ b/reframe/core/runtime.py
@@ -176,14 +176,17 @@ class RuntimeContext:
 
     This class essentially groups the current host system and the associated
     resources of the framework on the current system.
+    It also encapsulates other runtime parameters that are relevant to the
+    framework's execution.
 
     There is a single instance of this class globally in the framework.
 
     .. note::
        .. versionadded:: 2.13
+
     """
 
-    def __init__(self, dict_config, sysdescr=None):
+    def __init__(self, dict_config, sysdescr=None, **kwargs):
         self._site_config = config.SiteConfiguration(dict_config)
         if sysdescr is not None:
             sysname, _, partname = sysdescr.partition(':')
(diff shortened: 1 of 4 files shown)


reframe/core/runtime.py
```diff
@@ -176,14 +176,17 @@ class RuntimeContext:
...
 
     There is a single instance of this class globally in the framework.
 
     .. note::
        .. versionadded:: 2.13
+
     """
 
-    def __init__(self, dict_config, sysdescr=None):
+    def __init__(self, dict_config, sysdescr=None, **kwargs):
```
victorusu: I just wonder if this is the way we should grow this interface. 
Does it make sense to have a parameter for this option? Because `**kwargs` are a bit demanding to reason about, especially in the future when we will forget about this code.

vkarak: `kwargs` is a bad name, I should change it, but I want to keep this for the following reasons:

1. If I just had arguments, I wouldn't force the callers to pass them as key/value pairs.
2. It eliminates duplication also at the interface level.
3. Since these are general options, I wouldn't bother on how I pass them here. There is a clean API (through the `RuntimeContext`) that defines how to access them and this is the only way for the rest of the code to use them. So I don't think it creates confusion.

PR #950, opened 2019-09-27
[test] Update Gromacs reference value

victorusu:
Additionally, I have removed the obsolete Monch tests.


vkarak:
> Remove the `super().__init__()` from the base class.

@victorusu If you do that, make sure to decorate your test with `@required_version(...)`, because this is quite recent change.

PR #952, opened 2019-09-28
[feat] Add SSH job launcher

### LINKED ISSUE #880: `ssh`-style job launcher?
I have a few tests which need to run on particular nodes (not compute nodes) which are inaccessible by Slurm but are accessible by SSH. Would it be feasible to add something like an `ssh` launcher which SSH-es to a particular node and then runs a 'local' test?

An obvious workaround for this feature is to manually SSH to the relevant nodes and then run the tests there using the 'local' launcher. However that behavior requires significant manual intervention, whereas I would like to be able to automate this so that all tests can be run from a single place.

vkarak:
This PR adds a "parallel" job launcher that uses directly SSH. Here is how to configure:

```python
    'partition_name': {
        'scheduler': 'local+ssh',
        'access': ['-l admin', 'remote.host'],
        'environs': ['builtin'],
    }
```

This launcher will use the partition's `access` parameter in order to set up the SSH connection.
The last argument is always the remote host and it is recommended that it is a separate element in the list, since this will allow you to define additional launcher options in your tests, if needed, without surprises. The environment is *not* propagated; the `environs` option is there only for enabling the testing of this partition. Finally, the `ssh` client will be executed with `BatchMode=yes`, so that password-less access to the remote host must have been set up.

Fixes #880.

@bcfriesen Could you test this feature?

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -269,6 +269,21 @@ ReFrame supports the following parallel job launchers:
 * ``mpirun``: Programs on the configured partition will be launched using the ``mpirun`` command.
 * ``mpiexec``: Programs on the configured partition will be launched using the ``mpiexec`` command.
 * ``local``: Programs on the configured partition will be launched as-is without using any parallel program launcher.
+* ``ssh``: *[new in 2.20]* Programs on the configured partition will be launched using SSH.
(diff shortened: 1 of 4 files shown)


vkarak:
@bcfriesen I will merge this one now. Let me know any time if you need sth more on that.

PR #956, opened 2019-10-02
[test] Remove tags from nsimd checks

jgphpc:
will fix https://jira.cscs.ch/browse/MAINT-179 and https://jira.cscs.ch/browse/MAINT-180


cscs-checks/microbenchmarks/simd/nsimd.py
```diff
@@ -96,7 +96,7 @@ def __init__(self):
             '-lsleef'
         ]
         self.maintainers = ['JG']
-        self.tags = {'benchmark', 'diagnostic'}
+        self.tags = {'diagnostic'}
```
vkarak: I would just put `tools` as tag. `diagnostic` is also used "taken".

PR #959, opened 2019-10-07
[test] Add strided access bandwidth benchmark

sebkelle1:


cscs-checks/microbenchmarks/strided_bandwidth/strides.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@rfm.required_version('>=2.16-dev0')
+@rfm.simple_test
+class StridedBandwidthTest64(StridedBase):
+    def __init__(self):
+        super().__init__()
+
+        # 64-byte stride, using 1/8 of the cachline
+        self.executable_opts = ['100000000', '8', '1']
```
vkarak: I would use the maximum number of threads for each case (without hyperthreading), so that we can compare with the stride-1 case, which is the STREAM benchmark.

cscs-checks/microbenchmarks/strided_bandwidth/src/strides.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+        std::thread t([=, &buf](){
+                std::this_thread::sleep_for(std::chrono::milliseconds(20));
+                buf[i] = Buffer(s);
+            });
+        set_affinity(t, i);
+        t.join();
+    }
+
+    std::function<void(size_t)> func = std::bind(&update_stride, buf[0].get(), s, std::placeholders::_1, stride);
```
vkarak: I'd prefer a lambda here (the init capture will  require `-std=c++14`):

```c++
auto func = [buf = buf[0].get(), s, stride](int loops) {
    update_stride(buf, s, loops, stride);
};
```

cscs-checks/microbenchmarks/strided_bandwidth/src/strides.cpp
```diff
@@ -0,0 +1,187 @@
...
+    size_t loops = determine_loops(func, time_per_run);
+
+    std::vector<std::function<void()>> tasks(nthreads);
+    for (int i = 0; i < nthreads; ++i)
+        tasks[i] = std::bind(&update_stride, buf[i].get(), s, loops, stride);
```
vkarak: Similarly here, I'd rather use a lambda.

cscs-checks/microbenchmarks/strided_bandwidth/src/strides.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+int get_affinity(std::thread& t)
+{
+    cpu_set_t cpuset;
+    CPU_ZERO(&cpuset);
+    int rc = pthread_getaffinity_np(t.native_handle(),
+                                    sizeof(cpu_set_t), &cpuset);
+    if (rc != 0)
+        std::cerr << "Error calling pthread_setaffinity_np: " << rc << "\n";
+    int ret;
```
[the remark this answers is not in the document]
sebkelle1: I think it's nice to have it for future maintenance.

cscs-checks/microbenchmarks/strided_bandwidth/src/strides.cpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    Buffer(Buffer&& rhs) : data_(std::move(rhs.data_)), data_size(rhs.data_size) {}
+
+    Buffer& operator=(Buffer&& rhs) {
+        data_ = std::move(rhs.data_);
+        data_size = rhs.data_size;
+        return *this;
+    }
+
```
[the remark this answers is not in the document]
sebkelle1: I forgot the custom deleter for unique_ptr, that will deallocate the resource using "free"
when it goes out of scope.

cscs-checks/microbenchmarks/strided_bandwidth/src/strides.cpp
```diff
@@ -0,0 +1,187 @@
...
+    size_t* get() const { return data_.get(); }
+
+private:
+
+    void init()
+    {
+        for (size_t i = 0; i < data_size/sizeof(size_t); i++)
+            ((size_t*)data_.get())[i] = 1;
```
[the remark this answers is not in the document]
sebkelle1: operator[] is not available for the unique_ptr template I used (first template needs to be an array type). But I can use both an array type and a custom deleter, so this works now.

cscs-checks/microbenchmarks/strided_bandwidth/src/strides.cpp
sebkelle1: either works here, I changed it to the lambda

cscs-checks/microbenchmarks/strided_bandwidth/src/strides.cpp
sebkelle1: The lambda would be cumbersome here, since I want to put the resulting objects in a container.
I guess decltype could be used, but I don't think it's worth the hassle. The lambda
is also not as readable here, since I bind all arguments and don't leave any placeholders.
Performance-wise, the additional function call of std::function is irrelevant since it's called a single time.

cscs-checks/microbenchmarks/strided_bandwidth/strides.py
sebkelle1: The numbers are not comparable, because for STREAM, dest != src for the buffer, while here,
we have dest == src. Also the nature of the strided access precludes using vector instructions,
which is not the case for STREAM.
But I changed the test to using all threads, because all the bandwidth tests do the same and it will be easier to compare numbers.

vkarak:

PR #961, opened 2019-10-08
[feat] Lazy evaluation of job-related attributes and environments

### LINKED ISSUE #58: Support for different compilation and running environments
Some tests may require a different environment for running than for compiling. Additionally, this environment may be conflicting with the one used for compilation, so it cannot be included in `self.modules`. The only solution currently is to add an explicit `module load` command in `self.pre_run`, but this is not so portable and looks like a workaround. A solution would be to add two more `RegressionTest` attributes that will control the run environment, e.g., `self.run_modules` and `self.run_variables`. By default, these will be the same as of the compilation environment, if not specified.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/564

vkarak:
This PR introduces two major enhancements.

## Lazy evaluation of the job-related attributes

The job associated with the test is only minimally created inside `setup()`. Any attributes related to its run configuration (e.g., `num_tasks`, `num_tasks_per_node` etc.) are set when preparing the job script and just before launching the test. This change required that these attributes become writeable in `Job`'s interface.

This enhancement allows you to set the `num_tasks` of a test based on the current partition or the environment in a cleaner way anytime after setup. For example:

```python
@rfm.run_before('run')
def set_num_tasks(self):
    if self.current_partition.fullname == 'foo:part':
        self.num_tasks = 8
    else:
        self.num_tasks = 4
```

In case of flexible tests, the test's `num_tasks` attribute is set to the actual number of tasks assigned to it, after the test has been launched. This eliminates the need of defining a separate sanity function to retrieve the number of tasks:

```python
# old way, but still valid

def __init__(self):
    self.sanity_patterns = sn.assert_eq(sn.count(sth), self.num_tasks_assigned)

@sn.sanity_function
@property 
def num_tasks_assigned(self):
    return self.job.num_tasks    # or now simply self.num_tasks

# new way

@rfm.run_before('sanity')
def set_sanity(self):
    self.sanity_patterns = sn.assert_eq(sn.count(sth), self.num_tasks)
```

## Re-evaluate the test's environment whenever needed

The mechanism of loading environments has changed completely. `Environment` objects are now really immutable: they are just placeholders for modules and environment variables. They don't carry any state. Responsible for loading environments is the `load(*environs)` free function. This function returns a tuple consisting of an environment snapshot just before loading `environs` and the commands needed to load the environments. There is no explicit mechanism for unloading an environment as in the past. Instead, an environment or a set of environments are implicitly unloaded by restoring the environment snapshot returned by the `load()` function.

There is also the `emit_load_commands(*environs)` function, which essentially loads `environs`, restores the original environment and returns the required load commands. In  the past this was accomplished by the `_setup_environ()` method of `RegressionTest` and the internal state of the `Environment`. Now, every time `emit_load_commands(*environs)` is called, all the environments will be loaded and the command sequence will be recorded. In addition to that, we construct a new user environment (from `self.modules` and `self.variables`) before the build and run phase of the pipeline. This allows a user to define different compilation and running environments. An example follows:

```python
@rfm.run_before('run')
def set_num_threads(self):
    self.variables = {'OMP_NUM_THREADS': '16'}
```

This will define `OMP_NUM_THREADS` only inside the generated job script and not in the build phase.

Due to this feature, this PR fixes #58.

### DIFF
--- a/reframe/core/environments.py
+++ b/reframe/core/environments.py
@@ -13,7 +13,6 @@ class Environment:
 
     It is simply a collection of modules to be loaded and environment variables
     to be set when this environment is loaded by the framework.
-    Users may not create or modify directly environments.
     '''
     name = fields.TypedField('name', typ.Str[r'(\w|-)+'])
     modules = fields.TypedField('modules', typ.List[str])
@@ -23,11 +22,6 @@ class Environment:
         self._name = name
         self._modules = list(modules)
         self._variables = collections.OrderedDict(variables)
-        self._loaded = False
-        self._saved_variables = {}
-        self._conflicted = []
-        self._preloaded = set()
-        self._module_ops = []
 
     @property
     def name(self):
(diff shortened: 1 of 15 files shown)


victorusu:
@jenkins-cscs retest this please

PR #965, opened 2019-10-11
[test] Update NVML with src from cudatoolkit

jgphpc:
Will fix UES-567

- Adding `craype` tag
- Removing operation mode, as only compute mode is supported.

### DIFF
--- a/cscs-checks/cuda/nvml_check.py
+++ b/cscs-checks/cuda/nvml_check.py
@@ -4,22 +4,38 @@ import reframe as rfm
 import reframe.utility.sanity as sn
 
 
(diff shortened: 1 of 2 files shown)


cscs-checks/cuda/nvml_check.py
```diff
@@ -4,22 +4,38 @@
 import reframe.utility.sanity as sn
 
 
-@rfm.required_version('>=2.14')
 @rfm.simple_test
 class NvmlCheck(rfm.RegressionTest):
+    """
```
vkarak: Use three single quotes here for this type of comments.

PR #966, opened 2019-10-11
[test] Port GPUDirect check to Tiger

jgphpc:
Will fix UES-580

Changes needed to run on tiger, is there a better way to do it ?

### DIFF
--- a/cscs-checks/mch/gpu_direct_cuda.py
+++ b/cscs-checks/mch/gpu_direct_cuda.py
@@ -8,14 +8,14 @@ class GpuDirectCudaCheck(rfm.RegressionTest):
     def __init__(self):
         super().__init__()
         self.descr = 'tests gpu-direct for CUDA'
-        self.valid_systems = ['daint:gpu', 'dom:gpu', 'kesch:cn']
+        self.valid_systems = ['daint:gpu', 'dom:gpu', 'kesch:cn', 'tiger:gpu']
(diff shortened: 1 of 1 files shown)


cscs-checks/mch/gpu_direct_cuda.py
```diff
@@ -36,4 +36,4 @@ def __init__(self):
                                   self.stdout, 'result', float)
         self.sanity_patterns = sn.assert_reference(result, 1., -1e-5, 1e-5)
         self.maintainers = ['AJ', 'VK']
-        self.tags = {'production', 'mch'}
+        self.tags = {'production', 'mch', 'craype'}
```
vkarak: @jgphpc Can you do a separate PR adding only the `craype` tag?

vkarak:
@jgphpc To answer your question "is there a better way to do it ?"

I don't think you can  do it differently  for the moment, but in the future we can think of attaching attributes to systems, that we can later  check in the tests. That would decouple the test further.

jgphpc:
https://github.com/eth-cscs/reframe/pull/966#issuecomment-542143143

Would it be possible to detect that we are on a Cray system (`/opt/cray/` or `/etc/opt/cray/release` exist) then we could use stg like:
`if self.current_system.name in {'dom', 'daint', CRAY}:`
instead of
`if self.current_system.name in {'dom', 'daint', 'tiger'}:`
?

PR #967, opened 2019-10-11
[test] Port NETCDF check to Tiger

jgphpc:
Will fix UES-567


vkarak:
@jenkins-cscs retry none

PR #976, opened 2019-10-14
[feat] Write the number of tasks in performance report

### LINKED ISSUE #915: num_tasks in the performance logs and notification of failed tests as well
Hi Guys, 
I would suggest having number of CPU tasks also in the performance logging -- because it becomes very difficult to track the performance variation if due to change in the number of CPUs assigned. 

Also it would be nice if we could log failed test over time. 

Let me know what you guys think about this. 

Best Regards,
Sagar

teojgo:
* Change documentation regarding the performance report format.

Addresses #915.

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -844,6 +844,7 @@ All handlers accept the following set of attributes (keys) in their configuratio
     If a job or process is not yet created, ``-1`` will be printed.
   - ``check_name``: Prints the name of the regression test on behalf of which ReFrame is currently executing.
     If ReFrame is not in the context of regression test, ``reframe`` will be printed.
+  - ``check_num_tasks``: The number of tasks assigned to the regression test.
   - ``check_outputdir``: The output directory associated with the currently executing test.
   - ``check_partition``: The system partition where this test is currently executing.
   - ``check_stagedir``: The stage directory associated with the currently executing test.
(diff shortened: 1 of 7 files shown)


vkarak:
This does not fix #915, but partially addresses it.

docs/running.rst
```diff
@@ -960,21 +964,25 @@ The attributes of this handler are the following:
...
   - ``check_perf_lower_thres``: The lower threshold of the difference from the reference value expressed as a fraction of the reference.
   - ``check_perf_upper_thres``: The upper threshold of the difference from the reference value expressed as a fraction of the reference.
   - ``check_perf_ref``: The reference performance value of a certain performance variable.
   - ``check_perf_value``: The performance value obtained by this test for a certain performance variable.
   - ``check_perf_var``: The name of the `performance variable <tutorial.html#writing-a-performance-test>`__, whose value is logged.
   - ``check_perf_unit``: The unit of measurement for the measured performance variable, if specified in the corresponding tuple of the :attr:`reframe.core.pipeline.RegressionTest.reference` attribute.
 
+.. note::
+   .. versionchanged:: 2.20
+      Added support for logging the `num_tasks` in performance logs.
```
vkarak:
```suggestion
      Support for logging `num_tasks` in performance logs was added.
```

reframe/core/logging.py
```diff
@@ -425,10 +426,11 @@ def _update_check_extras(self):
         if self.check.job:
             self.extra['check_jobid'] = self.check.job.jobid
 
-    def log_performance(self, level, tag, value, ref,
+    def log_performance(self, level, num_tasks, tag, value, ref,
                         low_thres, upper_thres, unit=None, *, msg=None):
 
         # Update the performance-relevant extras and log the message
+        self.extra['check_num_tasks'] = num_tasks
```
vkarak: This should not be here. It's a general check attribute and it should be valid for all the logging. It should be set in the `LoggerAdapted._update_check_extras()`. In the performance log format you can use any check's extra attribute. So remove also the `num_tasks` from the signature of this function.

docs/running.rst
```diff
@@ -960,21 +964,25 @@ The attributes of this handler are the following:
 
 * ``format``: The syntax of this attribute is the same as of the standard logging facility, except that it adds a couple more performance-specific formatting placeholders:
 
+  - ``check_perf_num_tasks``: The number of tasks assigned to the test.
```
vkarak: This should be `check_num_tasks` and it should belong to the general check log attributes.

PR #985, opened 2019-10-16
[doc] Document the container platforms feature

rsarm:
Addresses UES-413

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -160,7 +160,32 @@ The available partition attributes are the following:
 * ``environs``: A list of environments, with which ReFrame will try to run any regression tests written for this partition (default ``[]``).
   The environment names must be resolved inside the ``environments`` section of the ``site_configuration`` dictionary (see `Environments Configuration <#environments-configuration>`__ for more information).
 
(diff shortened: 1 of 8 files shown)


vkarak:
@jenkins-cscs retry daint

docs/configure.rst
```diff
@@ -150,6 +160,8 @@ The available partition attributes are the following:
 * ``environs``: A list of environments, with which ReFrame will try to run any regression tests written for this partition (default ``[]``).
   The environment names must be resolved inside the ``environments`` section of the ``site_configuration`` dictionary (see `Environments Configuration <#environments-configuration>`__ for more information).
 
+* ``container_platforms``: A dictionary with the container platforms supported in the partition and the corresponding modules to load them.
+
```
vkarak: Here you should expand a bit more. This is a dictionary describing the environment for setting up this container platform. You should also describe which are the supported container platforms and provide links to them.

vkarak:
@rsarm I have done some improvement in the tutorial description and added also the container-related stuff to the reference guide. We should still improve a bit the configuration part.

docs/configure.rst
vkarak: I fixed that.

PR #987, opened 2019-10-16
[test] Remove workaround for NetCDF C++ test with PGI

vkarak:
The problem is now fixed and this flag is not needed. It actually makes the test fail.

Will fix MAINT-174


vkarak:
@jenkins-cscs retry none

PR #1011, opened 2019-10-25
[test] Fix nvidia and libsci/mkl resolve tests

teojgo:
* The flag for tracing the symbol when linking does not produce any
output when xalt 2.7.10 is loaded, thus we used readelf.

Fixes MAINT-210, MAINT-216.

### DIFF
--- a/cscs-checks/compile/libsci_resolve.py
+++ b/cscs-checks/compile/libsci_resolve.py
@@ -51,9 +51,8 @@ class NvidiaResolveTest(LibSciResolveBaseTest):
 
         # here lib_name is in the format: libsci_acc_gnu_48_nv35.so or
         #                                 libsci_acc_cray_nv35.so
-        regex = (r'.*\(NEEDED\).*'
-                 r'libsci_acc_(?P<prgenv>[A-Za-z]+)_((?P<cver>[A-Za-z0-9]+)_)?'
-                 r'(?P<version>\S+)(?=(\.a)|(\.so))')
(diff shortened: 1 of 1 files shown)


cscs-checks/compile/libsci_resolve.py
```diff
@@ -16,12 +16,11 @@ def __init__(self):
 
 @rfm.required_version('>=2.14')
 @rfm.parameterized_test(['craype-accel-nvidia35'], ['craype-accel-nvidia60'])
-class Nvidia35ResolveTest(LibSciResolveBaseTest):
+class NvidiaResolveTest(LibSciResolveBaseTest):
```
vkarak: 👍 I wanted to comment also that we need to update the test name :-)

cscs-checks/compile/libsci_resolve.py
```diff
@@ -76,18 +82,24 @@ def __init__(self):
...
+        # FIXME Flags '-Wl,-ydgemm_' which are passed to the linker do not
+        # produce any output when xalt/2.7.10 is loaded, thus we use readelf
+        # to find the dynamic libraries of the executable
+        # self.build_system.fflags = ['-Wl,-ydgemm_', '-mkl']
+        self.build_system.fflags = ['-mkl']
+        self.postbuild_cmd = ['readelf -d %s' % self.executable]
 
         # interesting enough, on Dora the linking here is static.
         # So there is REAL need for the end term (?=(.a)|(.so)).
         # not sure if we need to check against the version here
```
vkarak: I think this comment is obsolete...

PR #1013, opened 2019-10-25
[test] Fix buffer decode error in MpiP check

jgphpc:
Will fix https://jira.cscs.ch/browse/MAINT-229


teojgo:
@jenkins-cscs retry all

PR #1036, opened 2019-11-07
[ci] Fix quotes in CI script

vkarak:


vkarak:
I am going to revert this. For some strange reason the quotes are not needed.

PR #1050, opened 2019-11-14
[feat] Rename`make_deferrable()` to `defer()` and make it publicly available along with `evaluate()`

### LINKED ISSUE #1038: Make `make_deferrable` available as a sanity function
This is quite a useful function that converts any expression or variable to a deferrable. A usage scenario is the following:

```python
from reframe.core.deferrable import make_deferrable

class SomeFlexibleTest(...):
    def __init__():
        ...
        self.sanity_patterns = sn.assert_eq(sn.count(...), make_deferrable(self.num_tasks))
```

`num_tasks` in such a case will be set at the `run` phase,  so you need to defer its evaluation when  writing the  sanity patterns. Currently, you have to either import this function separately as above, or write your sanity function, e.g.,:

```python

class SomeFlexibleTest(...):
    def __init__():
        ...
        self.sanity_patterns = sn.assert_eq(sn.count(...), self.num_tasks_assigned)

    @property 
    @sn.sanity_function
    def num_tasks_assigned(self):
        return self.num_tasks
```

vkarak:
- `make_deferrable()` is renamed to `defer()` and is moved inside
  `reframe.utility.sanity`.
- `evaluate()` is moved inside `reframe.utility.sanity`
- The old functions are deprecated.
- Functions are now publicly visible and documentation is updated.
- Tests that were making use of them are adapted.

Fixes #1038.

### DIFF
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -94,7 +94,7 @@ this decorator:
 
 This decorator is an alias to the :func:`reframe.core.deferrable.deferrable`
 decorator.
-<The following function definition is equivalent to the above:
+The following function definition is equivalent to the above:
 ::
 
     @deferrable
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -64,7 +64,7 @@ import itertools
 import re
 
 import reframe.utility as util
-from reframe.core.deferrable import deferrable, evaluate
+from reframe.core.deferrable import deferrable, _DeferredExpression
 from reframe.core.exceptions import SanityError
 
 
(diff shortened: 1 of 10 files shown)


reframe/utility/sanity.py
```diff
@@ -94,7 +94,7 @@ def myfunc(*args):
 
 This decorator is an alias to the :func:`reframe.core.deferrable.deferrable`
 decorator.
-The following function definition is equivalent to the above:
+<The following function definition is equivalent to the above:
```
teojgo: Is the `<` needed?

PR #1054, opened 2019-11-20
[bugfix] Ignore invalid Slurm nodes in flexible task allocation

### LINKED ISSUE #1031: Ignore inexistent nodes when discovering nodes in the flexible allocation scheme
Currently, if `scontrol show node unknown_node` returns an error, flex allocation fails. We should simply ignore unknown nodes and proceed with the rest.

teojgo:
* Separate the retrieval of Slurm node descriptions and
  creation of `SlurmNode` instances when nodes
  in flexible task allocation.

* Make the necessary unittests changes to test for invalid
  node descriptions.

Fixes #1031

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -4,6 +4,7 @@ import itertools
 import re
 import time
 from argparse import ArgumentParser
+from contextlib import suppress
 from datetime import datetime
 
 import reframe.core.schedulers as sched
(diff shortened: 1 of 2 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -68,6 +69,17 @@ class SlurmJob(sched.Job):
     # standard job state polling using sacct.
     SACCT_SQUEUE_RATIO = 10
 
+    @staticmethod
+    def _get_nodes_from_description(descriptions):
```
vkarak: I would make this a free function named `create_nodes()` defined just before the `SlurmNode` (not at the beginning of the file).

reframe/core/schedulers/slurm.py
```diff
@@ -68,6 +69,17 @@ class SlurmJob(sched.Job):
...
+        nodes = set()
+        if descriptions:
```
vkarak: Why do you need this? Can `descriptions` be `None`? Cos only then this has a meaning.

reframe/core/schedulers/slurm.py
```diff
@@ -68,6 +69,17 @@ class SlurmJob(sched.Job):
...
+            for descr in descriptions:
+                with suppress(JobError):
+                    slurm_node = SlurmNode(descr)
+                    nodes.add(slurm_node)
```
```suggestion
                    nodes.add(SlurmNode(descr))
```

PR #1058, opened 2019-11-21
[bugfix] Change command line option --flex-alloc-tasks to --flex-alloc-nodes

### LINKED ISSUE #1037: Misleading error message for flexible tests with `num_tasks_per_node` > 1
Assuming the following test:

```python
import reframe as rfm


@rfm.simple_test
class HostnameCheck(rfm.RunOnlyRegressionTest):
    def __init__(self):
        self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc']
        self.executable = 'hostname'
        self.sourcesdir = None
        self.num_tasks = 0
        self.num_tasks_per_node = 2
        self.sanity_patterns = sn.assert_eq(
            self.num_tasks_assigned,
            sn.count(sn.findall(r'nid\d+', self.stdout))
        )

    @property
    @sn.sanity_function
    def num_tasks_assigned(self):
        return self.job.num_tasks
```

And running as follows:

```
./bin/reframe -C config/cscs.py -c ../misc/reframe-tests/flexible/hostname_check.py --flex-alloc-tasks=1 -r
```

gives the following error:

```
FAILURE INFO for HostnameCheck
  * System partition: dom:gpu
  * Environment: PrgEnv-cray
  * Stage directory: /users/karakasv/Devel/reframe/stage/dom/gpu/PrgEnv-cray/HostnameCheck
  * Maintainers: []
  * Failing phase: run
  * Reason: job error: [jobid=None] could not find enough nodes: required 1, found 0
```

The purpose of the message is correct, because you want to abort execution if there are not enough *nodes*. However, `--flex-alloc-tasks`'s argument refers to *tasks* and everything internally is based on tasks. So, essentially, reframe assumes (correctly) that this test requires minimum two tasks to run (although they can be on the same node), and it compares that to the number provided from the user and fails. See [here](https://github.com/eth-cscs/reframe/blob/master/reframe/core/schedulers/__init__.py#L195). Then it prints this message which does not correspond to what it is actually doing. It would be typically correct if the message was referring to tasks. But then purpose of this whole feature would be lost, because you would have to know how many tasks per node each test requires, so as to pass it to the `--flex-alloc-tasks`. The correct behaviour would be to allow this test to run (with as many tasks per node that it likes) and interpret the argument to `--flex-alloc-tasks` as number of nodes (and perhaps rename the option).

teojgo:
* Change the command line option `--flex-alloc-tasks` to `--flex-alloc-nodes` and make it work based on node numbers.

* Change classes and unittests.

* Update documentation.

Fixes #1037

### DIFF
--- a/reframe/core/schedulers/__init__.py
+++ b/reframe/core/schedulers/__init__.py
@@ -193,11 +193,10 @@ class Job(abc.ABC):
                                'this backend') from e
 
             if guessed_num_tasks < min_num_tasks:
-                nodes_required = min_num_tasks // num_tasks_per_node
-                nodes_found = guessed_num_tasks // num_tasks_per_node
-                raise JobError('could not find enough nodes: '
-                               'required %s, found %s' %
-                               (nodes_required, nodes_found))
+                raise JobError(
+                    'could not satisfy the minimum task requirement: '
+                    'required %s, requested %s' %
+                    (min_num_tasks, guessed_num_tasks))
 
             self.num_tasks = guessed_num_tasks
             getlogger().debug('flex_alloc_tasks: setting num_tasks to %s' %
@@ -215,10 +214,16 @@ class Job(abc.ABC):
         pass
 
     def guess_num_tasks(self):
+        num_tasks_per_node = self.num_tasks_per_node or 1
         if isinstance(self.sched_flex_alloc_tasks, int):
             if self.sched_flex_alloc_tasks <= 0:
                 raise JobError('invalid number of flex_alloc_tasks: %s' %
                                self.sched_flex_alloc_tasks)
+            elif self.sched_flex_alloc_tasks % num_tasks_per_node != 0:
+                raise JobError(
+                    'invalid number of flex_alloc_tasks: %s, it must be a '
+                    'multiple of num_tasks_per_node: %s' %
+                    (self.sched_flex_alloc_tasks, num_tasks_per_node))
 
             return self.sched_flex_alloc_tasks
 
@@ -237,7 +242,6 @@ class Job(abc.ABC):
                 'flex_alloc_tasks: selecting idle nodes: '
                 'available nodes now: %s' % len(available_nodes))
 
-        num_tasks_per_node = self.num_tasks_per_node or 1
         num_tasks = len(available_nodes) * num_tasks_per_node
         return num_tasks
 
--- a/reframe/core/schedulers/__init__.py
+++ b/reframe/core/schedulers/__init__.py
@@ -67,7 +67,7 @@ class Job(abc.ABC):
                  stderr=None,
                  pre_run=[],
                  post_run=[],
-                 sched_flex_alloc_tasks=None,
+                 sched_flex_alloc_nodes=None,
                  sched_access=[],
                  sched_account=None,
                  sched_partition=None,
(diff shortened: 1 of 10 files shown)


reframe/core/schedulers/__init__.py
```diff
@@ -215,10 +214,16 @@ def emit_preamble(self):
...
+        num_tasks_per_node = self.num_tasks_per_node or 1
         if isinstance(self.sched_flex_alloc_tasks, int):
             if self.sched_flex_alloc_tasks <= 0:
                 raise JobError('invalid number of flex_alloc_tasks: %s' %
                                self.sched_flex_alloc_tasks)
+            elif self.sched_flex_alloc_tasks % num_tasks_per_node != 0:
+                raise JobError(
+                    'invalid number of flex_alloc_tasks: %s, it must be a '
+                    'multiple of num_tasks_per_node: %s' %
+                    (self.sched_flex_alloc_tasks, num_tasks_per_node))
```
vkarak: This fixes the typicality of producing a correct error message, but I think we should do more than that, if possible, because essentially you need to know the `num_tasks_per_node` of a flexible test.

reframe/frontend/cli.py
```diff
@@ -183,9 +183,9 @@ def main():
         help='Specify the maximum number of times a failed regression test '
              'may be retried (default: 0)')
     run_options.add_argument(
-        '--flex-alloc-tasks', action='store',
-        dest='flex_alloc_tasks', metavar='{all|idle|NUM}', default='idle',
-        help="Strategy for flexible task allocation (default: 'idle').")
+        '--flex-alloc-nodes', action='store',
+        dest='flex_alloc_nodes', metavar='{all|idle|NUM}', default='idle',
+        help="Strategy for flexible node allocation (default: 'idle').")
```
vkarak: You should deprecate the other option, not delete it. If specified, you should print a deprecation warning and either exit or simply set the `flex_alloc_nodes`.

reframe/frontend/cli.py
```diff
@@ -562,20 +566,26 @@ def main():
             exec_policy.skip_sanity_check = options.skip_sanity_check
             exec_policy.skip_performance_check = options.skip_performance_check
             exec_policy.keep_stage_files = options.keep_stage_files
+
+            if options.flex_alloc_tasks:
+                printer.warning('--flex-alloc-tasks is deprecated, you should '
+                                'use --flex-alloc-nodes instead: Exiting...')
+                sys.exit(1)
```
vkarak: This warning is given too late. It should be issued before any action, so that it is shown even with the `-l` option.

vkarak: There is a conceptual problem here. We cannot issue a warning and exit. It should be an error, but deprecations are not errors. I suggest the following. Make the default value of both options `None` so that you know if they are specified (but don't change the documentation or the help message). You can then do sth like this:

```python
if options.flex_alloc_tasks:
    printer.warning('deprecation message')
    options.flex_alloc_nodes = options.flex_alloc_nodes or options.flex_alloc_tasks

options.flex_alloc_nodes = options.flex_alloc_nodes or 'idle'
```

vkarak:
@jenkins-cscs retry none

PR #1067, opened 2019-11-23
[feat] Emit job options to `aprun` command line

### LINKED ISSUE #1017: Improve Cray ALPS support
Current support is really basic. We just emit `aprun -B <command>`. We should emit the detailed command with the specific task allocation options.

vkarak:
Fixes #1017.


vkarak:
@jgphpc Can you test that on Tiger?

vkarak:
As suggested by @TheOneTrueScud, I should change the name of the launcher backend to `aprun` so as to match documentation + it feels more natural.

vkarak:
I will eventually not change the name of the scheduler. The documentation is correct. The `aprun` vs. `alps` was just an error in our private communications.

PR #1073, opened 2019-11-28
[feat] Improve error message when no action is specified

### LINKED ISSUE #1060: Improve error message when no action is specified
Quoting from @rtmclay :

> I think that your error message should change to something like:  No action specified.  Please specify '-l' for listing or '-r' for running

ekouts:
There is also the '-L' option for 'listing matched regression checks with a detailed description'. Message could also be `No action specified. Please specify '-l'/'-L' for listing or '-r' for running.` to be more accurate.

Closes #1060.

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -617,7 +617,8 @@ def main():
                     printer.info(runner.stats.performance_report())
 
         else:
-            printer.info('No action specified. Exiting...')
+            printer.info("No action specified. Please specify '-l' for "
+                         "listing or '-r' for running.")
             printer.info("Try `%s -h' for a list of available actions." %
                          argparser.prog)
             sys.exit(1)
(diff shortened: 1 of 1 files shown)


reframe/frontend/cli.py
```diff
@@ -617,7 +617,8 @@ def main():
                     printer.info(runner.stats.performance_report())
 
         else:
-            printer.info('No action specified. Exiting...')
+            printer.info("No action specified. Please specify '-l' for "
+                         "listing or '-r' for running.")
```
vkarak: I would use this type of quoting: `` `-l' `` and, yes use both `-l` and `-L` in the description, as you proposed.

reframe/frontend/cli.py
```diff
@@ -617,7 +617,8 @@ def main():
                     printer.info(runner.stats.performance_report())
 
         else:
-            printer.info('No action specified. Exiting...')
+            printer.info("No action specified. Please specify `-l'/`-L' for "
```
vkarak: I would also change both those to `printer.error()`, so as to get printed in red.

vkarak: I have also merged the message in a single line.

PR #1074, opened 2019-11-28
[refactor] Redesign job scheduler backends

### LINKED ISSUE #74: Consider using the delegation pattern in the Job scheduler internal interface
And perhaps in other places, too.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/489

vkarak:
Scheduler logic is now separated from the job information. The `Job` class still offers the same API to its clients, but it delegates all the scheduler-specific work to a scheduler backend. The advantage of this design is that we can easily and cleanly support custom configurations for the backend schedulers. All scheduler backends inherit from the `JobScheduler` base abstract class. This class is intentionally left without any implementation, because it needs to serve as a pure interface, which the scheduler backends must implement. The key difference of this interface compared to the current one is that these function take the user job as an argument instead referring to it through `self`.

I fixed also a minor bug in the local launcher, that actually didn't ignore `options`, if these were set after its creation (now this is also the only way to set the launcher options, which is how it is used in practice).

I have also enabled flexible allocation for the local scheduler, since it's trivial. It's always the current host. :-)

@ZQyou these changes will most probably break your PBS local scheduler adaptations, but the changes should be straightforward to adapt.

Fixes #74. 

- ~~I still need to check if the documentation shows up correctly.~~
- ~~I still need to fix the unit tests in CSCS CI.~~

### DIFF
--- a/reframe/core/schedulers/__init__.py
+++ b/reframe/core/schedulers/__init__.py
@@ -103,18 +103,9 @@ class Job:
     def __init__(self,
                  name,
                  workdir='.',
-                 num_tasks=1,
-                 num_tasks_per_node=None,
-                 num_tasks_per_core=None,
-                 num_tasks_per_socket=None,
-                 num_cpus_per_task=None,
-                 use_smt=None,
-                 time_limit=None,
                  script_filename=None,
                  stdout=None,
                  stderr=None,
-                 pre_run=[],
-                 post_run=[],
                  sched_flex_alloc_nodes=None,
                  sched_access=[],
                  sched_account=None,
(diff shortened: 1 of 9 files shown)


reframe/core/schedulers/__init__.py
```diff
@@ -222,75 +256,40 @@ def guess_num_tasks(self):
...
+        available_nodes = self.scheduler.allnodes()
         getlogger().debug('flex_alloc_nodes: total available nodes %s ' %
                           len(available_nodes))
 
         # Try to guess the number of tasks now
-        available_nodes = self.filter_nodes(
-            available_nodes, self.sched_access + self.options)
-
+        available_nodes = self.scheduler.filternodes(self, available_nodes)
+        print(available_nodes)
```
ekouts: This print should probably leave.

unittests/test_launchers.py
```diff
@@ -70,19 +76,24 @@ def expected_command(self):
...
+        emitted_command = self.run_command(self.minimal_job)
         self.assertEqual(self.expected_minimal_command, emitted_command)
 
 
 class TestSrunLauncher(_TestLauncher, unittest.TestCase):
     @property
     def launcher(self):
-        return getlauncher('srun')(options=['--foo'])
+        ret = getlauncher('srun')()
+        print('ret =', ret)
```
ekouts: This print should be removed as well.

vkarak:
@ekouts @teojgo @victorusu You can now review. All unit tests should be passing now. I have still the wip tag, because I need to check the documentation.

reframe/core/schedulers/__init__.py
```diff
@@ -14,12 +13,49 @@
...
+    @abc.abstractmethod
+    def cancel(self, job):
+        pass
+
+    @abc.abstractmethod
+    def finished(self, job):
+        pass
+
+
+class Job:
```
victorusu: I really liked the design. The separation of `Job`  into `JobScheduler` and `Job` allows a simpler implementation of the different schedulers and their functionalities. Now, `Job` is a struct that holds info about the job state, which is populated by the scheduler that schedules it.

PR #1077, opened 2019-11-29
[feat] Add nodelist in failure report

### LINKED ISSUE #1069: Print nodelist in failure report
A custom change from @ZQyou

rsarm:
Closes #1069

### DIFF
--- a/reframe/frontend/statistics.py
+++ b/reframe/frontend/statistics.py
@@ -84,7 +84,7 @@ class TestStats:
             report.append('  * System partition: %s' % partname)
             report.append('  * Environment: %s' % environ_name)
             report.append('  * Stage directory: %s' % check.stagedir)
-
+            report.append('  * Nodes: %s' % check.job._nodelist)
             job_type = 'local' if check.is_local() else 'batch job'
             jobid = check.job.jobid if check.job else -1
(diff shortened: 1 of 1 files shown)


reframe/frontend/statistics.py
```diff
@@ -84,6 +84,10 @@ def failure_report(self):
             report.append('  * System partition: %s' % partname)
             report.append('  * Environment: %s' % environ_name)
             report.append('  * Stage directory: %s' % check.stagedir)
+            if check.job:
+                report.append('  * Nodes: %s' % check.job._nodelist)
```
vkarak: You should use the public attribute, `nodelist`, not `_nodelist`.

vkarak: Also you should format the list with sth like the following: `','.join(check.job.nodelist)`.

reframe/frontend/statistics.py
```diff
@@ -84,6 +84,10 @@ def failure_report(self):
...
+            else:
+                report.append('  * Nodes: <%s>' % check.job)
```
vkarak: It be nicer if you can put everything in a single line:

```python
report.append('  * Node list: %s' % (','.join(check.job.nodelist) if check.job and check.job.nodelist else '<None>'))
```

Notice the `nodelist`  may also be `None`.

PR #1086, opened 2019-12-05
[test] Remove slurm switch option from dom

victorusu:
Dom doesn't have multiple switches. This option makes gromacs jobs on dom wait 5 min to start.

### DIFF
--- a/cscs-checks/apps/gromacs/gromacs_check.py
+++ b/cscs-checks/apps/gromacs/gromacs_check.py
@@ -36,11 +36,12 @@ class GromacsBaseCheck(rfm.RunOnlyRegressionTest):
         self.maintainers = ['VH']
         self.strict_check = False
         self.use_multithreading = False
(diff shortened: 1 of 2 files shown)


cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -36,11 +36,12 @@ def __init__(self, output_file):
...
         self.strict_check = False
         self.use_multithreading = False
-        self.extra_resources = {
-            'switches': {
-                'num_switches': 1
+        if self.current_system.name == 'daint':
+            self.extra_resources = {
+                'switches': {
+                    'num_switches': 1
+                }
```
vkarak: Why don't you just undefine the switches in Dom's configuration? You don't need to change the test then.

vkarak:
@jenkins-cscs retry daint

PR #1090, opened 2019-12-06
[feat] Make job completion time available as a logging variable

### LINKED ISSUE #1044: perflogging timestamps with async policy
All of the timestamps in perflogs are the same when tests are run with the async policy. Is it possible for the performance logging to be triggered closer to the completion time of each test? This complicates correlation with other sources with large or long running test suites.

```
1 2019-08-27T11:35:49.198249-07:00 gert01 ReFramePerf - - [nerscjson@42] {"reframeversion": "2.19-nersc", "name": "AllocSpeedTest_no", "jobid": 1546259, "env": "PrgEnv-cray", "partition": "haswell", "system": "gerty", "tags": "benchmarks", "var": "2MB", "val": 0.000522142, "unit": "seconds", "ref": 0.00389933, "lbound": -1, "ubound": 4}
1 2019-08-27T11:35:49.201164-07:00 gert01 ReFramePerf - - [nerscjson@42] {"reframeversion": "2.19-nersc", "name": "AllocSpeedTest_no", "jobid": 1546259, "env": "PrgEnv-cray", "partition": "haswell", "system": "gerty", "tags": "benchmarks", "var": "4MB", "val": 0.000434982, "unit": "seconds", "ref": 0.000519077, "lbound": -1, "ubound": 4}
1 2019-08-27T11:35:49.202126-07:00 gert01 ReFramePerf - - [nerscjson@42] {"reframeversion": "2.19-nersc", "name": "AllocSpeedTest_no", "jobid": 1546259, "env": "PrgEnv-cray", "partition": "haswell", "system": "gerty", "tags": "benchmarks", "var": "8MB", "val": 0.00112756, "unit": "seconds", "ref": 0.000557189, "lbound": -1, "ubound": 4}
1 2019-08-27T11:35:49.203101-07:00 gert01 ReFramePerf - - [nerscjson@42] {"reframeversion": "2.19-nersc", "name": "AllocSpeedTest_no", "jobid": 1546259, "env": "PrgEnv-cray", "partition": "haswell", "system": "gerty", "tags": "benchmarks", "var": "16MB", "val": 0.00204565, "unit": "seconds", "ref": 0.000976913, "lbound": -1, "ubound": 4}
1 2019-08-27T11:35:49.204043-07:00 gert01 ReFramePerf - - [nerscjson@42] {"reframeversion": "2.19-nersc", "name": "AllocSpeedTest_no", "jobid": 1546259, "env": "PrgEnv-cray", "partition": "haswell", "system": "gerty", "tags": "benchmarks", "var": "32MB", "val": 0.00385665, "unit": "seconds", "ref": 0.00172543, "lbound": -1, "ubound": 4}
```

ekouts:
The attribute is more accurate for the slurm scheduler that uses sacct (but not the squeue job scheduler). The accuracy for the rest is dependent on the polling rate of the running tasks.

Closes  #1044.


ekouts:
@jenkins-cscs retry kesch

reframe/core/schedulers/slurm.py
```diff
@@ -318,6 +326,10 @@ def _update_state(self, job):
         if slurm_state_completed(job.state):
             # Since Slurm exitcodes are positive take the maximum one
             job.exitcode = max(int(s.group('exitcode')) for s in state_match)
+            self._completion_time = max(
+                datetime.strptime(s.group('end'), '%Y-%m-%dT%H:%M:%S')
```
vkarak: Are you sure that a parsable `end` time is always there? Even if the job is still running or pending?

vkarak: To make things more Pythonic and compact I'd also rewrite this as follows:

```python
parse_time = functools.partial(datetime.strptime, format='...')
self._completion_time = max(parse_time(s.group('end')) for s in state_match)
```

reframe/core/logging.py
```diff
@@ -427,6 +428,15 @@ def _update_check_extras(self):
         if self.check.job:
             self.extra['check_jobid'] = self.check.job.jobid
 
+        if self.check.job:
```
vkarak: Why do you need this additional `if` here? The above one should suffice.

reframe/core/logging.py
```diff
@@ -427,6 +428,15 @@ def _update_check_extras(self):
...
+            if self.check.job.completion_time:
+                # Format is the same for every handler
+                fmt = self.logger.handlers[0].formatter.datefmt
+                self.extra['check_completion_time'] = (
+                    self.check.job.completion_time.strftime(fmt))
+            else:
+                self.extra['check_completion_time'] = "NA"
```
vkarak: You don't need the `else` here. We are fine with `None` being printed.

reframe/core/logging.py
```diff
@@ -427,6 +428,15 @@ def _update_check_extras(self):
         if self.check.job:
             self.extra['check_jobid'] = self.check.job.jobid
 
+        if self.check.job:
+            if self.check.job.completion_time:
+                # Format is the same for every handler
```
vkarak: This comment is a bit confusing. Rewrite as follows:

```python
# Use the logging handlers' date format to format completion_time
# NOTE: All handlers use the same date format
```

reframe/core/logging.py
```diff
@@ -427,6 +428,15 @@ def _update_check_extras(self):
...
+                fmt = self.logger.handlers[0].formatter.datefmt
+                self.extra['check_completion_time'] = (
+                    self.check.job.completion_time.strftime(fmt))
```
vkarak: Try assigning the `self.check.job.completion_time` to a different variable to avoid the line split here.

reframe/core/schedulers/__init__.py
```diff
@@ -321,7 +327,8 @@ def wait(self):
         if self.jobid is None:
             raise JobNotStartedError('cannot wait an unstarted job')
 
-        return self.scheduler.wait(self)
+        self.scheduler.wait(self)
+        self._completion_time = datetime.now()
```
vkarak: You should set the completion time also when `finished()` is called and set it here only if it is not already set.

reframe/core/schedulers/slurm.py
ekouts: > Are you sure that a parsable `end` time is always there?

I have tried it on daint and while it is running or pending the `end` time is `Unknown`. But the matched string is only parsed after the job is completed.

vkarak: So `strptime()` does not fail if you pass it a string that does not represent time then, right? Then it's fine.

ekouts: It does fail. But it is inside this if `if slurm_state_completed(job.state):`. It shouldn't be parsed while it is pending or running.

ekouts: Should I put this on a try? Because I don't know if I can test every state that is considered as completed.

vkarak: Ok, now I see it. That's fine.

vkarak: > Should I put this inside a try? Because I don't know if I can test every state that is considered as completed.

No, I haven't seen you were actually calling this inside the `if slurm_state_completed(job.state)`.

ekouts: > To make things more Pythonic and compact I'd also rewrite this as follows:
> 
> ```python
> parse_time = functools.partial(datetime.strptime, format='...')
> self._completion_time = max(parse_time(s.group('end')) for s in state_match)
> ```

Strptime doesn't have keyword arguments and I think I cannot set the second argument of the function.

ekouts:
@jenkins-cscs retry kesch

reframe/core/schedulers/slurm.py
vkarak: I think it's because it uses the C library behind the scenes. That's fine.

reframe/core/schedulers/__init__.py
```diff
@@ -340,7 +346,11 @@ def finished(self):
         if self.jobid is None:
             raise JobNotStartedError('cannot poll an unstarted job')
 
-        return self.scheduler.finished(self)
+        res = self.scheduler.finished(self)
```
vkarak: `res` -> `done`.

reframe/core/schedulers/slurm.py
```diff
@@ -297,10 +296,9 @@ def _set_nodelist(self, job, nodespec):
     def _update_state(self, job):
         '''Check the status of the job.'''
 
-        completed = _run_strict(
-            'sacct -S %s -P -j %s -o jobid,state,exitcode,nodelist,end' %
-            (datetime.now().strftime('%F'), job.jobid)
-        )
+        cmd = 'sacct -S %s -P -j %s -o jobid,state,exitcode,nodelist,end' % (
+            datetime.now().strftime('%F'), job.jobid)
```
vkarak: Put the closing parenthesis in a new line.

reframe/utility/os_ext.py
```diff
@@ -78,6 +80,7 @@ def run_command_async(cmd,
                             stderr=stderr,
                             universal_newlines=True,
                             shell=shell,
+                            env=env,
```
vkarak: Unfortunately, this is not equivalent to setting and unsetting the variable in `os.environ`. Quoting from `Popen`'s documentation:

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment.

vkarak: This explains the failure on Kesch and would lead to similar problems in other systems.

reframe/core/environments.py
```diff
@@ -144,6 +144,23 @@ def emit_load_commands(*environs):
     return commands
 
 
+class modify_env:
```
vkarak: This is not how I meant to implement this context manager. You are not actually using the `Environment` class which has all the machinery we want. Here is how:

```python
class temp_environment:
    def __init__(modules=[], variables=[]):
        self._modules = modules
        self._variables = variables

    def __enter__(self):
        self._environ_save, _ =  load(Environment(self._modules, self._variables))
        return self._environ_save

    def __exit__(self, exc_type, exc_value, traceback):
        self._environ_save.restore()
```

And you should add a unit test for this.

reframe/utility/os_ext.py
vkarak: Remove the `env` argument completely.

reframe/core/schedulers/slurm.py
```diff
@@ -88,6 +89,11 @@ def __init__(self):
         self._is_cancelling = False
         self._is_job_array = None
         self._update_state_count = 0
+        self._completion_time = None
+
+    @property
+    def completion_time(self):
```
vkarak: Allow this to get the `job` as argument (so no more property), and leave `_update_state` untouched.

reframe/core/schedulers/pbs.py
```diff
@@ -31,11 +31,15 @@
...
         self._prefix = '#PBS'
-        self._time_finished = None
+        self._completion_time = None
 
         # Optional part of the job id refering to the PBS server
         self._pbs_server = None
 
+    @property
+    def completion_time(self):
+        return self._completion_time
```
vkarak: The PBS scheduler should return `None`. It doesn't do anything more accurate that the `Job` does. So again, you can leave untouched the rest of its implementation.

reframe/core/schedulers/slurm.py
```diff
@@ -290,10 +296,12 @@ def _set_nodelist(self, job, nodespec):
     def _update_state(self, job):
         '''Check the status of the job.'''
 
-        completed = _run_strict(
-            'sacct -S %s -P -j %s -o jobid,state,exitcode,nodelist' %
-            (datetime.now().strftime('%F'), job.jobid)
-        )
+        with env.modify_env(variables={'SLURM_TIME_FORMAT': 'standard'}):
```
vkarak: Passing a dictionary would work, even with my context manager, but if you want to guarantee the order of the variables (here not the case, though), you should pass it as a list of tuples:

```python
[('SLURM_TIME_FORMAT', 'standard')]
```

reframe/core/schedulers/slurm.py
```diff
@@ -91,8 +91,35 @@ def __init__(self):
...
+                'sacct -S %s -P -j %s -o jobid,end' %
+                (datetime.now().strftime('%F'), job.jobid),
+                log=False
+            )
+
+        # This matches the format for both normal jobs as well as job arrays.
+        # For job arrays the job_id has one of the following formats:
+        #   * <job_id>_<array_task_id>
+        #   * <job_id>_[<array_task_id_start>-<array_task_id_end>]
+        # See (`Job Array Support<https://slurm.schedmd.com/job_array.html`__)
```
vkarak: Don't copy paste the comment. I would define the regex at the class level or at least this basic pattern that matches the job id and re use that both here and in `_update_state`. And I'd move the comment there as well. In that hoisted regex pattern, don't define any capturing group, such that you can write it as follows:

```python
class SlurmJobScheduler:
    # Comment from Theo
    _state_patt = r'\d+(?:_\d+|_\[\d+-\d+\])?'

    def completion_time(self, job):
        ...
        state_match = re.findall(r'^(?<jobid>%s)\|(?P<end>\S+)' % self._state_patt, ...)
```

reframe/core/schedulers/slurm.py
```diff
@@ -91,8 +91,35 @@ def __init__(self):
...
+        state_match = list(re.finditer(
```
vkarak: `re.findall()` should be equivalent to `list(re.finditer())`. Please update it in `_update_state()` as well.

reframe/core/schedulers/slurm.py
```diff
@@ -91,8 +91,35 @@ def __init__(self):
...
 
-    @property
-    def completion_time(self):
+    def completion_time(self, job):
+        if (self._completion_time or
+            not slurm_state_completed(job.state)):
+            return self._completion_time
+
+        slurm_variable = [('SLURM_TIME_FORMAT', 'standard')]
+        with env.temp_environment(variables=slurm_variable):
```
vkarak: Don't use a separate variable here.

reframe/core/schedulers/slurm.py
```diff
@@ -91,8 +91,35 @@ def __init__(self):
...
+        # This matches the format for both normal jobs as well as job arrays.
+        # For job arrays the job_id has one of the following formats:
+        #   * <job_id>_<array_task_id>
+        #   * <job_id>_[<array_task_id_start>-<array_task_id_end>]
+        # See (`Job Array Support<https://slurm.schedmd.com/job_array.html`__)
+        state_match = list(re.finditer(
+            r'^(?P<jobid>\d+(?:_\d+|_\[\d+-\d+\])?)\|(?P<end>\S+)',
+            completed.stdout, re.MULTILINE))
+        if not state_match:
+            return self._completion_time
```
vkarak: I'd put `return None` here, in order to be more explicit.

reframe/core/schedulers/slurm.py
```diff
@@ -91,8 +91,35 @@ def __init__(self):
...
-    @property
-    def completion_time(self):
+    def completion_time(self, job):
+        if (self._completion_time or
+            not slurm_state_completed(job.state)):
+            return self._completion_time
+
+        slurm_variable = [('SLURM_TIME_FORMAT', 'standard')]
+        with env.temp_environment(variables=slurm_variable):
+            completed = _run_strict(
```
vkarak: Don't run this strictly. If the command fails, simply return `None`, because otherwise logging will crash.

reframe/utility/os_ext.py
```diff
@@ -66,12 +66,14 @@ def run_command_async(cmd,
...
                       stderr=subprocess.PIPE,
                       shell=False,
-                      env=None,
+                      log=True,
                       **popen_args):
     # Import logger here to avoid unnecessary circular dependencies
-    from reframe.core.logging import getlogger
+    if log:
+        from reframe.core.logging import getlogger
+
```
vkarak: Remove this blank line.

reframe/core/schedulers/slurm.py
ekouts: It is not equivalent. `re.findall()` returns a list of tuples and I won't be able to use named variables anymore, like `s.group('state')`. I would have to access the tuple elements with their index.

vkarak: Ok, that's fine. I think that was the reason that we have used this construct then.

unittests/test_environments.py
```diff
@@ -88,6 +88,15 @@ def test_load_restore(self):
 
         assert not self.environ.is_loaded
 
+    @fixtures.switch_to_user_runtime
+    def test_temp_environment_context_manager(self):
+        with env.temp_environment(variables=[('_var0', 'val2'),
+                                             ('_var3', 'val3')],
+                                  modules=['testmod_foo']):
```
vkarak: For brevity, don't use kwargs here. You will just need to change the order of the arguments.

unittests/test_environments.py
```diff
@@ -88,6 +88,15 @@ def test_load_restore(self):
...
+            assert os.environ['_var0'] == 'val2'
+            assert os.environ['_var3'] == 'val3'
```
vkarak: It'd be easier to check that the environment is set inside the context manager, if `__enter__` returned the actual environment that was created, which btw makes more sense to be returned instead of the snapshot. Then you could do the following:

```python
with env.temp_environment(...) as environ:
    assert environ.is_loaded

assert not environ.is_loaded
```

vkarak:
Code-wise this PR is ready. Only remaining thing is to document the new logging variable.

vkarak:
I changed the name of the logging variable to `check_job_completion_time`, since it is more precise.

vkarak:
@brandongc I will merge this one. You should now be able to use `%(check_job_completion_time)s` inside the log format to get the actual completion time of the job. The timestamp will be formatted based on the value of `datefmt`.

PR #1099, opened 2019-12-10
[feat] Allow tests to timeout if their associated job is pending for too long

### LINKED ISSUE #881: Cancel job after a certain queueing time
This a feature request coming from @mand35:

> We want to submit our jobs daily. But in some cases nodes can be occupied for longer, than we would like to skin these data points and neglect these runs for nodes which are busy the whole day. We do not want to queue up more and more jobs for one node.

rsarm:
Closes #881


reframe/core/pipeline.py
```diff
@@ -670,6 +678,7 @@ def _rfm_init(self, name=None, prefix=None):
         self.num_tasks_per_socket = None
         self.use_multithreading = None
         self.exclusive_access = False
+        self.max_queuing_time = None
```
vkarak: What if we called the variable `max_queue_time`. "Queuing" and "queueing" are both correct spellings, so I am sure it will create confusion.

reframe/core/schedulers/slurm.py
```diff
@@ -386,7 +386,14 @@ def wait(self, job):
...
         intervals = itertools.cycle(settings().job_poll_intervals)
         self._update_state(job)
+        start_pending = time.time()
         while not slurm_state_completed(job.state):
+            if max_queuing_time:
+                if slurm_state_pending(job.state):
+                    if time.time() - start_pending > max_queuing_time:
+                        self.cancel(job)
+                        raise JobError('maximum queuing time exceeded')
+
```
vkarak: What I proposed yesterday was to change how wait is implemented at the `Job` level so that it could allow us to implement this in a scheduler-agnostic way. Now, you have to implement the same logic for `wait()` in all backends.

I can do a PR changing the logic and you can continue with this feature.

reframe/core/pipeline.py
```diff
@@ -421,6 +421,14 @@ class RegressionTest(metaclass=RegressionTestMeta):
     use_multithreading = fields.TypedField('use_multithreading',
                                            bool, type(None))
 
+    #: Specify the maximum time the job can be in the queue before running
+    #:
+    #: Ignored if :class:`None`.
+    #:
+    #: :type: integral or :class:`None`
+    #: :default: :class:`None
+    max_queuing_time = fields.TypedField('max_queuing_time', int, type(None))
```
vkarak: Also why not using a `TimeField` here as in `time_limit`? You do not document also what this integer represents.

reframe/core/schedulers/slurm.py
vkarak: I looked into the code in more detail. I don't think that this can be done. The `local` scheduler uses a completely different mentality when waiting. So passing a timeout argument to the `wait()` method is fine.

reframe/core/pipeline.py
```diff
@@ -421,6 +421,14 @@ class RegressionTest(metaclass=RegressionTestMeta):
     use_multithreading = fields.TypedField('use_multithreading',
                                            bool, type(None))
 
+    #: Specify the maximum time the job can be pending before starting running
+    #:
+    #: The time is specified with the same format of `time_limit`.
+    #:
+    #: :type: :class:`tuple[int]`
+    #: :default: :class:`None
+    max_pending_time = fields.TimerField('max_pending_time', type(None))
```
vkarak: I don't like very much the tuple syntax of `TimerField` , so I suggest fixing #1106 first before proceeding with this PR. Because otherwise we introduce a new variable with the same bad syntax.

reframe/core/schedulers/slurm.py
```diff
@@ -418,7 +419,15 @@ def wait(self, job):
 
         intervals = itertools.cycle(settings().job_poll_intervals)
         self._update_state(job)
+
         while not slurm_state_completed(job.state):
+            if max_pending_time:
+                if slurm_state_pending(job.state):
+                    if datetime.now() - self._submit_time > max_pending_time:
```
ekouts:
```suggestion
                    if datetime.now() - self._submit_time > job.max_pending_time:
```

reframe/core/schedulers/slurm.py
```diff
@@ -431,7 +440,7 @@ def cancel(self, job):
                     timeout=settings().job_submit_timeout)
         self._is_cancelling = True
 
-    def finished(self, job):
+    def finished(self, job, max_pending_time):
```
ekouts:
```suggestion
    def finished(self, job):
```

reframe/core/schedulers/slurm.py
```diff
@@ -443,6 +452,13 @@ def finished(self, job):
             getlogger().debug('ignoring error during polling: %s' % e)
             return False
         else:
+            if max_pending_time:
```
ekouts:
```suggestion
            if job.max_pending_time:
```

reframe/core/schedulers/slurm.py
```diff
@@ -443,6 +452,13 @@ def finished(self, job):
...
+                if slurm_state_pending(job.state):
+                    if datetime.now() - self._submit_time > max_pending_time:
```
ekouts:
```suggestion
                    if datetime.now() - self._submit_time > job.max_pending_time:
```

ekouts:
You don't have unittests that test this for slurm/pbs but I am not sure how you could test it. Maybe you could give a very low `max_pending_time` so that you make sure it cancels the job. I don't know whether @vkarak has better suggestions.

unittests/test_schedulers.py
```diff
@@ -32,6 +32,7 @@ def setUp(self):
             self.scheduler, self.launcher,
             name='testjob',
             workdir=self.workdir,
+            max_pending_time=None,
```
vkarak: Remove this.

unittests/test_schedulers.py
```diff
@@ -437,6 +439,20 @@ def test_submit_job_array(self):
...
+    # Monkey patch `self._update_state` to simulate that the job is
+    # pending on the queue for enough time so it can be canceled due
+    # to exceeding the maximum pending time
+    @fixtures.switch_to_user_runtime
+    def test_submit_max_pending_time(self):
+        self.setup_user()
+        self.parallel_cmd = 'sleep 30'
+        self.prepare()
+        self.testjob.scheduler._update_state = self._update_state
+        self.testjob._max_pending_time = timedelta(seconds=5)
```
vkarak: This will slow down the unit tests immensely. Wait only for one second.

unittests/test_schedulers.py
```diff
@@ -562,6 +578,23 @@ def test_submit_timelimit(self):
...
+        self.testjob.submit()
+        with pytest.raises(JobError):
+            self.testjob.wait()
```
vkarak: This test is copy-paste. It should be factored out in the base class.

unittests/test_schedulers.py
```diff
@@ -197,6 +197,20 @@ def test_guess_num_tasks(self):
...
+    # to exceeding the maximum pending time
+    @fixtures.switch_to_user_runtime
+    def test_submit_max_pending_time(self):
+        self.setup_user()
+        self.parallel_cmd = 'sleep 30'
+        self.prepare()
+        self.testjob.scheduler._update_state = self._update_state
+        self.testjob._max_pending_time = timedelta(seconds=5)
+        self.testjob.submit()
+        with pytest.raises(JobError):
```
vkarak: You should check here that the message you are getting is the correct one.

unittests/test_schedulers.py
```diff
@@ -197,6 +197,20 @@ def test_guess_num_tasks(self):
...
+    # Monkey patch `self._update_state` to simulate that the job is
+    # pending on the queue for enough time so it can be canceled due
+    # to exceeding the maximum pending time
+    @fixtures.switch_to_user_runtime
+    def test_submit_max_pending_time(self):
+        self.setup_user()
+        self.parallel_cmd = 'sleep 30'
+        self.prepare()
+        self.testjob.scheduler._update_state = self._update_state
+        self.testjob._max_pending_time = timedelta(seconds=5)
```
vkarak: Again 5 secs is too long. It will slow down the unit tests. Please reduce that to 1 sec or less if possible.

vkarak:
@rsarm I am going to fix this PR now.

PR #1100, opened 2019-12-11
[ci] Use `--flex-alloc-nodes` option to run user tests in CI

teojgo:


vkarak:
@jenkins-cscs retry dom

PR #1104, opened 2019-12-16
[test] Add check for Slurm error message when no constraint is given

ekouts:
The error in the sanity pattern may be too specific for daint and dom. It could also change to something more generic, like `error: You have to specify, at least, what sort of node you need:`.

Fixes UES-427


ekouts:
@jenkins-cscs retry

vkarak:
I am doing some fine tuning of the test and will push shortly.

vkarak:
@jenkins-cscs retry daint

PR #1107, opened 2019-12-18
[feat] Print repository's revision hash when ReFrame's version is requested

### LINKED ISSUE #1052: Add support for logging ReFrame git repo hash
FYI @bfriesen

teojgo:
Fixes #1052


vkarak:
@bcfriesen As soon as this is merged the `%(version)s` specifier in the log format will print also the revision hash, if ReFrame's install prefix contains the repo metadata. Otherwise, it will print the version as usual. Same rule applies also for `--version`.

PR #1124, opened 2020-01-17
[bugfix] Fix infinite loop bug when cleanup phase fails

### LINKED ISSUE #1121: Regression test not properly removed from tasks if it fails in the cleanup stage
If a regression test fails at the cleanup phase, then the cleanup phase is retried continuously. 
This is a log snippet showing the cleanup phase failure:
```
debug: HelloWorldTestMPIOpenMP_c_static on dom:gpu using PrgEnv-cray_classic: caught builtins.FileNotFoundError: [Errno 2] No such file or directory: '/scratch/snx3000tds/jenscscs/regression/production/stage/2020-01-17_08-57-54/dom/gpu/PrgEnv-cray_classic/HelloWorldTestMPIOpenMP_c_static'
[2020-01-17T09:54:01] debug: reframe: removing task from running list: HelloWorldTestMPIOpenMP_c_static on dom:gpu using PrgEnv-cray_classic
[2020-01-17T09:54:01] debug: reframe: not in running tasks
```

ekouts:
The task cleanup should be in a try block for both serial and asynchronous execution policies in order to have the correct _retired_tasks list.

The output of reframe for tests that fail on cleanup can also be misleading because both will be printed
```
[       OK ] CleanupFailTest on generic:login using builtin-gcc
[     FAIL ] CleanupFailTest on generic:login using builtin-gcc
```
The line `[       OK ]` is printed just after performance.

To avoid the confusion now the OK is printed after the performance stage in both serial and asynchronous policy and a new message `[    ERROR ]` is printed for failure in the cleanup stage. The error message is different because the failure is probably due to a reframe/system error. New output looks like that:

serial:
```
[----------] started processing CleanupFailTest (CleanupFailTest)
[ RUN      ] CleanupFailTest on generic:login using builtin-gcc
[       OK ] CleanupFailTest on generic:login using builtin-gcc
[    ERROR ] CleanupFailTest on generic:login using builtin-gcc
[----------] finished processing CleanupFailTest (CleanupFailTest)

[----------] started processing SleepTest (SleepTest)
[ RUN      ] SleepTest on generic:login using builtin-gcc
[       OK ] SleepTest on generic:login using builtin-gcc
[----------] finished processing SleepTest (SleepTest)
```

async:
```
----------] started processing CleanupFailTest (CleanupFailTest)
[ RUN      ] CleanupFailTest on generic:login using builtin-gcc
[----------] finished processing CleanupFailTest (CleanupFailTest)

[----------] started processing SleepTest (SleepTest)
[ RUN      ] SleepTest on generic:login using builtin-gcc
[----------] finished processing SleepTest (SleepTest)

[       OK ] CleanupFailTest on generic:login using builtin-gcc
[    ERROR ] CleanupFailTest on generic:login using builtin-gcc
[       OK ] SleepTest on generic:login using builtin-gcc
```

Fixes #1121


unittests/test_policies.py
```diff
@@ -248,6 +248,29 @@ def test_dependencies_with_retries(self):
...
+
+        self.assertRunall()
+        stats = self.runner.stats
+        assert stats.num_cases(0) == 2
+        assert len(stats.failures()) == 1
+
+        # Check that cleanup is executed properly for successful tests as well
+        for t in stats.tasks():
+            check = t.testcase.check
+            if t.failed:
```
teojgo: Add a line `self.runner.policy.keep_stage_files = False` because it is set to `True` in `setUp`.

reframe/frontend/executors/policies.py
```diff
@@ -293,7 +296,10 @@ def runcase(self, case):
     def _cleanup_all(self):
         for task in self._retired_tasks:
             if task.ref_count == 0:
-                task.cleanup(not self.keep_stage_files)
+                try:
```
teojgo: Same as above.

unittests/resources/checks_unlisted/cleanup_fail.py
```diff
@@ -0,0 +1,28 @@
+import os
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+@rfm.simple_test
+class CleanupFailTest(rfm.RunOnlyRegressionTest):
```
vkarak: Consider moving this inside the `unittests/resources/checks/frontend_checks.py`. You will not then need the `SleepTest`, which is duplicate, you won't need any other unit test, since the standards execution policy unit tests should cover you. The only thing you need is to adjust the expected failures in the corresponding unit tests.

unittests/test_policies.py
vkarak: If you had taken my approach above, you wouldn't need this.

PR #1125, opened 2020-01-17
[feat] Do not allow chaining of the `-c` option; pass multiple paths as a colon separated list 

### LINKED ISSUE #1120: Make `-c` option accept colon separated paths instead of appending to the check search path
This will address scenarios like the one reported in #1119.

teojgo:
* Deprecate using the checkpath command line option multiple times
  and issue a warning.

* Create unittest for colon separated paths.

* Add the change to the documentation.

Fixes #1120.

### DIFF
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -306,7 +306,7 @@ ReFrame the does not search recursively into directories specified with the ``-c
 
 The ``-c`` option completely overrides the default path.
 Currently, there is no option to prepend or append to the default regression path.
-However, you can build your own check path by specifying multiple times the ``-c`` option.
+However, you can build your own check path by specifying multiple colon separate paths to the ``-c`` option.
 The ``-c``\ option accepts also regular files. This is very useful when you are implementing new regression tests, since it allows you to run only your test:
 
 .. code-block:: bash
@@ -320,6 +320,17 @@ The ``-c``\ option accepts also regular files. This is very useful when you are 
    In this case, any conflicting test will not be loaded and a warning will be issued.
 
    .. versionadded:: 2.12
+
+.. note::
+   Using the command line ``-c`` or ``--checkpath`` multiple times is now deprecated and only the last option is going to be taken into account.
+   Multiple paths should be passed as a colon separated string to the above option:
+
+   .. code-block:: bash
+
+      ./bin/reframe -c /path/to/my/first/test.py:/path/to/my/second/test.py -r
+
+
+   .. versionchanged:: 2.22
 
 
 Filtering of Regression Tests
--- a/docs/running.rst
+++ b/docs/running.rst
@@ -323,7 +323,7 @@ The ``-c``\ option accepts also regular files. This is very useful when you are 
 
 .. warning::
    Using the command line ``-c`` or ``--checkpath`` multiple times is not supported anymore and only the last option will be considered.
-   Multiple paths should be passed as a colon separated string to the above option:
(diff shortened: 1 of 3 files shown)


reframe/frontend/cli.py
```diff
@@ -410,7 +410,11 @@ def main():
     # Setup the check loader
     if options.checkpath:
         load_path = []
-        for d in options.checkpath:
+        if len(options.checkpath) > 1:
+            printer.warning("using command line option `--checkpath' "
+                            "multiple times is deprecated, please use "
```
ekouts: Again, I would suggest to warn the user that only the last check path will be taken into account, unless we keep supporting multiple `--checkpath`.

docs/running.rst
```diff
@@ -321,6 +321,17 @@ The ``-c``\ option accepts also regular files. This is very useful when you are
...
+.. note::
+   Using the command line ``-c`` or ``--checkpath`` multiple times is now deprecated and only the last option is going to be taken into account.
+   Multiple paths should be passed as a colon separated string to the above option:
+
+   .. code-block:: bash
+
+      ./bin/reframe -c /path/to/my/first/test.py:/path/to/my/second/test.py -r
+
+
+   .. versionchanged:: 2.22
```
vkarak:
```suggestion
   .. versionchanged:: 3.0
```

docs/running.rst
```diff
@@ -321,6 +321,17 @@ The ``-c``\ option accepts also regular files. This is very useful when you are
 
    .. versionadded:: 2.12
 
+.. note::
+   Using the command line ``-c`` or ``--checkpath`` multiple times is now deprecated and only the last option is going to be taken into account.
```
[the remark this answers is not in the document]
vkarak: Also change the "is going to be taken into account." to "will be considered."

docs/running.rst
```diff
@@ -321,6 +321,17 @@ The ``-c``\ option accepts also regular files. This is very useful when you are
...
+   Multiple paths should be passed as a colon separated string to the above option:
```
vkarak:
```suggestion
   Multiple paths should be passed instead as a colon separated list:
```

docs/running.rst
```diff
@@ -321,6 +321,17 @@ The ``-c``\ option accepts also regular files. This is very useful when you are
...
+
+   .. code-block:: bash
+
+      ./bin/reframe -c /path/to/my/first/test.py:/path/to/my/second/test.py -r
```
vkarak: Let the second path be a directory, so as to show that you can mix both.

reframe/frontend/cli.py
```diff
@@ -94,7 +94,7 @@ def main():
     # Check discovery options
     locate_options.add_argument(
         '-c', '--checkpath', action='append', metavar='DIR|FILE',
```
vkarak: Make the action `store`.

reframe/frontend/cli.py
vkarak: Since we are moving to 3.0, just remove the warning.

reframe/frontend/cli.py
```diff
@@ -94,7 +94,7 @@ def main():
...
-        help='Search for checks in DIR or FILE')
+        help="Search for checks in DIR or FILE, using `:' for multiple paths")
```
```suggestion
        help="Search for checks in DIR or FILE; multiple paths can be separated with `:'")
```

vkarak:
@ekouts Can you also approve this, so that we can merge it?

Issue #66, opened 2018-02-15
System and partition initialisation are random


vkarak:
This is due to the fact that both systems and partitions are defined as dictionaries. This has several repercussions:

* No specific order that the regression goes through the partitions during the phase of test execution.
* Hostname matching rules must be unique. No last resort, catch-all pattern may be placed, since this may be tried earlier and lead to false system detection.

With Python 3.7 normal dictionaries will become ordered, but it's better not to rely on that.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/234

PR #1128, opened 2020-01-22
[feat] Introduce new configuration syntax and validation schema

### LINKED ISSUE #1113: Define a new and more flexible configuration syntax
The idea is to have everything under `site_configuration`, which itself should be formatted as a JSON object in Python. This will allow us to accept easily `.json` config files and later YAML. Additionally, all configuration sections must be allowed to be configured per system or partition (like the `environments`, but with a different syntax). Here is a tentative proposal for the syntax:

```python
site_configuration = {
    'systems': [
        {
            'name': 'generic',
            'descr': 'Generic example system',
            'hostnames': ['localhost'],
            'partitions': [
                {
                    'name': 'login',
                    'scheduler': 'local',
                    'modules': [],
                    'access':  [],
                    'environs': ['builtin-gcc'],
                    'descr': 'Login nodes'
                }
            ]
        }
    ],

    'environments': [
        {
            'name': 'builtin',
            'target_systems': ['*'],
            'cc':  'cc',
            'cxx': '',
            'ftn': '',
        },
        {
            'name': 'builtin-gcc',
            'target_systems': ['*'],
            'cc':  'gcc',
            'cxx': 'g++',
            'ftn': 'gfortran',
        }
    ],

    'schedulers': [
        {
            'name': 'slurm',
            'target_systems': ['*'],
            'wait_poll_intervals': [1, 2, 3],
            'submit_timeout': 60
        }
    ],

    'logging': [
        {
            'target_systems': ['*'],
            'level': 'DEBUG',
            'handlers': [
                {
                    'type': 'file',
                    'name': 'reframe.log',
                    'level': 'DEBUG',
                    'format': '[%(asctime)s] %(levelname)s: '
                    '%(check_info)s: %(message)s',
                    'append': False,
                },

                # Output handling
                {
                    'type': 'stream',
                    'name': 'stdout',
                    'level': 'INFO',
                    'format': '%(message)s'
                },
                {
                    'type': 'file',
                    'name': 'reframe.out',
                    'level': 'INFO',
                    'format': '%(message)s',
                    'append': False,
                }
            ]
        }
    ],

    'perf_logging': [
        {
            'target_systems': ['*'],
            'level': 'DEBUG',
            'handlers': [
                {
                    'type': 'filelog',
                    'prefix': '%(check_system)s/%(check_partition)s',
                    'level': 'INFO',
                    'format': (
                        '%(asctime)s|reframe %(version)s|'
                        '%(check_info)s|jobid=%(check_jobid)s|'
                        '%(check_perf_var)s=%(check_perf_value)s|'
                        'ref=%(check_perf_ref)s '
                        '(l=%(check_perf_lower_thres)s, '
                        'u=%(check_perf_upper_thres)s)|'
                        '%(check_perf_unit)s'
                    ),
                    'append': True
                }
            ]
        }
    ],

    'general': [
        {
            'target_systems': ['*'],
            'check_path': ['checks/'],
            'check_path_recurse': True
        }
    ]
}

```

The `target_systems` will be optional and if not specified differently, it will default to `['*']`.

This issue should take care of defining the JSON schema for the configuration and of providing a first implementation that will read and load the configuration.

vkarak:
This PR introduces only the syntax of the new (upcoming) configuration. The new configuration will be fully JSON compliant and will provide more flexibility.

The whole configuration is now stored (in its Python version) as a JSON object in the `site_configuration` variable. Systems, partitions and environments are now defined as JSON objects inside a JSON array. In the past, these were entries of Python dictionary, whose keys corresponded to the names of the represented objects (systems, partitions, etc.). The name is now a property instead. This will solve the issue of random initialization of systems (see #66). During system auto-detection, ReFrame will now try the system definitions in order and pick the configuration entry that was first matched.

All logging configuration parameters will be moved inside the same `site_configuration` JSON object, encoded inside different properties.

New properties were added for scheduler-specific options and for general framework options, such as the search path for checks.

All base configuration entries (except the `systems`) define a `target_systems` attribute, which is essentially a reference to the system or system/partition combination that this entry refers to. If not specified, the entry is valid for any system and system partitions. This replaces the scoped dictionaries exposed in the current configuration (although internally scoped dictionaries may be used).

### Default values

JSON schema allows to define default values of properties, but they play no role in the JSON validation process; the validated JSON instance is not changed, and there are some good reasons for that (see [here](https://python-jsonschema.readthedocs.io/en/stable/faq/#why-doesn-t-my-schema-s-default-property-set-the-default-on-my-instance)). In order to access the default values, if specified, you would have to locate them inside the schema, which it would tie the code setting the defaults very  much to the schema itself (not the JSON instance). For this reason, I introduce a `defaults` property in the schema that stores the defaults organized hierarchically per property as this should appear in the JSON instance. This will lead to a much more intuitive code when we read the configuration and we need to figure out any default values.

A final note on the default values: I only specify default values if they are not `None` or not empty lists.

Finally, this PR provides an example configuration file (`schemas/settings.py`) which corresponds to the old configuration found in `unittests/resources/settings.py`.

For validating a configuration file, you can use this online JSON validator, where you can feed him with the schema: https://www.jsonschemavalidator.net/

Fixes #1113.

### DIFF
--- a/schemas/config.json
+++ b/schemas/config.json
@@ -1,0 +1,367 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/eth-cscs/reframe/master/schemas/config.json",
+    "title": "Validation schema for ReFrame's configuration file",
+    "defs": {
+        "alphanum_string": {
+            "type": "string",
+            "pattern": "([a-zA-Z0-9_]|-)+"
+        },
+        "system_ref": {
+            "type": "array",
+            "items": {"type": "string"}
+        },
+        "envvar_list": {
+            "type": "array",
+            "items": {
+                "type": "array",
+                "items": [
+                    {
+                        "type": "string",
+                        "pattern": "([a-zA-Z_][a-zA-Z0-9_]*)"
+                    },
+                    {"type": "string"}
+                ],
+                "additionalProperties": false
+            }
+        },
+        "modules_list": {
+            "type": "array",
+            "items": {"type": "string"}
+        },
+        "loglevel": {
+            "type": "string",
+            "enum": ["critical", "error", "warning",
+                     "info", "verbose", "debug"]
+        },
+        "handler_common": {
+            "type": "object",
+            "properties": {
+                "type": {"type": "string"},
+                "level": {"$ref": "#/defs/loglevel"},
+                "format": {"type": "string"},
+                "datefmt": {"type": "string"}
+            },
+            "required": ["type"]
+        },
+        "file_handler": {
+            "allOf": [
+                {"$ref": "#/defs/handler_common"},
+                {
+                    "properties": {
+                        "name": {"type": "string"},
+                        "append": {"type": "boolean"},
+                        "timestamp": {
+                            "anyOf": [
+                                {"type": "boolean"},
+                                {"type": "string"}
+                            ]
+                        }
+                    },
+                    "required": ["name"]
+                }
+            ]
+        },
+        "filelog_handler": {
+            "allOf": [
+                {"$ref": "#/defs/handler_common"},
+                {
+                    "properties": {
+                        "prefix": {"type": "string"}
+                    }
+                }
+            ]
+        },
+        "graylog_handler": {
+            "allOf": [
+                {"$ref": "#/defs/handler_common"},
+                {
+                    "properties": {
+                        "address": {"type": "string"},
+                        "extras": {"type": "object"}
+                    },
+                    "required": ["address"]
+                }
+            ]
+        },
+        "stream_handler": {
+            "allOf": [
+                {"$ref": "#/defs/handler_common"},
+                {
+                    "properties": {
+                        "name": {
+                            "type": "string",
+                            "enum": ["stdout", "stderr"]
+                        }
+                    }
+                }
+            ]
+        },
+        "syslog_handler": {
+            "allOf": [
+                {"$ref": "#/defs/handler_common"},
+                {
+                    "properties": {
+                        "socktype": {
+                            "type": "string",
+                            "enum": ["tcp", "udp"]
+                        },
+                        "facility": {"type": "string"},
+                        "address": {"type": "string"}
+                    },
+                    "required": ["address"]
+                }
+            ]
+        }
+    },
+    "type": "object",
+    "properties": {
+        "systems": {
+            "type": "array",
+            "items": {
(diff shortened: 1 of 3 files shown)


vkarak:
Any comments on this, guys?

teojgo:
@vkarak I am still looking into it.

schemas/config.json
```diff
@@ -0,0 +1,367 @@
...
+        "file_handler": {
+            "allOf": [
+                {"$ref": "#/defs/handler_common"},
+                {
+                    "properties": {
+                        "name": {"type": "string"},
+                        "append": {"type": "boolean"},
+                        "timestamp": {
+                            "anyOf": [
+                                {"type": "boolean"},
```
victorusu: I do not understand why the `timestamp` can be a boolean.

schemas/config.json
```diff
@@ -0,0 +1,367 @@
...
+                                    "items": {"type": "string"}
+                                },
+                                "container_platforms": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "object",
+                                        "properties": {
+                                            "name": {
+                                                "type": "string",
+                                                "enum": ["Docker", "Sarus",
```
victorusu: why are we removing shifter?

schemas/config.json
```diff
@@ -0,0 +1,367 @@
...
+                    "target_systems": {"$ref": "#/defs/system_ref"}
+                },
+                "additionalProperties": false
+            }
+        }
+    },
+    "required": ["systems", "environments", "logging", "perf_logging"],
+    "additionalProperties": false,
+    "defaults": {
+        "environments/cc": "cc",
```
victorusu: The defaults are not vanilla Linux. Should we make it 
```
"environments/cc": "cc",
"environments/cxx": "c++",
"environments/ftn": "gfortran",
...
```

schemas/config.json
vkarak: This is exactly what we specify in documentation. It's similar to a command-line option with an optional argument. If `true`, you will use the default format, if you pass the format, you will use that one.

schemas/config.json
vkarak: It is according to what we have in the documentation.

schemas/config.json
vkarak: I agree, but again this is what we have in the documentation. I think this should be a different issue. Can you please open it?

vkarak: @victorusu Can you make a feature request for this?

PR #1129, opened 2020-01-22
[feat] Evaluate `extra_resources` just before job submission

### LINKED ISSUE #1089: Evaluate also `extra_resources` lazily just before submitting the job
This is along the lines of #961. 

@bcfriesen This should allow you to define the `extra_resources` after the setup in order to use `self.stagedir`.

teojgo:
* Move the handling of the `extra_resources` during the
  `run` phase.

* Change the corresponding unit test to reflect the change.

Fixes #1089

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -621,12 +621,6 @@ class RegressionTest(metaclass=RegressionTestMeta):
     #:
     #:    A new more powerful syntax was introduced
     #:    that allows also custom job script directive prefixes.
-    #:
-    #: .. note::
-    #:    .. versionchanged:: 2.22
-    #:
-    #:    The attr:`extra_resources` is now used during the generation of
-    #:    allowing its definition after the :func:`setup` phase.
     extra_resources = fields.TypedField('extra_resources',
                                         typ.Dict[str, typ.Dict[str, object]])
 
@@ -1189,7 +1183,6 @@ class RegressionTest(metaclass=RegressionTestMeta):
                 self._current_partition.get_resource(r, **v))
 
         self._job.options = resources_opts + self._job.options
-
         with os_ext.change_dir(self._stagedir):
             try:
                 self._job.prepare(commands, environs)
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -622,6 +622,11 @@ class RegressionTest(metaclass=RegressionTestMeta):
     #:    A new more powerful syntax was introduced
     #:    that allows also custom job script directive prefixes.
     #:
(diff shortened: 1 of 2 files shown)


reframe/core/pipeline.py
```diff
@@ -1185,6 +1174,22 @@ def run(self):
...
+        # Get job options from managed resources and prepend them to
+        # job_opts. We want any user supplied options to be able to
+        # override those set by the framework.
+        resources_opts = []
+        for r, v in self.extra_resources.items():
+            resources_opts.extend(
+                self._current_partition.get_resource(r, **v))
+
+        self._job.options = resources_opts + self._job.options
+
```
vkarak: Remove this blank line.

unittests/test_pipeline.py
```diff
@@ -2,6 +2,7 @@
 import pytest
 import re
 import tempfile
+import types
```
vkarak: This is not used.

unittests/test_pipeline.py
```diff
@@ -343,21 +344,35 @@ def test_sourcepath_upref(self):
...
+            'gpu': {'num_gpus_per_node': 2},
+            'datawarp': {'capacity': '100GB', 'stagein_src': test.stagedir}
+        }
         test.job.options += ['--foo']
+
+        test.run()
+        test.wait()
+        test.check_sanity()
+        test.check_performance()
+        test.cleanup(remove_files=True)
```
vkarak: You should use the `_run()` function defined in this test file, instead of running yourself the test like that.

reframe/core/pipeline.py
```diff
@@ -622,6 +622,11 @@ class RegressionTest(metaclass=RegressionTestMeta):
     #:    A new more powerful syntax was introduced
     #:    that allows also custom job script directive prefixes.
     #:
+    #: .. note::
+    #:    .. versionchanged:: 2.22
+    #:
+    #:    The attr:`extra_resources` is now used during the generation of
+    #:    allowing its definition after the :func:`setup` phase.
```
vkarak: I don't think that this clarification is needed, since nothing relevant was mentioned in the documentation. This is an implementation detail.

unittests/test_pipeline.py
teojgo: I want to set the `extra_resources` between `setup` and `run` while `run` does both.

vkarak: Do it properly with the `run_after` decorator. Actually if you define the test the right way. See a bit above how this is done.

teojgo:
@jenkins-cscs retry all

PR #1138, opened 2020-01-25
[test] Remove stale FIXME and NOTE tags from tests

vkarak:
I also  updated the syntax in the touched tests.

Fixes UES-649.

### DIFF
--- a/cscs-checks/tools/profiling_and_debugging/perftools.py
+++ b/cscs-checks/tools/profiling_and_debugging/perftools.py
@@ -29,8 +29,8 @@ class PerftoolsCheck(rfm.RegressionTest):
             self.modules += ['craype-accel-nvidia60']
 
         self.build_system = 'Make'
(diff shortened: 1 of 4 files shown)


cscs-checks/tools/profiling_and_debugging/perftools.py
```diff
@@ -50,10 +50,6 @@ def __init__(self, lang):
                 'LIB=-lstdc++']
 
         self.executable = 'jacobi'
-        # NOTE: Reduce time limit because for PrgEnv-pgi even if the output
-        # is correct, the batch job uses all the time.
-        self.time_limit = (0, 5, 0)
```
victorusu: Are we letting the job run for 10 min? or the issue was resolved?

vkarak: The job finishes in time. This comment is no more valid, that's why I removed it.

PR #1150, opened 2020-02-04
[feat] Support RFC3339 compliant time zone formatting when logging time fields

### LINKED ISSUE #1061: Add a fully RFC3339 conformant date/time logging formatter
The `%z` timezone specifier of `strftime()` produces a valid ISO8601 formatting, but not a valid RFC3339:

```pycon
>>> time.strftime('%FT%T%z')
'2019-11-21T19:25:04-0700'
```

The RFC3339 conformant result would be `'2019-11-21T19:25:04-07:00'` (note that this is valid also under ISO8601).

Original request from @brandongc.

vkarak:
This PR adds a new time formatting directive for log record timestamps, namely the GNU `date` extension `%:z`. This directive will print the time zone difference in fully RFC3339 compliant way, i.e., `+/-HH:MM` instead or `+/-HHMM`, which `%z` prints.

This PR also replaces how `Job`'s completion time is stored and retrieved, also inside the Slurm backend. We use `time.time()` to retrieve the time in seconds from the Epoch instead of `datetime.now()`, because `datetime` does not store time zone information. Also all backends are now required to provide the completion time in seconds from the Epoch, so Slurm backend was modified accordingly.

Fixes #1061.

@brandongc FYI

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -8,6 +8,7 @@ import pprint
 import shutil
 import sys
 import socket
+import time
 from datetime import datetime
 
 import reframe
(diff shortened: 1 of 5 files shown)


reframe/core/logging.py
```diff
@@ -132,6 +133,37 @@ def close(self):
...
+
+    def format(self, record):
+        datefmt = self.datefmt or self.default_time_format
+        try:
+            if record.check_job_completion_time is not None:
+                ct = self.converter(record.check_job_completion_time)
+                record.check_job_completion_time = _format_time_rfc3339(
+                    ct, datefmt
+                )
+        except AttributeError:
```
vkarak: Most probably, I don't need the `try/except` here, because we always define all the extras.

PR #1152, opened 2020-02-04
[bugfix] Fix failure when check search path contains duplicates

### LINKED ISSUE #1119: Duplicated check path leads to error
Hi,
I had aliased the search path with 
`alias reframe='reframe -C /cvmfs/soft.computecanada.ca/custom/reframe/config/settings.py -c /cvmfs/soft.computecanada.ca/custom/reframe/checks'` 

Then, I wanted to use just a subdirectory, so I used : 
`reframe -R -r -c /cvmfs/soft.computecanada.ca/custom/reframe/checks/advanced/osu` 

but it failed because it found a duplicate test (i.e. the same one) : 
```
Command line: /home/mboisson/.local/easybuild/software/2017/Core/reframe/2.21/bin/reframe -C /cvmfs/soft.computecanada.ca/custom/reframe/config/settings.py -c /cvmfs/soft.computecanada.ca/custom/reframe/checks -R -r -c /cvmfs/soft.computecanada.ca/custom/reframe/checks/advanced/osu
Reframe version: 2.21
Launched by user: mboisson
Launched on host: build-node.computecanada.ca
Reframe paths
=============
    Check prefix      :
(R) Check search path : '/cvmfs/soft.computecanada.ca/custom/reframe/checks:/cvmfs/soft.computecanada.ca/custom/reframe/checks/advanced/osu'
    Stage dir prefix     : /home/mboisson/tmp/reframe/stage/
    Output dir prefix    : /home/mboisson/tmp/reframe/output/
    Perf. logging prefix : /home/mboisson/tmp/reframe/perflogs
/home/mboisson/.local/easybuild/software/2017/Core/reframe/2.21/bin/reframe: /cvmfs/soft.computecanada.ca/custom/reframe/checks/advanced/osu/osu_benchmarks.py: test `OSULatencyTest' already defined in `/cvmfs/soft.computecanada.ca/custom/reframe/checks/advanced/osu/osu_benchmarks.py'
``` 

When there are multiple `-c` argument, it would be more intuitive to overwrite the previous value, rather than concatenate them ?

teojgo:
* Duplicate check path is not added many times to the load path.

* If the recursive option is passed, only the parent path of the
checkpaths is used in the load path.

* Create the corresponding unit tests.

Fixes #1119


reframe/frontend/cli.py
```diff
@@ -410,15 +411,27 @@ def main():
...
+        load_path = OrderedSet()
         for d in options.checkpath.split(':'):
             d = os_ext.expandvars(d)
             if not os.path.exists(d):
                 printer.warning("%s: path `%s' does not exist. Skipping..." %
                                 (argparser.prog, d))
                 continue
 
-            load_path.append(d)
+            d = os.path.abspath(os.path.realpath(d))
```
victorusu: Why do we need to get the `abspath` from `realpath`? why not take `abspath(d)`?

teojgo: In case it is a symbolic link.

vkarak:
I have some comments on that one, I'll submit my review shortly.

reframe/frontend/cli.py
```diff
@@ -410,15 +411,27 @@ def main():
...
+            if d not in load_path:
+                load_path.add(d)
+
+        if options.recursive:
```
vkarak: You should do that while constructing the real load path and then you avoid the extra loop and creation of the list below.

reframe/frontend/cli.py
```diff
@@ -410,15 +411,27 @@ def main():
...
+            for d in list(load_path):
+                d_parent = os.path.dirname(d)
+                while d_parent not in {'', '/'}:
```
vkarak: Why do you check for the empty string? You have converted all paths to absolute.

reframe/frontend/cli.py
```diff
@@ -410,15 +411,27 @@ def main():
...
             d = os_ext.expandvars(d)
             if not os.path.exists(d):
                 printer.warning("%s: path `%s' does not exist. Skipping..." %
                                 (argparser.prog, d))
                 continue
 
-            load_path.append(d)
+            d = os.path.abspath(os.path.realpath(d))
+            if d not in load_path:
+                load_path.add(d)
```
```python
if d in load_path:
     continue
```

And then add it to the path after you have done the analysis for the recursive option below, which should be done in this loop, on `d`, not in a separate one.

reframe/frontend/cli.py
vkarak: I guess you have  put it for paths ending in `/`. You should normalize them before processing them.

unittests/test_cli.py
```diff
@@ -281,6 +281,51 @@ def test_checkpath_colon_separated(self):
             r'Found (\d+) check', stdout, re.MULTILINE).group(1)
         self.assertEqual(num_checks, '2')
 
+    def test_checkpath_duplicate(self):
```
vkarak: The whole functionality about removing duplicates should be put in a separate `os_ext` utility function and tested separately. Invoking ReFrame for testing such a stand-alone feature is overkill. Also in terms of test code + execution time.

reframe/frontend/cli.py
teojgo: The problem with this is that the shorter path may come after the longer one. So if '/a/b/c' preceeds '/a/'b' then I will not have considered it in the search for the parent paths of '/a/b/c'.

reframe/utility/os_ext.py
```diff
@@ -408,3 +409,36 @@ def concat_files(dst, *files, sep='\n', overwrite=False):
...
+        for p in list(unique_paths):
+            p_parent = os.path.dirname(p)
+            while p_parent != '/':
+                if p_parent in unique_paths:
+                    unique_paths.remove(p)
+                    break
+
+                p_parent = os.path.dirname(p_parent)
+
+    return list(unique_paths)
```
vkarak: Instead of removing the path from the `unique_paths`, it'd be more efficient if you just kept track of the paths to remove in a set. Then you could avoid the copy in for loop. Then you could simply return the `list(unique_paths - children)`

unittests/test_utility.py
```diff
@@ -1004,3 +1004,23 @@ def test_concat_files(self):
                 with open(concat_file) as cf:
                     out = cf.read()
                     assert out == 'Hello1\nHello2\n'
+
+    def test_unique_abs_paths(self):
+        p1 = 'a/b/c'
+        p1_copy = p1[:]
+        p1_parent = 'a/b'
+        p2 = '/d/e//'
+        p3 = '/d/e/f'
```
vkarak: It's not easy to follow what the unit tests are doing when you are using these variables for the paths. Just hard code them as list arguments. A bit of duplication here serves much more here, because the tests will be self-explaining.

unittests/test_utility.py
```diff
@@ -765,111 +763,111 @@ def test_update(self):
...
-        with self.assertRaises(AttributeError):
+        with pytest.raises(AttributeError):
             l.extend([3, 4])
 
-        with self.assertRaises(AttributeError):
+        with pytest.raises(AttributeError):
             l.insert(1, 4)
 
-        with self.assertRaises(AttributeError):
+        with pytest.raises(AttributeError):
```
vkarak: Can you introduce these changes separately with you upcoming PR about moving to pytest syntax? It's too hard to follow and off topic.

teojgo: oops, I retrieved the wrong file.

Issue #1162, opened 2020-02-14
Setup hooks don't work for CompileOnlyRegressionTest


ekouts:
The setup hooks are not called at all before/after the `setup` phase when a test is `CompileOnlyRegressionTest`. The reason is that the `setup()` function is overwritten in the `CompileOnlyRegressionTest` and the decorator is not inherited. 

Similarly in the `RunOnlyRegressionTest` the hooks for the `run` phase are called but not where we expect it. It is called before/after the call to the base class of `run` inside `RunOnlyRegressionTest.run()`.

Here is a simple example. When the test is `RegressionTest`.
```
import reframe as rfm
import reframe.utility.sanity as sn

@rfm.simple_test
class Example1Test(rfm.RegressionTest):
    def __init__(self):
        self.descr = 'Simple matrix-vector multiplication example'
        self.valid_systems = ['*']
        self.valid_prog_environs = ['*']
        self.sourcepath = 'example_matrix_vector_multiplication.c'
        self.executable_opts = ['1024', '100']
        self.sanity_patterns = sn.assert_found(
            r'time for single matrix vector multiplication', self.stdout)
        self.maintainers = ['you-can-type-your-email-here']
        self.tags = {'tutorial'}

    @rfm.run_before('setup')
    def test_hook(self):
        assert(0 == 1)
```

Output is failing (as it should) because of the hook:
```
Command line: ./bin/reframe -c tutorial/example1.py --skip-sanity -r
Reframe version: 3.0-dev1
Launched by user: ekoutsaniti
Launched on host: dhcp-133-245.cscs.ch
Reframe paths
=============
    Check prefix      : 
    Check search path : 'tutorial/example1.py'
    Stage dir prefix     : /Users/ekoutsaniti/repos/reframe-fork/stage/
    Output dir prefix    : /Users/ekoutsaniti/repos/reframe-fork/output/
    Perf. logging prefix : /Users/ekoutsaniti/repos/reframe-fork/perflogs
[==========] Running 1 check(s)
[==========] Started on Fri Feb 14 08:39:30 2020 

[----------] started processing Example1Test (Simple matrix-vector multiplication example)
[ RUN      ] Example1Test on generic:login using builtin-gcc
[     FAIL ] Example1Test
[  FAILED  ] Ran 1 test case(s) from 1 check(s) (1 failure(s))
[==========] Finished on Fri Feb 14 08:39:30 2020 

==============================================================================
SUMMARY OF FAILURES
[...]

PR #1163, opened 2020-02-14
[bugfix] Execute setup hooks for `CompileOnlyRegressionTest` and make the run hooks for the `RunOnlyRegressionTest` execute at the right place

ekouts:
Fix CompileOnlyRegressionTest setup hooks.

- Fix should be simple for the setup hooks, I just also need to find where to add a unittest for that.
- The `pre_run` hook in `RunOnlyRegressionTest` is not called exactly before the `run` but only because there is a call to the base class function.
- Finally I don't know if we want the hooks to be called in no-op functions, like the `run` in `CompileOnlyRegressionTest`. With the current implementation they won't be called.

Fixes issue #1162.


ekouts:
@jenkins-cscs retry tsa

reframe/core/pipeline.py
```diff
@@ -1516,7 +1517,7 @@ def run(self):
                 self._copy_to_stagedir(os.path.join(self._prefix,
                                                     self.sourcesdir))
 
-        super().run()
+        super().run.__wrapped__(self)
```
vkarak: Brava! Didn't know about that, but that's exactly what we want! I think it works, because we're using `@functools.wraps` in the `@_run_hooks` decorator.

unittests/test_pipeline.py
```diff
@@ -457,6 +457,37 @@ def postfoo(self):
...
+            def __init__(self):
+                self.name = 'hellocheck_compile'
+                self.descr = 'C Hello World test'
+
+                # All available systems are supported
+                self.valid_systems = ['*']
+                self.valid_prog_environs = ['*']
+                self.sourcepath = 'hello.c'
+                self.tags = {'foo', 'bar'}
+                self.name = type(self).__name__
```
vkarak: Which is the name you want. This or that in L464.

unittests/test_pipeline.py
```diff
@@ -457,6 +457,37 @@ def postfoo(self):
         assert '_RFM_PRE_SETUP' in os.environ
         assert '_RFM_POST_SETUP' in os.environ
 
+    def test_setup_hooks_in_compile_only_test(self):
+        @fixtures.custom_prefix('unittests/resources/checks')
+        class MyTest(rfm.CompileOnlyRegressionTest):
+            def __init__(self):
+                self.name = 'hellocheck_compile'
+                self.descr = 'C Hello World test'
```
vkarak: No need for description.

unittests/test_pipeline.py
```diff
@@ -457,6 +457,37 @@ def postfoo(self):
...
+
+                # All available systems are supported
```
vkarak: No need for this comment, too. And remove it the blank line before as well. Let's keep this inline test as short and simple as possible.

unittests/test_pipeline.py
```diff
@@ -457,6 +457,37 @@ def postfoo(self):
...
+                self.valid_systems = ['*']
+                self.valid_prog_environs = ['*']
+                self.sourcepath = 'hello.c'
+                self.tags = {'foo', 'bar'}
```
vkarak: Why tags?

unittests/test_pipeline.py
```diff
@@ -457,6 +457,37 @@ def postfoo(self):
...
+                self.name = type(self).__name__
+                self.executable = os.path.join('.', self.name)
```
vkarak: You should not care about this.

unittests/test_pipeline.py
```diff
@@ -457,6 +457,37 @@ def postfoo(self):
...
+                self.sanity_patterns = sn.any(sn.glob(self.executable))
```
vkarak: And for this as well, so you should simply put something like the following:

```suggestion
                self.sanity_patterns = sn.assert_found('.*', self.stdout)
```

unittests/test_pipeline.py
```diff
@@ -457,6 +457,37 @@ def postfoo(self):
...
+
+            @rfm.run_before('setup')
+            def prefoo(self):
+                assert self.current_environ is None
+                os.environ['_RFM_PRE_SETUP_COMPILE'] = 'foo'
```
vkarak: Better increment a counter or put a value in a list, rather than modifying the environment.

unittests/test_pipeline.py
```diff
@@ -498,6 +529,32 @@ def check_executable(self):
...
+                self.executable_opts = ['Hello, World!']
+                self.local = True
+                self.valid_prog_environs = ['*']
+                self.valid_systems = ['*']
+                self.sanity_patterns = sn.assert_found(
+                    r'Hello, World\!', self.stdout)
+
+            @rfm.run_before('run')
+            def setflags(self):
+                self.post_run = ['echo hello > greetings.txt']
```
vkarak: This should be enough:

```suggestion
                self.post_run = ['touch greetings.txt']
```

unittests/test_pipeline.py
```diff
@@ -498,6 +529,32 @@ def check_executable(self):
...
+
+            @rfm.run_after('run')
+            def check_executable(self):
```
vkarak:
```suggestion
            def check_greetings(self):
```

unittests/test_pipeline.py
```diff
@@ -498,6 +529,32 @@ def check_executable(self):
...
+                self.executable = './hello.sh'
+                self.executable_opts = ['Hello, World!']
+                self.local = True
+                self.valid_prog_environs = ['*']
+                self.valid_systems = ['*']
+                self.sanity_patterns = sn.assert_found(
+                    r'Hello, World\!', self.stdout)
+
+            @rfm.run_before('run')
+            def setflags(self):
```
vkarak: The name is wrong.

unittests/test_pipeline.py
vkarak: I think that you can go without defining the executable.

unittests/test_pipeline.py
```diff
@@ -462,31 +462,24 @@ def test_setup_hooks_in_compile_only_test(self):
...
-                # All available systems are supported
                 self.valid_systems = ['*']
                 self.valid_prog_environs = ['*']
                 self.sourcepath = 'hello.c'
-                self.tags = {'foo', 'bar'}
-                self.name = type(self).__name__
                 self.executable = os.path.join('.', self.name)
-                self.sanity_patterns = sn.any(sn.glob(self.executable))
+                self.sanity_patterns = sn.assert_found('.*', self.stdout)
+                self.var = 0
```
vkarak:
```suggestion
                self.count = 0
```

unittests/test_pipeline.py
```diff
@@ -542,15 +535,13 @@ def __init__(self):
...
+            def set_post_run(self):
+                self.sourcesdir = None
 
             @rfm.run_after('run')
-            def check_executable(self):
-                outfile = os.path.join(self.stagedir, 'greetings.txt')
-
-                # Make sure that this hook is executed after wait()
-                assert os.path.exists(outfile)
+            def check_greetings(self):
```
vkarak: The name has no meaning any more.

unittests/test_pipeline.py
```diff
@@ -542,15 +535,13 @@ def __init__(self):
                     r'Hello, World\!', self.stdout)
 
             @rfm.run_before('run')
-            def setflags(self):
-                self.post_run = ['echo hello > greetings.txt']
+            def set_post_run(self):
```
vkarak:
```suggestion
            def unset_sourcesdir(self):
```

unittests/test_pipeline.py
```diff
@@ -542,15 +535,13 @@ def __init__(self):
...
 
             @rfm.run_after('run')
-            def check_executable(self):
-                outfile = os.path.join(self.stagedir, 'greetings.txt')
-
-                # Make sure that this hook is executed after wait()
-                assert os.path.exists(outfile)
+            def check_greetings(self):
+                # Make sure nothing was copied to the stage directory
+                assert len(os.listdir(self.stagedir)) == 3
```
vkarak: This test is not so robust. It will fail if the number of files generated by reframe increases. This should run before the `run` phase and you should make sure that the stage directory is empty.

ekouts: So I replaced the `unset_sourcesdir` hook with the check that the stage directory is empty before run. This was failing before the change.

unittests/test_pipeline.py
ekouts: I have to define the executable name because otherwise the build script has this line `gcc hello.c -o ./TestHooks.test_setup_hooks_in_compile_only_test.<locals>.MyTest` and fails with the error `.../unittests/tmp52k816ku/stage/generic/login/builtin-gcc/hellocheck_compile/rfm_hellocheck_compile_build.sh: line 12: locals: No such file or directory`. Do you have any suggestion how I could avoid it?

vkarak: Ok, not a problem then to leave it as of now.

PR #1168, opened 2020-02-14
[test] CDO tests for Tsa 19.04

omlins:


vkarak:
@omlins Can you do this PR on @lucamar's branch for the Tsa tests instead? It is not yet merged that one.

Issue #1070, opened 2019-11-25
Add a Torque scheduler backend


vkarak:
The current PBS backend targets essentially PBS pro. FYI @ZQyou.

vkarak:
@smoors is working on a Torque backend. @ZQyou, since you have done some local modifications in our PBS backend (which is actually tailored to PBS pro rather than Torque), you two may want to sync.

PR #1169, opened 2020-02-14
[feat] allow grep_command_output() to accept a list or tuple of patterns

smoors:


vkarak:
Hi @smoors and thanks for contributing. I am wondering why would you need that? These `os_ext` utilities are only meant for internal framework use. You are not supposed to use them in your test. For `grep`-ing patterns in your test output, there are the sanity functions, which are much more powerful.

smoors:
hi @vkarak ! thanks for your quick reply.

it's not meant for testing.
currently there is no `torque` scheduler implemented yet, so I am implementing one based on the (very rudimentary) `pbs` scheduler.
to check the status of a job it would be nice if we can search for multiple regexes in one go, avoiding unneeded querying of the scheduler.

vkarak:
Good to know that you are working on a Torque backend! In fact, we have issue #1070 for that. I could assign it to you, if you wish. It'd be better to see the changes you propose in this PR as part of a larger PR about the Torque backend, where we could see the context better. Is this ok with you?

smoors:
sure :)
I've got a basic version working, but want to refine it a bit.

vkarak:
Great! I'll close this PR then and wait for your other with the full implementation. If you want to share your work in progress, you can always open a PR and mark with `[wip]` in the title.

PR #1170, opened 2020-02-14
[feat] Allow Sarus backend to load local images

rsarm:
Fixes UES-676


vkarak:
@rsarm Does it simply address UES-676 or does it fix it? If it only addresses it, why does it not fix it?

reframe/core/containers.py
```diff
@@ -120,10 +120,14 @@ class Sarus(ContainerPlatform):
     def __init__(self):
         super().__init__()
         self.with_mpi = False
+        self.pull_image = True
```
vkarak: Isn't this a more generic feature? Couldn't it work the same for Singularity? If yes, you could move the attribute to the base class. You will also need to document this variable and type check it.

rsarm: With Singularity we set something like
```
self.container_platform.image = 'docker://ubuntu:18.04'
```
Then the image pull and the command execution happen on a single command
```
singularity exec docker://ubuntu:18.04 <command>
```
If we put a path to an image, Singularity realizes the given image is already there and automatically skips the pull. I think as we have it now it covers both cases (needs to pull and image exists).

Since the image is stored with a different name (for `docker://ubuntu:18.04` it  would be `ubuntu_18.04.sif`), to do the pull and exec on two different steps (*), we would need to change a bit more the code than what we did for Shifter/Sarus. We would have to change the launch command to use `ubuntu_18.04.sif` if the test gives something like `docker://ubuntu:18.04`, or change the prepare command if a path to an image is given.

(*) The pull step looks like this
```
singularity pull docker://ubuntu:18.04
```
and creates an image `ubuntu_18.04.sif`. The run step looks like this
```
singularity exec ubuntu_18.04.sif <cmd>
```

What do you think?

vkarak: I agree with your comment about Singularity. There is no need to change the backend. Regarding Sarus, though, could we avoid the `pull_image` attribute? What if the backend recognized when it should or should not pull and do that transparently to the user?

teojgo: The alternative would be to always pull before running. If the layers are found in cache, it just extracts the layers. I am not sure if this can be skipped.

vkarak: I am against changing the Singularity backend. They know better when to pull or not and since it's working for both arguments, I'd leave it like that. The problem is with Sarus. My question is essentially, whether ReFrame could figure out automatically when to pull or not and issue the command accordingly without having to specify it in the test.

rsarm: ReFrame could figure out that an image doesn't come from DockerHub since local images start with `load/library`. On the test we write something like this
`self.container_platform.image = 'load/library/lammps:22Aug2018-gpu'`
I'm not sure thought if that's portable. I don't know if `load/libary` is configurable parameter of the Sarus registry that might change.

vkarak: Why the decisive factor couldn't be that the image name starts with `docker://` or not?

rsarm: no no, that's only for singularity

vkarak: @rsarm I've just talked to Alberto. When the `load/` prefix (not the `load/libraries/`) is present, it is safe to assume that the container is local. The `load/` prefix serves the role of a server name, just as `docker.index.io`, which is the default. So I would suggest to remove the `pull_image` attribute completely.

vkarak: Another question that pops up is whether we need to allow tweaking of the pull command so as to support [this](https://sarus.readthedocs.io/en/latest/user/user_guide.html#pulling-images-from-private-or-3rd-party-registries). But we can skip it from this PR and open a separate issue if we think we need to support it.

rsarm: ok, thanks! I will write it using the `/load/library`.

vkarak: No, just the `load/` prefix.

vkarak:
@rsarm Can you please also update the PR title accordingly?

reframe/core/containers.py
```diff
@@ -123,6 +123,10 @@ def __init__(self):
         self._command = 'sarus'
 
     def emit_prepare_commands(self):
+        repo_name, *_ = self.image.split('/')
+        if repo_name == 'load':
```
vkarak: I would simply check if `self.image.startswith('load/')`.

unittests/test_containers.py
```diff
@@ -104,6 +104,17 @@ def expected_cmd_with_run_opts(self):
                 "--foo --bar name:tag bash -c 'cd /stagedir; cmd'")
 
 
+class TestShifterNGLocalImage(TestShifterNG):
+    @property
+    def expected_cmd_prepare(self):
+        return []
+
+    def test_prepare_command(self):
+        self.container_platform.image = 'load/library/name:tag'
```
vkarak:
```suggestion
        self.container_platform.image = 'load/library/image:tag'
```

PR #1172, opened 2020-02-15
[feat] Add support for the Torque scheduler

### LINKED ISSUE #1070: Add a Torque scheduler backend

smoors:
Fixes #1070.


smoors:
@vkarak this works at least on our cluster.
my knowledge of the ReFrame code is limited, so I'll be happy to get suggestions for improvement.

vkarak:
@smoors Thanks a lot for the PR. I will have a look at it soon and give you feedback.

vkarak:
@smoors You may also join the Slack channel, cos there is already some interest in this backend :-)

smoors:
@vkarak I already tried to join the Slack channel via https://reframe-slack.herokuapp.com/ but somehow it did not work.

maybe you could invite me?

vkarak:
@smoors Can you register to the mailing list, so that I can get your email from there and invite you manually?

smoors:
done, thanks.

vkarak:
@smoors Let me know if you got the invitation to Slack.

ZQyou:
@smoors  @vkarak Thanks. This is great.  I will test this branch and get back to you soon.

smoors:
@ZQyou if you have made additional changes to the pbs backend,
that apply also to the torque backend,
I'd be interested in those too.

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
+import re
```
vkarak: Please add the following copyright notice in this file:

```
# Copyright 2016-2020 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
# ReFrame Project Developers. See the top-level LICENSE file for details.
#
```

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
...
+
+from reframe.core.config import settings
+from reframe.core.exceptions import JobError
+from reframe.core.logging import getlogger
+from reframe.core.schedulers.pbs import PbsJobScheduler, _run_strict
+from reframe.core.schedulers.registry import register_scheduler
+import reframe.utility.os_ext as os_ext
```
vkarak: Move this before the `from ... import`.

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
...
+        # Needed for clusters that have the keep_completed parameter set to 0
+        if completed.returncode == 153:
+            getlogger().debug(
+                'jobid not known by scheduler, assuming job completed'
+            )
+            job.state = 'COMPLETED'
+            return
+
+        if completed.returncode != 0:
+            raise JobError(completed.returncode, completed.stderr)
```
vkarak: `JobError` takes a message and the job id.

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
...
+    'C': 'COMPLETED',
+}
+
+
+@register_scheduler('torque')
+class TorqueJobScheduler(PbsJobScheduler):
+    def _update_state(self, job):
+        '''Check the status of the job.'''
+
+        completed = os_ext.run_command('qstat -f %s' % job.jobid, log=False)
```
vkarak: Why do you disable logging of the command here? It can be useful for debugging.

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        if not state_match:
+            getlogger().debug(
+                'job state not found (stdout follows)\n%s' % completed.stdout
+            )
+            return
+
+        state = state_match.group('state')
+        job.state = JOB_STATES[state]
+
```
vkarak: You may remove this blank line.

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
...
+        if job.state == 'COMPLETED':
+            code_match = re.search(
+                r'^\s*exit_status = (?P<code>\d+)',
+                completed.stdout,
+                re.MULTILINE,
+            )
+            if not code_match:
+                return
```
vkarak: Please leave a blank line after `return`.

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
...
+            job.exitcode = int(code_match.group('code'))
+
+    def _emit_lselect_option(self, job):
+        num_tasks_per_node = job.num_tasks_per_node or 1
+        num_cpus_per_task = job.num_cpus_per_task or 1
+        num_nodes = job.num_tasks // num_tasks_per_node
+        num_cpus_per_node = num_tasks_per_node * num_cpus_per_task
+
+        select_opt = '-l nodes=%s:ppn=%s' % (num_nodes, num_cpus_per_node)
```
vkarak: I think that this is the only difference with the PBS `_emit_lselect_option`, right? We should find then a better way to avoid all this code duplication.

reframe/core/schedulers/torque.py
```diff
@@ -0,0 +1,117 @@
...
+            job.script_filename,
+        )
+        completed = _run_strict(cmd, timeout=settings().job_submit_timeout)
+        jobid_match = re.search(r'^(?P<jobid>\S+)', completed.stdout)
+        if not jobid_match:
+            raise JobError(
+                'could not retrieve the job id ' 'of the submitted job'
+            )
+
+        jobid, *info = jobid_match.group('jobid').split('.', maxsplit=1)
```
vkarak: The `maxsplit` is the only difference from PBS' `submit()`. Can't we remove completely your `submit()`?

smoors:
> Hi @smoors, thanks for your PR. Apart from my other comments, we will need a unit test. Have a look at the `test_schedulers.py`. There is already there a basic framework to test the backends, so it should not be difficult to add them. Another question I have is about `qstat`. Does it always produce an output for a job that has finished (like `sacct` from Slurm for example), or does its result perishes after a while?

@vkarak thanks for the review. I'll look into a unit test. the time that a completed job stays in the list depends on the cluster configuration, that's exactly why I added the first if clause in the update_state method. I'll update the comment to make it more clear.

reframe/core/schedulers/torque.py
smoors: no particular reason, I'll enable it again if you prefer.

reframe/core/schedulers/torque.py
smoors: ok, I'll try to factor it out.

reframe/core/schedulers/torque.py
smoors: well, `maxsplit` has to be equal to 1 to work, at least in our site. so I'm ok to remove it if it can be changed to 1 in the PBS (pro) scheduler.

smoors: another option is to not split it at all..

vkarak:
@jenkins-cscs ok to test

unittests/test_schedulers.py
```diff
@@ -530,6 +530,52 @@ def test_submit_timelimit(self):
...
+            '#PBS -q %s' % self.testjob.sched_partition,
+            '#PBS --gres=gpu:4',
+            '#DW jobdw capacity=100GB',
+            '#DW stage_in source=/foo'
+        ])
+        with open(self.testjob.script_filename) as fp:
+            found_directives = set(re.findall(r'^\#\w+ .*', fp.read(),
+                                              re.MULTILINE))
+
+        self.assertEqual(expected_directives, found_directives)
```
vkarak: Use pytest syntax here:

```suggestion
        assert expected_directives == found_directives
```

unittests/test_schedulers.py
```diff
@@ -530,6 +530,52 @@ def test_submit_timelimit(self):
...
+
+    def test_submit_timelimit(self):
+        # Skip this test for PBS, since we the minimum time limit is 1min
+        self.skipTest("Torque minimum time limit is 60s")
```
vkarak:
```suggestion
        pytest.skip("Torque minimum time limit is 60s")
```

unittests/test_schedulers.py
```diff
@@ -530,6 +530,52 @@ def test_submit_timelimit(self):
...
+        super().test_prepare()
+        num_nodes = self.testjob.num_tasks // self.testjob.num_tasks_per_node
+        num_cpus_per_node = (self.testjob.num_cpus_per_task *
+                             self.testjob.num_tasks_per_node)
+        expected_directives = set([
+            '#PBS -N "testjob"',
+            '#PBS -l walltime=0:5:0',
+            '#PBS -o %s' % self.testjob.stdout,
+            '#PBS -e %s' % self.testjob.stderr,
+            '#PBS -l nodes=%s:ppn=%s:haswell' % (num_nodes, num_cpus_per_node),
```
vkarak: Since this is the only difference to the PBS scheduler, I'd factor this out as a separate dynamic property and have the test class derive from the `TestPbsJob`.

unittests/test_schedulers.py
```diff
@@ -530,6 +530,52 @@ def test_submit_timelimit(self):
...
 
 
+class TestTorqueJob(_TestJob, unittest.TestCase):
+    @property
+    def sched_name(self):
+        return 'torque'
+
+    @property
+    def launcher_name(self):
+        return 'local'
```
vkarak: You may remove this completely if you inherit from `TestPbsJob` as I suggest below.

vkarak: Also in order to make sure that the unit tests pass for the Torque backend, you should run them as follows:

```
./test_reframe.py --rfm-user-config=/path/to/config/using/torque.py
```

unittests/test_schedulers.py
smoors: actually this is not the only difference; there's also the testjob options.
i've tried to factor out as much as possible, not sure if this is what you had in mind.

unittests/test_schedulers.py
smoors: i've tested it with a user config and all torque tests pass (test_submit_timelimit skipped).

unittests/test_schedulers.py
vkarak: It's good overall, I will push some minor changes to that.

unittests/test_schedulers.py
vkarak: Yep, I've just adapted our PBS config and works fine also with Slurm wrappers. That's good news, because we can CI regularly this backend. I will push the CI changes to this PR as well.

vkarak:
@ZQyou Can you check if the Torque backend submitted here works out-of-the-box for your installation? It works fine for me with the Slurm Torque wrappers.

ZQyou:
@vkarak  I have tested @smoors Torque backend for our regular regression testing.   I simply replaced `pbs` partition with `torque` in my settings and checks  and  It works pretty well.  

I like the way handling code 153. It saves my time to rerun some failures.

Some features I have in my local changes but missing here are:
1. The option `-A ACCOUNT, --account ACCOUNT`:I add these in my `pbs.py`
```
    def emit_preamble(self, job):
[ ... skipped .. ]
        if job.sched_account:
            preamble.append(
                job._format_option('-A %s' % job.sched_account))
```
2. The job property `job.nodelist`: I call this function when the job is in completed state:
```
    def _get_nodelist(self, job, jobstat):
        # exec_host = o0580/0-27+o0444/0-27+o0345/0-27
        nodelist_match = re.search('exec_host = (?P<nodelist>\S+)', jobstat.stdout)
        if nodelist_match:
            nodelist = [ x.split('/')[0] for x in nodelist_match.group('nodelist').split('+') ]
            nodelist.sort()
            job.nodelist = nodelist
```
3. The option ` --nodelist NODELIST`: Unlike Slurm Torque specify nodes via the resource  `nodes=`.  The changes I made can be found in https://github.com/ZQyou/reframe/blob/osc/2.21/reframe/core/schedulers/pbs.py

-ZQ

vkarak:
Thanks @ZQyou for testing! I would suggest to merge this PR for now and we can add this functionality in a separate PR. I will create two issues to keep track of that, meanwhile. Would you mind submitting a PR with your changes as well?

vkarak:
I have opened two separate feature requests about these modifications: #1212, #1213.

PR #1182, opened 2020-02-19
[test] Update GROMACS maintenance reference values

victorusu:
Removed the FIXME from GROMACS test and set the value to 0.0 for the `daint:mc` partition. This is due to the fact that we do not run this test anymore under this circumstance.


vkarak:
@jenkins-cscs retry daint

PR #1183, opened 2020-02-19
[bugfix] Fix TypeError crash in the performance checking phase when the extracted performance variable value is not a number

### LINKED ISSUE #1146: performance report does not accept str
Running this check [src](https://gist.github.com/jgphpc/11fc212945ef1beb7e44c2998afe1e25)
works:

```
PERFORMANCE REPORT
ExampleRunOnlyTest
- dom:login
   - PrgEnv-cray
      * num_tasks: 1
      * elapsed_time: 8 s
      * filename: 9.c <---------- yes
```

but fails with:

```
FAILURE INFO for ExampleRunOnlyTest
  * System partition: dom:login
  * Environment: PrgEnv-cray
  * Stage directory: /apps/common/UES/sandbox/jgp/hpctools.git/reframechecks/common/sphexa/DEBUGJG/JIRA1/stage/dom/login/PrgEnv-cray/ExampleRunOnlyTest
  * Node list: dom101
  * Job type: local (id=4458)
  * Maintainers: ['x']
  * Failing phase: performance
  * Reason: unexpected error: '>=' not supported between instances of 'str' and 'float'
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/frontend/executors/__init__.py", line 163, in _safe_call
    return fn(*args, **kwargs)
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/core/pipeline.py", line 96, in _fn
    func(obj, *args, **kwargs)
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/core/pipeline.py", line 1235, in performance
    self.check_performance()
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/core/pipeline.py", line 1319, in check_performance
    'expected {1} (l={2}, u={3})' % tag))
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/utility/sanity.py", line 727, in evaluate
    return expr.evaluate()
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/core/deferrable.py", line 53, in evaluate
    ret = self._fn(*fn_args, **fn_kwargs)
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/utility/sanity.py", line 546, in assert_reference
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/utility/sanity.py", line 727, in evaluate
    return expr.evaluate()
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/core/deferrable.py", line 53, in evaluate
    ret = self._fn(*fn_args, **fn_kwargs)
  File "/apps/common/UES/sandbox/jgp/reframe.git.jgphpc/reframe/utility/sanity.py", line 492, in assert_bounded
    if val >= lower and val <= upper:
TypeError: '>=' not supported between instances of 'str' and 'float'
```

Would be nice to have reframe support this. 
Would be super nice to have this (being able to use a string for the unit) supported:

```
      * elapsed_time: 8 (9.c)
````

### PR DESCRIPTION
**Pull request by victorusu**
Instead of failing when comparing reference of different types with
```
TypeError: '>=' not supported between instances of 'str' and 'float'
```
now, ReFrame fails with a much nicer message like (*edited by @vkarak*)

```
the value extracted for performance variable 'generic:login:filename' is not a number: 9.c
```

Closes #1146

### CODE REVIEW

File: reframe/utility/sanity.py [hunk 1]
```diff
@@ -547,6 +547,8 @@ def calc_bound(thres):
     except SanityError:
         error_msg = msg or '{0} is beyond reference value {1} (l={2}, u={3})'
         raise SanityError(_format(error_msg, val, ref, lower, upper))
+    except TypeError:
```
vkarak: The fix should go inside the actual comparison in the [assert_bounded()](https://github.com/eth-cscs/reframe/blob/d38da74943e3317ae1fc6f541e2aaf837307c2ef/reframe/utility/sanity.py#L492) function.

reframe/utility/sanity.py
```diff
@@ -547,6 +547,8 @@ def calc_bound(thres):
...
+        raise SanityError(_format('cannot compare {0} with reference value {1}', val, ref))
```
vkarak:
```suggestion
        raise SanityError(_format('cannot compare {0} with reference value {1}', val, ref)) from None
```

vkarak:
@victorusu I have changed the implementation. I am quoting the comment of my last commit:

```
    Move checking of the perf var extracted variable inside the pipeline

    The reason for that is that we don't do any argument type checking in any other
    sanity function, so it would be strange to do it in either `assert_reference()`
    or `assert_bounded()` only. The reason for that is that type and value errors
    are handled specially if they are raised from user test context. In this
    particular case, though, the `assert_reference()` is called from the framework
    context as a result of user input. So it makes more sense then to do the type
    checking inside `check_performance()`, where we call it.

    We also don't type check the reference and lower/upper thresholds, because this
    should be addressed correctly by the updated declaration of the `reference`
    field as soon as https://github.com/eth-cscs/reframe/issues/1147 is fixed (see
    also related FIXME note for the `reference` field).

    Finally, this commit adds also a pipeline unit test to check for this condition.
```

vkarak:
@teojgo Can you review this PR?

PR #1185, opened 2020-02-19
[feat] Set default value of `sourcesdir` to `'src'` only if such a directory exists, otherwise set to `None`

### LINKED ISSUE #1094: test fail if 'src' directory is not present, even if no sources are required
When running tests that don't need any sources (`sourcepath` is left unspecified), they fail hard if no `src` directory is present (because that's the default value for `sourcesdir`).

```
  * Failing phase: run
  * Reason: pipeline error: virtual copying of files failed: [Errno 2] No such file or directory: '/tmp/test_reframe/src'
```

Either the non-existing `src` directory should be ignored if no sources are required for the test, or the default value for `sourcesdir` should be change to `None`?

victorusu:
~~If the test is derived from RunOnlyRegressionTest, then it does not have
to create a `src/` dir in the test folder anymore. By default, if there is
no `src/` dir, then ReFrame will set the `self.sourcesdir` to `None`.~~

Close #1094


reframe/core/pipeline.py
```diff
@@ -1459,6 +1459,11 @@ class RunOnlyRegressionTest(RegressionTest):
     module.
     '''
 
+    def _rfm_init(self, *args, **kwargs):
+        super()._rfm_init(*args, **kwargs)
+        if not os.path.exists(os.path.join(self._prefix, 'src')):
+            self.sourcesdir = None
+
```
vkarak: As pointed out by @teojgo, it is wrong to set it here. Also why don't we generalise this behaviour and implement it once in the `RegressionTest()`? if `src/` exists, it becomes the default otherwise `sourcesdir` is set to `None`. This can be done in the `_rfm_init()` of `RegressionTest`. We will need to update the documentation likewise and perhaps a new unit test or adapt an existing one.

vkarak:
@teojgo Please review this one.

reframe/core/pipeline.py
vkarak: Oh, I should remove this.

vkarak:
@jenkins-cscs retry daint

PR #1186, opened 2020-02-19
[feat] Add a tool to convert from the old to the new configuration syntax

### LINKED ISSUE #1114: Create a tool to convert from the old to the new configuration syntax
The conversion functionality should be provided by ReFrame, which will use it itself to automatically convert any given old configuration file to the new syntax. The tool will just invoke this conversion function.

ekouts:
The PR, so far, contains only the function that converts the old configuration to a python json object. It also updates the validation schema and the example of the new configuration, which is based on `unittests/resources/settings.py`.

The validation schema still lacks support for the variables:
- `job_poll_intervals`
- `job_submit_timeout`

Closes #1114.


reframe/core/config.py
```diff
@@ -217,3 +218,106 @@ def create_env(system, partition, name):
                 system.add_partition(part)
 
             self._systems[sys_name] = system
+
+
+def create_json_object():
```
vkarak: The interface I want is this one:

```python
def convert_old_config(filename: str) -> str:
```
This should read the file `filename` convert it to the new file and store it in a temporary file, whose name it will return.

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+    }
+    old_systems = old_config.site_configuration['systems'].items()
+    for sys_name, sys_specs in old_systems:
+        sys_dict = {'name': sys_name}
+        sys_dict.update(sys_specs)
+        # Make variables dictionary into a list of lists
+        if 'variables' in sys_specs:
+            sys_dict['variables'] = []
+            for vname, v in sys_specs['variables'].items():
+                sys_dict['variables'].append([vname, v])
```
vkarak: Leave a blank line here.

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+        'logging': [],
+        'perf_logging': [],
+    }
+    old_systems = old_config.site_configuration['systems'].items()
+    for sys_name, sys_specs in old_systems:
+        sys_dict = {'name': sys_name}
+        sys_dict.update(sys_specs)
+        # Make variables dictionary into a list of lists
+        if 'variables' in sys_specs:
+            sys_dict['variables'] = []
```
vkarak:
```suggestion
            sys_dict['variables'] = list(sys_specs['variables'].items())
```

vkarak: If the validation is not happy with the list of tuples, then just do it in a list comprehension as follows:

```suggestion
            sys_dict['variables'] = [[vname, v] for vname, v in sys_dict[ 'variables']]
```

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+                new_p.update(p)
+                if p['scheduler'] == 'nativeslurm':
+                    new_p['scheduler'] = 'slurm'
+                    new_p['launcher'] = 'srun'
+                elif p['scheduler'] == 'local':
+                    new_p['scheduler'] = 'local'
+                    new_p['launcher'] = 'local'
+                else:
+                    sched = p['scheduler'].split('+')[0]
+                    launch = p['scheduler'].split('+')[1]
```
vkarak:
```suggestion
                    sched, launch, *_ = p['scheduler'].split('+')
```

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+                    new_p['scheduler'] = sched
+                    new_p['launcher'] = launch
+
+                # Make resources dictionary into a list
+                if 'resources' in p:
+                    new_p['resources'] = []
```
vkarak:
```suggestion
                    new_p['resources'] = [{'name': rname, 'options': r} for rname, r in p['resources'].items()]
```

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+                    for rname, r in p['resources'].items():
+                        new_r = {'name': rname, 'options': r}
+                        new_p['resources'].append(new_r)
+
+                # Make variables dictionary into a list of lists
+                if 'variables' in p:
+                    new_p['variables'] = []
+                    for vname, v in p['variables'].items():
```
vkarak: Same comment as before.

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+
+                sys_dict['partitions'].append(new_p)
+
+        converted['systems'].append(sys_dict)
+
+    old_environs = old_config.site_configuration['environments'].items()
+    for env_target, env_entries in old_environs:
+        for ename, e in env_entries.items():
+            new_env = {'name': ename}
+            if (env_target != '*'):
```
vkarak:
```suggestion
            if env_target != '*':
```

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+        converted[log_name].append(
+            {
+                'level': original_log['level'].lower(),
+                'handlers': new_handlers
+            }
+        )
+
+    update_config('logging', old_config.logging_config)
+    update_config('perf_logging', old_config.perf_logging_config)
+
```
vkarak: Remove this line.

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+    if (hasattr(old_config, 'checks_path') or
+        hasattr(old_config, 'checks_path_recurse')):
```
vkarak: To avoid this check here, I would put an empty entry for `general` in the new dictionary in any case and just fill it in as you have it below. At the end I would check if it's still empty and if it is, I'd delete it.

schemas/settings.py
```diff
@@ -63,7 +63,7 @@
         },
         {
             'name': 'sys0',
-            'descr': 'System for testing check dependencies',
+            'descr': 'System for test dependencies unit tests',
```
vkarak:
```suggestion
            'descr': 'System for checking test dependencies',
```

schemas/settings.py
```diff
@@ -174,4 +174,10 @@
             ]
         }
     ]
+    'general': [
+        {
+            'check_search_path': ['checks/'],
+            'check_search_recursive': True
+        }
+    ],
```
vkarak: Aren't these the default ones?

reframe/core/config.py
```diff
@@ -217,3 +220,121 @@ def create_env(system, partition, name):
...
+                'check_search_path'
+            ] = old_config.checks_path
+
+        if hasattr(old_config, 'checks_path_recurse'):
+            converted['general'][0][
+                'check_search_recursive'
+            ] = old_config.checks_path_recurse
+
+    with tempfile.NamedTemporaryFile(mode='w', delete=False) as fp:
+        json.dump(converted, fp, indent=4)
```
vkarak: The only problem with that is that it generates a JSON config file from a Python one, whereas the user would perhaps expect again a Python one. Some users have code in their configuration file for dynamically generating some configuration options. Although we cannot reproduce this dynamic part, at least we can provide a reasonable starting point. I suggest that you write a formatting function for Python dictionaries as well. The `debug.repr()` function does a similar thing as well; it should not be too difficult.

schemas/settings.py
ekouts: I initially changed that to match the description in `unittests/resources/settings.py`. So I will also update that file with the recommended description.

schemas/settings.py
ekouts: They are, so I will remove it

vkarak:
@ekouts How your `ppretty()` function differs to Python's [pprint()](https://docs.python.org/3.8/library/pprint.html?highlight=pretty)?

ekouts:
I tried `pprint()` with `indent=2` but the output didn't look that much like the example. On the other hand if you think it's good enough then I can use this instead. The best I could get is this:
```
{ 'systems': [ { 'descr': 'Generic example system',
                 'hostnames': ['localhost'],
                 'name': 'generic',
                 'partitions': [ { 'access': [],
                                   'descr': 'Login nodes',
                                   'environs': ['builtin-gcc'],
                                   'launcher': 'local',
                                   'modules': [],
                                   'name': 'login',
                                   'scheduler': 'local'}]},
               { 'descr': 'Fake system for unit tests',
                 'hostnames': ['testsys'],
                 'modules': ['foo/1.0'],
                 'name': 'testsys',
                 'partitions': [ { 'descr': 'Login nodes',
                                   'environs': [ 'PrgEnv-cray',
                                                 'PrgEnv-gnu',
                                                 'builtin-gcc'],
                                   'launcher': 'local',
                                   'name': 'login',
                                   'resources': [],
                                   'scheduler': 'local'},
                                 { 'access': [],
                                   'descr': 'GPU partition',
                                   'environs': ['PrgEnv-gnu', 'builtin-gcc'],
                                   'launcher': 'srun',
                                   'modules': ['foogpu'],
                                   'name': 'gpu',
                                   'resources': [ { 'name': 'gpu',
                                                    'options': [ '--gres=gpu:{num_gpus_per_node}']},
                                                  { 'name': 'datawarp',
                                                    'options': [ '#DW jobdw '
                                                                 'capacity={capacity}',
                                                                 '#DW stage_in '
                                                                 'source={stagein_src}']}],
                                   'scheduler': 'slurm',
                                   'variables': [['FOO_GPU', 'yes']]}],
                 'perflogdir': '.rfm_testing/perflogs',
                 'prefix': '.rfm_testing',
                 'resourcesdir': '.rfm_testing/resources',
                 'variables': [['FOO_CMD', 'foobar']]},
               { 'descr': 'System for checking test dependencies',
                 'hostnames': ['sys\\d+'],
                 'name': 'sys0',
                 'partitions': [ { 'environs': ['e0', 'e1'],
                                   'launcher': 'local',
                                   'name': 'p0',
                                   'scheduler': 'local'},
                                 { 'environs': ['e0', 'e1'],
                                   'launcher': 'local',
                                   'name': 'p1',
                                   'scheduler': 'local'}]}]}

```
And the output from the one I wrote is this:
```
{
    'systems': [
        {
            'name': 'generic',
            'descr': 'Generic example system',
            'hostnames': [
                'localhost'
            ],
            'partitions': [
                {
                    'name': 'login',
                    'scheduler': 'local',
                    'modules': [],
                    'access': [],
                    'environs': [
                        'builtin-gcc'
                    ],
                    'descr': 'Login nodes',
                    'launcher': 'local'
                }
            ]
        },
        {
            'name': 'testsys',
            'descr': 'Fake system for unit tests',
            'hostnames': [
                'testsys'
            ],
            'prefix': '.rfm_testing',
            'resourcesdir': '.rfm_testing/resources',
            'perflogdir': '.rfm_testing/perflogs',
            'modules': [
                'foo/1.0'
            ],
            'variables': [
                [
                    'FOO_CMD',
                    'foobar'
                ]
            ],
            'partitions': [
                {
                    'name': 'login',
                    'scheduler': 'local',
                    'resources': [],
                    'environs': [
                        'PrgEnv-cray',
                        'PrgEnv-gnu',
                        'builtin-gcc'
                    ],
                    'descr': 'Login nodes',
                    'launcher': 'local'
                },
                {
                    'name': 'gpu',
                    'scheduler': 'slurm',
                    'modules': [
                        'foogpu'
                    ],
                    'variables': [
                        [
                            'FOO_GPU',
                            'yes'
                        ]
                    ],
                    'resources': [
                        {
                            'name': 'gpu',
                            'options': [
                                '--gres=gpu:{num_gpus_per_node}'
                            ]
                        },
                        {
                            'name': 'datawarp',
                            'options': [
                                '#DW jobdw capacity={capacity}',
                                '#DW stage_in source={stagein_src}'
                            ]
                        }
                    ],
                    'access': [],
                    'environs': [
                        'PrgEnv-gnu',
                        'builtin-gcc'
                    ],
                    'descr': 'GPU partition',
                    'launcher': 'srun'
                }
            ]
        },
        {
            'name': 'sys0',
            'descr': 'System for checking test dependencies',
            'hostnames': [
                'sys\\d+'
            ],
            'partitions': [
                {
                    'name': 'p0',
                    'scheduler': 'local',
                    'environs': [
                        'e0',
                        'e1'
                    ],
                    'launcher': 'local'
                },
                {
                    'name': 'p1',
                    'scheduler': 'local',
                    'environs': [
                        'e0',
                        'e1'
                    ],
                    'launcher': 'local'
                }
            ]
        }
    ],
}
```

vkarak:
@ekouts Way better your formatting. I think we should have this as a separate function in `utility`. I will do my full review shortly.

ekouts:
> @ekouts Way better your formatting. I think we should have this as a separate function in `utility`.

Yes it makes sense because it's more general and actually handles tuples as well that are not part of the configuration as far as I could tell.

reframe/core/config.py
```diff
@@ -217,3 +220,151 @@ def create_env(system, partition, name):
                 system.add_partition(part)
 
             self._systems[sys_name] = system
+
+
+def ppretty(value, htchar=' ', lfchar='\n', indent=4, current_depth=0):
+    nlch = lfchar + htchar * indent * (current_depth + 1)
+    if type(value) is dict:
```
vkarak:
```suggestion
    if isinstance(value, collections.abc.Mapping):
```

reframe/core/config.py
```diff
@@ -217,3 +220,151 @@ def create_env(system, partition, name):
...
+        if value == {}:
+            return '{}'
+        items = [
+            nlch + repr(key) + ': ' +
+            ppretty(value[key], htchar, lfchar, indent, current_depth + 1)
+            for key in value
+        ]
+        return '{%s}' % (','.join(items) + lfchar +
```
vkarak: If you decide using `repr` and not `str`, you should use `%r` in format strings.

```suggestion
        return '{%r}' % (','.join(items) + lfchar +
```

reframe/core/config.py
```diff
@@ -217,3 +220,151 @@ def create_env(system, partition, name):
...
+                         htchar * indent * current_depth)
+    elif type(value) is list:
```
vkarak:
```suggestion
    elif isinstance(value, collections.abc.Sequence):
```

reframe/core/config.py
```diff
@@ -217,3 +220,151 @@ def create_env(system, partition, name):
...
+        if value == []:
+            return '[]'
+        items = [
+            nlch + ppretty(item, htchar, lfchar, indent, current_depth + 1)
+            for item in value
+        ]
+        return '[%s]' % (','.join(items) + lfchar +
+                         htchar * indent * current_depth)
+    elif type(value) is tuple:
```
vkarak:
```suggestion
    elif isinstance(value, tuple):
```

And put this check first, before the check for the `Sequence`.

reframe/core/config.py
```diff
@@ -217,3 +220,151 @@ def create_env(system, partition, name):
...
+        old_modes = old_config.site_configuration['modes'].items()
+        for target_mode, mode_entries in old_modes:
+            for mname, m in mode_entries.items():
+                new_mode = {'name': mname, 'options': m}
+                if target_mode != '*':
+                    new_mode['target_systems'] = [target_mode]
+
+                converted['modes'].append(new_mode)
+
+    def update_config(log_name, original_log):
```
vkarak:
```suggestion
    def update_logging_config(log_name, original_log):
```

reframe/core/config.py
```diff
@@ -217,3 +220,151 @@ def create_env(system, partition, name):
...
+        converted['general'][0][
+            'check_search_recursive'
+        ] = old_config.checks_path_recurse
+
+    if converted['general'] == [{}]:
+        del converted['general']
+
+    with tempfile.NamedTemporaryFile(mode='w', delete=False) as fp:
+        fp.write('#\n# This file was automatically generated '
+                 'by ReFrame based on `{}\'.\n#\n\n'.format(filename))
```
vkarak: We are moving to Python 3.6, so use an f-string here:

```suggestion
                 f"by ReFrame based on `{filename}'.\n#\n\n")
```

reframe/core/config.py
```diff
@@ -217,3 +220,151 @@ def create_env(system, partition, name):
...
+        fp.write('site_configuration = ')
+        fp.write(ppretty(converted))
+        fp.write('\n')
```
vkarak:
```suggestion
        fp.write(f'site_configuration = {ppretty(converted)}\n')
```

reframe/core/config.py
ekouts: The problem is that strings are also recognized as a `collections.abc.Sequence`, which leads to an infinite loop. So I think I should leave it as
```suggestion
    elif isinstance(value, list):
```

vkarak: Ok, so keep your original checks everywhere (dict, list, tuple) for consistency. You should also have a case for sets.

vkarak:
@ekouts Don't write a unit tests for the conversion. I will include one in my upcoming PR for the new configuration mechanism. Just move your pretty print function to a `utility` and unit test that one only.

vkarak:
Moving over to the next sprint.

reframe/core/config.py
```diff
@@ -217,3 +223,160 @@ def create_env(system, partition, name):
...
+                sys_dict['partitions'].append(new_p)
+
+        converted['systems'].append(sys_dict)
+
+    old_environs = old_config.site_configuration['environments'].items()
+    for env_target, env_entries in old_environs:
+        for ename, e in env_entries.items():
+            new_env = {'name': ename}
+            if env_target != '*':
+                new_env['target_systems'] = [env_target]
```
vkarak: The new environments must not have the `type`, so the validation fails:

```
Failed validating 'additionalProperties' in schema['properties']['environments']['items']:
    {'additionalProperties': False,
     'properties': {'cc': {'type': 'string'},
                    'cflags': {'items': {'type': 'string'},
                               'type': 'array'},
                    'cppflags': {'items': {'type': 'string'},
                                 'type': 'array'},
                    'cxx': {'type': 'string'},
                    'cxxflags': {'items': {'type': 'string'},
                                 'type': 'array'},
                    'fflags': {'items': {'type': 'string'},
                               'type': 'array'},
                    'ftn': {'type': 'string'},
                    'ldflags': {'items': {'type': 'string'},
                                'type': 'array'},
                    'modules': {'$ref': '#/defs/modules_list'},
                    'name': {'type': 'string'},
                    'target_systems': {'$ref': '#/defs/system_ref'},
                    'variables': {'$ref': '#/defs/envvar_list'}},
     'required': ['name'],
     'type': 'object'}

On instance['environments'][9]:
    {'cc': 'pgcc',
     'cxx': 'pgc++',
     'ftn': 'pgf90',
     'modules': ['PrgEnv-pgi/19.9'],
     'name': 'PrgEnv-pgi-nompi',
     'target_systems': ['arolla'],
     'type': 'ProgEnvironment'}
```

reframe/core/config.py
ekouts: What should I print in the case of an empty set? `set()`?

vkarak: Yes.

reframe/utility/__init__.py
```diff
@@ -126,6 +126,61 @@ def toalphanum(s):
...
+            return 'set()'
+
+        items = [
+            nlch + ppretty(item, htchar, lfchar, ind, basic_offset + 1)
+            for item in value
+        ]
+        return '{%s}' % (','.join(items) + lfchar +
+                         htchar * ind * basic_offset)
+    else:
+        return repr(value)
```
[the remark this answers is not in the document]
ekouts: I think it is correct as it is for the way we are using it, to write in the configuration file. Now we get:
```
assert util.ppretty(1) == '1'
assert util.ppretty(1.2) == '1.2'
assert util.ppretty('a string') == "'a string'"
assert (util.ppretty(['a string', 2, 'another string']) ==
        "[\n"
        "    'a string',\n"
        "    2,\n"
        "    'another string'\n"
        "]")
```
If we change the `str` function in the last return statement, we would get:
```
assert util.ppretty(1) == '1'
assert util.ppretty(1.2) == '1.2'
assert util.ppretty('a string') == 'a string'
assert (util.ppretty(['a string', 2, 'another string']) ==
        "[\n"
        "    a string,\n"
        "    2,\n"
        "    another string\n"
        "]")
```
And if we use `%r` everywhere else:
```
assert util.ppretty(1) == '1'
assert util.ppretty(1.2) == '1.2'
assert util.ppretty('a string') == "'a string'"
assert util.ppretty(['a string']) == '["\\n    \'a string\'\\n"]'
```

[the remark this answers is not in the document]
vkarak: Ok, I see it. You are right. The terminal nodes must use repr. Just make sure this is true in the current form.

PR #1191, opened 2020-02-21
[test] Temporarily fix HPCG reference version

sebkelle1:


cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
```diff
@@ -25,6 +25,9 @@ def __init__(self):
         self.build_system.options = ['arch=MPI_GCC_OMP']
         self.sourcesdir = 'https://github.com/hpcg-benchmark/hpcg.git'
 
+        # FIXME: Remove this after the OpenMP pragma gets fixed in hpcg master
+        self.prebuild_cmd = ['git checkout 9484cd7f2c4744c783abbdcfd4f5cc34807b42b1']
```
vkarak: Don't they have versions? If they do, I'd prefer to have a version number here.

cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
```diff
@@ -25,6 +25,9 @@ def __init__(self):
...
+
```
vkarak: Remove this line, too.

cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
vkarak: Since it's a workaround we can keep the hash.

PR #1192, opened 2020-02-24
[misc] Use `check_job_completion_time` with timezone info for performance logging in all settings files

ekouts:
Use `check_job_completion_time` with timezone info for performance logging formatting instead of `asctime` attribute.

The changes are applied in the cscs configuration files, the example and the tutorial configuration.

Fixes UES-709.

### DIFF
--- a/config/cscs-ci.py
+++ b/config/cscs-ci.py
@@ -172,7 +172,7 @@ class ReframeSettings:
                 'prefix': '%(check_system)s/%(check_partition)s',
                 'level': 'INFO',
                 'format': (
-                    '%(asctime)s|reframe %(version)s|'
+                    '%(check_job_completion_time)s|reframe %(version)s|'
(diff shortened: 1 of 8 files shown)


ekouts:
> I think it also need to be changed here:
> 
> https://github.com/eth-cscs/reframe/blob/423320ebf9c08d57e51e431f1bc6da7d3a8c5ae3/unittests/resources/settings.py#L157
> 
> and
> 
> https://github.com/eth-cscs/reframe/blob/423320ebf9c08d57e51e431f1bc6da7d3a8c5ae3/schemas/settings.py#L169

I wasn't sure whether we want to change the unittests' configuration, but I changed it. Maybe @vkarak can confirm that we want this change in the unittests as well.

vkarak:
@ekouts, Yes everything, everywhere. These were just reduced versions of the original one.

vkarak:
@teojgo Can you approves this one?

PR #1193, opened 2020-02-24
[bugfix] Do not leave stale test jobs when ReFrame receives a TERM signal

### LINKED ISSUE #1143: ReFrame does not cancel the spawned test jobs if it receives a TERM signal
When Jenkins kills a job it sends the TERM signal. This will obviously cause ReFrame to terminate, but it will leave stale test jobs behind. We should treat it the same way we treat keyboard interrupts.

teojgo:
* Allow ReFrame to exit with grace when it receives a SIGTERM.

* Add unittest to simulate a SIGTERM handle.

Fixes #1143


unittests/test_policies.py
```diff
@@ -7,8 +7,11 @@
 import itertools
 import os
 import pytest
+import signal
+import time
 import tempfile
 import unittest
+from multiprocessing import Pipe, Process
```
vkarak:
```suggestion
import multiprocessing
```
And use fully qualified names later on.

unittests/test_policies.py
```diff
@@ -257,6 +261,30 @@ def test_dependencies(self):
             if t.ref_count == 0:
                 assert os.path.exists(os.path.join(check.outputdir, 'out.txt'))
 
+    def test_sigterm(self):
+        # Wrapper of self.runall which is used from a child process and
+        # passes the thrown exception to the parent process
```
vkarak:
```suggestion
        # passes any exception to the parent process
```

unittests/test_policies.py
```diff
@@ -257,6 +261,30 @@ def test_dependencies(self):
...
+        def _runall(checks, conn):
+            ex = None
+            try:
+                self.runall(checks)
+            except ReframeForceExitError as e:
```
vkarak:
```suggestion
            except BaseException as e:
```

unittests/test_policies.py
```diff
@@ -257,6 +261,30 @@ def test_dependencies(self):
             if t.ref_count == 0:
                 assert os.path.exists(os.path.join(check.outputdir, 'out.txt'))
 
+    def test_sigterm(self):
+        # Wrapper of self.runall which is used from a child process and
+        # passes the thrown exception to the parent process
+        def _runall(checks, conn):
+            ex = None
```
vkarak:
```suggestion
            exc = None
```

unittests/test_policies.py
```diff
@@ -257,6 +261,30 @@ def test_dependencies(self):
...
+        with pytest.raises(ReframeForceExitError, match='Received SIGTERM'):
+            parent_conn, child_conn = Pipe(False)
+            p = Process(target=_runall, args=([SleepCheck(3)], child_conn))
+            p.start()
+            child_conn.close()
+            # Allow some time so that the SleepCheck is submitted
+            time.sleep(0.5)
+            os.kill(p.pid, signal.SIGTERM)
+            p.join(5)
+            raise parent_conn.recv()
```
vkarak: What happens if no exception is thrown in the child?

unittests/test_policies.py
```diff
@@ -257,6 +261,30 @@ def test_dependencies(self):
...
+            try:
+                self.runall(checks)
+            except ReframeForceExitError as e:
+                ex = e
+
+            conn.send(ex)
+            conn.close()
+
+        with pytest.raises(ReframeForceExitError, match='Received SIGTERM'):
+            parent_conn, child_conn = Pipe(False)
```
vkarak: Is there a way to use a `Pipe` with a `with` statement? According to the documentation, `Connection` objects are context managers.

unittests/test_policies.py
```diff
@@ -257,6 +261,30 @@ def test_dependencies(self):
...
+            p = Process(target=_runall, args=([SleepCheck(3)], child_conn))
+            p.start()
+            child_conn.close()
+            # Allow some time so that the SleepCheck is submitted
+            time.sleep(0.5)
+            os.kill(p.pid, signal.SIGTERM)
+            p.join(5)
```
vkarak: Why a timeout here?

unittests/test_policies.py
vkarak:
```suggestion
            rd_endpoint, wr_endpoint = Pipe(duplex=False)
```

unittests/test_policies.py
```diff
@@ -257,6 +260,38 @@ def test_dependencies(self):
...
+
+        with pytest.raises(ReframeForceExitError,
+                           match='received TERM signal'):
+            rd_endpoint, wr_endpoint = multiprocessing.Pipe(False)
+            p = multiprocessing.Process(target=_runall,
+                                        args=([SleepCheck(3)], wr_endpoint))
+            p.start()
+
+            # The unused write endpoint has to be closed from the parent to
+            # ensure that the `recv` method of `rd_endpoint` returns
```
vkarak:
```suggestion
            # ensure that the `recv()` method of `rd_endpoint` returns
```

unittests/test_policies.py
```diff
@@ -257,6 +260,38 @@ def test_dependencies(self):
...
+                self.runall(checks)
+            except BaseException as e:
+                exc = e
+
+            conn.send(exc)
+            conn.close()
+
+        with pytest.raises(ReframeForceExitError,
+                           match='received TERM signal'):
+            rd_endpoint, wr_endpoint = multiprocessing.Pipe(False)
```
vkarak: Pass the `duplex` keyword so that the code is more self explanatory:

```suggestion
            rd_endpoint, wr_endpoint = multiprocessing.Pipe(duplex=False)
```

unittests/test_policies.py
vkarak:
```suggestion
            # ensure that the `recv()` method of `rd_endpoint` returns
```

unittests/test_policies.py
vkarak: Pass the `duplex` keyword so that the code is more self explanatory:

```suggestion
            rd_endpoint, wr_endpoint = multiprocessing.Pipe(duplex=False)
```

vkarak: Pass the `duplex` keyword so that the code is more self explanatory:

```suggestion
            rd_endpoint, wr_endpoint = multiprocessing.Pipe(duplex=False)
```

vkarak:
@teojgo Strangely CI has finished successfully, but here it appears as still running.

vkarak:
@teojgo The test fails on Kesch. Moving over to the next sprint.

vkarak:
@jenkins-cscs retry kesch

unittests/test_policies.py
```diff
@@ -257,6 +260,38 @@ def test_dependencies(self):
...
+
+            # The unused write endpoint has to be closed from the parent to
+            # ensure that the `recv()` method of `rd_endpoint` returns
+            wr_endpoint.close()
+
+            # Allow some time so that the SleepCheck is submitted
+            time.sleep(0.5)
+            p.terminate()
+            p.join()
+            exc = rd_endpoint.recv()
```
vkarak: Move the `with pytest.raises()` here.

vkarak: And you should also test that the test is canceled.

unittests/test_policies.py
```diff
@@ -257,6 +260,38 @@ def test_dependencies(self):
...
+            p = multiprocessing.Process(target=_runall,
+                                        args=([SleepCheck(3)], wr_endpoint))
+            p.start()
+
+            # The unused write endpoint has to be closed from the parent to
+            # ensure that the `recv()` method of `rd_endpoint` returns
+            wr_endpoint.close()
+
+            # Allow some time so that the SleepCheck is submitted
+            time.sleep(0.5)
```
vkarak: Perhaps increase this to 1.

unittests/test_policies.py
```diff
@@ -273,8 +273,6 @@ def _runall(checks, conn):
             conn.send(exc)
             conn.close()
 
-        with pytest.raises(ReframeForceExitError,
-                           match='received TERM signal'):
             rd_endpoint, wr_endpoint = multiprocessing.Pipe(duplex=False)
```
vkarak: Indent out this.

PR #1199, opened 2020-02-26
[test] Fix deprecated `time_limit` syntax in checks

teojgo:


cscs-checks/apps/cpmd/cpmd_check.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, scale):
             self.valid_systems += ['dom:gpu']
         else:
             self.num_tasks = 16
-            self.time_limit = (0, 20, 0)
```
vkarak: Is this a realistic time limit? It looks like a workaround. Can you check please the execution time and then possibly remove this completely?

cscs-checks/microbenchmarks/spec-accel/spec.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, prg_envs):
 
         self.num_tasks = 1
         self.num_tasks_per_node = 1
-        self.time_limit = (0, 30, 0)
+        self.time_limit = '30m'
```
vkarak: Same here. I don't think they run for so long.

vkarak:
Can you please check also the CI failures?

cscs-checks/microbenchmarks/spec-accel/spec.py
teojgo: @vkarak this one needs this timelimit since the run takes longer.

vkarak:
@jenkins-cscs retry none

PR #1200, opened 2020-02-26
[test] Fix misleading variant description in the affinity check

mand35:


teojgo:
ok to test

mand35:
Thanks @teojgo for the hint. change applied.

Issue #1159, opened 2020-02-12
Parametrizing tests based on module discovery


mboisson:
It would be a really great feature if ReFrame could parametrize tests based on the available modules. For example, something like this : 

```
@rfm.parameterized_test(*utility.all_modules('gromacs',min_version,max_version))
```

vkarak:
An optional callable argument that would allow custom version comparisons would be useful, I guess, too. We don't want to implement all version conventions.

davidedelvento:
Yes, this would be fantastic!!!

For sites running lmod, this could be easily and relatively robustly implemented using [spider json format](https://lmod.readthedocs.io/en/latest/136_spider.html)

For everyone else, I guess the output of `module available` will need to be parsed. I've already implemented such a parser for our system., it's just 60 lines of (dense) python code. If it can be of any help, I'll be happy to clean it up, improve the unit tests I have written (which are not well done) and contribute it to this project.

vkarak:
> For sites running lmod, this could be easily and relatively robustly implemented using spider json format

Exactly!

> If it can be of any help, I'll be happy to un-densify it, make it clearer for other to read, improve the unit tests I have written (which are not well done) and contribute it to this project.

Contributions are always welcome ;-)

vkarak:
This one turns out to be less straightforward than it seems. We should answer what are the modules over which we want to parametrize? The decorator will run at test loading, before even the test is filtered per partition and programming environment. If you have the same package built for arch A and B, which is exposed by loading different base modules, then for which of the two would you parametrize the test? You would end up having different tests for each programming environment (compiler/toolchain) as well as target system. @victorusu who's working on that may have more details.

victorusu:
I started to implement this feature, but because decorators are executed before the class instantiation. I have two questions/difficulties:

1. Qhat happens if one needs to load a module before having access to the module in question? e.g. at CSCS in order to load GROMACS one needs to 
```
module load daint-gpu # or module load daint-mc
module load GROMACS
```
ReFrame is aware of those `daint-gpu` and `daint-mc` modules but at the virtual partition level.


2. What happens if the test needs to preload the programming environment before accessing the module in question? e.g. if one uses hierarchical module schemes, then one might need to first load the compiler, then load the MPI implementation, etc before the module becomes available.



Can you guys give me suggestions?

mboisson:
Can't we rely on module spider ? It tells you what you need to load before getting the module

victorusu:
[...]

PR #1203, opened 2020-02-27
[wip] [feat] Module helper

davidedelvento:
**Contex**
Per discussion on #1159 
Not adding [feat] to the PR title since there is no feature just a helper file which (hopefully) will later contribute to the feature implementation described in #1159 

**Actual PR description**
This is what NCAR uses in house, with minimal cleaning, for the very reason of #1159. Before spending more time on it to clean it up and integrating into ReFrame, I wanted to get some feedback.

Note that there is one skipped test (empty, mostly a reminder of something we might want to add later) and a failed one. The one that fails is there to demonstrate the assumptions that the method makes which may cause unexpected behavior. I think that rather than fixing the parsing for all possible edge cases, it'd be better, easier and more robust if the users avoid combinations of modules which cause them (see the test case for detail, in my opinion is very clear what is happening and why). The failing test is there just to grab your attention and talk about options. I am pretty sure before merging this test needs at least to be marked as skipped.

Besides that, not sure what you'd prefer to do, since this is a standalone file dropped into the tree, which at the moment has no connection with the rest of the framework. Perhaps if there's a branch were #1159 is being worked on, it may be better merged there?

### DIFF
--- a/reframe/utility/lmod_helper.py
+++ b/reframe/utility/lmod_helper.py
@@ -1,0 +1,67 @@
+import subprocess
+
+class PopenCommunicate():
+    def exec(self, command):
+        proc=subprocess.Popen(command,
+                              shell = True,
+                              stderr=subprocess.STDOUT,
+                              stdout=subprocess.PIPE)
+        return str(proc.communicate()[0]) # None-terminated, removed None
+
+
+def _all_avail_mods(thing_to_load, ml_av_output):
+    all_versions = []
+    look_for =  thing_to_load + "/"
+    for item in ml_av_output.replace("\n", " ").split(" "):
+        if item.strip().startswith(look_for):
+            all_versions.append(item.strip().split("/")[1])
+
+    return all_versions
+
+def _loop_among_one(module_glob, pc, env_context=""):
+    glob_to_load = module_glob.split()[-1]
+    if not glob_to_load.endswith("/*"):
+        raise Exception("Environment bit '" + module_glob +
+                        "' is incorrect. Must be like 'module load gnu/*'")
+
+    glob_to_load = glob_to_load[0:len(glob_to_load)-2]
+
+    ml_av_out = pc.exec(env_context + "module avail " + glob_to_load)
+    all_versions = _all_avail_mods(glob_to_load, ml_av_out)
+    all_mods = []
+    for a_version in all_versions:
+        all_mods.append(module_glob.replace("*", a_version))
+    return all_mods
+
+def _expand_module_set(module_set, module_glob, pc):
+    new_module_set = []
+    if len(module_set) == 0:
+        new_module_set = _loop_among_one(module_glob, pc)
+    else:
+        # combinatorial explosion of all that is found
+        for item in module_set:
+            for new_module in _loop_among_one(module_glob, pc, item + "; "):
+                new_module_set.append(item + "; " + new_module)
+    return new_module_set
+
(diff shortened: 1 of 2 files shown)


reframe/utility/lmod_helper.py
```diff
@@ -0,0 +1,67 @@
+import subprocess
+
+class PopenCommunicate():
```
victorusu, line 3: It would be great if this code could have a tighter integration with ReFrame.
The `PopenCommunicate()` should not be needed since ReFrame provides methods to run [external processes](
https://github.com/eth-cscs/reframe/blob/c9e3406c044843f570223f76d313912cc27795cc/reframe/utility/os_ext.py#L30).
Having said that, I am in the opinion that one needs to use the Python bindings for Lmod, which is the way that all module systems are implemented in ReFrame (https://github.com/eth-cscs/reframe/blob/master/reframe/core/modules.py). So, this class should be dropped.

reframe/utility/lmod_helper.py
```diff
@@ -0,0 +1,67 @@
...
+    def exec(self, command):
+        proc=subprocess.Popen(command,
+                              shell = True,
+                              stderr=subprocess.STDOUT,
+                              stdout=subprocess.PIPE)
+        return str(proc.communicate()[0]) # None-terminated, removed None
+
+
+def _all_avail_mods(thing_to_load, ml_av_output):
```
victorusu, line 12: I think an extended version of this function should go to the Lmod backend. But it does require an extension of the [Modules' System](https://github.com/eth-cscs/reframe/blob/c9e3406c044843f570223f76d313912cc27795cc/reframe/core/modules.py#L80).

reframe/utility/lmod_helper.py
```diff
@@ -0,0 +1,67 @@
...
+    all_versions = []
+    look_for =  thing_to_load + "/"
+    for item in ml_av_output.replace("\n", " ").split(" "):
+        if item.strip().startswith(look_for):
+            all_versions.append(item.strip().split("/")[1])
+
+    return all_versions
+
+def _loop_among_one(module_glob, pc, env_context=""):
```
victorusu, line 21: I also think this should go to the Lmod backend. 
The python bindings for Lmod are preferred over calling `ml avail`.
And the function name should be slightly improved to better explain what it does.

reframe/utility/lmod_helper.py
davidedelvento: Hi @victorusu 
Thanks for the suggestions. I see your points. In fact, I pushed here what I had done in another context, so it's no surprise that it looks a bit like duct-taped ;-)

I'll try to get to your suggestions as soon as I have time.
Cheers,
Davide

davidedelvento:
I haven't forgotten about this, quite the contrary. Just gotten caught in more urgent things. Hope to find the time for this (and overall reframe work) in the not-too-far future...

vkarak:
@davidedelvento Nice. It will be much easier to merge this if it targets specifically #1159. For that we'd need simply to extend the module systems API of ReFrame (and accordingly its backends) to obtain a list of module names based on a given pattern. Each backend may implement that in its own way. We'd also need a utility function (pretty simple) that would expose that functionality in a convenient way to be used in a paremeterized test. Such a PR would be much simpler than this one and would make it easier to merge upstream.

vkarak:
This is replaced by #1417.

PR #1207, opened 2020-03-03
[test] Implement workaround for PETSc test when using PrgEnv-intel and static linking

ajocksch:
for static linking the C++ compiler is used until Cray fixes the problem

Cray Bug #255701


vkarak:
@jenkins-cscs retry trsa

PR #1211, opened 2020-03-03
[test] Correct the test for number of nodes in DGEMM

victorusu:
Currently, the number of nodes test is not being evaluated a run time due to the fact that it is a lazily evaluated function. 
Therefore, I am adding a call to force the evaluation at sanity time.


vkarak:
@jenkins-cscs retry none

PR #1214, opened 2020-03-04
[bugfix] Fix crash when creating a regression test interactively

### LINKED ISSUE #1194: ReFrame crashes when trying to create a regression test interactively
```
>>> import reframe as rfm
>>>
>>> @rfm.simple_test
... class MyTest(rfm.RegressionTest):
...     def __init__(self):
...             pass
...
>>> t = MyTest()
  File "<stdin>", line 1, in <module>
  File "/Users/karakasv/Repositories/reframe/reframe/core/pipeline.py", line 665, in __new__
    os.path.abspath(os.path.dirname(inspect.getfile(cls))))
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/inspect.py", line 653, in getfile
    raise TypeError('{!r} is a built-in class'.format(object))
TypeError: <module '__main__' (built-in)> is a built-in class
```

The problem is that `inspect.getfile()` raises a `TypeError` if invoked from an interactive shell.

teojgo:
* Also create `is_interactive` function in `os_ext` to check whether
  the session is interactive.

* Add unittest for the above function.

Fixes #1194

### DIFF
--- a/reframe/utility/os_ext.py
+++ b/reframe/utility/os_ext.py
@@ -229,7 +229,7 @@ def inpath(entry, pathvar):
 
 def is_interactive():
     '''Returns whether the given Python session is interactive'''
-    return bool(getattr(sys, 'ps1', sys.flags.interactive))
+    return hasattr(sys, 'ps1') or sys.flags.interactive
 
 
 def subdirs(dirname, recurse=False):
--- a/reframe/utility/os_ext.py
+++ b/reframe/utility/os_ext.py
@@ -227,6 +227,11 @@ def inpath(entry, pathvar):
     return entry in set(pathvar.split(':'))
 
 
+def is_interactive():
+    '''Returns whether the given Python session is interactive'''
+    return bool(getattr(sys, 'ps1', sys.flags.interactive))
+
+
 def subdirs(dirname, recurse=False):
     '''Returns a list of dirname + its subdirectories. If recurse is True,
     recursion is performed in pre-order.'''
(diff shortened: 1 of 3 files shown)


unittests/test_utility.py
```diff
@@ -173,6 +173,11 @@ def test_samefile(self):
                                os.path.join(prefix, 'broken1'))
         shutil.rmtree(prefix)
 
+    def test_is_interactive(self):
+        # Set `sys.ps1` to immitate an interactive session
+        sys.ps1 = 'rfm>>> '
+        assert os_ext.is_interactive()
```
vkarak: You should restore `sys.ps1` to its original value or delete the attribute if it didn't exist. You might break other tests otherwise!

reframe/utility/os_ext.py
```diff
@@ -227,6 +227,11 @@ def inpath(entry, pathvar):
     return entry in set(pathvar.split(':'))
 
 
+def is_interactive():
+    '''Returns whether the given Python session is interactive'''
+    return bool(getattr(sys, 'ps1', sys.flags.interactive))
```
vkarak: In the extreme case that `sys.ps1 == ''`, this will fail. I suggest doing it as follows:

```suggestion
    return hasattr(sys, 'ps1') or sys.flags.interactive
```

unittests/test_utility.py
vkarak: To be complete you should also check the case where `sys.flags.interactive` is set to True. Again you should restore the values after you do your test.

teojgo: The `sys.flags.interactive` is readonly and cannot be adjusted. It's value is set based on the `-i` command line option when running `python`.

teojgo:
@jenkins-cscs retry all

PR #1220, opened 2020-03-06
[bugfix] Improve unit tests for the handling of the TERM signal

### LINKED ISSUE #1216: Unit test `TestAsynchronousExecutionPolicy.test_sigterm` fails randomly on master
I think the unit test is not so robust. We should definitely revise it asap. Check here:

https://travis-ci.org/eth-cscs/reframe/branches?utm_medium=notification&utm_source=github_status

teojgo:
* Increase the time of the actual submitted SleepTest.

* Use `multiprocessing.Manager` and communicate with `Namespace`
  objects of the `Manager` instance between parent-child processes.

Fixes #1216

### DIFF
--- a/unittests/test_policies.py
+++ b/unittests/test_policies.py
@@ -288,17 +288,14 @@ class TestSerialExecutionPolicy(unittest.TestCase):
             time.sleep(0.2)
             p.terminate()
             p.join()
-            exc = ns.exc
-            num_cases = ns.num_cases
-            num_failures = ns.num_failures
 
-        # Either the test is submitted and it fails due to the termination
-        # or it is not yet submitted when the termination signal is sent
-        assert (num_cases, num_failures) in {(1, 1), (0, 0)}
-        with pytest.raises(ReframeForceExitError,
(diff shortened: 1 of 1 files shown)


unittests/test_policies.py
```diff
@@ -262,32 +262,35 @@ def test_dependencies(self):
...
+
+            # Allow some time so that the SleepCheck is submitted.
+            # The sleep time of the submitted test is much larger to
+            # ensure that it does not finish before the termination signal
+            time.sleep(0.2)
+            p.terminate()
+            p.join()
+            exc = ns.exc
+            num_cases = ns.num_cases
+            num_failures = ns.num_failures
```
vkarak: Remove these three lines and work directly with the namespace object in the following.

teojgo: These fields are not accessible outside of the context of the manager that created the `Namespace` object. I will put the `assert` inside the `with` as discussed.

vkarak:
@jenkins-cscs retry tsa

PR #1227, opened 2020-03-11
[bugfix] Make unit test for TERM signal handling more robust

teojgo:
* Explicitly call `os.kill` from a regression check.

Related to #1217

### DIFF
--- a/unittests/resources/checks_unlisted/selfkill.py
+++ b/unittests/resources/checks_unlisted/selfkill.py
@@ -8,6 +8,7 @@
 #
 import os
 import signal
+import time
 
 import reframe as rfm
 import reframe.utility.sanity as sn
@@ -24,6 +25,7 @@ class SelfKillCheck(rfm.RunOnlyRegressionTest):
         self.tags = {type(self).__name__}
         self.maintainers = ['TM']
 
-    @rfm.run_after('run')
-    def kill(self):
+    def run(self):
+        super().run()
+        time.sleep(0.5)
         os.kill(os.getpid(), signal.SIGTERM)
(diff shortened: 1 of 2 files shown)


unittests/resources/checks_unlisted/selfkill.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self):
+        self.local = True
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
+        self.executable = 'echo hello'
+        self.sanity_patterns = sn.assert_found('hello', self.stdout)
+        self.tags = {type(self).__name__}
+        self.maintainers = ['TM']
+
+    @rfm.run_after('run')
```
vkarak: This will run after the associated job has finished. That's why you have to override the `run()` method in this case:

```python
def run(self):
    super().run()
    os.kill(os.getpid(), signal.SIGTERM)
```

vkarak: I would also make the test a "sleep check", and wait here a bit before committing suicide.

unittests/test_policies.py
```diff
@@ -261,41 +261,16 @@ def test_dependencies(self):
...
+        self.loader = RegressionCheckLoader(
+            ['unittests/resources/checks_unlisted/selfkill.py']
+        )
+        checks = self.loader.load_all()
+        with pytest.raises(ReframeForceExitError,
+                           match='received TERM signal'):
+            self.runall(checks)
+
+        assert self.runner.stats.num_cases() == 1
+        assert len(self.runner.stats.failures()) == 1
```
vkarak: Here you have also to assert that the associated job has indeed finished.

unittests/test_policies.py
```diff
@@ -261,41 +261,16 @@ def test_dependencies(self):
...
-            try:
-                self.runall(checks)
-            except BaseException as e:
-                exc = e
-            finally:
-                ns.exc = exc
-                ns.num_cases = self.runner.stats.num_cases()
-                ns.num_failures = len(self.runner.stats.failures())
-
-        with multiprocessing.Manager() as manager:
```
vkarak: Make sure to remove any unnecessary imports you have added.

PR #1228, opened 2020-03-11
[feat] Deprecate test syntax that overrides `RegressionTest` methods

### LINKED ISSUE #1130: Deprecate test syntax that overrides `RegressionTest` methods

ekouts:
This feature is introduced to discourage users to override `RegressionTest` methods and use the hooks instead. If it is necessary, the user can still override the original methods but he has to use a new decorator that is also introduced in this pr.

Fixes #1130 

Things to be done:
- [x] implement feature
- [x] unittests for the new feature
- [x] fixing the original unittests, so that they don't throw warnings
- [x] update documentation


vkarak:
Moving to next sprint; it's still a wip.

ekouts:

reframe/core/decorators.py
```diff
@@ -72,6 +74,42 @@ def _validate_test(cls):
...
+        else:
+            test_class = RegressionTest
+
+        for func in funcs:
+            if getattr(cls, func) != getattr(test_class, func):
+                msg = (f'Trying to override method {func}. The syntax '
+                       f'that overrides Regression test methods is '
+                       f'deprecated. Consider using the reframe hooks '
+                       f'instead or the decorator `extend_test`.')
+                user_deprecation_warning(msg)
```
vkarak: I'm not sure about this implementation. The things I don't like are that you have to list the "official" subclasses of `RegressionTest` here, the fact that you go arbitrarily in anything defined under `RegressionTest`, and that this check is triggered with the `_validate_test()`. We do need the `@extended_test` decorator, but I'm not entirely sure about the implementation. One thing I'm currently thinking is perhaps to decorate some of the internal functions as "final", meaning that you will not be able to override them unless you have explicitly decorated the class. But this check should better go in the `RegressionTestMeta`. These are just thoughts, I have not proof of concept.

reframe/core/decorators.py
```diff
@@ -72,6 +74,42 @@ def _validate_test(cls):
...
+
+
+def extend_test(cls):
+    '''Class decorator for allowing method overriding.
```
vkarak: We will need to elaborate more here.

unittests/resources/checks/frontend_checks.py
```diff
@@ -12,6 +12,7 @@
 from reframe.core.exceptions import ReframeError, PerformanceError
 
 
+@rfm.extend_test
```
vkarak: No need for extension in any of these tests. They should be adapted to use the hooks.

unittests/resources/checks_unlisted/extend_class.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@rfm.simple_test
+@rfm.extend_test
+class TestExtendedCompileOnly(rfm.CompileOnlyRegressionTest):
+    def __init__(self):
+        pass
+
+    def setup(self, partition, environ, **job_opts):
+        super().setup(system, environ, **job_opts)
```
vkarak: No need for this file. Create the tests inline inside the corresponding unit tests.

unittests/resources/checks_unlisted/good.py
```diff
@@ -31,6 +32,7 @@ def __repr__(self):
 
 
 @rfm.parameterized_test(*({'a': x, 'b': y} for x in range(3) for y in range(2)))
+@rfm.extend_test
```
vkarak: I guess this is needed for the `__eq__()`, right? That's why we need to mark final only specific framework methods.

unittests/resources/checks_unlisted/kbd_interrupt.py
```diff
@@ -14,6 +14,7 @@
 
 
 @rfm.simple_test
+@rfm.extend_test
```
vkarak: Again this should use hooks.

vkarak:
We will need to update the documentation as well.

reframe/core/decorators.py
vkarak: I have prepared a proof-of-concept that works nicely:

```python
import functools


class Meta(type):
    def __init__(cls, name, bases, namespace, **kwargs):
        super().__init__(name, bases, namespace, **kwargs)
        cls._final_methods = {v.__name__ for v in namespace.values()
                              if hasattr(v, '_final')}

        if hasattr(cls, '_is_special') and cls._is_special:
            return

        for v in namespace.values():
            for b in bases:
                if callable(v) and v.__name__ in b._final_methods:
                    print(f"'{cls.__qualname__}.{v.__name__}' attempts to override "
                          f"final method '{b.__qualname__}.{v.__name__}'")


def final(fn):
    fn._final = True

    @functools.wraps(fn)
    def _wrapped(*args, **kwargs):
        return fn(*args, **kwargs)

    return _wrapped


def special_test(cls):
    print('special_test deco')
    cls._special_test = True
    return cls


class C(metaclass=Meta):
    @classmethod
    def __init_subclass__(cls, *, special=False, **kwargs):
        super().__init_subclass__(**kwargs)
        cls._is_special = special

    def foo(self):
        pass

    @final
    def bar(self):
        pass

    @final
    def barx(self):
        pass


class D(C):
    def foo(self):
        pass

    def bar(self):
        pass


if __name__ == '__main__':
    pass
```

This will issue warnings for the class `D` since it tries to override final methods. However, if you define it as `class D(C, special=True):`, the warnings will be suppressed. In this POC we don't need the class decorator and in fact, the decorator cannot serve us, since it gets evaluated after the class has been constructed. The real reason for using `special=True` is to extend the functionality of the basic `RegressionTest`, for example, the `CompileOnlyRegressionTest` and the `RunOnlyRegressionTest` should do that. Users my use this when they want to create their own custom extensions or as a transitional measure for their old-style tests.

vkarak: Also an additional note. Any overrided method in a special test must be marked again as `final` in order for the warnings to be issued for its subclasses:

```python
class D(C, special=True):
    def foo(self):
        pass

    @final
    def bar(self):
        pass


class E(D):
    def bar(self):
        pass
```

ekouts: Thanks, I will have a look and change my implementation!

vkarak: As `Meta` you can use the `RegressionTestMeta` that we have already.

ekouts: I agree, this is a nicer solution, but we have to decide which are the final methods. The only not-so-nice thing I see is that all the final functions should be defined in `CompileOnlyRegressionTest` and `RunOnlyRegressionTest` even when they just call the super. For example `RunOnlyRegressionTest` should define a final `setup()` and call the `super().setup()`, so that the user gets a deprecation warning in case they override it.

In the commit I will push, I have only made final the methods that are part of the pipeline but maybe it makes sense to make final the "private" `_function_name` as final since the user might overwrite them by accident and change the behavior of the framework.

reframe/core/meta.py
```diff
@@ -33,3 +35,27 @@ def __init__(cls, name, bases, namespace, **kwargs):
...
+        # Add the final functions from its parents
+        cls._final_methods.update(*[b._final_methods for b in bases
+                                    if hasattr(b, '_final_methods')])
+
+        # Filter the methods that are defined in this class but are not final
+        for v in namespace.values():
+            if callable(v) and not hasattr(v, '_final'):
+                cls._final_methods.discard(v.__name__)
+
+        if hasattr(cls, '_extended_test') and cls._extended_test:
```
vkarak, line 51: Call this attribute `_rfm_special_test`.

reframe/core/meta.py
```diff
@@ -33,3 +35,27 @@ def __init__(cls, name, bases, namespace, **kwargs):
...
+            return
+
+        for v in namespace.values():
+            for b in bases:
+                if callable(v) and v.__name__ in b._final_methods:
+                    msg = (f"'{cls.__qualname__}.{v.__name__}' attempts to "
+                           f"override final method "
+                           f"'{b.__qualname__}.{v.__name__}'. "
+                           f"Consider using the reframe hooks instead.")
```
vkarak, line 60:
```suggestion
                           f"'{b.__qualname__}.{v.__name__}'; "
                           f"consider using the pipeline hooks instead")
```

reframe/core/pipeline.py
```diff
@@ -681,6 +691,11 @@ def __new__(cls, *args, **kwargs):
     def __init__(self):
         pass
 
+    @classmethod
+    def __init_subclass__(cls, *, extended_test=False, **kwargs):
```
vkarak, line 695: I think I like more the `special` keyword, since it's a test treated specially:

```suggestion
    def __init_subclass__(cls, *, special=False, **kwargs):
```

reframe/core/pipeline.py
```diff
@@ -681,6 +691,11 @@ def __new__(cls, *args, **kwargs):
...
+        super().__init_subclass__(**kwargs)
+        cls._extended_test = extended_test
```
vkarak, line 697:
```suggestion
        cls._rfm_special_test = special
```

reframe/core/meta.py
```diff
@@ -33,3 +35,27 @@ def __init__(cls, name, bases, namespace, **kwargs):
...
+                              if hasattr(v, '_final')}
+
+        # Add the final functions from its parents
+        cls._final_methods.update(*[b._final_methods for b in bases
+                                    if hasattr(b, '_final_methods')])
+
+        # Filter the methods that are defined in this class but are not final
+        for v in namespace.values():
+            if callable(v) and not hasattr(v, '_final'):
+                cls._final_methods.discard(v.__name__)
```
vkarak, line 49: Why this is needed? You want essentially to override the `final` keyword? I don't think we need that and it's not correct conceptually. Normally, the `final` keyword would have effect over all subclasses. We just want to insert an exception to that with the special test classes, so the exception should be true only for the current class and not for its subclasses.

reframe/core/pipeline.py
```diff
@@ -1487,26 +1511,29 @@ def __str__(self):
...
 
-class RunOnlyRegressionTest(RegressionTest):
+class RunOnlyRegressionTest(RegressionTest, extended_test=True):
     '''Base class for run-only regression tests.
 
     This class is also directly available under the top-level :mod:`reframe`
     module.
     '''
 
+    @final
```
vkarak, line 1521: Avoiding the discard above, only the final keyword at the base class will be needed.

unittests/test_loader.py
```diff
@@ -73,9 +74,88 @@ def test_load_bad_init(self):
...
-            tests = self.loader.load_from_file(
-                'unittests/resources/checks_unlisted/extend_class.py')
-
-        assert len(record) == 3
+            @rfm.simple_test
+            class TestSimple(rfm.RegressionTest):
+                # The test should not raise a deprecation warning even though
+                # it overrides __init__
+                def __init__(self):
+                    pass
```
vkarak, line 84: I would only wrap with `pytest.warns()` only those that we expect them to issue a warning.

unittests/test_loader.py
```diff
@@ -73,9 +74,88 @@ def test_load_bad_init(self):
...
+
+                # Should not raise a warning
+                def setup(self, partition, environ, **job_opts):
+                    super().setup(system, environ, **job_opts)
+
+                # Should not raise a warning
+                def run(self):
+                    super().run()
+
+        assert len(record) == 4
```
vkarak, line 161: This is not enough, because you don't have control over what was actually raised.

unittests/test_loader.py
```diff
@@ -73,9 +74,88 @@ def test_load_bad_init(self):
...
+
+            @rfm.simple_test
+            class TestDeprecated(rfm.RegressionTest):
+                # Should raise a warning
+                def setup(self, partition, environ, **job_opts):
+                    super().setup(system, environ, **job_opts)
+
+            @rfm.simple_test
+            class TestDeprecatedRunOnly(rfm.RunOnlyRegressionTest):
+                # Should raise a warning
```
vkarak, line 94: You will not need these comments if you use `pytest.warns()` for each tests that needs to issue a warning.

vkarak: Also, I think you need to test the final decorator as well.

reframe/core/meta.py
ekouts: I will remove it then. I did it so that we are able to override the behavior of final for special classes.

unittests/test_loader.py
```diff
@@ -75,41 +75,43 @@ def test_load_bad_init(self):
...
+        with pytest.warns(ReframeDeprecationWarning):
             @rfm.simple_test
             class TestDeprecatedCompileOnlyDerived(TestDeprecatedCompileOnly):
-                # Should not raise a warning because the setup of the parent
-                # was not set as final
+                # Should raise a warning
                 def setup(self, partition, environ, **job_opts):
                     super().setup(system, environ, **job_opts)
 
+        with pytest.warns(None) as warnings:
```
vkarak, line 105: I think that this is not needed. The test would fail in any case.

unittests/test_loader.py
```diff
@@ -75,41 +75,43 @@ def test_load_bad_init(self):
         assert 0 == len(tests)
 
     def test_extended_test(self):
```
vkarak, line 77:
```suggestion
    def test_special_test(self):
```

unittests/test_loader.py
```diff
@@ -75,41 +75,43 @@ def test_load_bad_init(self):
...
             @rfm.simple_test
-            class TestExtended(rfm.RegressionTest, extended_test=True):
+            class TestSimple(rfm.RegressionTest):
+                # The test should not raise a deprecation warning even though
+                # it overrides __init__
+                def __init__(self):
+                    pass
+
+            @rfm.simple_test
+            class TestExtended(rfm.RegressionTest, special=True):
```
vkarak, line 114:
```suggestion
            class TestSpecial(rfm.RegressionTest, special=True):
```

unittests/test_loader.py
```diff
@@ -118,21 +120,22 @@ def setup(self, partition, environ, **job_opts):
                     super().setup(system, environ, **job_opts)
 
             @rfm.simple_test
-            class TestExtendedDerived(TestExtended):
+            class TestExtendedRunOnly(rfm.RunOnlyRegressionTest,
```
vkarak, line 123:
```suggestion
            class TestSpecialRunOnly(rfm.RunOnlyRegressionTest,
```

unittests/test_loader.py
```diff
@@ -118,21 +120,22 @@ def setup(self, partition, environ, **job_opts):
...
 
-                # Should raise a warning
+                # Should not raise a warning
                 def run(self):
                     super().run()
 
             @rfm.simple_test
-            class TestExtendedRunOnly(rfm.RunOnlyRegressionTest,
-                                      extended_test=True):
+            class TestExtendedCompileOnly(rfm.CompileOnlyRegressionTest,
```
vkarak, line 137:
```suggestion
            class TestSpecialCompileOnly(rfm.CompileOnlyRegressionTest,
```

unittests/test_loader.py
```diff
@@ -144,18 +147,37 @@ def setup(self, partition, environ, **job_opts):
...
                     super().run()
 
+        assert not any(isinstance(w.message, ReframeDeprecationWarning)
+                       for w in warnings)
+
+        with pytest.warns(ReframeDeprecationWarning) as warnings:
             @rfm.simple_test
-            class TestExtendedCompileOnly(rfm.CompileOnlyRegressionTest,
-                                          extended_test=True):
+            class TestExtendedDerived(TestExtended):
```
vkarak, line 155:
```suggestion
            class TestSpecialDerived(TestExtended):
```

unittests/test_loader.py
ekouts: The test doesn't fail without it, the warnings only appear in the summary like this when I remove the special (...11 passed, 2 warnings in 0.16s...) I think it will look as if it is successful in jenkins, even if it wasn't working correctly

vkarak: Ok, that's fine then. Leave it as is.

PR #1229, opened 2020-03-12
[test] Move MpiP and VTune checks to the eth-cscs/hpctools repo

jgphpc:
moving mpip and vtune checks to https://hpctools.readthedocs.io


vkarak:
@jgphpc There is a conflict with the master here. Can you please fix it?

PR #1230, opened 2020-03-13
[test] Update GROMACS test

victorusu:


cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -73,20 +73,20 @@ def __init__(self, scale, variant):
         references = {
             'maint': {
                 'small': {
-                    'dom:gpu': {'perf': (29.3, -0.05, None, 'ns/day')},
-                    'daint:gpu': {'perf': (30.3, -0.10, None, 'ns/day')}
+                    'dom:gpu': {'perf': (0.0, -0.05, None, 'ns/day')},
```
teojgo: Why `0.0` here?

vkarak: @victorusu Why zeros here?

victorusu: Because we do not run the dom maintenance test ever

vkarak: We can simply remove it then. ReFrame automatically inserts a `*` reference for all unknown systems.

vkarak: This, it will, but it's better if you address that rather than add another reference, because then your test is not generic enough.

victorusu: Can you propose a solution?

vkarak: Yes, you can simply suppress the `KeyError` and don't define `self.reference` in that case. A very similar case is [this](https://github.com/eth-cscs/reframe/blob/255d9709ddeae2b14c7d6beac218e437a582ec8e/cscs-checks/cuda/multi_gpu.py#L129) one.

cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -73,20 +73,20 @@ def __init__(self, scale, variant):
...
+                    'daint:gpu': {'perf': (0.0, -0.10, None, 'ns/day')}
                 },
                 'large': {
-                    'daint:gpu': {'perf': (42.0, -0.10, None, 'ns/day')}
+                    'daint:gpu': {'perf': (0.0, -0.10, None, 'ns/day')}
```
vkarak, line 80: I can understand the comment about Dom, but not for this one. Why `0.0` also here?

vkarak:
@jenkins-cscs retry none

PR #1232, opened 2020-03-16
[test] Update P2P latency tests

ajocksch:
I oriented the reference values for the P2P OSU latency test on the values for daint and the last outputs from

/apps/dom/UES/jenscscs/regression/production/logs/dom/gpu/


vkarak:
@jenkins-cscs retry none

PR #1236, opened 2020-03-17
[test] osu alltoall removed kesch:pn

ajocksch:
The default MPI library requires GPU components which are not available on the post processing nodes of kesch


vkarak:
Closing. Duplicate of #1241.

PR #1237, opened 2020-03-18
[feat] Print test dependencies in output of `--list-detailed`

### LINKED ISSUE #1091: --list-detailed output does not mention test dependencies
It would be nice if the `-L` output also mentions test dependencies, now it doesn't (using current master `64bab9d`)

ekouts:
This is an example of the new expected output:

```
List of matched checks
======================
  * T0 (found in /path/to/reframe/unittests/resources/checks_unlisted/deps_complex.py)
      - description: T0
      - systems: *
      - environments: *
      - modules: 
      - task allocation: standard
      - dependencies:
      - tags: 
      - maintainers: 
  * T1 (found in /path/to/reframe/unittests/resources/checks_unlisted/deps_complex.py)
      - description: T1
      - systems: *
      - environments: *
      - modules: 
      - task allocation: standard
      - dependencies: T4, T5
      - tags: 
      - maintainers:
...
```

Fixes #1091


vkarak:
We could have another option for showing the full dependency tree, but that's a different issue.

PR #1239, opened 2020-03-24
[test] Remove unnecessary tags in the HaloExchangeTest 

kraushm:
Based on Jira issue https://jira.cscs.ch/browse/UES-772

### DIFF
--- a/cscs-checks/mch/collectives_halo.py
+++ b/cscs-checks/mch/collectives_halo.py
@@ -108,7 +108,7 @@ class CollectivesBaseTest(rfm.RegressionTest):
         }
 
         self.maintainers = ['AJ', 'MKr']
-        self.tags = {'production', 'mch', 'craype'}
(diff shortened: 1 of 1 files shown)


cscs-checks/mch/collectives_halo.py
```diff
@@ -108,7 +108,7 @@ def __init__(self, variant, bench_reference):
         }
 
         self.maintainers = ['AJ', 'MKr']
-        self.tags = {'production', 'mch', 'craype'}
+        self.tags = {'mch'}
```
vkarak: The tags should be removed only for Tsa/Arolla, right?

kraushm: Indeed, my bad.

vkarak:
@jenkins-cscs retry daint

vkarak:
@jenkins-cscs retry none

PR #1241, opened 2020-03-25
[test] Remove kesch:pn from supported systems of FlexAlltoallTest

lucamar:
Removing `kesch:pn` from `FlexAlltoallTest`, as suggested by @vkarak in https://jira.cscs.ch/browse/UES-740. 
Please note that if we want to keep it, I can just export the correct path for ` MV2_USE_GPUDIRECT_GDRCOPY` in the check, as I have reported in https://jira.cscs.ch/browse/UES-533.
As a matter of fact, `daint:mc`, `dom:mc`, `arolla:pn` and `tsa:pn` are also listed in the check.


kraushm:
I just ran the test on the RH7.7 partition (although with limited resources) in order to validate, whether the modified setup of the software stack (CUDA - non-CUDA) would have any impact. It passed, hence it seems ok from that perspective to keep the tests for `tsa:pn` and `arolla:pn`.

Issue #1157, opened 2020-02-11
Revisit the structure of the documentation


vkarak:
For example, logging's configuration is not addressed at all in "Configuring ReFrame For Your Site" section, but it is only addressed in "Running ReFrame". Also how the user navigates through the documentation should be rethought as well. There could be multiple levels of tutorials that will build gradually on more complex use cases, but we should also have complete references for the configuration and the command-line interface as well.

PR #1242, opened 2020-03-25
[feat] Introduce a new powerful configuration mechanism

### LINKED ISSUE #66: System and partition initialisation are random

### LINKED ISSUE #1115: Allow configuration to be modified from environment variables and the command line

### LINKED ISSUE #1116: Implement a configuration mechanism to support the new configuration syntax.
- Configuration file lookup. We want ReFrame to look into different standard locations for a configuration file.
- Update framework's internals and glue with the new configuration infrastructure.

vkarak:
# Summary of this PR

- Support for the new-style configuration files
  - Configuration files can be either Python or JSON
- All configuration files have been migrated to the new syntax.
- More configuration options
- Ability to associate command line options to environment and/or configuration variables.
- Configuration component is completely refactored
- Framework's runtime context component refactored and simplified
- Internal representations for system and system partitions adapted
- The bulk of unit tests using a modified runtime have been migrated to pytest using fixtures and CI scripts have been adapted accordingly.
- Better mechanism for registering scheduler and launcher backends that avoids spurious imports
- Free functions of the environments module that manipulate Environments have been moved to the runtime module, in order to avoid circular dependencies and ensure a better conceptual fit

# Changes of this PR from the user's point of view

- The new configuration syntax introduced in #1128 is now the default. The old configuration style is still accepted, but a deprecation warning is issued and it is converted to the new syntax in a temporary file as implemented by #1186.
- The new syntax allows users to define configuration parameters per system or per system/partition combination. However some configuration options, such as the modules to load before starting a regression test run, can only be treated at the system level, even if a user defines them at a finer level.
- Several command line options are associated with environment variables and/or configuration variables. There is a precedence order, such that, command-line options are considered first, environment variables second and, finally, configuration variables. See `cli.py` for all the associations.
- Configuration files can be either in JSON format or a Python file that defines a `site_configuration` variable which holds essentially the JSON configuration formatted in Python syntax.
- ReFrame now looks for configuration files in the following locations and in that order:

  1. `${HOME}/.reframe/settings.{py,json}`
  2. `${REFRAME_INSTALL_PREFIX}/settings.{py,json}`
  3. `/etc/reframe.d/settings.{py,json}`

  If both a `.py` and `.json` configuration file are found, the Python one will be preferred.

- If no configuration file is found or none is supplied, ReFrame will use a generic fallback configuration file, that is located in `reframe/core/settings.py` and will allow ReFrame to run minimally on any system. This is *NOT* meant to be altered by users. If the user supplies a configuration file, ReFrame will not fall back to the generic configuration. This is different from the current behavior.
- ReFrame starts always using the generic fallback configuration and then switches to the one provided by the user. This is for allowing ReFrame to enable logging right from the beginning, so that early messages are nicely formatted and colored.
- The very old logging configuration syntax (prior to 2.13) is no more recognized and the configuration conversion tool does not take it into account.
- The `--show-env-config` option is eliminated and the functionality of the `--show-config` option is expanded. You can use that option to query any configuration parameter of the framework. By default, without an argument, it will show the entire ReFrame configuration for the *current* system, but you can request a configuration parameter using the syntax described below for the `_SiteConfig.get()` method. The output of this command is in JSON forrmat.
- The default value for the different flags in build systems is now the empty list `[]` instead of `None`. The behavior of the default flags is still the same, though, such that no `XFLAGS` is emitted if flags' list is empty. If the user wants to have `XFLAGS` explicitly emitted even if empty, `['']` has to be specified as the flags' value.
- The default values for the different compiler name variables is the empty string `''` instead of `None`. This and the previous changes were made in order for the default values to be consistent with the JSON schema requirements.
- `ShifterNG` container backend is changed to `Shifter`.
- The changes of #1125 are reverted and `-c` option's behaviour is as before. The reason for that is that now the check search path can be completely overriden through an environment variable.
- Printing of "ReFrame paths" section has changed as well as how list of checks are printed. IMHO look more modern now.

# Implementation details

The configuration component has been redesigned entirely and implemented from ground up. All other components attached to it have also been profoundly changed and redesigned. Here are the core concepts of the new configuration mechanism implementation.

The heart of the configuration is the `reframe.core.config._SiteConfig`. This is the ultimate reference for any configuration option throughout the lifetime of ReFrame run and is responsible for supplying the default values for any configuration parameter. Defaults are no more scattered among the different internal APIs of the framework. 

The `reframe.core.systems.System` and `reframe.core.systems.SystemPartition` abstractions still remain and the pipeline interfaces with them as in the past. A `System` and all its partitions and environments are created from a `_SiteConfig` objects through the `System.create()` factory method. 

The `reframe.core.runtime.RuntimeContext` is also considerably changed. It serves its purpose as a wrapper to the site configuration providing some additional logic on top of it, e.g., computed stage/output directory prefixes and the information about the current test run (used in the retries feature). It also offers the facility to directly access a configuration option through its `get_option()` method. The `HostResources` and `HostSystem` objects are eliminated and any related functionality has been moved to either the `RuntimeContext` or the `System`. Also the system auto-detection is no more part of the `RuntimeContext`.

## How the `_SiteConfig` works

The `_SiteConfig` objects stores the raw site configuration as this was extracted from the configuration file.
It also holds the JSON schema for validating the configuration, and some other variables, relating to selecting the sub-configuration for a specific system or system/partition combination.

### Loading and validating the configuration

This is pretty much simple. We just load the JSON file in one of the `_create_from_xxx()` methods and we check for any JSON decoding error, which we propagate upward as a `ConfigError`. The validation against the schema happens in the `validate()` method, which clients must call separately.

### Selecting a sub-configuration for a system or system/partition combination

This happens in the `select_subconfig()` method which optionally takes a system name in the `system:partition` format. If a system name is not provided, the auto-detection takes place as in the past.

The purpose of this method is to construct a valid configuration just for the requested system or system/partition combination. This configuration is kept separately in the `_local_config` attribute, while the `_local_system` attribute acts as a flag in order to avoid unnecessary re-evaluations of the sub-configuration.

In the new syntax, all configuration sections can have a `target_systems` property that allows them to be configured differently per system and partition. In order to correctly address this, we create a `ScopedDict()` for each section and use that to resolve the configuration values against the requested system/partition combination.

After the new local configuration is built, we perform some additional validation steps to ensure that this configuration is valid: (a) all required fields are present in the local configuration as well, (b) all environments of the current system are defined in the local configuration. We perform these checks manually, not through the schema validation, because we don't want to do a full validation (this is just a subset of the original configuration) and step (b) cannot be validated through the schema anyway.

You may call `select_subconfig()` several times on the same `_SiteConfig` object in order to retrieve different sub-configurations.

### Retrieving configuration options

The method for retrieving configuration options is the `get()` method of `_SiteConfig`. This will take care also of returning the default values defined in the schema. The format of the `option` argument can be one of the following:

- `section/subsection/.../property`
- `section/<index>/subsection/.../property`
- `section/@<name>/.../property`

If a section contains a list of elements, you will have to use and index to retrieve the exact element or `@name` if the element are name addressable. An element is considered name addressable if it has a `name` property. Here are some examples of calling `get()`:

```python
site_config.get('environments/0/cxx')
site_config.get('environments/@PrgEnv-gnu/cc')
```

See the `test_config.py` unit test for more use cases.

If any option does not exist in the current configuration, its default value will be looked up in the schema.
The lookup key is the same as the option name with any indexes or `@name` accessors removed. So for the above `get()` calls the following keys will be looked up in the schema:

```
defaults/environments/cxx
defaults/environments/cc
```

If a default value cannot be retrieved from the schema the `default` argument passed to `get()` will be returned (which defaults to `None`). This argument is also returned in case of index errors or if an element cannot by found by its name.

The idea is that any other component of the framework that needs a special configuration variable asks it from the site configuration through the runtime. This is for example the case of the scheduler backends, which do `get('schedulers/@backend_name/job_submit_timeout')` in order to retrieve the job submit timeout parameter. And this parameter can be configured differently for different backends and for different system/partitions combinations!

### Sticky options

You can add and remove sticky options to the site configuration through the `add/remove_stick_option()` methods. A sticky option overrides any other reference to that option in the configuration, including the defaults. This is the mechanism that the frontend uses to impose command line options and/or environment variables over their configuration option counterparts. The format for setting a sticky option is similar to the defaults described above but without the `defaults/` prefix. So to override completely the `cxx` value in the configuration file, one has to call `site_config.add_sticky_option('environments/cxx', 'clang')`.

## Associating command line options with environment variables and configuration parameters

You can now associate a command-line argument with an environment variable and/or a configuration parameter. This is achieved by extending the interface of the `reframe.frontend.argparse.ArgumentParser` and specifically the `add_argument()` method. This method now accepts two more keyword arguments for specifying the association:

1. `envvar`: The name of the associated environment variable.
2. `configvar`: The name of the associated configuration variable (without indexes or name accessors).

When you retrieve a value of an option of the namespace object returned by `ArgumentParser.parse_args()`, if the option was not specified in the command-line (i.e., has a value of `None`), the associated environment variable will be queried if it exists. We explicitly do *not* query the associated configuration variable here, because the namespace object is only visible to the `cli.py` and, as discussed previously, we want the `_SiteConfig` to be the ultimate reference of configuration parameters throughout the framework. For this reason, the `ArgumentParser` is equipped with an `update_config()` method, whose main purpose is to add sticky options to the site configuration based on the namespace's options.

# Other changes and improvements

## Unit tests

Unit tests that required switching the runtime configuration have been all migrated to pure pytest syntax using fixtures. This makes each test more self contained and avoids excessive monkey patching. There is still some duplication of common fixtures among different files, but I leave that as future work.

Notice also the changes required to run the unit tests on the different systems. Since I am using the `tmp_path` fixture wherever I need a temporary directory, you have to pass the `--basetemp` option to pytest in order to run on Daint/Dom etc, since $TMPDIR is not mounted on the compute nodes.

## Registration of scheduler and parallel job launcher backends

The logic of registration has been consolidated and moved inside the `reframe.core.backends` module. Also the modules of the backends are imported programmatically upon request from the `getscheduler()` and `getlauncher()` methods. This avoids nasty circular import problems. Nothing changes in terms of the registration itself or retrieval of the scheduler and launchers except the module to import. For this reason, I had to change all the checks that make use of any of these methods.

## Environment loading and snapshots

I have moved the logic of loading an environment and taking an environment snapshot inside the `runtime` module. The reason for that is that these operations need the runtime due to the modules system, which led to intricate import patterns.

# Open Issues / Todos

- [x] Fine tune printing of the "ReFrame directories"
- [x] Fix `FIXME` notes
- [x] Remove stale code if any
- [x] Allow `add_argument()` to define environment variable/configuration options combinations only
- [x] Convert environment variable values to the correct type based on the `const` argument of the argument parser.
- [x] Update comments in `ArgumentParser`
- [x] Fix crash reported below.
- [x] Logging is still configured the old way, ignoring the defaults set in the schema.
- [x] Verify Graylog backend and consider using environment variables for specifying the server address.
- [x] Update the documentation
- [x] Fix crash when trying to initialize non-configured modules systems.

Fixes #1116.
Fixes #1115.
Fixes #66.
Addresses #1157.


vkarak:
Code crashes if you run it from a separate directory:


```python
$ ../reframe/bin/reframe -C ../reframe/config/cscs.py -c checks/mch/ncl/ncl.py -r --performance-report
  File "../reframe/bin/reframe", line 11, in <module>
    cli.main()
  File "/users/karakasv/Devel/reframe/reframe/frontend/cli.py", line 311, in main
    site_config = config.load_config('reframe/core/settings.py')
  File "/users/karakasv/Devel/reframe/reframe/core/config.py", line 480, in load_config
    return _SiteConfig.create(filename)
  File "/users/karakasv/Devel/reframe/reframe/core/config.py", line 128, in create
    return cls._create_from_python(filename)
  File "/users/karakasv/Devel/reframe/reframe/core/config.py", line 137, in _create_from_python
    mod = util.import_module_from_file(filename)
  File "/users/karakasv/Devel/reframe/reframe/utility/__init__.py", line 72, in import_module_from_file
    return _do_import_module_from_file(filename, module_name)
  File "/users/karakasv/Devel/reframe/reframe/utility/__init__.py", line 54, in _do_import_module_from_file
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 674, in exec_module
  File "<frozen importlib._bootstrap_external>", line 780, in get_code
  File "<frozen importlib._bootstrap_external>", line 832, in get_data
FileNotFoundError: [Errno 2] No such file or directory: '/users/karakasv/Devel/reframe-tests/reframe/core/settings.py'
```

vkarak:
@teojgo Yes, it should.

teojgo:
@vkarak ok, don't change it for the moment it does not affect the review in any case.

vkarak:
I removed the documentation from the open issues of this PR. It will be addressed in a separate issue.

reframe/core/environments.py
```diff
@@ -4,13 +4,13 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import collections
+import collections.abc
```
teojgo, line 7: This is not needed

reframe/core/environments.py
```diff
@@ -4,13 +4,13 @@
...
 import os
 
 import reframe.core.fields as fields
 import reframe.utility as util
 import reframe.utility.os_ext as os_ext
```
teojgo, line 12: This is not needed

reframe/core/runtime.py
```diff
@@ -16,107 +16,30 @@
 import reframe.core.config as config
 import reframe.core.fields as fields
 import reframe.utility.os_ext as os_ext
-from reframe.core.exceptions import (ConfigError,
-                                     ReframeFatalError,
-                                     SpawnedProcessError,
-                                     SystemAutodetectionError,
-                                     UnknownSystemError)
-from reframe.core.modules import ModulesSystem
+from reframe.core.environments import (Environment, snapshot)
```
teojgo, line 19: Remove the import of `re` and `socket` above

unittests/test_cli.py
```diff
@@ -6,20 +6,20 @@
 import copy
 import itertools
 import os
+import pathlib
```
teojgo, line 9: Importing `copy`, `re`, `tempfile` and `unittest` is not needed.

unittests/test_config.py
```diff
@@ -3,153 +3,331 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
-import copy
+import json
+import os
```
teojgo, line 7: Importing `unittest` is no longer needed

vkarak:
@jenkins-cscs retry all

reframe/core/logging.py
```diff
@@ -243,101 +207,88 @@ def _create_syslog_handler(handler_config):
...
 
-    host = handler_config.get('host', None)
-    port = handler_config.get('port', None)
-    extras = handler_config.get('extras', None)
-    if host is None:
-        raise ConfigError('graylog handler: no host specified')
-
-    if port is None:
+    address = site_config.get(f'{config_prefix}/address')
+    host, *port = address.split(':', maxsplit=1)
```
teojgo, line 250: Why do you need `*port` since `maxsplit=1`?

vkarak: It is correct:

```pycon
>>> 'server:123'.split(':', maxsplit=1)
['server', '123']
```

vkarak: It is correct:

```pycon
>>> 'server:123'.split(':', maxsplit=1)
['server', '123']
```

teojgo: Yes, but why do you need the starred `port`?

vkarak: Yes, because the user may not specify the port and the Graylog handler needs it. We could enforce this at the schema level, but I just did as before.

teojgo: So if the user does not specify, you need the `port` to be an empty list?

vkarak: No, I need the port to be a number, but I use the `*` notation to avoid handling exceptions and be more concise. Later on, I do `port = port[0]`.

vkarak: The only case that I'm not handling and I should is when the input is `server:`.

teojgo: Ok, I get it now, thanks.

unittests/test_logging.py
```diff
@@ -3,23 +3,27 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+import pdb
```
teojgo, line 6: Remove ths `pdb` import

unittests/test_logging.py
```diff
@@ -3,23 +3,27 @@
...
 import logging
 import logging.handlers
 import os
 import pytest
-import sys
 import re
+import sys
 import tempfile
 import time
 import unittest
```
teojgo, line 16: Convert also the `TestLogger` class to pytest syntax

vkarak: I did only the necessary, because I want to leave that task for the #1149.

vkarak:
Thanks @teojgo for the review! Have you tried playing with the new configuration to get the feeling? I'm interested of things that may not be so intuitive or corner cases that are not caught.

teojgo:
@varak I played a little bit. I will dig in now and comment on the way.

vkarak:
Just to remember to fix. Code crashes if it can't find a sane installation of the modules system requested:

```
karakasv@tresa reframe [feat/new-config-mechanism] $ ./bin/reframe -C config/cscs.py --system=dom -l
  File "./bin/reframe", line 11, in <module>
    cli.main()
  File "/Users/karakasv/Repositories/reframe/reframe/frontend/cli.py", line 353, in main
    runtime.init_runtime(site_config)
  File "/Users/karakasv/Repositories/reframe/reframe/core/runtime.py", line 164, in init_runtime
    _runtime_context = RuntimeContext(site_config)
  File "/Users/karakasv/Repositories/reframe/reframe/core/runtime.py", line 41, in __init__
    self._system = System.create(site_config)
  File "/Users/karakasv/Repositories/reframe/reframe/core/systems.py", line 262, in create
    partitions=partitions
  File "/Users/karakasv/Repositories/reframe/reframe/core/systems.py", line 176, in __init__
    self._modules_system = ModulesSystem.create(modules_system)
  File "/Users/karakasv/Repositories/reframe/reframe/core/modules.py", line 97, in create
    return ModulesSystem(TModImpl())
  File "/Users/karakasv/Repositories/reframe/reframe/core/modules.py", line 408, in __init__
    raise ConfigError('could not find a sane TMod installation')
reframe.core.exceptions.ConfigError: could not find a sane TMod installation
```

unittests/test_cli.py
vkarak: `re` is needed here.

reframe/frontend/argparse.py
```diff
@@ -20,57 +18,186 @@
...
-# fields that forward the set/get actions to the internal argument holder.
+# parser) was to programmaticallly export them by implementing the
+# `__getattr__()` method, such as to delegate any lookup of unknown public
+# attributes to the underlying `argparse.ArgumentParser`.
 #
+# Finally, the functionality of the ArgumentParser is extended to support
+# associations of command-line arguments with environment variables and/or
+# configuration parameters. Additionally, we allow to define pseudo-arguments
+# that essentially associate environment variables with configuration
+# arguments, without having to have to define a corresponding command line
```
ekouts, line 37:
```suggestion
# arguments, without having to define a corresponding command line
```

docs/config_reference.rst
```diff
@@ -0,0 +1,1215 @@
...
+
+.. envvar:: RFM_PREFIX
+
+.. option:: --prefix
+
+   :required: No
+   :default: ``"."``
+
+   Directory prefix for a ReFrame run on this system.
+   Any directories or files produced by ReFrame will be placed use this prefix, if not specified otherwise.
```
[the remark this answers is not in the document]
vkarak: I used your first suggestion.

docs/config_reference.rst
```diff
@@ -0,0 +1,1215 @@
...
+
+.. envvar:: RFM_OUTPUT_DIR
+
+.. option:: -o DIR | --output DIR
+
+   :required: No
+   :default: ``"${RFM_PREFIX}/output"``
+
+   Output directory prefix for this system.
+   This is the directory prefix, where ReFrame will save information about the succeeded tests.
```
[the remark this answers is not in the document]
vkarak: I used "successful."

docs/config_reference.rst
```diff
@@ -0,0 +1,1215 @@
...
+              'name': 'datawarp',
+              'options': [
+                  '#DW jobdw capacity={capacity} access_mode={mode} type=scratch',
+                  '#DW stage_out source={out_src} destination={out_dst} type={stage_filetype}'
+              ]
+          }
+      ]
+
+
+   A regression test that wants to make use of that resource, it can set its :attr:`extra_resources` as follows:
```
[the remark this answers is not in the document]
vkarak: I think my original version is correct here. The subject of the sentence is the whole clause "A regression test that wants to make use of that resource." So the "it" does not sound correct.

docs/configure.rst
```diff
@@ -2,437 +2,419 @@
...
-           '#DW stage_out source={out_src} destination={out_dst} type={stage_filetype}'
-       ]
-   }
+Each environment is associated with a name.
+This name will be used to reference this environment in different contexts, as for example in the ``environs`` property of the system partitions.
+This environment definition is minimal, since the default values for the rest of the properties serve our purpose.
+For a complete list of the environment properties, see the `configuration reference <config_reference.html#environment-configuration>`__.
 
-  A regression test that wants to make use of that resource, it can set its :attr:`extra_resources` as follows:
+An important feature in ReFrame's configuration, is that you can define section objects differently for different systems or system partitions.
```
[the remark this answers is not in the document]
vkarak: "differently" expresses more explicitly what I want to say. "separately" just implies it.

docs/configure.rst
```diff
@@ -2,437 +2,419 @@
...
+1. A file handler to print debug messages in the ``reframe.log`` file using a more extensive message format that contains a timestamp, the level name etc.
+2. A stream handler to  print any informational messages (and warnings and errors) from ReFrame to the standard output.
+   This handles essentially the actual output of ReFrame.
+3. A file handler to print the framework's output in the ``reframe.out`` file.
 
-* ``nativeslurm``: This is equivalent to ``slurm+srun``.
-* ``local``: This is equivalent to ``local+local``.
+It might initially be confusing the fact that there are two ``level`` properties: one at the logger level and one at the handler level.
+This works hierarchically.
+When ReFrame logs a message, this first goes into its internal logger, is checked again the logger's level (here  ``debug``) and if it's above it, it is forwarded to the handlers.
```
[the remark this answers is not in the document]
vkarak: Haha, yes! I will rephrase that.

docs/running.rst
```diff
@@ -787,180 +782,24 @@ By default, the output in ``reframe.log`` looks like the following:
...
-
-The type of this handler is ``filelog`` and logs the performance of a regression test in one or more files.
-The attributes of this handler are the following:
-
-* ``prefix``: This is the directory prefix (usually dynamic) where the performance logs of a test will be stored.
-  This attribute accepts any of the check-specific formatting placeholders described `above <#common-log-handler-attributes>`__.
-  This allows you to create dynamic paths based on the current system, partition and/or programming environment a test executes.
-  This dynamic prefix is appended to the "global" performance log directory prefix, configurable through the ``--perflogdir`` option or the ``perflogdir`` attribute of the `system configuration <configuring.html#system-configuration>`__.
-  The default configuration of ReFrame for performance logging (shown in the previous listing) generates the following files:
+Performance testing is an essential part of ReFrame and performance data from tests is handled specially.
```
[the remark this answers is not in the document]
vkarak: Do you propose this because you are a "data is plural" proponent? Because there two different independent clauses here, so the "is" refers to the "performance data." I suggest sticking to the singular form of data, since this is what we are using throughout.

reframe/core/exceptions.py
```diff
@@ -283,8 +283,8 @@ def format_user_frame(frame):
     if isinstance(exc_value, KeyboardInterrupt):
         return 'cancelled by user'
 
-    if isinstance(exc_value, OSError):
-        return 'OS error: %s' % exc_value
+    # if isinstance(exc_value, OSError):
+    #     return 'OS error: %s' % exc_value
```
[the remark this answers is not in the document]
vkarak: Good catch! This was temporary in order to check something...

vkarak:
@jenkins-cscs retry trsa

unittests/test_schedulers.py
```diff
@@ -16,966 +17,914 @@
...
+            'AllocTRES= CapWatts=n/a CurrentWatts=100 '
+            'LowestJoules=100000000 ConsumedJoules=0 '
+            'ExtSensorsJoules=n/s ExtSensorsWatts=0 '
+            'ExtSensorsTemp=n/s Reason=Foo/ ',
+
+            'NodeName=nid00005 Arch=x86_64 CoresPerSocket=12 '
+            'CPUAlloc=0 CPUErr=0 CPUTot=24 CPULoad=0.00 '
+            'AvailableFeatures=f5 ActiveFeatures=f5 '
+            'Gres=gpu_mem:16280,gpu:1 NodeAddr=nid00003'
+            'NodeHostName=nid00003 Version=10.00 OS=Linux '
```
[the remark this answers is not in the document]
vkarak: Can you explain a bit the changes in the node ids?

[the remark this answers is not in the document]
teojgo: It's not something important, it's just that you have the same value of `NodeAdd`r, `NodeHostName` and `NodeName` on its node.

ci-scripts/ci-runner.bash
```diff
@@ -173,55 +160,58 @@ elif [ $CI_TUTORIAL -eq 1 ]; then
...
-    # Performing the unittests
-    echo "=================="
-    echo "Running unit tests"
-    echo "=================="
+    # Run unit tests with the scheduler backends
+    PATH_save=$PATH
+    tempdir=$(mktemp -d -p $SCRATCH)
+    if [[ $(hostname) =~ dom ]]; then
+        export PATH=/apps/dom/UES/karakasv/slurm-wrappers/bin:$PATH
+        # for backend in slurm pbs torque; do
```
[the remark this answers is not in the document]
vkarak: You are right, sth is fishy here. I will try to remember what's going on here.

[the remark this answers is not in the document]
vkarak: Yes, I need to uncomment this.

unittests/test_schedulers.py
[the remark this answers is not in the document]
vkarak: This is how we have it in master and this is what Slurm gives in any case. I believe it will be confusing to change them.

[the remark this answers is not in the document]
vkarak: Got it now, `NodeName`, `NodeHostName` and `NodeAddr` should all be the same.

vkarak:
@teojgo I will fix them as soon as I get all the approvals. I have fixed conflicts multiple times for this PR so far.

docs/migration_2_to_3.rst
```diff
@@ -2,14 +2,56 @@
...
+
+- Your code formatting will be lost.
+  ReFrame will use its own, which is PEP8 compliant nonetheless.
+- Any comments will be lost.
+- Any code that was used to dynamically generate configuration parameters will be lost.
+  ReFrame will generate the new configuration based on what was the actual old configuration after any dynamic generation.
+
+
+
+Updating Your Tests
```
vkarak, line 48: I need to a subsection here about the change in location of `getscheduler()` and `getlauncher()` functions.

docs/migration_2_to_3.rst
```diff
@@ -2,14 +2,56 @@
...
 
-Updating your tests
+At the same time, ReFrame 3 deprecates some common pre-2.20 test syntax in favor of the more modern and intuitive pipeline hooks.
+
+This guide details the necessary steps in order to easily migrate to ReFrame 3.
+
+
+Updating Your Site Configuration
+--------------------------------
+
```
vkarak, line 15: I need to mention that the default settings file has changed. It's no more `reframe/settings.py`.

vkarak:
@jenkins-cscs retry daint

PR #1246, opened 2020-03-30
[test] Update nvprof test and add maintenance tag

jgphpc:
Will fix [UES-721](https://jira.cscs.ch/browse/UES-721)

### DIFF
--- a/cscs-checks/tools/profiling_and_debugging/nvprof.py
+++ b/cscs-checks/tools/profiling_and_debugging/nvprof.py
@@ -20,15 +20,6 @@ class NvprofCheck(rfm.RegressionTest):
         self.sourcesdir = 'src/Cuda'
         self.executable = 'nvprof'
         self.target_executable = './jacobi'
-        self.executable_opts = [self.target_executable]
-        self.sanity_patterns = sn.all([
-            sn.assert_found('Profiling application: %s' %
-                            self.target_executable, self.stderr),
-            sn.assert_found('[CUDA memcpy HtoD]', self.stderr),
-            sn.assert_found('[CUDA memcpy DtoH]', self.stderr),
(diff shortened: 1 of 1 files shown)


jgphpc:
It fails because the fix is not on daint (yet). What's recommended here: exclude it from jenkins with`-x NvprofCheck` or add some python in the check ?

teojgo:
@jgphpc we can exclude it from the daily runs.

cscs-checks/tools/profiling_and_debugging/nvprof.py
```diff
@@ -45,10 +35,20 @@ def __init__(self):
...
+        self.executable_opts = [self.target_executable]
+        self.post_run = ['cat /etc/modprobe.d/nvidia.conf']
+        self.sanity_patterns = sn.all([
+            sn.assert_found('Profiling application: %s' %
+                            self.target_executable, self.stderr),
+            sn.assert_found('[CUDA memcpy HtoD]', self.stderr),
+            sn.assert_found('[CUDA memcpy DtoH]', self.stderr),
+            sn.assert_found(r'\s+100(\s+\S+){3}\s+jacobi_kernel', self.stderr),
+            sn.assert_found(r'NVreg_RestrictProfilingToAdminUsers=0',
+                            self.stdout)
```
vkarak: I am not sure this test is needed. What is important for this test is that nvprof works. How it works should not the test's problem. In its current form, it caught the problem, so I don't see the reason for putting this and temporarily disabling the test on Daint.

jgphpc: I will remove that assert.
The check will still fail because the rfc is not applied yet on daint.

jgphpc: https://github.com/eth-cscs/reframe/pull/1246/commits/6c252acd2337e769329c3897ee9738994d69fedc

PR #1249, opened 2020-03-30
[test] Update CP2K CPU small check

lucamar:
I have used the performance data stored in `/apps/daint/UES/jenscscs/regression/production/logs/daint/mc` starting November 7th 2019, after the last upgrade of Piz Daint:
- `Cp2kCpuCheck_small_prod.log`, 138 data points, average `180.9 s`, 95% within ~7.5% threshold (rounded to 8%)


vkarak:
@jenkins-cscs retry none

PR #1250, opened 2020-03-30
[test] Update QE CPU checks

lucamar:
I have used the performance data stored in `/apps/daint/UES/jenscscs/regression/production/logs/daint/mc` starting November 7th 2019, after the last upgrade of Piz Daint:
- `QECheck_large.log`, 145 data points, average `149.7 s`, 95% within 52% threshold
- `QECheck_small.log`, 164 data points, average `147.3 s`, 95% within 41% threshold

The thresholds required to reach the 95th percentile are huge, since the standard deviation of the samples is high. 
However, the thresholds would be even higher if I select only recent data points after the last maintenance (January 29th 2020).


vkarak:
@jenkins-cscs retry none

PR #1253, opened 2020-04-03
[feat] Support `*:<partition name>` name pattern in the list of valid systems

### LINKED ISSUE #1126: Supporting valid_systems = ['*:partition_name']
Hi,
We have multiple systems with the same partitions. Currently, in our tests, we need to specify a whole list for valid_systems, such as : 

``` 
self.valid_systems = ['cedar:cpu', 'graham:cpu', 'beluga:cpu', 'helios:cpu', 'niagara:cpu']
``` 

it would be great if it could be supported to have `['*:cpu']` in `valid_systems`

mboisson:
Fixes #1126.


vkarak:
test this please

reframe/core/pipeline.py
```diff
@@ -957,11 +957,18 @@ def supports_system(self, partition_name):
             return True
 
         # Check if this is a relative name
-        if partition_name.find(':') == -1:
-            partition_name = '%s:%s' % (self.current_system.name,
-                                        partition_name)
+        if partition_name.find(':') != -1:
+            system_partition = partition_name.split(':')
+            system_name = system_partition[0]
+            partition_name = system_partition[1]
```
teojgo, line 963:
```suggestion
            system_name, partition_name = partition_name.split(':')
```

reframe/core/pipeline.py
```diff
@@ -957,11 +957,18 @@ def supports_system(self, partition_name):
...
+        else:
+            system_name = self.current_system.name
+            partition_name = partition_name
+
+        specific = '%s:%s' % (system_name, partition_name)
+        generic = '*:%s' % partition_name
 
-        return partition_name in self.valid_systems
+        return any(p in self.valid_systems for p in [specific, generic])
```
teojgo, line 971:
```suggestion
        return specific in self.valid_systems or generic in self.valid_systems
```

reframe/core/pipeline.py
```diff
@@ -158,7 +158,7 @@ class RegressionTest(metaclass=RegressionTestMeta):
                                             typ.List[str])
 
     #: List of systems supported by this test.
-    #: The general syntax for systems is ``<sysname>[:<partname]``.
+    #: The general syntax for systems is ``<sysname|*>[:<partname]``.
```
vkarak, line 161: Better keep the `sysname` reference here as before and add a separate sentence explaining what `sysname` and `partname` could be. I am saying  also `partname` here, because I think it'd be better if all of the following were supported:

- `*:*`: equivalent to `*`
- `*:gpu`: what you implement in this PR
- `system:*`: equivalent to `system`.
- `system:part`: as always.

Saying that check my comments on the implementation later on.

reframe/core/pipeline.py
```diff
@@ -957,11 +957,16 @@ def supports_system(self, partition_name):
...
+            system_partition, partition_name = partition_name.split(':')
+        else:
+            system_name = self.current_system.name
+            partition_name = partition_name
+
+        specific = '%s:%s' % (system_name, partition_name)
+        generic = '*:%s' % partition_name
 
-        return partition_name in self.valid_systems
+        return specific in self.valid_systems or generic in self.valid_systems
```
vkarak, line 969: First, I would change the formal argument name to something more generic: `partition_name` is a bit misleading. I would suggest to use just `name`. Second, you could avoid the hassle of checking the different formats if we had a `normalize_partname()` function (even nested inside `supports_system()`) that would normalize the the partition names and then try to match them. Something along the following lines:

```python
    def supports_system(self, name):
        def normalize_partname(name):
            pass

        def match_system(patt, normalized_name):
            pass

        return any(match_system(normalize_partname(n), normalize_partname(name)) for n in self.valid_systems)
```

The `match_system()` would do the simple matching we are looking for. We don't want neither regex matches nor fnmatches.

unittests/test_pipeline.py
```diff
@@ -287,6 +287,12 @@ def test_supports_system(self):
         assert not test.supports_system('testsys:gpu')
         assert not test.supports_system('testsys:login')
 
+        test.valid_systems = ['*:gpu']
+        assert test.supports_system('testsys:gpu')
+        assert test.supports_system('foo:gpu')
+        assert not test.supports_system('testsys:cpu')
+        assert not test.supports_system('testsys:login')
```
vkarak, line 294: You could add some additional assertions for testing the `*:*` and `system:*` alternatives.

reframe/core/pipeline.py
mboisson: implemented differently, but should achieve the same idea

reframe/core/pipeline.py
mboisson: done

mboisson:
Looks like I need to think about this a bit more... 🤔

vkarak:
Actually you can try to run the unit tests yourself by running `./test_reframe.py`. It accepts all the pytest options.

mboisson:
Ok, should be good now.

reframe/core/pipeline.py
```diff
@@ -949,19 +950,20 @@ def info(self):
...
+        if name.find(':') != -1:
+            system_name, partition_name = name.split(':')
+        else:
+            system_name, partition_name = self.current_system.name, name
+
+        names_to_test = ['*', '*:*']
+        names_to_test.append(system_name)
+        names_to_test.append('%s:*' % system_name)
+        names_to_test.append('*:%s' % partition_name)
+        names_to_test.append('%s:%s' % (system_name, partition_name))
```
vkarak, line 964: Just create the list all at once and use f-strings. Also better call it `valid_matches`:

```suggestion
        valid_matches = ['*', '*:*', f'{system_name}:*', ...]
```

It could also be a set since depending on the input `name`, you may have duplicates.

unittests/test_pipeline.py
```diff
@@ -287,6 +287,23 @@ def test_supports_system(self):
...
+
+        test.valid_systems = ['*:*']
+        assert test.supports_system('gpu')
+        assert test.supports_system('login')
+        assert test.supports_system('testsys:gpu')
+        assert test.supports_system('testsys:login')
```
vkarak, line 300: Better move this up to where the test for `['*']` is.

unittests/test_pipeline.py
```diff
@@ -287,6 +287,23 @@ def test_supports_system(self):
...
+
+        test.valid_systems = ['testsys:*']
+        assert test.supports_system('testsys:login')
+        assert test.supports_system('gpu')
+        assert not test.supports_system('foo:gpu')
+
```
vkarak, line 306: Can you also add a test for testing a system that is of the form `x:y:z`?

reframe/core/pipeline.py
```diff
@@ -949,19 +950,20 @@ def info(self):
...
-
+    def supports_system(self, name):
         # Check if this is a relative name
-        if partition_name.find(':') == -1:
-            partition_name = '%s:%s' % (self.current_system.name,
-                                        partition_name)
+        if name.find(':') != -1:
+            system_name, partition_name = name.split(':')
+        else:
+            system_name, partition_name = self.current_system.name, name
```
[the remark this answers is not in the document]
mboisson: This breaks the current meaning. The current meaning is that the argument is the *partition* name by default, not the system name.

unittests/test_pipeline.py
mboisson: What should be the result of this test ?

vkarak: This is entirely up to us, but it should definitely not crash. I would suggest `testsys:login:foo` to compare to false, which is what the current implementation would do in any case if you use `maxsplit=1`.

reframe/core/pipeline.py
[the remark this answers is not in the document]
vkarak: Oh yes, you're right. It's some time I have written this... :-) So you can ignore this comment. Just address the case of malformed input with two `:`.

unittests/test_pipeline.py
mboisson: Mmm, I would argue that the logical result should be that it does give an error, not a false result

vkarak: It's not a false result. We only consider a single `:` separator and whatever is left to it is the system name, whatever is right to it is the partition name, then it's consistent if in the case of `testsys:login:foo` you assume that `testsys` is the system name and `login:foo` is the partition name. So the test should compare false.

mboisson: So, `login:foo` would be a valid partition name ?

vkarak: Although I'm not entirely convinced of this logic, since we are really talking about a corner case, I figured out that we can quite easily validate the in `valid_systems`, which I think is the best solution and we will be both happy. If you change the `str` [here](https://github.com/eth-cscs/reframe/blob/master/reframe/core/pipeline.py#L165) to `typ.Str[r'proper_regex']`, we're good to go. If `valid_systems` are not correct, you'll get an error right from the beginning. So you can skip this unit test completely.

vkarak: This regex seems to do the job: `^[^:]+(:[^:]+)?$`. I don't know if we could write in a simpler way.

vkarak:
@mboisson my plan is to merge this today, so I will go on with the solution of changing the type of the `valid_systems` field.

unittests/test_pipeline.py
vkarak: I've just done the change and pushed.

vkarak: I checked the code a bit more in detail. the `valid_systems` definitions does not have to change. The argument to the `supports_system` comes essentially from the configuration file through the `SystemPartition.fullname` property and there is no way it can be wrongly formatted due to the user. If it is, then it's a framework bug. So I will revert my change above and will approve the PR as is.

vkarak:
@teojgo Can you approve this one?

PR #1259, opened 2020-04-10
[bugfix] Fix `AttributeError` in flexible node allocation from the Slurm backend

### LINKED ISSUE #1258: AttributeError is raised in 'filternodes' method of slurm scheduler
The exception is raised when `job.sched_access` is a `SequenceView` object and subsequent statements try to append values to an immutable object.

teojgo:
* Fix `AttributeError` raised when trying to mutate a `SequenceView`
  object in the `filternodes` method of the slurm scheduler.

* Add a corresponding unittest.

Fixes #1258

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -231,19 +231,19 @@ class SlurmJobScheduler(sched.JobScheduler):
         # Collect options that restrict node selection
         options = job.sched_access + job.options
         if job.sched_partition:
-            options.append('--partition=%s' % job.sched_partition)
+            options += ['--partition=%s' % job.sched_partition]
 
         if job.sched_account:
-            options.append('--account=%s' % job.sched_account)
+            options += ['--account=%s' % job.sched_account]
 
         if job.sched_nodelist:
-            options.append('--nodelist=%s' % job.sched_nodelist)
+            options += ['--nodelist=%s' % job.sched_nodelist]
 
         if job.sched_exclude_nodelist:
-            options.append('--exclude=%s' % job.sched_exclude_nodelist)
+            options += ['--exclude=%s' % job.sched_exclude_nodelist]
 
         if job.sched_reservation:
-            options.append('--reservation=%s' % job.sched_reservation)
+            options += ['--reservation=%s' % job.sched_reservation]
 
         option_parser = ArgumentParser()
         option_parser.add_argument('--reservation')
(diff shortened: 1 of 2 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -231,19 +231,19 @@ def filternodes(self, job, nodes):
         # Collect options that restrict node selection
         options = job.sched_access + job.options
         if job.sched_partition:
-            options.append('--partition=%s' % job.sched_partition)
+            options += ['--partition=%s' % job.sched_partition]
```
ekouts, line 234:
```suggestion
            options += [f'--partition={job.sched_partition}']
```
Should you change this to use f-strings?

reframe/core/schedulers/slurm.py
```diff
@@ -231,19 +231,19 @@ def filternodes(self, job, nodes):
...
 
         if job.sched_account:
-            options.append('--account=%s' % job.sched_account)
+            options += ['--account=%s' % job.sched_account]
```
ekouts, line 237: Same for all the options

vkarak:
@teojgo I can't understand where this `SequenceView` object comes from.

reframe/core/schedulers/slurm.py
```diff
@@ -231,19 +231,19 @@ def filternodes(self, job, nodes):
         # Collect options that restrict node selection
         options = job.sched_access + job.options
```
vkarak, line 232: It's more correct conceptually if you change just this line:

```suggestion
        # Create a mutable list out of the SequenceView object of `job.sched_access`
        options = list(job.sched_access + job.options)
```

unittests/test_schedulers.py
```diff
@@ -743,6 +744,16 @@ def test_sched_access_idle(self):
         self.prepare_job()
         assert self.testjob.num_tasks == 8
 
+    def test_sched_access_idle_sequence_view(self):
```
vkarak, line 747: I don't think this is needed. Just change the `_sched_access` assignment in one of the `sched_access` tests.

vkarak:
It seems that this unit test is not enough. It does not break if you revert the fix. Since I'm fixing the PR, I will have a look at it as well.

teojgo:
> It seems that this unit test is not enough. It does not break if you revert the fix. Since I'm fixing the PR, I will have a look at it as well.

@varak yes, that's why I had the separate unit test because you need both `self.testjob._sched_access` and one of the other `self.test_job._sched...` in order to try to mutate the SequenceView.

vkarak:
@teojgo Yes, just realized that. I will cherry pick that commit.

PR #1262, opened 2020-04-14
[test] Update OSU Allreduce check

ekouts:
Fixes UES-785.

### DIFF
--- a/cscs-checks/microbenchmarks/osu/osu_tests.py
+++ b/cscs-checks/microbenchmarks/osu/osu_tests.py
@@ -108,23 +108,23 @@ class AllreduceTest(rfm.RegressionTest):
             self.num_tasks = 6
             self.reference = {
                 'dom:gpu': {
-                    'latency': (6.0, None, 0.10, 'us')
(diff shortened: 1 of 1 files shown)


cscs-checks/microbenchmarks/osu/osu_tests.py
```diff
@@ -111,10 +111,10 @@ def __init__(self, variant):
                     'latency': (6.0, None, 0.10, 'us')
                 },
                 'daint:gpu': {
-                    'latency': (7.81, None, 0.25, 'us')
+                    'latency': (13.75, None, 0.25, 'us')
```
rsarm, line 114: The range that covers ~90% of the values is more or less 6us to 15us, with the most repeated values around 6-7us. I would put the reference at the most repeated value (~6.5us), but in that case the max would have to be close to 200%. Would that be an option?

rsarm:
I think the refrence of the `large` test also need to be updated.

vkarak:
@jenkins-cscs retry dom

PR #1264, opened 2020-04-15
[test] Implement Cray linker workaround for CLE7.0.UP01.PS09

vkarak:
Fixes UES-829.

### DIFF
--- a/cscs-checks/compile/libsci_resolve.py
+++ b/cscs-checks/compile/libsci_resolve.py
@@ -76,6 +76,12 @@ class NvidiaResolveTest(LibSciResolveBaseTest):
                 sn.extractsingle(regex, self.stdout, 'version'), mod_name)
         ])
 
+    @rfm.run_before('compile')
+    def cray_linker_workaround(self):
+        if (self.current_environ.name.startswith('PrgEnv-cray') and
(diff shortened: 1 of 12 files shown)


vkarak:
@jenkins-cscs retry all

cscs-checks/compile/libsci_resolve.py
```diff
@@ -76,6 +76,12 @@ def set_sanity(self):
                 sn.extractsingle(regex, self.stdout, 'version'), mod_name)
         ])
 
+    @rfm.run_before('compile')
+    def cray_linker_workaround(self):
```
ekouts, line 80: You haven't included the FIXME note in all the tests. Is this on purpose?

vkarak: Good point, I forgot it and I need to update the comment as well in the other fixme notes.

victorusu: @vkarak what do you think about adding this to XALT?

vkarak: @victorusu I don't understand this very well.

vkarak:
@jenkins-cscs retry none

PR #1270, opened 2020-04-17
[test] Update tests for QuantumESPRESSO 6.5

lucamar:
I propose an updated check for the release 6.5 of `QuantumESPRESSO`, which now features a GPU enabled version for `pw.x`: I have created the tests based on `small` and `large` checks of `CP2K`.
Please note that the reference performance values are still tentative, as I have run few times only: I use the wall time reported in the program output, since the cpu time reports the sum over all threads.

I have changed the folder and filename from `espresso` to `quantumespresso`, since all other checks of scientific applications are listed with the full name; furthermore, `ESPRESSO` is a different scientific package and using this name can create some confusion (see http://espressomd.org/wordpress).

I have created the subfolder `src` with the input file and the pseudopotential file, since they don't take much more space with respect to other checks (321K with respect to 241K of `CP2K`) and it's better to track changes: therefore, I think that `/apps/common/UES/regression/resources/Espresso` could be be removed after the obsolete check `QECheck_small` that still uses that path is removed as well.


victorusu:
@lucamar, the PR `lgtm` but I was wondering if we have any licensing issues by having the pseudopotential in the ReFrame project.

lucamar:
@victorusu The pseudopotentials used by QuantumESPRESSO are publicly available both on the QuantumESPRESSO web site and on the GitHub project space. The user needs to give credit to the author: see https://www.quantum-espresso.org/pseudopotentials/about and also the User Portal at
https://user.cscs.ch/computing/applications/quantumespresso/#licensing-terms-and-conditions

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@rfm.parameterized_test(*([s, v]
+                          for s in ['small', 'large']
+                          for v in ['maint', 'prod']))
+class QuantumESPRESSOCpuCheck(QuantumESPRESSOCheck):
+    def __init__(self, scale, variant):
+        super().__init__()
+        self.descr = 'QuantumESPRESSO CPU check (version: %s, %s)' \
+                     % (scale, variant)
```
vkarak, line 43: Rewrite this using f-strings:

```suggestion
        self.descr = f'QuantumESPRESSO CPU check (version: {scale}, {variant})'
```

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -0,0 +1,139 @@
...
+
+class QuantumESPRESSOCheck(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        self.valid_prog_environs = ['builtin']
+        self.executable = 'pw.x'
+        self.executable_opts = ['-in', 'ausurf.in']
+
+        self.sanity_patterns = sn.all([
+            sn.assert_found(r'convergence has been achieved', self.stdout),
+        ])
```
vkarak, line 19: You override `sanity_patterns` completely in both subclasses. So what's the point of having this one?

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -0,0 +1,139 @@
...
+
+
+@rfm.parameterized_test(*([s, v]
+                          for s in ['small', 'large']
+                          for v in ['maint', 'prod']))
+class QuantumESPRESSOGpuCheck(QuantumESPRESSOCheck):
+    def __init__(self, scale, variant):
+        super().__init__()
+        self.descr = 'QuantumESPRESSO GPU check (version: %s, %s)' \
+                     % (scale, variant)
```
vkarak, line 95:
```suggestion
        self.descr = f'QuantumESPRESSO GPU check (version: {scale}, {variant})'
```

vkarak:
@jenkins-cscs retry dom

cscs-checks/apps/quantumespresso/quantumespresso_check.py
lucamar: Thanks for the comment @vkarak: I'm sorry, I did not intend to override this `sanity_patterns`, but to append the ones in the subclasses to this first one, which should be a prerequisite

lucamar:
@jenkins-cscs retry dom

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -56,7 +55,7 @@ def __init__(self, scale, variant):
         energy = sn.extractsingle(r'!\s+total energy\s+=\s+(?P<energy>\S+) Ry',
                                   self.stdout, 'energy', float)
         energy_diff = sn.abs(energy-energy_reference)
-        self.sanity_patterns = sn.all([
+        self.sanity_patterns += sn.all([
             sn.assert_lt(energy_diff, 1e-8)
         ])
```
vkarak: This will not have the desired result and it will always succeed if the final list has at least one element. The correct way to write this is the following:

```suggestion
        self.sanity_patterns = sn.all(
            self.sanity_patterns,
            [sn.assert_lt(energy_diff, 1e-8)]
        )
```

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -110,7 +108,7 @@ def __init__(self, scale, variant):
         energy = sn.extractsingle(r'!\s+total energy\s+=\s+(?P<energy>\S+) Ry',
                                   self.stdout, 'energy', float)
         energy_diff = sn.abs(energy-energy_reference)
-        self.sanity_patterns = sn.all([
+        self.sanity_patterns += sn.all([
```
vkarak, line 111: Same comment here.

cscs-checks/apps/quantumespresso/quantumespresso_check.py
lucamar: Thanks @vkarak, I have changed the code accordingly

cscs-checks/apps/quantumespresso/quantumespresso_check.py
lucamar: Thanks @vkarak, I have changed the code accordingly

lucamar:
@jenkins-cscs retry dom

vkarak:
@lucamar I've just noticed that on of test is still named the  old way `QECheck_small`.

vkarak:
The problem is that the old file is not removed. I will fix that now and merge this.

PR #1279, opened 2020-04-23
[bugfix] More robust completion check for the Torque scheduler

### LINKED ISSUE #1238: Torque job failed if standard output and error files are hosted until the job is completed
This line is not sufficient to determine the job completeness if standard output and error files are hosted until the job is completed 

https://github.com/eth-cscs/reframe/blob/007bc0b2b477a0ef86eeb4bb3b7eb00e4c0301ec/reframe/core/schedulers/torque.py#L95

We still need these line before return:
```
        with os_ext.change_dir(job.workdir):
            done = os.path.exists(job.stdout) and os.path.exists(job.stderr)
```

victorusu:
I added the code proposed by @ZQyou to the torque scheduler. @ZQyou can you please check if it works for you?

Fixes #1238

### DIFF
--- a/reframe/core/schedulers/torque.py
+++ b/reframe/core/schedulers/torque.py
@@ -93,7 +93,7 @@ class TorqueJobScheduler(PbsJobScheduler):
                     raise JobError('maximum pending time exceeded',
                                    jobid=job.jobid)
 
-            with os_ext.change_dir(job.workdir):
-               done = os.path.exists(job.stdout) and os.path.exists(job.stderr)
-
+            stdout = os.path.join(job.workdir, job.stdout)
+            stderr = os.path.join(job.workdir, job.stderr)
+            done = os.path.exists(stdout) and os.path.exists(stderr)
             return job.state == 'COMPLETED' and done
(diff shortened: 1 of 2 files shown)


reframe/core/schedulers/torque.py
```diff
@@ -92,4 +93,6 @@ def finished(self, job):
                     raise JobError('maximum pending time exceeded',
                                    jobid=job.jobid)
 
-            return job.state == 'COMPLETED'
+            with os_ext.change_dir(job.workdir):
+               done = os.path.exists(job.stdout) and os.path.exists(job.stderr)
```
vkarak, line 97: Leave a blank line here.

vkarak:
@jenkins-cscs retry daint

unittests/test_cli.py
```diff
@@ -234,7 +234,7 @@ def test_failure_stats(self):
         returncode, stdout, stderr = self._run_reframe()
 
         assert r'FAILURE STATISTICS' in stdout
-        assert r'sanity      1     [SanityFailureCheck' in stdout
+        assert r'sanity        1     [SanityFailureCheck' in stdout
```
vkarak, line 237: I don't understand this change. This should be from another PR.

victorusu: Me neither. I have merged from master

vkarak:
@victorusu There is only a PEP8 issue remaining for this.

PR #1287, opened 2020-05-05
[bugfix] Fix deprecation warnings in unit tests

### LINKED ISSUE #1285: Fix deprecation warnings in unit tests

ekouts:
It fixes the deprecation warnings on the unittests and changes jenkins script so that it treats `ReframeDeprecationWarning` as error.

Fixes #1285

### DIFF
--- a/unittests/resources/checks_unlisted/selfkill.py
+++ b/unittests/resources/checks_unlisted/selfkill.py
@@ -25,7 +25,7 @@ class SelfKillCheck(rfm.RunOnlyRegressionTest):
         self.tags = {type(self).__name__}
         self.maintainers = ['TM']
 
-    def run(self):
-        super().run()
+    @rfm.run_before('run')
+    def self_kill(self):
         time.sleep(0.5)
         os.kill(os.getpid(), signal.SIGTERM)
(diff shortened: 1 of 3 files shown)


unittests/resources/checks_unlisted/selfkill.py
```diff
@@ -25,7 +25,7 @@ def __init__(self):
         self.tags = {type(self).__name__}
         self.maintainers = ['TM']
 
-    def run(self):
-        super().run()
+    @rfm.run_before('run')
+    def self_kill(self):
```
vkarak, line 29: I think this test should be simply marked as "special." Using the hooks here changes its logic. The idea was to immediately kill ReFrame after a job is submitted. Here you rely too much on the sleeping time. In the past the sleepinh time was just to give it enough time to submit the job, but at least we knew that the kill would happen after the submit command.

teojgo: Then instead of using the `@rfm` decorator you could use:
`@pytest.mark.filterwarnings('ignore::reframe.core.exceptions.ReframeDeprecationWarning')`
on the `test_sigterm` test method

vkarak: @teojgo, no there is no need to do so. That's why we can mark a test as "special" with `special=True` in its definition.

teojgo: @vkarak yeap I have missed that

ekouts: I marked it as special

PR #1293, opened 2020-05-07
[bugfix] Fix the import of `getlauncher()` function in Greasy test

teojgo:


vkarak:
I will merge that. The CI failure will be addressed in #1294.

PR #1296, opened 2020-05-07
[test] Fix Slurm constraint check for cabinets

vkarak:
I also removed unnecessary programming environments from `valid_prog_environs`.

Fixes UES-859.

### DIFF
--- a/cscs-checks/system/slurm/slurm.py
+++ b/cscs-checks/system/slurm/slurm.py
@@ -16,7 +16,7 @@ class SlurmSimpleBaseCheck(rfm.RunOnlyRegressionTest):
                               'kesch:cn', 'kesch:pn',
                               'arolla:cn', 'arolla:pn',
                               'tsa:cn', 'tsa:pn']
-        self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu']
+        self.valid_prog_environs = ['PrgEnv-cray']
         self.tags = {'slurm', 'maintenance', 'ops',
                      'production', 'single-node'}
         self.num_tasks_per_node = 1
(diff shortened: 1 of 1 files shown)


cscs-checks/system/slurm/slurm.py
```diff
@@ -16,7 +16,7 @@ def __init__(self):
                               'kesch:cn', 'kesch:pn',
                               'arolla:cn', 'arolla:pn',
                               'tsa:cn', 'tsa:pn']
-        self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu']
+        self.valid_prog_environs = ['PrgEnv-cray']
```
teojgo, line 19:
```suggestion
        self.valid_prog_environs = ['builtin']
```

cscs-checks/system/slurm/slurm.py
```diff
@@ -35,7 +35,7 @@ def __init__(self):
...
```
teojgo, line 38:
```suggestion
        self.valid_prog_environs = ['builtin']
```

vkarak:
@teojgo I had the same thought, but I was reluctant to do it, because some tests do have a build step, so it's better not to rely on the `builtin` environment. But I have not a strong feeling about this. I could use `builtin` for the run-only tests and `PrgEnv-cray` for the rest.

teojgo:
> @teojgo I had the same thought, but I was reluctant to do it, because some tests do have a build step, so it's better not to rely on the `builtin` environment. But I have not a strong feeling about this. I could use `builtin` for the run-only tests and `PrgEnv-cray` for the rest.

Ah ok, then it is fine for me.

vkarak:
CI is blocking due to #1298.

PR #1301, opened 2020-05-07
[bugfix] Fix possible deadlock in PBS-based scheduler backends when a job is cancelled immediately after submission

### LINKED ISSUE #1298: Unittest `test_cancel` deadlocks for PBS-based schedulers
The unittest deadlocks since the test might by cancelled before the job starts executing. In that case, no stdout or stderr is produced and therefore `done` == False:
https://github.com/eth-cscs/reframe/blob/625ce43d724c13343c6c693066a0db3ee4fb3c5e/reframe/core/schedulers/torque.py#L97

teojgo:
* Create the stdout, stderr files if the don't exist
  to make the torque scheduler treat the job as finished.

Fixes #1298

### DIFF
--- a/reframe/core/schedulers/pbs.py
+++ b/reframe/core/schedulers/pbs.py
@@ -51,6 +51,7 @@ class PbsJobScheduler(sched.JobScheduler):
         self._job_submit_timeout = rt.runtime().get_option(
             f'schedulers/@{self.registered_name}/job_submit_timeout'
         )
+        self._cancelled = False
 
         # Optional part of the job id refering to the PBS server
         self._pbs_server = None
@@ -140,7 +141,7 @@ class PbsJobScheduler(sched.JobScheduler):
             time.sleep(next(intervals))
 
     def cancel(self, job):
-        self._is_cancelling = True
+        self._cancelled = True
 
         # Recreate the full job id
         jobid = str(job.jobid)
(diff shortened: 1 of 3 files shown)


teojgo:
> I think the problem is more fundamental, because this deadlock could happen in normal operation. I guess that a better solution would be to fix that in the PBS/Torque backends, such that if the job is cancelled, we should not do the additional check for its stdout/stderr files to mark it as done.

Then maybe open a new issue to find a fundamental solution, while temporarily have this one so that it does not block the CI?

vkarak:
>  Then maybe open a new issue to find a fundamental solution, while temporarily have this one so that it does not block the CI?

No, the solution is quite easy. Set an `is_cancelling` flag that will be set when `cancel()` is called and the `finished()` should simply do the additional check only if this flag is not set.

reframe/core/schedulers/pbs.py
```diff
@@ -156,6 +158,8 @@ def finished(self, job):
         with os_ext.change_dir(job.workdir):
             done = os.path.exists(job.stdout) and os.path.exists(job.stderr)
 
+        done = done or getattr(self, '_is_cancelling', False)
+
```
vkarak, line 162:
```suggestion
            output_ready = os.path.exists(job.stdout) and os.path.exists(job.stderr)

        done = self._cancelled or ouput_ready
```

reframe/core/schedulers/torque.py
```diff
@@ -95,4 +95,5 @@ def finished(self, job):
             stdout = os.path.join(job.workdir, job.stdout)
             stderr = os.path.join(job.workdir, job.stderr)
             done = os.path.exists(stdout) and os.path.exists(stderr)
+            done = done or getattr(self, '_is_cancelling', False)
```
vkarak, line 98: Use a similar approach as the suggestion above.

PR #1304, opened 2020-05-08
[test] New Slurm setup for Cp2kGpuCheck

lucamar:
I propose to run the `Cp2kGpuCheck` with 6 tasks per node and 2 cpus per task, as it is slightly more efficient with respect to the current Slurm setup.
I would also like to test the `CP2K` modules recently provided with the latest stable release 7.1: however, the new modules are not yet default.
Should we change directly the modules loaded in the existing checks or should we rather create a new set of checks for the new versions with a new variant (e.g. `dev`) and keep the existing ones for the default modules only?
The latter would be probably a better solution to keep track of the performance of the new vs. old release of applications: the new variant would result in having `Cp2kCpuCheck_small_dev`, etc...

Please let me know your opinion and I would adjust the checks accordingly.


lucamar:
@jenkins-cscs retry dom

lucamar:
The check on `dom:mc` failed:
```
Reason: job blocked error: [jobid=1015576] job cancelled because it was blocked due to a perhaps non-recoverable reason: ReqNodeNotAvail,  UnavailableNodes:nid00[002,038-040,043,404-405,418-419]
```

lucamar:
@jenkins-cscs retry dom

lucamar:
No chance to have 6 `mc` nodes on Dom:
```
dom101:~$ scontrol show nodes nid00[000-011,032-043,052-055,404-407,416-419,401-402] | grep -A 5 ActiveFeatures=mc | grep State
   State=MAINT+DRAIN ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
   State=MAINT+DRAIN ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
   State=RESERVED ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
   State=IDLE ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
   State=IDLE ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
   State=IDLE ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
   State=IDLE ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
   State=MAINT ThreadsPerCore=2 TmpDisk=0 Weight=1 Owner=N/A MCS_label=N/A
```

vkarak:
Hi @lucamar, is this setup valid also for the default CP2K version? If yes, I would suggest to keep the default module and invoke reframe with `-M` to map to the new version.

lucamar:
@vkarak Thanks for your comment: yes, the new setup is also valid for the current default module.
Of course we can use the module mapping to test the new module, as I have being doing so far: the only disadvantage is that you need to issue two different commands for `gpu` and `mc` checks, since the modules to be mapped are different. 
I thought that it would be interesting to keep track of the performance of the new modules of supported applications before changing the defaults: maybe we can find a different way to do that...

vkarak:
@lucamar What if we parametrized the test on the module name?

PR #1311, opened 2020-05-11
[bugfix] Fix formatting of search path info line in output

vkarak:


vkarak:
@jenkins-cscs retry all

PR #1315, opened 2020-05-13
[ci] Create deployment Jenkinsfile

teojgo:


JenkinsfileDeploy
```diff
@@ -0,0 +1,42 @@
...
+    node('dom') {
+        dir('production_repo') {
+            sh(returnStatus: true,
+               script: """#!/bin/bash -l
+                          module load daint-gpu
+                          module load hub
+                          git clone git@github.com:eth-cscs/production.git
+                          cd production/easybuild/easyconfigs/r/reframe
+                          git checkout -b reframe/${version}
+                          sed 's/3.0-dev5/${version}/g' reframe-3.0-dev5.eb > reframe-${version}.eb
```
vkarak, line 33: I wouldn't hardcode the base version here, since we will need to change it from time to time.
Also why are you doing a sed across the whole file and don't just change the version? If we have the version hardcoded elsewhere in the easyconfig, it's just badly written.

JenkinsfileDeploy
```diff
@@ -0,0 +1,42 @@
...
+                          git add reframe-${version}.eb
+                          git commit -m "Add recipe for Reframe version ${version}"
```
vkarak, line 35:
```suggestion
                          git commit -m "Add recipe for ReFrame version ${version}"
```

teojgo:
@jenkins-cscs retry daint

PR #1324, opened 2020-05-18
[feat] Insert 'noqa' annotations in long lines generated by the configuration conversion tool 

### LINKED ISSUE #1284: Insert 'noqa' annotations in long lines generated by the configuration conversion tool

rsarm:
Closes #1284

### DIFF
--- a/reframe/core/config.py
+++ b/reframe/core/config.py
@@ -528,7 +528,7 @@ def convert_old_config(filename, newfilename=None):
                 f"by ReFrame based on '{filename}'.\n#\n\n"
                 f"site_configuration = {util.ppretty(converted)}\n")
 
-    contents = '\n'.join(l if len(l) < 80 else f'{l}  # noqa'
+    contents = '\n'.join(l if len(l) < 80 else f'{l}  # noqa: E501'
                          for l in contents.split('\n'))
 
     if newfilename:
--- a/reframe/core/config.py
+++ b/reframe/core/config.py
@@ -528,6 +528,9 @@ def convert_old_config(filename, newfilename=None):
                 f"by ReFrame based on '{filename}'.\n#\n\n"
                 f"site_configuration = {util.ppretty(converted)}\n")
 
+    contents = '\n'.join(l if len(l) < 80 else f'{l}  # noqa'
+                         for l in contents.split('\n'))
+
     if newfilename:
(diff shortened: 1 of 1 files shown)


reframe/core/config.py
```diff
@@ -528,6 +528,9 @@ def handler_list(handler_config):
                 f"by ReFrame based on '{filename}'.\n#\n\n"
                 f"site_configuration = {util.ppretty(converted)}\n")
 
+    contents = '\n'.join(l if len(l) < 80 else f'{l}  # noqa'
+                         for l in contents.split('\n'))
```
vkarak, line 532: Why not construct the `contents` above directly as a list so as not have to split it here? Also the line limit is 79 according to PEP8. Being pedantic, I would suggest to put the comment on the next multiple-of-indentation-depth column, if that does not make the code overly complicated.

vkarak: Also the annotation should be `# noqa: E501

ekouts: > Why not construct the contents above directly as a list so as not have to split it here?

You mean change `util.ppretty`? Do we want to return a list of lines, instead of the "pretty" string?

vkarak: Oh, ok you're right! I've just seen the content being set above as a string and didn't noticed the `ppretty`. Leave it as is then this part.

rsarm: if the next multiple-of-indentation position is separated by a single space from the end of the line, should we have the comment starting only after a single space? normally it's two spaces after.

rsarm: I think the limit is 79, including it https://www.python.org/dev/peps/pep-0008/#maximum-line-length

vkarak: > if the next multiple-of-indentation position is separated by a single space from the end of the line, should we have the comment starting only after a single space? normally it's two spaces after.

The minimum should be two as the standard says. But we can keep it simple and just add two spaces and that's it.

> I think the limit is 79, including it https://www.python.org/dev/peps/pep-0008/#maximum-line-length

You have it right, you have used `< 80` instead of `<= 80`. That's fine.

vkarak: @rsarm Just fix the `noqa` annotation with the error code and we're ready to merge it.

PR #1325, opened 2020-05-18
[feat] Add execution time profiling for the framework's pipeline

### LINKED ISSUE #97: Add execution time profiling for the framework
Internal issue: https://madra.cscs.ch/scs/reframe/issues/105

ekouts:
Before merging we have to figure out
1. What is the output going to look like. Here is an example of the output now

*Async Policy*
```
[       OK ] ( 1/10) T0 on generic:default using builtin
[          ] setup: 0.012 compile: 0.020 run: 0.290 sanity: 0.010 performance: 0.010 total: 0.341 
[       OK ] ( 2/10) T4 on generic:default using builtin
[          ] setup: 0.011 compile: 0.021 run: 0.177 sanity: 0.009 performance: 0.009 total: 0.227 
[       OK ] ( 3/10) T5 on generic:default using builtin
[          ] setup: 0.011 compile: 0.019 run: 0.173 sanity: 0.010 performance: 0.009 total: 0.223 
[       OK ] ( 4/10) T1 on generic:default using builtin
[          ] setup: 0.011 compile: 0.019 run: 0.178 sanity: 0.011 performance: 0.010 total: 0.230 
[     FAIL ] ( 5/10) T8 on generic:default using builtin
[          ] setup: 0.012 compile: - run: - sanity: - performance: - total: 0.012 
[     FAIL ] ( 6/10) T9
[          ] setup: - compile: - run: - sanity: - performance: - total: - 
[       OK ] ( 7/10) T6 on generic:default using builtin
[          ] setup: 0.012 compile: 0.021 run: 0.175 sanity: 0.010 performance: 0.009 total: 0.227 
[       OK ] ( 8/10) T3 on generic:default using builtin
[          ] setup: 0.012 compile: 0.020 run: 0.175 sanity: 0.010 performance: 0.009 total: 0.226 
[     FAIL ] ( 9/10) T2 on generic:default using builtin
[          ] setup: 0.011 compile: 0.019 run: 0.176 sanity: 0.010 performance: - total: 0.216 
[     FAIL ] (10/10) T7
[          ] setup: - compile: - run: - sanity: - performance: - total: - 
```
*Serial Policy*
```
...
[----------] started processing T6 (T6)
[ RUN      ] T6 on generic:default using builtin
[       OK ] ( 5/10) T6 on generic:default using builtin
[          ] setup: 0.012 compile: 0.021 run: 0.199 sanity: 0.014 performance: 0.011 total: 0.257 
[----------] finished processing T6 (T6)

[----------] started processing T2 (T2)
[ RUN      ] T2 on generic:default using builtin
[     FAIL ] ( 6/10) T2 on generic:default using builtin
[          ] setup: 0.013 compile: 0.022 run: 0.197 sanity: 0.012 performance: - total: 0.244 
[----------] finished processing T2 (T2)

[----------] started processing T3 (T3)
[ RUN      ] T3 on generic:default using builtin
[       OK ] ( 7/10) T3 on generic:default using builtin
[          ] setup: 0.012 compile: 0.022 run: 0.196 sanity: 0.012 performance: 0.010 total: 0.253 
[----------] finished processing T3 (T3)

[----------] started processing T7 (T7)
[ RUN      ] T7 on generic:default using builtin
[     FAIL ] ( 8/10) T7
[          ] setup: - compile: - run: - sanity: - performance: - total: - 
[----------] finished processing T7 (T7)
...
```
2. at what level the information should be collected. The timestamps are now in the level of `RegressionTask`, but maybe they should be in `RegressionTest`, or even the scheduler level if we want more accurate information in the future (waiting time, actual execution time, etc)

Closes #97


vkarak:
@ekouts I'd like the output to be more compact. Something like the following, where we print only the build, run and total time in the OK line?

```
[       OK ] (1/2) HelloMakeTest on generic:default using builtin [compile: 0.649s run: 0.167s total: 0.838s]
[       OK ] (2/2) hellocheck on generic:default using builtin [compile: 0.149s run: 0.154s total: 0.323s]
```

You can print the full timings addition at verbose level:

```
[       OK ] (1/2) HelloMakeTest on generic:default using builtin [compile: 0.649s run: 0.167s total: 0.838s]
==> setup: 0.010s compile: 0.649s run: 0.167s sanity: 0.006s performance: 0.005s total: 0.838s
[       OK ] (2/2) hellocheck on generic:default using builtin [compile: 0.149s run: 0.154s total: 0.323s]
==> setup: 0.009s compile: 0.149s run: 0.154s sanity: 0.005s performance: 0.005s total: 0.323s
```

vkarak:
We will also need a similar summary line at the end.

reframe/frontend/executors/__init__.py
```diff
@@ -130,6 +131,37 @@ def __init__(self, case, listeners=[]):
...
+            'start_sanity': None,
+            'finish_sanity': None,
+            'start_performance': None,
+            'finish_performance': None,
+            'start_cleanup': None,
+            'finish_cleanup': None
+        }
+
+    @property
+    def duration(self):
```
vkarak, line 151: I would also make that accept the phase.

reframe/frontend/executors/policies.py
```diff
@@ -26,6 +26,18 @@ def _cleanup_all(tasks, *args, **kwargs):
     # Remove cleaned up tests
     tasks[:] = [t for t in tasks if t.ref_count]
 
+def time_profiling(duration):
```
vkarak, line 29: I prefer sth like the following:

```suggestion
def format_duration(task):
```

reframe/frontend/executors/__init__.py
ekouts: If I change the timestamps to be taken in the `_safe_call()` function, what would `duration('run')` return? `timestamp['finish_run'] - timestamp['start_run' ]` or `timestamps['finish_wait'] - timestamps['start_run' ]`?

ekouts: Because in the output run is not the same as the internal `run` phase of the pipeline.

vkarak: I think that's entirely up to us. You can return duration of the whole if the phase is for example `'run+wait'`. We can give it any name.

ekouts: The `'run+wait'` is maybe inconsistent because if it is what is called `run` now.
```
timestamp['run+wait'] != timestamp['run'] + timestamp['wait']
```
But we print it as `run` duration.

vkarak: Don't confuse what you report and how you get it. We can name it as we like when we are reporting it. I can see where the confusion is coming from, since the "run" phase from the [user perspective](https://reframe-hpc.readthedocs.io/en/latest/pipeline.html#the-run-phase) is the whole thing (run_spawn + run_wait) but inside the framework the `run` is essentially the run_spawn. Implementation-wise now, it's easier if the `phase` argument refers to what we do internally, i.e., the run_spawn subphase.

ekouts: Ok then, if you agree I will name:
```
duration['run'] = timestamp['finish_run'] - timestamp['start_run']
duration['wait'] = timestamp['finish_wait'] - timestamp['start_wait']
duration['run_pipeline'] = timestamp['finish_wait'] - timestamp['start_run']
duration['compile_pipeline'] = timestamp['finish_compile_wait'] - timestamp['start_compile']
```
`run_pipeline` and `compile_pipeline` could also be `'running'` and `'waiting'`

vkarak: I would rather call the combined phases as `run_complete` and `compile_complete`.

ekouts: Ok, changing them to this, thanks. There a couple more small issues.
1. What should the `total` be after these changes?
Before it was the sum of all the individual "phases". Should we also make this `timestamp['finish_last_stage'] - timestamp['start_setup']` or as it was before (summing only the phases we report)?
2. If I insert the timestamps in `_safe_call()` instead of each individual function and add directly an end timestamp based on the _failed_stage in fail(), then (in case the test fails in the run_spawn) `duration['run_complete']` should be:
`duration['run_complete'] = timestamp['finish_run'] - timestamp['start_run']` instead of `timestamp['finish_wait'] - timestamp['start_run']`

vkarak: 1. The total should *not* be the sum. I wanted to comment that before. We lose the "other" in such a case. So it should be as you propose, `last_stage - setup_start`. But for last stage why not adding just a new `pipeline_end` timestamp, so that the total will always be `pipelilne_end - setup_start`? Can you also inverse the timestamp names? `setup_start` instead of `start_setup`?

2. Using my suggestion above, you can always pick the `pipeline_end` if you can't find the end-of-phase timestamp.

reframe/frontend/executors/policies.py
```diff
@@ -27,6 +27,28 @@ def _cleanup_all(tasks, *args, **kwargs):
     tasks[:] = [t for t in tasks if t.ref_count]
 
 
+def format_duration(duration):
```
vkarak, line 30: I would make this a method of the task and rename it `pipeline_timings(phases)`, where `phases` is the list of phases or pseudo-phases whose timings you want to format. The idea is that you can call this a second time and print all the timings with the verbose log level.

reframe/frontend/executors/policies.py
```diff
@@ -27,6 +27,28 @@ def _cleanup_all(tasks, *args, **kwargs):
...
+    mssg = '[compile: '
+    if duration['compile_complete']:
+        mssg += f"{duration['compile_complete']:.3f} "
+    else:
+        mssg += 'n/a '
```
vkarak, line 35: I would avoid all these ifs here by defining a nested function to do this custom formating:

```python
def _tf(t):
    return f'{t:.3f}' if t else 'n/a'

msg = f"[compile: {_tf(duration['compile_complete'])}"
```

reframe/frontend/executors/policies.py
```diff
@@ -204,14 +228,16 @@ def on_task_run(self, task):
...
+            self.printer.status('ERROR', mssg, just='right')
         else:
             self._remove_from_running(task)
-            self.printer.status('FAIL', task.check.info(), just='right')
+            self.printer.status('FAIL', mssg, just='right')
 
     def on_task_success(self, task):
-        self.printer.status('OK', task.check.info(), just='right')
+        mssg = f'{task.check.info()} {format_duration(task.duration)}'
+        self.printer.status('OK', mssg, just='right')
```
vkarak, line 240: You will need to  print another status line here with all the timings at verbose level.

docs/pipeline.rst
```diff
@@ -129,3 +129,14 @@ ReFrame tries to keep concurrency high by maintaining as many test cases as poss
 When the `concurrency limit <config_reference.html#.systems[].partitions[].max_jobs>`__ is reached, ReFrame will first try to free up execution slots by checking if any of the spawned jobs have finished, and it will fill that slots first before throttling execution.
 
 ReFrame uses polling to check the status of the spawned jobs, but it does so in a dynamic way, in order to ensure both responsiveness and avoid overloading the system job scheduler with excessive polling.
+
+Time Profiling of the Pipeline
+------------------------------
+
+Since version 3.0, ReFrame keeps track of the time a test spends in each phases of the pipeline, but it has some limitations.
```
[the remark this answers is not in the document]
vkarak: I will fine tune a bit this section.

reframe/frontend/executors/__init__.py
```diff
@@ -162,15 +220,23 @@ def _notify_listeners(self, callback_name):
...
             self._current_stage = fn.__name__
+            self._timestamp[f'{self._current_stage}_start'] = time.time()
 
         try:
             with logging.logging_context(self.check) as logger:
                 logger.debug('entering stage: %s' % self._current_stage)
-                return fn(*args, **kwargs)
+                ret = fn(*args, **kwargs)
+                self._timestamp[f'{self._current_stage}_finish'] = time.time()
+                self._timestamp['pipeline_end'] = time.time()
```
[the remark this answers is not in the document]
ekouts: Because I think it doesn't work correctly for the cases that we want to raise an error, and not just return the result.

reframe/frontend/executors/__init__.py
```diff
@@ -132,61 +132,50 @@ def __init__(self, case, listeners=[]):
...
+            return f'{t:.3f}' if t else 'n/a'
 
-        try:
-            res['total'] = (
-                self._timestamp['pipeline_end'] -
-                self._timestamp['setup_start']
-            )
-        except KeyError:
-            res['total'] = None
+        mssg = ''
```
vkarak, line 166: Better use everywhere `msg` instead of `mssg`. This is the convention we use throughout the framework.

reframe/frontend/executors/__init__.py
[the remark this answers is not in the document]
vkarak: Strange, I'll check that.

reframe/frontend/executors/policies.py
```diff
@@ -116,15 +94,36 @@ def on_task_exit(self, task):
         pass
 
     def on_task_failure(self, task):
-        mssg = f'{task.check.info()} {format_duration(task.duration)}'
+        timings = task.pipeline_timings(['compile_complete',
+                                         'run_complete',
+                                         'total'])
+        mssg = f'{task.check.info()} [{timings}]'
```
vkarak, line 100: I don't like very much that we repeat this thing three times.

reframe/frontend/executors/__init__.py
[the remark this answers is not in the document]
vkarak: Can  you post an example? I can't see the reason.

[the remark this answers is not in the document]
vkarak: This, which is very similar to what we do here works as expected:

```python
import os
import sys


def foo():
    try:
        os.stat(sys.argv[1])
        return 'hello'
    except OSError as e:
        raise Exception('error') from e
    finally:
        print('finally')


if __name__ == '__main__':
    print(foo())
```

`finally` will be printed in any case.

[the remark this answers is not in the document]
ekouts: You are right, I was getting an error and I thought it was because finally doesn't work as expected. After all, the problem is that if I put the lines
```
self._timestamp[f'{self._current_stage}_finish'] = time.time()
self._timestamp['pipeline_end'] = time.time()
```
in the finally section we will get the timestamp after the `self.fail()`. Fail will print the final message `[     FAIL ] ... [compile: ...]` and this will cause a reframe error.

reframe/frontend/executors/policies.py
ekouts: We could have pipeline_timings get just an argument `verbose: boolean` and return the normal/verbose message instead of passing the list.

docs/tutorial_basic.rst
```diff
@@ -177,51 +177,51 @@ If everything is configured correctly for your system, you should get an output
...
-   [==========] Finished on Fri May 29 10:31:15 2020
+  [ReFrame Setup]
+   version:           3.0-dev7 (rev: 85ca676f)
+   command:           './bin/reframe -C tutorial/config/settings.py -c tutorial/example1.py -r'
+   launched by:       user@daint104
+   working directory: '/path/to/reframe'
+   settings file:     'tutorial/config/settings.py'
+   check search path: (R) '/path/to/reframe/tutorial/example1.py'
+   stage directory:   '/path/to/reframe/stage'
+   output directory:  '/path/to/reframe/output'
```
vkarak, line 188: This should be indented on space more.

docs/tutorial_basic.rst
```diff
@@ -177,51 +177,51 @@ If everything is configured correctly for your system, you should get an output
...
+  [ RUN      ] Example1Test on daint:gpu using PrgEnv-intel
+  [ RUN      ] Example1Test on daint:gpu using PrgEnv-pgi
+  [ RUN      ] Example1Test on daint:mc using PrgEnv-cray
+  [ RUN      ] Example1Test on daint:mc using PrgEnv-gnu
+  [ RUN      ] Example1Test on daint:mc using PrgEnv-intel
+  [ RUN      ] Example1Test on daint:mc using PrgEnv-pgi
+  [----------] finished processing Example1Test (Simple matrix-vector multiplication example)
+
+  [----------] waiting for spawned checks to finish
+  [       OK ] ( 1/12) Example1Test on daint:login using PrgEnv-intel [compile: 1.940s run: 20.747s total: 23.778s]
```
vkarak, line 209: Did you run them on Daint or on Dom?

reframe/frontend/executors/__init__.py
```diff
@@ -162,15 +209,23 @@ def _notify_listeners(self, callback_name):
...
+            self._timestamps[f'{self._current_stage}_start'] = time.time()
 
         try:
             with logging.logging_context(self.check) as logger:
                 logger.debug('entering stage: %s' % self._current_stage)
-                return fn(*args, **kwargs)
+                ret = fn(*args, **kwargs)
+                self._timestamps[f'{self._current_stage}_finish'] = time.time()
+                self._timestamps['pipeline_end'] = time.time()
+                return ret
```
vkarak, line 220: We can avoid the duplication for each case, by using a context manager to do this thing. It doesn't have to be public, just define it nested in this function. Then you can simply write:

```python
with update_timestamps():
    return fn(*args, **kwargs)
```

reframe/frontend/executors/policies.py
```diff
@@ -94,13 +94,36 @@ def on_task_exit(self, task):
...
-            self.printer.status('FAIL', task.check.info(), just='right')
+            self.printer.status('FAIL', msg, just='right')
+
+        timings = task.pipeline_timings(['setup',
+                                         'compile_complete',
+                                         'run_complete',
+                                         'sanity',
+                                         'performance',
+                                         'total'])
+        getlogger().verbose(f"==> {timings}")
```
vkarak, line 112: A shortcut function like that would be better, similar to what you have proposed:

```suggestion
        getlogger().verbose(f"==> {task.pipeline_timings_all()}")
```

reframe/frontend/executors/policies.py
```diff
@@ -94,13 +94,36 @@ def on_task_exit(self, task):
         pass
 
     def on_task_failure(self, task):
+        timings = task.pipeline_timings(['compile_complete',
+                                         'run_complete',
+                                         'total'])
+        msg = f'{task.check.info()} [{timings}]'
```
vkarak, line 100:
```suggestion
        msg = f'{task.check.info()} [{task.pipeline_timings_basics()}]'
```

docs/tutorial_basic.rst
ekouts: On dom, I didn't change  the hostname on purpose but I can change it if you prefer it.

vkarak: No, you did it right. I was just wondering about the large `run` times. It means that your jobs might have been queueing for some time...

Issue #1318, opened 2020-05-13
Print configuration file that was loaded in the "ReFrame setup" section of the output


vkarak:
This is quite important now that the config file can be loaded from multiple locations or set with environment variables.

PR #1326, opened 2020-05-19
[feat] Print ReFrame environment variables in debug output

### LINKED ISSUE #1319: Print all ReFrame related environment variables in debug output

teojgo:
Fixes #1319


reframe/frontend/cli.py
```diff
@@ -32,6 +32,31 @@
 from reframe.frontend.printer import PrettyPrinter
 
 
+REFRAME_ENV = [
```
ekouts, line 35: I also found these RFM_* variables, should they be added?
```
RFM_MODULES_PRELOAD
RFM_RECURSIVE
RFM_TIMESTAMP
```
If you took the list from the manpage maybe it should be updated there as well.

teojgo: Yes I got from the manpage

vkarak: I don't like this duplication. This mapping is stored already inside the `ArgumentParser`. Also we could instead print the configuration parameters as stored in the site configuration and mark in parenthesis where did the configuration parameter got its value. NB: I don't know how easy is this currently, but we should investigate. I do believe that this duplication can be avoided.

teojgo: @vkarak, I see that it is possible to retrieve the env variables from the argparser._options_map. Should I proceed like that? Or see the alternative that you are saying?

vkarak: @teojgo I think you should just print the environment variables. If somebody needs to check the exact value of a configuration option, she can use the `--show-config` option. If we also fix #1318 we will have the complete information needed to debug. As for the debug output regarding this, I'd suggest to print also the environment variables that are not set by showing them as `<not set>`, e.g., `RFM_PREFIX=<not set>`

vkarak: And also remove the current "Command-line configuration" debug message.

reframe/frontend/cli.py
```diff
@@ -53,6 +53,13 @@ def format_check(check, detailed):
     return '\n'.join(lines)
 
 
+def format_env(envvars):
+    ret = 'ReFrame environment variables:\n'
```
vkarak, line 57:
```suggestion
    ret = '[ReFrame Environment]\n'
```

So as to match the `[ReFrame setup]`.

reframe/frontend/cli.py
```diff
@@ -53,6 +53,13 @@ def format_check(check, detailed):
...
+    notset = "<not set>"
+    ret += '\n'.join([f'    {e}={os.environ.get(e, notset)}' for e in envvars])
```
vkarak, line 59:
```suggestion
    ret += '\n'.join([f'  {e}={os.getenv(e, notset)}' for e in envvars])
```

To match the  following section.

reframe/frontend/cli.py
```diff
@@ -53,6 +53,13 @@ def format_check(check, detailed):
     return '\n'.join(lines)
 
 
+def format_env(envvars):
+    ret = 'ReFrame environment variables:\n'
+    notset = "<not set>"
```
vkarak, line 58:
```suggestion
    notset = '<not set>'
```

reframe/frontend/cli.py
vkarak: Print also the `RFM_INSTALL_PREFIX`. This is not in the environment variables returned by the argument parser.

reframe/frontend/cli.py
```diff
@@ -416,13 +423,14 @@ def main():
...
 
+    printer.debug(format_env(options.env_vars))
+
     # Setup the check loader
     loader = RegressionCheckLoader(
         load_path=site_config.get('general/0/check_search_path'),
         recurse=site_config.get('general/0/check_search_recursive'),
         ignore_conflicts=site_config.get('general/0/ignore_check_conflicts')
     )
-    printer.debug(argparse.format_options(options))
```
vkarak, line 425: I think  you can delete the  `format_options()` altogether.

reframe/frontend/argparse.py
```diff
@@ -57,6 +57,12 @@ def cmd_options(self):
         '''Options filled in by command-line'''
         return self.__namespace
 
+    @property
+    def env_vars(self):
+        '''Environment variables related to ReFrame'''
+        return sorted(
+            [v[0].split()[0] for v in self.__option_map.values() if v[0]])
```
vkarak, line 64:
```suggestion
        return sorted(
            v[0].split()[0] for v in self.__option_map.values() if v[0]
        )

```

reframe/frontend/cli.py
teojgo: @vkarak should I keep the order or just print it above every env variable?

vkarak: Since you sort them, it'd be betted if that one was in the right order. I guess it's more correct in any case to sort them here (the strings) instead of sorting them in `env_vars`. Also you don't need the list here. You can call `join()` on a generator.

PR #1327, opened 2020-05-19
[doc] Update the required Sphinx version in conf.py

teojgo:
Fixes the build failure on readthedocs


docs/conf.py
```diff
@@ -38,7 +38,7 @@
 
 # If your documentation needs a minimal Sphinx version, state it here.
 #
-needs_sphinx = '1.6.3'
+needs_sphinx = '3.0.0'
```
vkarak, line 41: Aha, and I was always wondering why rtd was not using  a more recent version of Sphinx :-)

PR #1331, opened 2020-05-21
[bugfix] Fix early abort when compile step fails with async policy

### LINKED ISSUE #1330: Tasks waiting to run are silently aborted if a later task fails compile stage, with async policy
This might be related to #1121 and/or #507 

When using the AsynchronousExecutionPolicy, a task failure at the compile if a tasks fails at the compile_wait step can lead to reframe exiting while earlier queued jobs are still waiting to run.

I think the root cause is in `reframe/frontend/executors/policies.py` -> `class AsynchronousExecutionPolicy(ExecutionPolicy, TaskEventListener)` ->   `def _reschedule(self, task):`, `compile_wait` can throw a variety of compile-time errors that should fail the task, but not abort the reframe run.

I have a patch that I think fixes this, will post an PR shortly

sleak-lbl:
Fixes #1330 

Exceptions raised from build tasks within `compile_wait` with the async policy are not caught by `reschedule_task`, and so bubble up and stop the reframe process. Any queued jobs then get abandoned.

This PR catches exceptions that should cause the test, but not the whole reframe run, to fail

### DIFF
--- a/reframe/frontend/executors/policies.py
+++ b/reframe/frontend/executors/policies.py
@@ -11,8 +11,7 @@ import time
 
 from datetime import datetime
 
-from reframe.core.exceptions import (TaskDependencyError, TaskExit, 
-                                     BuildError, PipelineError)
+from reframe.core.exceptions import (TaskDependencyError, TaskExit)
 from reframe.core.logging import getlogger
 from reframe.frontend.executors import (ExecutionPolicy, RegressionTask,
                                         TaskEventListener, ABORT_REASONS)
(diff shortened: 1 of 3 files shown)


sleak-lbl:
I've just had an early abort with this patch in place, so it's clearly not the soltion (or at least, the full solution) .. I've added a WIP marker to the PR, digging a bit more now

vkarak:
Check my comment in the issue for a possible cause of this bug.

vkarak:
@sleak-lbl Can you test this PR now with a real workload?

sleak-lbl:
tested with real workload and verified it worked (and, as far as I can tell, worked for the right reasons) - so I think it is good to go

vkarak:
Thanks for testing and confirming @sleak-lbl !

PR #1332, opened 2020-05-21
[feat] Add test description to failure report

sleak-lbl:
Sometimes the test name is enough to identify a failed test, but in other cases I've found it helpful to have the test description available too - the PR adds the test description to the failure report output, if a description is present


reframe/frontend/statistics.py
```diff
@@ -86,6 +86,7 @@ def failure_report(self):
 
             report.append(line_width * '-')
             report.append('FAILURE INFO for %s %s' % (check.name, retry_info))
+            if check.descr: report.append('  * Test Description: %s' % check.descr)
```
vkarak, line 89: Just print the description unconditionally. `check.descr` will always be non-empty currently, since by default it's the test's name.

PR #1333, opened 2020-05-21
[feat] Add support for UPC language and launchers

sleak-lbl:
testing functionality of UPC and UPC++ code requires reframe to know how to compile and run UPC code. This PR adds those features

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -352,7 +352,7 @@ class SingleSource(BuildSystem):
     #: The automatically detected extensions are the following:
     #:
     #:   - C: `.c` and `.upc`.
-    #:   - C++: `.cc`, `.cp`, `.cxx`, `.cpp`, `.CPP`, `.c++`, and `.C`.
+    #:   - C++: `.cc`, `.cp`, `.cxx`, `.cpp`, `.CPP`, `.c++` and `.C`.
     #:   - Fortran: `.f`, `.for`, `.ftn`, `.F`, `.FOR`, `.fpp`, `.FPP`, `.FTN`,
     #:     `.f90`, `.f95`, `.f03`, `.f08`, `.F90`, `.F95`, `.F03` and `.F08`.
     #:   - CUDA: `.cu`.
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -351,8 +351,8 @@ class SingleSource(BuildSystem):
     #: automatically based on the extension of the source file.
     #: The automatically detected extensions are the following:
     #:
-    #:   - C: `.c`.
-    #:   - C++: `.cc`, `.cp`, `.cxx`, `.cpp`, `.CPP`, `.c++` and `.C`.
+    #:   - C: `.c`, `upc`.
+    #:   - C++: `.cc`, `.cp`, `.cxx`, `.cpp`, `.CPP`, `.c++`, and `.C`.
     #:   - Fortran: `.f`, `.for`, `.ftn`, `.F`, `.FOR`, `.fpp`, `.FPP`, `.FTN`,
     #:     `.f90`, `.f95`, `.f03`, `.f08`, `.F90`, `.F95`, `.F03` and `.F08`.
     #:   - CUDA: `.cu`.
@@ -433,7 +433,7 @@ class SingleSource(BuildSystem):
 
     def _guess_language(self, filename):
         _, ext = os.path.splitext(filename)
-        if ext in ['.c']:
+        if ext in ['.c', '.upc']:
             return 'C'
 
         if ext in ['.cc', '.cp', '.cxx', '.cpp', '.CPP', '.c++', '.C']:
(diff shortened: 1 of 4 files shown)


reframe/core/buildsystems.py
```diff
@@ -351,8 +351,8 @@ class SingleSource(BuildSystem):
     #: automatically based on the extension of the source file.
     #: The automatically detected extensions are the following:
     #:
-    #:   - C: `.c`.
-    #:   - C++: `.cc`, `.cp`, `.cxx`, `.cpp`, `.CPP`, `.c++` and `.C`.
+    #:   - C: `.c`, `upc`.
```
vkarak, line 354:
```suggestion
    #:   - C: `.c`, `.upc`.
```

reframe/core/buildsystems.py
```diff
@@ -351,8 +351,8 @@ class SingleSource(BuildSystem):
...
+    #:   - C++: `.cc`, `.cp`, `.cxx`, `.cpp`, `.CPP`, `.c++`, and `.C`.
```
vkarak, line 355: I would skip the comma here for consistency with the rest.

reframe/core/launchers/mpi.py
```diff
@@ -22,6 +22,34 @@ def command(self, job):
...
+@register_launcher('upcrun')
+class UpcrunLauncher(JobLauncher):
+    '''Launcher for UPC applications.'''
+
+    def command(self, job):
+        cmd = ['upcrun']
+        if job.num_tasks:
+            if job.num_tasks_per_node:
+                num_nodes = job.num_tasks // job.num_tasks_per_node
+                cmd += ['-N', str(num_nodes)]
```
vkarak, line 34: We need a blank line here after the if block.

reframe/core/launchers/mpi.py
```diff
@@ -22,6 +22,34 @@ def command(self, job):
...
+            cmd += ['-n', str(job.num_tasks)]
```
vkarak, line 35: And a blank line here.

reframe/core/launchers/mpi.py
```diff
@@ -22,6 +22,34 @@ def command(self, job):
...
+        return cmd
+
+
+@register_launcher('upcxx-run')
+class UpcxxrunLauncher(JobLauncher):
```
vkarak, line 40: Why not derive from `UpcrunLauncher` and just change the basic command?

vkarak:
@sleak-lbl We will also need an entry about these launchers in documentation [here](https://reframe-hpc.readthedocs.io/en/latest/config_reference.html#systems-.partitions-.launcher).

reframe/core/buildsystems.py
sleak-lbl: add missing dot to upc extension

reframe/core/buildsystems.py
sleak-lbl: sounds good, done

reframe/core/launchers/mpi.py
sleak-lbl: sounds good, done

reframe/core/launchers/mpi.py
sleak-lbl: sounds good, done

reframe/core/launchers/mpi.py
sleak-lbl: good question, I see arguments each way:

 - UPC and UPC++ are similar-but-distinct UPC is a language, an extension of C, while UPC++ is a C++ library. I suspect that the compilers and launchers for each share a lot of code underneath, but they're logically distinct
 - on the other hand, the launchers have (as far as I can tell) the same syntax, so from an implementation perspective they could be considered as belonging to a common class

I can be happy with either implementation, which do you prefer?

vkarak:
@sleak-lbl I think something got wrong with your updating this PR. It seems that you merged with your own modified master and pushed here.

reframe/core/launchers/mpi.py
vkarak: Thinking a bit more about it, I believe it's better to leave it as is. Other launchers, too, may end up using the same arguments, but it doesn't mean that they have to derive from a common ancestor, since they're conceptually distinct.

reframe/core/launchers/mpi.py
```diff
@@ -22,6 +22,38 @@ def command(self, job):
...
+    '''Launcher for UPC++ applications.'''
+
+    def command(self, job):
+        cmd = ['upcxx-run']
+        if job.num_tasks:
```
vkarak, line 47: You don't need this if. It's guaranteed at this point that `job.num_tasks` will be a positive integer.

vkarak:
I forced push to this PR and fixed the commits. It seems that you have rebased locally over master and then you pushed here doing essentially a merge.

vkarak:
@jenkins-cscs test this please

sleak-lbl:
oops, I was trying to rebase to the eth-cscs master, must have used the wrong one .. thanks for fixing

PR #1334, opened 2020-05-25
[feat] Print the config file loaded by ReFrame

### LINKED ISSUE #1318: Print configuration file that was loaded in the "ReFrame setup" section of the output

teojgo:
Fixes #1318

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -444,7 +444,7 @@ def main():
     print_infoline('launched by',
                    f"{os_ext.osuser() or '<unknown>'}@{socket.gethostname()}")
     print_infoline('working directory', repr(os.getcwd()))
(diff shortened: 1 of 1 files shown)


reframe/frontend/cli.py
```diff
@@ -444,6 +444,7 @@ def print_infoline(param, value):
     print_infoline('launched by',
                    f"{os_ext.osuser() or '<unknown>'}@{socket.gethostname()}")
     print_infoline('working directory', repr(os.getcwd()))
+    print_infoline('configuration', f"'{site_config.filename}'")
```
vkarak, line 447:
```suggestion
    print_infoline('settings file', f'{site_config.filename!r}')
```

PR #1338, opened 2020-05-26
[feat] Replace scheduler-specific command-line options

### LINKED ISSUE #486: Replace scheduler-specific command-line options
The `--partition`, `--account`, `--nodelist` etc. options are actually scheduler-specific and, in my opinion, they should not be part of the command-line interface. Instead, the should be passed through a more generic mechanism such as the `--job-option` switch. There are advantages and disadvantages to what I suggest here:

Pros:
- Cleaner command-line
- Less implementation overhead for scheduler backends, since the user is responsible for passing the correct backend options in the special case that he wants to modify how an access to a specific partition is gained. A backend implementor need not be an expert of the scheduler.
- More uniform behaviour for the scheduler backends. For example, now we assume, at the supposedly scheduler-agnostic frontend, a SLURM-specific syntax for the node list.
- Avoid subtle bugs such as the one described in #75.

Cons:
- We lose the automatic mapping of the command-line options to the different backends. This is only true for the `--partition` one. The rest are simply ignored by all other schedulers except Slurm.
- Perhaps more typing

rsarm:
Closes #486


ekouts:
@jenkins-cscs retry all

docs/manpage.rst
```diff
@@ -289,32 +289,33 @@ Options controlling job submission
 
 .. option:: -A, --account=NAME
 
-   Submit test-related jobs using the account ``NAME``.
-   This option is relevant only for the Slurm backend and translates to Slurm's ``--account`` option and it precedes any options specified in the :js:attr:`access` system partition configuration parameter.
```
vkarak, line 293: All these descriptions should remain. You should add the the deprecation node + what this option is equivalent to. For example, for Slurm, this is equivalent to `-J account=NAME`.

reframe/frontend/cli.py
```diff
@@ -197,26 +197,21 @@ def main():
     # Run options
     run_options.add_argument(
         '-A', '--account', action='store',
-        help='Use ACCOUNT for submitting jobs (Slurm)'
-    )
+        help='*deprecated*, please use --job-option instead')
```
vkarak, line 200: We should keep the help message and add also the deprecation highlight saying how to replace it.

```suggestion
        help="*deprecated*, please use '-J account=ACCOUNT'")
```

reframe/frontend/cli.py
```diff
@@ -618,7 +613,41 @@ def print_infoline(param, value):
             exec_policy.sched_reservation = options.reservation
             exec_policy.sched_nodelist = options.nodelist
             exec_policy.sched_exclude_nodelist = options.exclude_nodes
-            exec_policy.sched_options = options.job_options
+            if options.account:
```
vkarak, line 616: All these deprecation warnings should go a bit up, because now they appear only when you try to run the tests. They should appear always, even when you are listing them.

vkarak:
I agree with @ekouts. Documentation of the `-J` option must be elaborated. I don't know what kind of unit tests we could add though.

reframe/frontend/cli.py
rsarm: At this point we can't know which scheduler is being used, right? If we put `-J account=ACCOUNT`, it would be slurm syntax, even if the user is running PBS. That's why I preferred a generic message.

vkarak: I'd put both: `-J account=NAME or -J q=NAME`. To save space you can also set `metavar='NAME'` so that you can use that name in your help message.

reframe/frontend/cli.py
```diff
@@ -564,6 +570,32 @@ def print_infoline(param, value):
...
         options.flex_alloc_nodes = options.flex_alloc_nodes or 'idle'
+        if options.account:
+            printer.warning(f"`--account' is deprecated and "
+                            f"will be removed in the future; you should "
+                            f"use `-J account={options.account}'")
+
+        if options.partition:
+            printer.warning(f"`--partition' is deprecated and "
+                            f"will be removed in the future; you should "
+                            f"use `-J partition={options.partition}'"
```
vkarak, line 581:
```suggestion
                            f"use `-J partition={options.partition}' "
```

reframe/frontend/cli.py
```diff
@@ -197,26 +197,27 @@ def main():
     # Run options
     run_options.add_argument(
         '-A', '--account', action='store',
-        help='Use ACCOUNT for submitting jobs (Slurm)'
-    )
+        help="Use ACCOUNT for submitting jobs (Slurm) "
+             "*deprecated*, please use '-J account=NAME'")
```
vkarak, line 201:
```suggestion
             "*deprecated*, please use '-J account=ACCOUNT'")
```

PR #1342, opened 2020-05-27
[feat] Add configuration option to disable the ReqNodeNotAvail check

### LINKED ISSUE #1049: ReqNodeNotAvail reason code causes ReFrame to incorrectly kill jobs on Cori at NERSC
The Slurm scheduler on the Cori system at NERSC takes a long time to compute backfill - up to 15 minutes per backfill cycle, due to having a huge number of enqueued jobs (several 10,000s). During this process, Slurm will often temporarily change the reason code for an enqueued job to `ReqNodeNotAvail, UnavailableNodes:nid[...]`, where the list of nodes is essentially the entire system, around 11,000 nodes. Although ReFrame attempts to check if the nodes are healthy (cf. #303 and resulting PR), this check seems to fail on Cori while the backfill cycle is running. I don't understand why it fails, but wonder if Slurm refuses to service the `scontrol show node` command with such a large list of nodes.

I have worked around this problem by simply removing `ReqNodeNotAvail` from the `_cancel_reasons` list in `reframe/core/schedulers/slurm.py` in NERSC's fork of ReFrame, but this is probably not the best long-term solution. If it is true that Slurm is refusing to service the large `scontrol` command, can you think of a way to work around this problem?

vkarak:
Fixes #1049.

FYI @bcfriesen you can now either `export RFM_IGNORE_REQNODENOTAVAIL=yes` or set the `ignore_reqnodenotavail` configuration parameter to `True`.


bcfriesen:
Great! Thanks @vkarak. I'll test this this week.

vkarak:
@bcfriesen Did you have time to test this one? If not, since it'd require you to move to 3.0 right away essentially, I could merge it anyway and if you see problems we can fix them in a later PR.

bcfriesen:
Hi @vkarak, no I'm still migrating to the 3.0 series. Please feel free to merge this and I'll open an issue if it doesn't work as expected. Thanks!

PR #1343, opened 2020-05-27
[feat] Add configuration option for using a login shell for jobs

### LINKED ISSUE #1198: Enable bash login option per system partition
In one of the partitions of my system, all jobs are required to have `-l` option in bash, as in `#!/bin/bash -l` (see [our documentation](https://www2.cisl.ucar.edu/resources/computational-systems/casper/example-scripts-casper-users)).

I see that the job scheduler and shell support that option (probably because of #882), but unfortunately the RegressionTest class does not 

https://github.com/eth-cscs/reframe/blob/d17ff88486096ef285727151ba3462146f54fc78/reframe/core/pipeline.py#L1212

That means that all ReFrame jobs submitted to that partition fail. Is this a feature that could get implemented?

vkarak:
Closes #1198. 

@davidedelvento FYI


davidedelvento:
I haven't tested it, but looks good. Thanks!

PR #1344, opened 2020-05-28
[test] Convert JacobiNoToolHybrid to pipeline hook format

jgphpc:
Will fix [UES-842](https://jira.cscs.ch/browse/UES-842)


cscs-checks/tools/profiling_and_debugging/notool.py
```diff
@@ -78,29 +82,36 @@ def __init__(self, lang):
         elif lang == 'F90':
             self.reference_lang = (0.17, -0.6, None, 's')
 
-    def setup(self, partition, environ, **job_opts):
-        super().setup(partition, environ, **job_opts)
+    @rfm.run_before('compile')
+    def set_(self):
```
vkarak, line 86:
```suggestion
    def setflags(self):
```

Perhaps better?

PR #1347, opened 2020-05-29
[feat] Add node list information for jobs submitted with the Torque backend

### LINKED ISSUE #1213: Get node list information for jobs submitted with the Torque backend
Sample implementation from @ZQyou:

```python
    def _get_nodelist(self, job, jobstat):
        # exec_host = o0580/0-27+o0444/0-27+o0345/0-27
        nodelist_match = re.search('exec_host = (?P<nodelist>\S+)', jobstat.stdout)
        if nodelist_match:
            nodelist = [ x.split('/')[0] for x in nodelist_match.group('nodelist').split('+') ]
            nodelist.sort()
            job.nodelist = nodelist
```

rsarm:
Close #1213


reframe/core/schedulers/torque.py
```diff
@@ -35,6 +35,14 @@
 class TorqueJobScheduler(PbsJobScheduler):
     TASKS_OPT = '-l nodes={num_nodes}:ppn={num_cpus_per_node}'
 
+    def _get_nodelist(self, job, stdout):
+        nodelist_match = re.search(r'exec_host = (?P<nodelist>\S+)', stdout)
+        if nodelist_match:
+            nodelist = [x.split('/')[0]
+                        for x in nodelist_match.group('nodelist').split('+')]
+            nodelist.sort()
```
vkarak, line 43: @ZQyou Do you want the list sorted here? Because in the Slurm backend we don't do that.

reframe/core/schedulers/torque.py
```diff
@@ -35,6 +35,14 @@
 class TorqueJobScheduler(PbsJobScheduler):
     TASKS_OPT = '-l nodes={num_nodes}:ppn={num_cpus_per_node}'
 
+    def _get_nodelist(self, job, stdout):
```
vkarak, line 38: For symmetry with the Slurm backend rename this as follows:

```suggestion
    def _set_nodelist(self, job, nodespec):
```

Also you should do the `re.search()` outside the function and check here if the nodelist is already set. Have a look at the corresponding function in the Slurm backend.

reframe/core/schedulers/torque.py
```diff
@@ -53,6 +61,8 @@ def _update_state(self, job):
         if completed.returncode != 0:
             raise JobError('qstat failed: %s' % completed.stderr, job.jobid)
 
+        self._get_nodelist(job, completed.stdout)
+
```
vkarak, line 65: Remove this blank line.

PR #1356, opened 2020-06-02
[feat] Add configuration option for always emitting `--nodes` in Slurm job scripts

### LINKED ISSUE #1057: Emit `-N` in Slurm job script
TACC requires the `-N` option to be emitted in the job script, but ReFrame does not emit it currently. We should give the option to enable that. The workaround currently is the following:

```python
@rfm.run_after('setup')
def set_num_nodes(self):
    self.job.options = [f'-N {self.num_tasks // self.num_tasks_per_node}']
```

FYI @rtmclay

rsarm:
Closes #1057


reframe/schemas/config.json
```diff
@@ -274,6 +274,7 @@
                 "type": "object",
                 "properties": {
                     "name": {"type": "string"},
+		    "emit_num_nodes": {"type": "boolean"},
```
ekouts, line 277:
```suggestion
                    "emit_num_nodes": {"type": "boolean"},
```
You should use spaces instead of tabs.

reframe/schemas/config.json
```diff
@@ -419,6 +420,7 @@
         "modes/target_systems": ["*"],
         "schedulers/job_submit_timeout": 60,
         "schedulers/target_systems": ["*"],
+	"schedulers/emit_num_nodes": false,
```
ekouts, line 423: Same as above.

vkarak:
I will remove the environment variable. It's a bit confusing. I will only keep the configuration parameter.

reframe/schemas/config.json
```diff
@@ -273,6 +273,8 @@
             "items": {
                 "type": "object",
                 "properties": {
+                    "name": {"type": "string"},
```
ekouts, line 276: name shouldn't be here twice, right?

vkarak: Good catch! Bad conflict resolution :-)

PR #1358, opened 2020-06-02
[doc] Add documentation about the cleanup stage for tests with dependencies

ekouts:
The cleaning up of tests with dependent tests was not explained in the documentation, so I added a small section.


docs/pipeline.rst
```diff
@@ -91,6 +91,7 @@ The Cleanup Phase
 During this final stage of the pipeline, the test's resources are cleaned up.
 More specifically, if the test has finished successfully, all interesting test files (build/job scripts, build/job script output and any user-specified files) are copied to ReFrame's output directory and the stage directory of the test is deleted.
 
+.. note:: This phase might be deferred in case a test has dependent test (see `Cleaning up stage files <dependencies.html#cleaning-up-stage-files>`__ for more details).
```
ekouts, line 94: @vkarak Since you are fine tuning this, I just noticed, this could be better like this I think
```suggestion
.. note:: This phase might be deferred in case a test has dependent tests (see `Cleaning up stage files <dependencies.html#cleaning-up-stage-files>`__ for more details).
```
or "a dependent test"

vkarak: I was about to do the same :-) Or just "dependents." I'm now trying to figure out a better way in Sphinx to reference subsections, because for full pages you can nicely do

```
:doc:`dependencies`
```

PR #1363, opened 2020-06-08
[bugfix] Combine access and job options for slurm '--constraint'

### LINKED ISSUE #1348: Passing a `--constraint` job option should be combined with the  `access` one for Slurm
If a `--contraint=A` option is used in the `access` field of the `SystemPartition` then a corresponding line is emitted inside the sbatch script.
Passing additionally a constraint in the `options` field of the `Job` class, such as  `--contstraint=B` emits a single line in the sbatch script after the line corresponding to `--constraint=A`.
Therefore, slurm takes into account only the `B` constraint.
The two constraint options should be combined and passed in a single line as: `--constraint=A,B`

teojgo:
* The `--constraint` options passed via `access` and `job.options`
  are combined in the resulting job script.

* Add unittests to test the expected behavior.

Fixes #1348


reframe/core/schedulers/slurm.py
```diff
@@ -195,15 +195,34 @@ def emit_preamble(self, job):
...
+                self._format_option(','.join(constraints), '--constraint={0}'))
 
         preamble.append(self._format_option(hint, '--hint={0}'))
         prefix_patt = re.compile(r'(#\w+)')
         for opt in job.options:
-            if not prefix_patt.match(opt):
-                preamble.append('%s %s' % (self._prefix, opt))
-            else:
-                preamble.append(opt)
+            if not opt.strip().startswith(('-C', '--constraint')):
```
ekouts, line 221: I am not sure about this, but shouldn't you also check whether it starts with `#SBATCH -C` or `#SBATCH --constraint`?

teojgo: No, because the `#SBATCH` prefix is added by ReFrame

reframe/core/schedulers/slurm.py
```diff
@@ -195,15 +195,34 @@ def emit_preamble(self, job):
...
 
         for opt in job.sched_access:
-            preamble.append('%s %s' % (self._prefix, opt))
+            if not opt.strip().startswith(('-C', '--constraint')):
+                preamble.append('%s %s' % (self._prefix, opt))
+
+        constraints = []
+        option_parser = ArgumentParser()
+        option_parser.add_argument('-C', '--constraint')
+        parsed_args, _ = option_parser.parse_known_args(job.sched_access)
```
vkarak, line 204:
```suggestion
        parsed_options, _ = constraint_parser.parse_known_args(job.sched_access)
```

reframe/core/schedulers/slurm.py
```diff
@@ -195,15 +195,34 @@ def emit_preamble(self, job):
...
+        if parsed_args.constraint:
+            constraints += [parsed_args.constraint]
```
vkarak, line 206:
```suggestion
            constraints.append(parsed_options.constraint)
```

reframe/core/schedulers/slurm.py
```diff
@@ -195,15 +195,34 @@ def emit_preamble(self, job):
...
+
+        # NOTE: Here last of the passed --constraint job options is taken
+        # into account in order to respect the behavior of slurm.
+        parsed_args, _ = option_parser.parse_known_args(job.options)
+        if parsed_args.constraint:
+            constraints += [parsed_args.constraint]
```
vkarak, line 212: Same comments as above.

reframe/core/schedulers/slurm.py
```diff
@@ -195,15 +195,34 @@ def emit_preamble(self, job):
...
+
+        if constraints:
+            preamble.append(
+                self._format_option(','.join(constraints), '--constraint={0}'))
```
vkarak, line 216: Move the closing parenthesis to the next line.

reframe/core/schedulers/slurm.py
vkarak, line 221:
```suggestion
            if opt.strip().startswith(('-C', '--constraint')):
                # Constraints are already processed
                continue
            
            # all the rest unindented here
```

vkarak: Also are you sure that passing a tuple to `startswith()` works as you intend here?? It's not documented.

unittests/test_schedulers.py
```diff
@@ -370,6 +370,26 @@ def test_no_empty_lines_in_preamble(minimal_job):
         assert line != ''
 
 
+def test_combined_access_constraint(make_job, slurm_only):
+    job = make_job(sched_access=['--constraint=c1'])
+    job.options = ['--constraint=c2,c3']
```
vkarak, line 375: Use the `-C` option here to exercise that path as well.

```suggestion
    job.options = ['-C c2,c3']
```

unittests/test_schedulers.py
```diff
@@ -370,6 +370,26 @@ def test_no_empty_lines_in_preamble(minimal_job):
...
+    prepare_job(job)
+    with open(job.script_filename) as fp:
+        assert re.search(r'--constraint=c1,c2,c3$', fp.read(), re.MULTILINE)
+        assert re.search(r'--constraint=(c1|c2,c3)$', fp.read(),
+                         re.MULTILINE) is None
```
vkarak, line 380:
```suggestion
        assert re.search(r'(?m)--constraint=c1,c2,c3$', fp.read())
        assert re.search(r'(?m)--constraint=(c1|c2,c3)$', fp.read()) is None
```

unittests/test_schedulers.py
```diff
@@ -370,6 +370,26 @@ def test_no_empty_lines_in_preamble(minimal_job):
...
+
+
+def test_combined_access_multiple_constraints(make_job, slurm_only):
+    job = make_job(sched_access=['--constraint=c1'])
+    job.options = ['--constraint=c2', '--constraint=c3']
```
vkarak, line 385:
```suggestion
    job.options = ['--constraint=c2', '-C c3']
```

unittests/test_schedulers.py
```diff
@@ -370,6 +370,26 @@ def test_no_empty_lines_in_preamble(minimal_job):
...
+    prepare_job(job)
+    with open(job.script_filename) as fp:
+        assert re.search(r'--constraint=c1,c3$', fp.read(), re.MULTILINE)
+        assert re.search(r'--constraint=(c1|c2|c3)$', fp.read(),
+                         re.MULTILINE) is None
```
vkarak, line 390: See my comments above.

reframe/core/schedulers/slurm.py
teojgo: Yes, I did not know either, it is mentioned here: https://docs.python.org/3/library/stdtypes.html#str.startswith that the prefix can be a tuple of prefixes to try.

teojgo:
@vkarak in which section do you think I should add the a documentation note about this?

vkarak:
@teojgo The point that @ekouts suggested above should be ok.

docs/manpage.rst
```diff
@@ -341,6 +341,10 @@ Options controlling job submission
    If ``key`` starts with ``-`` or ``#``, the option will be passed verbatim to the job script.
```
vkarak, line 341: This sentence is also not accurate.

PR #1371, opened 2020-06-11
[feat] Add bootstrap script to install dependencies of ReFrame

### LINKED ISSUE #1365: Install and run reframe directly from Github
The idea of this installation is to have ReFrame using the system Python and make it ready to use as soon as you clone it.

vkarak:
This really facilitates the processes of getting ready to run when you clone ReFrame from Github. You need only to do

```
./bootstrap.sh
```

or

```
./bootstrap.sh +docs
```

This will `pip install` all the dependencies inside the ReFrame source directory under `external/`. ReFrame's main "binaries" are adapted to look into this directory first for their dependencies. This solution does not preclude any of the other installation methods.

Fixes #1365.

Still to do:

- [x] Update README
- [x] Update documentation
- [x] Update deployment script
- [x] Update CI script


vkarak:
To test this PR, start from an empty virtual environment, so that ReFrame cannot find any of its dependencies, in case you have already installed them.

bootstrap.sh
```diff
@@ -0,0 +1,18 @@
...
+#
+# Bootstrap script for running ReFrame from source
+#
+# Run once before the first run.
+
+pip3 install --upgrade pip
+pip3 install -r requirements.txt --prefix=external/
+
+if [ x"$1" == x"+docs" ]; then
+    pip3 install -r docs/requirements.txt --prefix=external/
```
victorusu, line 17: I think we miss the actual documentation creation or a mechanism to also build the documentation.
For me
```
./bootstrap.sh +docs
```
means bootstrap reframe with or plus documentation.
But the documentation is not built.

vkarak: Good point, I'll do that.

vkarak: Done

vkarak:
@teojgo That should be possible, but I will need to separate where I install the external dependencies, e.g., `external/py36` and `external/py38`. Do you have any other suggestion?

teojgo:
> @teojgo That should be possible, but I will need to separate where I install the external dependencies, e.g., `external/py36` and `external/py38`. Do you have any other suggestion?

You can have the default python command as `python3` and the user can specify another if needed. Then since you get `pyver` and use it for external dependencies it will get the correct one.

vkarak:
@teojgo The `pyver` is used exclusively for retrieving the right `pip` that was installed by `ensurepip`. All ReFrame's dependencies are then placed unconditionally under `external/`. Notice that I'm using `--target` and not `--prefix`. Prefix differentiates over the python version through the `site-packages`. I could do that, but that would require me to update all the files that update the `sys.path`. Why do you need this distinction? With your solution, you would have to re-bootstrap ReFrame for every Python version, whereas with the current one, you don't need to do so. The only problem I see is that when one of the packages installed does not run with a different Python version than the one used to bootstrap ReFrame, which in this case we should take it into account when writing the `requirements.txt`.

teojgo:
> @teojgo The `pyver` is used exclusively for retrieving the right `pip` that was installed by `ensurepip`. All ReFrame's dependencies are then placed unconditionally under `external/`. Notice that I'm using `--target` and not `--prefix`. Prefix differentiates over the python version through the `site-packages`. I could do that, but that would require me to update all the files that update the `sys.path`. Why do you need this distinction? With your solution, you would have to re-bootstrap ReFrame for every Python version, whereas with the current one, you don't need to do so. The only problem I see is that when one of the packages installed does not run with a different Python version than the one used to bootstrap ReFrame, which in this case we should take it into account when writing the `requirements.txt`.

It's ok. The problem is when you have an alternative installation of python on your system that the standard python3. But I believe this is specific to my local installation where I have to make the differentiation. Otherwise it's fine.

vkarak:
@teojgo From our offline discussion I understood what you actually want and I'm gonna do it.

Issue #1087, opened 2019-12-05
Flexible node allocation fails with local scheduler


vkarak:
With #1074 we allowed local scheduler to support flexible allocation, but this is actually crashing now:

```diff
diff --git a/unittests/resources/checks/hellocheck.py b/unittests/resources/checks/hellocheck.py
index 62abd1d..7956eb8 100644
--- a/unittests/resources/checks/hellocheck.py
+++ b/unittests/resources/checks/hellocheck.py
@@ -9,6 +9,7 @@ class HelloTest(rfm.RegressionTest):
         self.descr = 'C Hello World test'

         # All available systems are supported
+        self.num_tasks = 0
         self.valid_systems = ['*']
         self.valid_prog_environs = ['*']
         self.sourcepath = 'hello.c'
```

```
Command line: ./bin/reframe -c unittests/resources/checks/hellocheck.py -r
Reframe version: 2.21-dev2
Launched by user: karakasv
Launched on host: dhcp-133-23.cscs.ch
Reframe paths
=============
    Check prefix      :
    Check search path : 'unittests/resources/checks/hellocheck.py'
    Stage dir prefix     : /Users/karakasv/Repositories/reframe/stage/
    Output dir prefix    : /Users/karakasv/Repositories/reframe/output/
    Perf. logging prefix : /Users/karakasv/Repositories/reframe/perflogs
[==========] Running 1 check(s)
[==========] Started on Thu Dec  5 18:45:48 2019

[----------] started processing hellocheck (C Hello World test)
[ RUN      ] hellocheck on generic:login using builtin-gcc
[     FAIL ] hellocheck on generic:login using builtin-gcc
[----------] finished processing hellocheck (C Hello World test)

[  FAILED  ] Ran 1 test case(s) from 1 check(s) (1 failure(s))
[==========] Finished on Thu Dec  5 18:45:48 2019

==============================================================================
SUMMARY OF FAILURES
------------------------------------------------------------------------------
FAILURE INFO for hellocheck
  * System partition: generic:login
  * Environment: builtin-gcc
  * Stage directory: /Users/karakasv/Repositories/reframe/stage/generic/login/builtin-gcc/hellocheck
  * Node list: <None>
  * Job type: local (id=None)
  * Maintainers: ['VK']
  * Failing phase: run
  * Reason: unexpected error: 'str' object has no attribute 'is_available'
Traceback (most recent call last):
[...]

PR #1374, opened 2020-06-15
[feat] Allow flexible node allocation to select nodes in any state

### LINKED ISSUE #1032: Add a `maint` mode in `--flex-alloc-nodes`
This will essentially select nodes that are in `MAINT` state.

teojgo:
Fixes #1032


vkarak:
@teojgo Can you please fix the title of the PR?

docs/manpage.rst
```diff
@@ -341,6 +341,10 @@ Options controlling job submission
    If ``key`` starts with ``-`` or ``#``, the option will be passed verbatim to the job script.
    Otherwise, ReFrame will add ``-`` or ``--`` as well as the directive corresponding to the current scheduler.
    This option will be emitted after any options specified in the :js:attr:`access` system partition configuration parameter.
+   Especially for the Slurm scheduler, constraint options, i.e ``-J constraint=value``, ``-J C=value``, ``-J --constraint=value``, ``-J -C=value` are going to be combined with the corresponding ones specified in the :js:attr:`access` system partition configuration parameter.
+   If multiple constraint options are specified with separate key-value pairs, only the last one is going to be taken into account.
+   For multiple combined constraints use the ``-J constraint=value1,value2`` syntax.
+   Note that the above is not valid if ``key`` starts with ``#`` in which case the option is going to be passed verbatim to the job script.
```
vkarak, line 347: This is from another PR.

docs/manpage.rst
```diff
@@ -364,6 +368,8 @@ If no node can be selected, the test will be marked as a failure with an appropr
      It is therefore possible that the number of tasks assigned does not correspond to the actual idle nodes.
 
      This is the default policy.
+
+   - ``maint``: Flexible tests will be assigned as many tasks as needed in order to span over the nodes of the node pool which are currently under maintenance.
```
vkarak, line 372: You should be more precise by mentioning the actual state that the nodes will be in.

reframe/core/schedulers/slurm.py
```diff
@@ -591,6 +615,10 @@ def is_available(self):
         return all([self._states == {'IDLE'}, self._partitions,
                     self._active_features, self._states])
 
+    def under_maintenance(self):
+        return all([self._states == {'MAINT'}, self._partitions,
+                    self._active_features, self._states])
```
vkarak, line 620: Couldn't we have a more generic method that would pick nodes that are in a specific state? I don't like this very much. It's too verbose and very specific.

teojgo:
> This PR is mixed with commits from other PRs. Can you fix it?

Yes I made a mistake when pushing because it have forked from another branch.

reframe/core/schedulers/slurm.py
teojgo: I wrote it like that so that it's more generic if we need to add different nodes that might use a different convention for maintenance. I forgot to add it to the abc `Node` class. Should I keep it like that?

vkarak: I think the most generic would be to use the argument passed `--flex-alloc-nodes` to denote the state. You just convert that to uppercase. So for example, one could also write `--flex-alloc-nodes=maint+drain`

teojgo: Yes it can be like that. I will have to change the logic which currently relies on sets and uses `+` as a separator for the list of states os a node.

vkarak: It can still rely on those I guess, we can discuss this offline.

reframe/core/schedulers/local.py
```diff
@@ -182,5 +182,5 @@ class _LocalNode(sched.Node):
     def __init__(self, name):
         self._name = name
 
-    def is_available(self):
-        return True
+    def in_state(self, state):
+        return NotImplemented
```
vkarak, line 186: This will break the local scheduler. The idea behind returning `True` previously was that the flex alloc nodes would work with `state=idle` for the local scheduler as well. I think we had a PR specifically addressing this in the past.

reframe/core/schedulers/slurm.py
```diff
@@ -611,13 +611,9 @@ def __eq__(self, other):
...
 
-    def is_available(self):
-        return all([self._states == {'IDLE'}, self._partitions,
-                    self._active_features, self._states])
-
-    def under_maintenance(self):
-        return all([self._states == {'MAINT'}, self._partitions,
-                    self._active_features, self._states])
+    def in_state(self, state):
+        return all([self._states == set(state.upper().split('+')),
```
vkarak, line 615:
```suggestion
        return all([self._states >= set(state.upper().split('+')),
```
A node in `MAINT+DRAIN` is both in `MAINT` and `DRAIN` states, so both queries should return true.

reframe/frontend/cli.py
```diff
@@ -260,8 +260,8 @@ def main():
     )
     run_options.add_argument(
         '--flex-alloc-nodes', action='store',
-        dest='flex_alloc_nodes', metavar='{all|idle|maint|NUM}', default=None,
-        help='Set strategy for the flexible node allocation (default: "idle").'
+        dest='flex_alloc_nodes', metavar='{STATE|ALL||NUM}', default=None,
+        help='Set strategy for the flexible node allocation (default: "IDLE").'
```
vkarak, line 264: I liked the lowercase here. The backends should convert them to uppercase or generally take care of the actual comparison.

unittests/test_schedulers.py
```diff
@@ -1070,14 +1070,14 @@ def test_str(slurm_node_allocated):
...
-    assert not slurm_node_drained.is_available()
-    assert not slurm_node_nopart.is_available()
+def test_slurm_node_in_state(slurm_node_allocated,
+                             slurm_node_idle,
+                             slurm_node_drained,
+                             slurm_node_nopart):
+    assert slurm_node_allocated.in_state('allocated')
+    assert slurm_node_idle.in_state('Idle')
+    assert slurm_node_drained.in_state('IDLE+Drain')
+    assert not slurm_node_nopart.in_state('IDLE')
```
vkarak, line 1080: You should check also the case when a node is in two states.

reframe/core/schedulers/local.py
teojgo: @vkarak Yes it was related to issue https://github.com/eth-cscs/reframe/issues/1087. Do you think that I should return True for idle and False otherwise?

reframe/core/schedulers/slurm.py
teojgo: I was thinking of this as well. The problem is that then you lose the option of asking for nodes that satisfy both states.

reframe/core/schedulers/local.py
vkarak: Yes, please do that.

reframe/core/schedulers/slurm.py
vkarak: Can you give an example?

teojgo: If you want to get the nodes that are both in `MAINT` and `DRAIN` then you cannot do it since according to the suggestion you will get also the ones that are either only  `MAINT` or `DRAIN`.

vkarak: Are you sure? Wouldn't `self._states` be `{'MAINT'}` or `{'DRAIN'}` in that case and what the user gave wouldn't be `{'MAINT', 'DRAIN'}`, so the `>=` will return false?

teojgo: You are right, my bad. I will implement this.

teojgo:
> I can still see stale changes from the constraint PR. Plus there is another error that exists also in master most probably. If you pass `--flex-alloc-nodes` without an argument, it will complain, whereas it should just assume `idle`.

But isn't the point of having it to pass an argument. There is no default if the argument is used. Should I put one?

vkarak:
@teojgo I think we say that the default is `idle` and [here](https://reframe-hpc.readthedocs.io/en/stable/manpage.html#cmdoption-flex-alloc-nodes) we imply that passing an argument is optional, which is not the case as it is now.

teojgo:
Yeah not using the option at all implies idle

teojgo:
But if you use the option an argument is required

vkarak:
Ok, now I get it. But we should still change the documentation of the command line option:

```
--flex-alloc-nodes=POLICY
blah blah blah

If this option is not specified, the default allocation policy for flexible tests is 'idle'.
```

PR #1377, opened 2020-06-18
[feat] Produce detailed JSON report for a regression testing session

### LINKED ISSUE #667: Create a JSON output report file
Similar to JUnit format, we can also create a JSON report file. I don't know if there's a standard schema for this, but in any case we can create our own.

rsarm:
Closes #667 

EDIT from @vkarak:

This PR adds to ReFrame the capability of producing a detailed JSON report whenever it runs. By default, the report is placed under `$HOME/.reframe/reports`. Each new ReFrame run generates a new report. A command line argument, configuration parameter and environment variable controlling this are added. All other reports generated from ReFrame should use this JSON. This is done already for the standard failure report but it is left for future work for the current performance report (the performance data of tests is already in the JSON). Here is an example output running `StreamTest`:

```javascript
{
  "session_info": {
    "cmdline": "./bin/reframe -c config/tresa.py -c cscs-checks/microbenchmarks/stream/stream.py --skip-prgenv-check --skip-system-check --performance-report -r",
    "config_file": "<builtin>",
    "data_version": "1.0",
    "hostname": "tresa.local",
    "prefix_output": "/Users/karakasv/Repositories/reframe/output",
    "prefix_stage": "/Users/karakasv/Repositories/reframe/stage",
    "user": "karakasv",
    "version": "3.1-dev2 (rev: 4c650425)",
    "workdir": "/Users/karakasv/Repositories/reframe",
    "time_start": "2020-07-24T00:04:00+0200",
    "time_end": "2020-07-24T00:04:02+0200",
    "time_elapsed": 1.8402886390686035,
    "num_cases": 1,
    "num_failures": 0
  },
  "runs": [
    {
      "num_cases": 1,
      "num_failures": 0,
      "runid": 0,
      "testcases": [
        {
          "build_stderr": "rfm_StreamTest_build.err",
          "build_stdout": "rfm_StreamTest_build.out",
          "description": "STREAM Benchmark",
          "environment": "builtin",
          "fail_reason": null,
          "fail_phase": null,
          "jobid": 75868,
          "job_stderr": "rfm_StreamTest_job.err",
          "job_stdout": "rfm_StreamTest_job.out",
          "name": "StreamTest",
          "maintainers": [
            "RS",
            "SK"
          ],
          "nodelist": [
            "tresa.local"
          ],
          "outputdir": "/Users/karakasv/Repositories/reframe/output/generic/default/builtin/StreamTest",
          "perfvars": [
            {
              "name": "triad",
              "reference": 0,
              "thres_lower": null,
              "thres_upper": null,
              "unit": "MB/s",
              "value": 17160.2
            }
          ],
          "result": "success",
          "stagedir": null,
          "scheduler": "local",
          "system": "generic:default",
          "tags": [
            "craype",
            "production"
          ],
          "time_compile": 0.7874908447265625,
          "time_performance": 0.0019030570983886719,
          "time_run": 0.988239049911499,
          "time_sanity": 0.0004889965057373047,
          "time_setup": 0.014341115951538086,
          "time_total": 1.8320262432098389
        }
      ]
    }
  ]
}
```

TODO: 

- [ ] Update documentation.


rsarm:
It looks something like this
```
[
    {
        "test": "UlimitCheck",
        "description": "Checking the output of ulimit -s in node.",
        "result": "fail",
        "system": "dom:gpu",
        "environment": "PrgEnv-gnu",
        "stagedir": "/users/sarafael/git/reframe/cscs-checks/prgenv/stage/dom/gpu/PrgEnv-gnu/UlimitCheck",
        "outputdir": "/users/sarafael/git/reframe/cscs-checks/prgenv/output/dom/gpu/PrgEnv-gnu/UlimitCheck",
        "nodelist": "nid00039",
        "jobtype": "batch job",
        "jobid": 1054524,
        "maintainers": [
            "RS",
            "CB"
        ],
        "tags": [
            "scs",
            "craype",
            "production"
        ],
        "retries": 0
    },
    {
        "test": "UlimitCheck",
        "description": "Checking the output of ulimit -s in node.",
        "result": "fail",
        "system": "dom:mc",
        "environment": "PrgEnv-gnu",
        "stagedir": "/users/sarafael/git/reframe/cscs-checks/prgenv/stage/dom/mc/PrgEnv-gnu/UlimitCheck",
        "outputdir": "/users/sarafael/git/reframe/cscs-checks/prgenv/output/dom/mc/PrgEnv-gnu/UlimitCheck",
        "nodelist": "nid00405",
        "jobtype": "batch job",
        "jobid": 1054525,
        "maintainers": [
            "RS",
            "CB"
        ],
        "tags": [
            "scs",
            "craype",
            "production"
        ],
        "retries": 0
    }
]
```

reframe/frontend/statistics.py
```diff
@@ -70,6 +72,75 @@ def retry_report(self):
...
+                    "tags": {
+                        "type": "array",
+                        "items": {"type": "string"}
+                    },
+                    "retries": {"type": "number"}
+                }
+            },
+        }
+        report = self.output_dict()
+        jsonschema.validate(instance=report, schema=schema)
```
ekouts, line 140: Why do we validate the output based on the schema? Maybe I am missing something, but since we are creating this it should be correct, right? Would be useful maybe as just a unittest?

reframe/frontend/statistics.py
```diff
@@ -70,6 +72,75 @@ def retry_report(self):
...
+                        if check.job and check.job.nodelist else '<None>')
+            job_type = 'local' if check.is_local() else 'batch job'
+            jobid = check.job.jobid if check.job else -1
+            if t.failed:
+                result = 'fail'
+            else:
+                result = 'success'
+
+            report.append({
+                'test': check.name,
```
ekouts, line 94:
```suggestion
                'name': check.name,
```
I would put `name` as the key, but the `test` key is also fine if you prefer it.

vkarak:
@rsarm Can you put an example of the output in case of a successful test?

rsarm:
@vkarak I didn't make too much distinction. The only difference is that the key `result` has the values `success` or `fail`.

vkarak:
Apart from @ekouts' comments, here are some additional ones:

* `nodelist` should be a list.
* Replace `jobtype` with `scheduler` and use the registered_name of the scheduler registered with this partition.

We need also the following fields:

* `failing_phase`: the phase that the test has failed or `none` if it hasn't.
* `failure_reason`: the failure reason.
* `build_stdout`: the basename of build's stdout file.
* `build_stderr`: the basename of build's stderr file.
* `job_stdout`: the basename of job's stdout file.
* `job_stderr`: the basename of job's stderr file.

rsarm:
Success:
```
[
    {
        "name": "UlimitCheck",
        "description": "Checking the output of ulimit -s in node.",
        "system": "dom:gpu",
        "environment": "PrgEnv-gnu",
        "tags": [
            "scs",
            "craype",
            "production"
        ],
        "maintainers": [
            "RS",
            "CB"
        ],
        "scheduler": "slurm",
        "job_stdout": "rfm_UlimitCheck_job.out",
        "job_stderr": "rfm_UlimitCheck_job.err",
        "jobid": 1061294,
        "nodelist": [
            "nid00038"
        ],
        "build_stdout": "rfm_UlimitCheck_build.out",
        "build_stderr": "rfm_UlimitCheck_build.err",
        "result": "success",
        "outputdir": "/users/sarafael/git/reframe/cscs-checks/prgenv/output/dom/gpu/PrgEnv-gnu/UlimitCheck"
    }
]
```

Fail:
```
[
    {
        "name": "UlimitCheck",
        "description": "Checking the output of ulimit -s in node.",
        "system": "dom:gpu",
        "environment": "PrgEnv-gnu",
        "tags": [
            "production",
            "scs",
            "craype"
        ],
        "maintainers": [
            "RS",
            "CB"
        ],
        "scheduler": "slurm",
        "job_stdout": "rfm_UlimitCheck_job.out",
        "job_stderr": "rfm_UlimitCheck_job.err",
        "jobid": 1061295,
        "nodelist": [
            "nid00038"
        ],
        "build_stdout": "rfm_UlimitCheck_build.out",
        "build_stderr": "rfm_UlimitCheck_build.err",
        "result": "fail",
        "failing_reason": "sanity error: pattern `Thr soft limit is unlimited' not found in `rfm_UlimitCheck_job.out'",
        "failing_phase": "sanity",
        "stagedir": "/users/sarafael/git/reframe/cscs-checks/prgenv/stage/dom/gpu/PrgEnv-gnu/UlimitCheck"
    }
]
```

ekouts:
@jenkins-cscs retry daint

ekouts:
@jenkins-cscs retry daint

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
 
         return '\n'.join(report)
 
+    @property
+    def report_dict(self):
+        _report = []
+        current_run = rt.runtime().current_run
+        for tl, tr in zip(self._tasks, range(current_run + 1)):
+            for t in tl:
```
vkarak, line 82: First, I suggest renaming `self._tasks` to `self._alltasks` and then rewrite the loop as follows:

```suggestion
        for run_no, run in enumerate(self._alltasks):
            for task in run:
```

reframe/schemas/report.json
```diff
@@ -0,0 +1,41 @@
...
+                "items": {"type": ["string", "null"]}
+            },
+            "try": {"type": "number"}
+        },
+        "additionalProperties": false,
+	"required": ["name", "description", "system", "environment", "stagedir",
+		     "outputdir", "nodelist", "scheduler", "jobid", "result",
+	             "failing_phase", "failing_reason", "build_stdout", "build_stderr",
+	             "job_stdout", "job_stderr", "maintainers", "tags", "try"]
+    }
```
vkarak, line 40: Also list the attributes in alphabetical order.

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
+                check = t.check
+                partition = check.current_partition
+                partname = partition.fullname if partition else None
+                environ_name = (check.current_environ.name
+                                if check.current_environ else None)
+                report = {
```
vkarak, line 88:
```suggestion
                entry = {
```

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
 
+    @property
+    def report_dict(self):
+        _report = []
+        current_run = rt.runtime().current_run
+        for tl, tr in zip(self._tasks, range(current_run + 1)):
+            for t in tl:
+                check = t.check
+                partition = check.current_partition
+                partname = partition.fullname if partition else None
```
vkarak, line 85:
```suggestion
                partfullname = partition.fullname if partition else None
```

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
+                    'job_stderr': None,
+                    'build_stdout': None,
+                    'build_stderr': None,
+                    'failing_reason': None,
+                    'failing_phase': None,
+                    'outputdir': None,
+                    'stagedir': None
+                }
+                if check.job:
+                    report['scheduler'] = check.job.scheduler.registered_name
```
vkarak, line 108: This you should get it from the partition, not the job and it should be there regardless if `check.job` is `None` or not.

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
+                    report['jobid'] = check.job.jobid
+                    report['nodelist'] = (check.job.nodelist
+                                          if check.job.nodelist
+                                          else [])
```
vkarak, line 112: I think `nodelist` is by default empty, isn't it? If not you should write this as follows:

```suggestion
                    entry['nodelist'] = check.job.nodelist or []
```

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
+                    report['job_stdout'] = check.job.stdout
+                    report['job_stderr'] = check.job.stderr
```
vkarak, line 114:
```suggestion
                    report['job_stdout'] = check.stdout.evaluate()
                    report['job_stderr'] = check.stderr.evaluate()
```

And they should go outside this `if`.

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
+
+                if check._build_job:
+                    report['build_stdout'] = check._build_job.stdout
+                    report['build_stderr'] = check._build_job.stderr
```
vkarak, line 118: We never access "private" fields! Rewrite as:

```suggestion
                    report['build_stdout'] = check.build_stdout.evaluate()
                    report['build_stderr'] = check.build_stderr.evaluate()
```

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
+
+                if t.failed:
+                    report['result'] = 'fail'
+                    if t.exc_info is not None:
+                        from reframe.core.exceptions import format_exception
```
vkarak, line 123: Import that in the beginning of the file.

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
+
+                        report['failing_reason'] = format_exception(
+                            *t.exc_info)
+                        report['failing_phase'] = t.failed_stage
+                        report['stagedir'] = check.stagedir
+                else:
+                    report['result'] = 'success'
+                    report['outputdir'] = check.outputdir
+
+                report['try'] = tr
```
vkarak, line 133: Replace with `run_no` as suggested below:

```suggestion
                report['run_no'] = run_no
```

reframe/schemas/report.json
```diff
@@ -0,0 +1,41 @@
...
+            "job_stderr": {"type": ["string", "null"]},
+            "maintainers": {
+                "type": "array",
+                "items": {"type": ["string", "null"]}
+            },
+            "tags": {
+                "type": "array",
+                "items": {"type": ["string", "null"]}
+            },
+            "try": {"type": "number"}
```
vkarak, line 33:
```suggestion
            "run_no": {"type": "number"}
```

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
         report.append('SUMMARY OF FAILURES')
         current_run = rt.runtime().current_run
-        for tf in (t for t in self.tasks(current_run) if t.failed):
-            check = tf.check
-            partition = check.current_partition
-            partname = partition.fullname if partition else 'None'
-            environ_name = (check.current_environ.name
-                            if check.current_environ else 'None')
+        for tf in (t for t in self.report_dict
+                   if t['result'] == 'fail' and t['try'] == current_run):
```
vkarak, line 149: First, rename `current_run` to `last_run`.

```suggestion
        for r in self.json():
            if r['result'] == 'success' or r['run_no'] != last_run:
                continue
```

reframe/frontend/statistics.py
```diff
@@ -70,46 +74,103 @@ def retry_report(self):
...
             retry_info = ('(for the last of %s retries)' % current_run
-                          if current_run > 0 else '')
+                          if 'retries' in tf.keys() else '')
```
vkarak, line 151: Why not:

```suggestion
                          if last_run > 0 else '')
```

unittests/test_policies.py
```diff
@@ -104,10 +107,28 @@ def num_failures_stage(runner, stage):
...
+    return schema
+
+
 def test_runall(make_runner, make_cases, common_exec_ctx):
     runner = make_runner()
     runner.runall(make_cases())
     stats = runner.stats
+    json_out = runner.stats.report_dict
+    schema = open_json_schema()
+    jsonschema.validate(json_out, schema)
```
vkarak, line 131:
```suggestion
    validate_report(runner.stats.json())
```

And do all necessary things in `validate_report()`.

unittests/test_policies.py
```diff
@@ -104,10 +107,28 @@ def num_failures_stage(runner, stage):
...
+    # Open and store the JSON schema for later validation
+    schema_filename = os.path.join(reframe.INSTALL_PREFIX, 'reframe',
+                                   'schemas', 'report.json')
+    with open(schema_filename) as fp:
+        try:
+            schema = json.loads(fp.read())
+        except json.JSONDecodeError as e:
+            raise ReframeFatalError(
+                f"invalid configuration schema: '{schema_filename}'"
+            ) from e
```
vkarak, line 120: You should not do any exception handling here, since you're in a unit test. If it crashes, the test will fail anyway.

unittests/test_policies.py
```diff
@@ -104,10 +107,28 @@ def num_failures_stage(runner, stage):
     return len([t for t in stats.failures() if t.failed_stage == stage])
 
 
+def open_json_schema():
+    # Open and store the JSON schema for later validation
+    schema_filename = os.path.join(reframe.INSTALL_PREFIX, 'reframe',
+                                   'schemas', 'report.json')
```
vkarak, line 113: Unit tests are meant to run from the top-level directory, so the following should be enough:

```suggestion
    schema_filename = 'reframe/schemas/runreport.json'
```

reframe/frontend/statistics.py
```diff
@@ -102,75 +98,74 @@ def report_dict(self):
...
+                        entry['failing_phase'] = t.failed_stage
+                        entry['stagedir'] = check.stagedir
                 else:
-                    report['result'] = 'success'
-                    report['outputdir'] = check.outputdir
+                    entry['result'] = 'success'
+                    entry['outputdir'] = check.outputdir
 
-                report['try'] = tr
+                entry['run_no'] = run_no
```
vkarak, line 127: I think it's better to call this `runid` instead. What do you think?

reframe/frontend/statistics.py
```diff
@@ -74,21 +71,20 @@ def retry_report(self):
...
                 environ_name = (check.current_environ.name
                                 if check.current_environ else None)
-                report = {
-                    'name': check.name,
+                entry = {
+                    'testname': check.name,
                     'description': check.descr,
-                    'system': partname,
+                    'system': partfullname,
                     'environment': environ_name,
```
vkarak: Set those too initially to `None` and reset their values later:

```python
partition, environ = check.current_partition, check.current_environ
if check.current_partition:
    entry['system'] = partition.fullname
    entry['scheduler'] = partition.scheduler.registered_name

if check.current_environ:
    entry['environment'] = environ.name
```

reframe/frontend/statistics.py
```diff
@@ -102,75 +98,74 @@ def report_dict(self):
...
                 }
                 if check.job:
-                    report['scheduler'] = check.job.scheduler.registered_name
-                    report['jobid'] = check.job.jobid
-                    report['nodelist'] = (check.job.nodelist
-                                          if check.job.nodelist
-                                          else [])
-                    report['job_stdout'] = check.job.stdout
-                    report['job_stderr'] = check.job.stderr
+                    entry['scheduler'] = partition.scheduler.registered_name
```
vkarak, line 106: This should not be here. Check my comment above.

reframe/frontend/statistics.py
```diff
@@ -102,75 +98,74 @@ def report_dict(self):
...
-        current_run = rt.runtime().current_run
-        for tf in (t for t in self.report_dict
-                   if t['result'] == 'fail' and t['try'] == current_run):
-            retry_info = ('(for the last of %s retries)' % current_run
-                          if 'retries' in tf.keys() else '')
-
+        last_run = rt.runtime().current_run
+        for r in self.json():
+            if r['result'] == 'success' or r['run_no'] != last_run:
+                continue
```
vkarak, line 144: Leave a blank line here.

reframe/frontend/statistics.py
```diff
@@ -102,75 +98,74 @@ def report_dict(self):
...
-                          (tf['name'], tf['environment'], tf['system']))
-            report.append("  * Reason: %s" % tf['failing_reason'])
+                          (r['testname'], r['environment'], r['system']))
+            report.append("  * Reason: %s" % r['failing_reason'])
 
-            if tf['failing_phase'] == 'sanity':
+            if r['failing_phase'] == 'sanity':
                 report.append('Sanity check failure')
-            elif tf['failing_phase'] == 'performance':
+            elif r['failing_phase'] == 'performance':
```
vkarak: Since you are rewriting this anyway, use f-strings everywhere.

unittests/test_policies.py
```diff
@@ -109,26 +109,23 @@ def num_failures_stage(runner, stage):
 
 def open_json_schema():
```
vkarak, line 110: I don't see the purpose of this function. I would merge it with the `validate_report()`.

vkarak:
@rsarm We should also report the times of the different pipeline stages in the report. You can get those from the `duration()` method of the `RegressionTask`.

rsarm:
> * Do you think we should cache the json report internally, so as not to recompute it every time we call `json()`?

@vkarak yes!

vkarak:
@rsarm In this case we should also have a way to reproduce it forcefully.

reframe/frontend/statistics.py
```diff
@@ -100,10 +97,24 @@ def json(self):
...
+                    'time_setup': t.duration('setup'),
+                    'time_compile': t.duration('complete'),
+                    'time_run': t.duration('t.duration'),
+                    'time_sanity': t.duration('sanity'),
+                    'time_performance': t.duration('performance'),
+                    'time_total': t.duration('total')
                 }
-                if check.job:
+                partition, environ = (check.current_partition,
+                                      check.current_environ)
```
vkarak, line 109: Better write it as follows, since it doesn't fit in single line:

```suggestion
                partition = check.current_partition
                environ = check.current_environ
```

reframe/frontend/statistics.py
```diff
@@ -100,10 +97,24 @@ def json(self):
                     'outputdir': None,
                     'stagedir': None,
                     'job_stdout': None,
-                    'job_stderr': None
+                    'job_stderr': None,
+                    'time_setup': t.duration('setup'),
+                    'time_compile': t.duration('complete'),
+                    'time_run': t.duration('t.duration'),
```
vkarak, line 103:
```suggestion
                    'time_run': t.duration('run_complete'),
```

reframe/frontend/statistics.py
```diff
@@ -100,10 +97,24 @@ def json(self):
                     'outputdir': None,
                     'stagedir': None,
                     'job_stdout': None,
-                    'job_stderr': None
+                    'job_stderr': None,
+                    'time_setup': t.duration('setup'),
+                    'time_compile': t.duration('complete'),
```
vkarak, line 102:
```suggestion
                    'time_compile': t.duration('complete_complete'),
```

reframe/frontend/statistics.py
```diff
@@ -100,10 +97,24 @@ def json(self):
...
+                    'time_run': t.duration('t.duration'),
+                    'time_sanity': t.duration('sanity'),
+                    'time_performance': t.duration('performance'),
+                    'time_total': t.duration('total')
                 }
-                if check.job:
+                partition, environ = (check.current_partition,
+                                      check.current_environ)
+                if check.current_partition:
```
vkarak, line 110:
```suggestion
                if partition:
```

reframe/frontend/statistics.py
```diff
@@ -100,10 +97,24 @@ def json(self):
...
+                    entry['system'] = partition.fullname
                     entry['scheduler'] = partition.scheduler.registered_name
+
+                if check.current_environ:
```
vkarak, line 114:
```suggestion
                if environ:
```

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,53 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://raw.githubusercontent.com/eth-cscs/reframe/master/schemas/runreport.json",
+    "title": "Validation schema for ReFrame's run report",
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "testname": {"type": "string"},
+            "description": {"type": ["string"]},
```
vkarak, line 10:
```suggestion
            "description": {"type": "string"},
```

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,53 @@
...
+            "system": {"type": ["string", "null"]},
```
vkarak, line 11: System can't be null.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,53 @@
...
+            "environment": {"type": ["string", "null"]},
+            "stagedir": {"type": ["string", "null"]},
+            "outputdir": {"type": ["string", "null"]},
+            "nodelist": {
+                "type": "array",
+                "items": {"type": "string"}
+            },
+            "scheduler": {"type": ["string", "null"]},
+            "jobid": {"type": ["number", "null"]},
+            "result": {"type": "string"},
```
vkarak, line 21: This should be an `enum` with the accepted values.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,53 @@
...
+            "failing_phase": {"type": ["string", "null"]},
+            "failing_reason": {"type": ["string", "null"]},
+            "build_stdout": {"type": ["string", "null"]},
+            "build_stderr": {"type": ["string", "null"]},
+            "job_stdout": {"type": ["string", "null"]},
+            "job_stderr": {"type": ["string", "null"]},
+            "maintainers": {
+                "type": "array",
+                "items": {"type": ["string", "null"]}
```
vkarak, line 30: The items can't be null.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,53 @@
...
+            },
+            "tags": {
+                "type": "array",
+                "items": {"type": ["string", "null"]}
```
vkarak, line 34: Same here.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,53 @@
...
+            },
+            "run_no": {"type": "number"},
```
vkarak, line 36:
```suggestion
            "runid": {"type": "number"},
```

But remember to change the code that generates the report.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,53 @@
...
+            "time_setup": {"type": "number"},
```
vkarak, line 37: @ekouts Is it guaranteed that the setup time is always there?

reframe/frontend/statistics.py
```diff
@@ -70,46 +71,110 @@ def retry_report(self):
...
+                entry['run_no'] = run_no
+
+                records.append(entry)
+
+        return records
+
+    def json_report(self):
+        with open('report.json', 'w') as fp:
+            json.dump(self.json(), fp, indent=4)
+
```
vkarak, line 147: Do that directly in the CLI.

reframe/frontend/statistics.py
```diff
@@ -70,46 +71,110 @@ def retry_report(self):
 
         return '\n'.join(report)
 
+    def json(self):
+        records = []
```
vkarak, line 75: As discussed, you can cache the json report.

reframe/schemas/runreport.json
ekouts: No, a testcase could fail even before the setup (because of a dependency). This could also be null.

reframe/schemas/runreport.json
rsarm: I think we need to allow null for cases like `BadSetupCheckEarly` of `unittests/resources/checks/frontend_checks.py`, that raise an exception after the setup. There ,`check.current_partition` is `None`.

vkarak: Yes, but `current_system` is never `None`. I think that you set the `system` first to the value of `current_system` instead of `None`.

reframe/frontend/cli.py
```diff
@@ -136,6 +137,10 @@ def main():
         help='Save ReFrame log files to the output directory',
         envvar='RFM_SAVE_LOG_FILES', configvar='general/save_log_files'
     )
+    output_options.add_argument(
+        '--keep-runreport', action='store_true', default=False,
+        help="Do not overwrite runreport.json",
+    )
```
vkarak, line 143: We still need to think about the way we should create the report files.

reframe/schemas/runreport.json
```diff
@@ -27,25 +30,25 @@
             "job_stderr": {"type": ["string", "null"]},
             "maintainers": {
                 "type": "array",
-                "items": {"type": ["string", "null"]}
+                "items": {"type": ["string"]}
```
vkarak, line 33:
```suggestion
                "items": {"type": "string"}
```

reframe/schemas/runreport.json
```diff
@@ -27,25 +30,25 @@
...
             },
             "tags": {
                 "type": "array",
-                "items": {"type": ["string", "null"]}
+                "items": {"type": ["string"]}
```
vkarak, line 37:
```suggestion
                "items": {"type": "string"}
```

reframe/schemas/runreport.json
```diff
@@ -27,25 +30,25 @@
...
-            "run_no": {"type": "number"},
-            "time_setup": {"type": "number"},
+            "runid": {"type": "number"},
+            "time_setup": {"type": ["number", "null"]},
             "time_compile": {"type": ["number", "null"]},
             "time_run": {"type": ["number", "null"]},
             "time_sanity": {"type": ["number", "null"]},
             "time_performance": {"type": ["number", "null"]},
-            "time_total": {"type": "number"}
+            "time_total": {"type": ["number", "null"]}
```
vkarak, line 45: @ekouts Can total be `None`??

reframe/utility/__init__.py
```diff
@@ -16,6 +16,24 @@
 from collections import UserDict
 
 
+def get_next_runreport_index():
```
vkarak, line 19: This should be in the statistics if we keep it.

reframe/schemas/runreport.json
ekouts: yes, when it fails because of a dependency it is `None`.

vkarak: Hmm, I think we should fix that in the runner. It doesn't make too much sense.

ekouts: It's because it gets the difference between the first and last timestamp:
https://github.com/eth-cscs/reframe/blob/master/reframe/frontend/executors/__init__.py#L144
Would you prefer to return 0 for `total` in that case? When it fails directly from a dependency what would we count as total time? The time in the runcase, since it was processed? Or the time that it was waiting?

rsarm:
@jenkins-cscs retry all

reframe/schemas/runreport.json
vkarak: @ekouts We could get a timestamp at the time when the task is created.

vkarak:
@rsarm Even the generic unit tests from Travis are failing.

vkarak:
The same if I run them locally. Can you please fix them?

vkarak:
Ok, I see the problem. You don't handle correctly the case where the `$HOME/.local/reframe` directory does not exist:

```
./bin/reframe: OS error: [Errno 2] No such file or directory: '/Users/user/.local/reframe'
```

vkarak:
But the unit tests must not create artifacts outside the temp directory that they are using. So just creating that directory is not acceptable. It needs further thinking.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,96 @@
...
+                                },
+                                "failing_phase": {"type": ["string", "null"]},
+                                "failing_reason": {"type": ["string", "null"]},
+                                "build_stdout": {"type": ["string", "null"]},
+                                "build_stderr": {"type": ["string", "null"]},
+                                "job_stdout": {"type": ["string", "null"]},
+                                "job_stderr": {"type": ["string", "null"]},
+                                "maintainers": {
+                                    "type": "array",
+                                    "items": {"type": ["string"]}
```
vkarak, line 64:
```suggestion
                                    "items": {"type": "string"}
```

reframe/frontend/cli.py
```diff
@@ -712,6 +736,24 @@ def print_infoline(param, value):
...
+                    runreport_file = os.path.join(runreport_dir,
+                                                  runreport_name)
+
+                try:
+                    with open(runreport_file, 'w') as fp:
+                        json.dump(runner.stats.json(reframe_info, force=True),
+                                  fp, indent=4)
+                except OSError:
+                    printer.error(f'invalid path: {runreport_file}')
+                    sys.exit(1)
```
vkarak, line 755: Parts of this should be abstracted in a function.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,96 @@
...
+    "type": "object",
+    "properties": {
+        "reframe_info": {
+            "type": "object",
+            "properties": {
+                "version": {"type": "string"},
+                "command": {"type": "string"},
+                "user": {"type": "string"},
+                "host": {"type": "string"},
+                "check_search_path": {"type": "string"},
```
vkarak, line 14: I am not sure we need this, because on the same ground why not store all the configuration parameters. We do need to store the configuration file instead.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,96 @@
...
+                "working_directory": {"type": "string"},
+                "recursive_search_path": {"type": "boolean"},
```
vkarak, line 16: Same comment here. Not needed.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,96 @@
...
+    "properties": {
+        "reframe_info": {
+            "type": "object",
+            "properties": {
+                "version": {"type": "string"},
+                "command": {"type": "string"},
+                "user": {"type": "string"},
+                "host": {"type": "string"},
+                "check_search_path": {"type": "string"},
+                "working_directory": {"type": "string"},
```
vkarak, line 15: A bit too verbose as a name.

reframe/schemas/runreport.json
```diff
@@ -0,0 +1,96 @@
...
+                                "job_stderr": {"type": ["string", "null"]},
+                                "maintainers": {
+                                    "type": "array",
+                                    "items": {"type": ["string"]}
+                                },
+                                "tags": {
+                                    "type": "array",
+                                    "items": {"type": "string"}
+                                },
+                                "runid": {"type": "number"},
```
vkarak, line 70: Not needed any more here.

reframe/frontend/statistics.py
```diff
@@ -66,46 +67,113 @@ def retry_report(self):
 
         return '\n'.join(report)
 
+    def json(self, reframe_info=None, force=False):
```
vkarak, line 70: This function should only be responsible for constructing the json report for each run. It should not be involved with the `reframe_info` at all.

vkarak:
@rsarm We will need also some documentation for this. For the moment we won't need a detailed description of what the report contains.

PR #1389, opened 2020-06-25
[test] PrgEnv-pgi workaround for CDT 20.06

vkarak:
Addresses UES-928.

### DIFF
--- a/cscs-checks/mch/automatic_arrays_acc.py
+++ b/cscs-checks/mch/automatic_arrays_acc.py
@@ -84,3 +84,8 @@ class AutomaticArraysCheck(rfm.RegressionTest):
         if (self.current_system.name == 'dom' and
             self.current_environ.name == 'PrgEnv-cray'):
             self.variables['LINKER_X86_64'] = '/usr/bin/ld'
+
+    @rfm.run_before('compile')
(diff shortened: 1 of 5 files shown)


cscs-checks/mch/automatic_arrays_acc.py
```diff
@@ -84,3 +84,8 @@ def cray_linker_workaround(self):
         if (self.current_system.name == 'dom' and
             self.current_environ.name == 'PrgEnv-cray'):
             self.variables['LINKER_X86_64'] = '/usr/bin/ld'
+
+    @rfm.run_before('compile')
+    def cdt2006_pgi_workaround(self):
+        if self.current_system.name == 'dom':
```
vkarak, line 90: I need to further restrict it here to apply the workaround only for PrgEnv-pgi.

victorusu: But this should be harmless to the other PE's no?

vkarak: Yes, in fact, the problem is not due to the PR.

vkarak:
@jenkins-cscs retry dom

PR #1393, opened 2020-06-28
[bugfix] Add missing launchers `upcrun` and `upcxx-run` in configuration schema

bcfriesen:
ReFrame 3 upstreamed support for UPC and UPC++ launchers from Berkeley Lab. This PR fixes a missed change to the launcher schemes. Without this, tests which request the `upcrun` or `upcxx-run` launchers will fail because ReFrame cannot find them in this list.


vkarak:
ok to test

PR #1403, opened 2020-07-07
[test] TrilinosTest_static build workaround for cdt/20.06

ChristopherBignamini:
Addresses UES-845
Addresses UES-925


ChristopherBignamini:
@jenkins-cscs retry dom

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -49,3 +51,13 @@ def __init__(self, linkage):
     def set_cxxflags(self):
         flags = self.prgenv_flags[self.current_environ.name]
         self.build_system.cxxflags = flags
+
+    @rfm.run_before('compile')
+    def set_prebuild_cmd(self):
+        if self.current_environ.name.startswith('PrgEnv-intel'):
+            if '20.06' in os.getenv('MODULERCFILE', ''):
```
vkarak, line 58:
```suggestion
        if self.current_environ.name == 'PrgEnv-intel' and cray_cdt_version() ==  '20.06':
```

Where the `cray_cdt_version()` should be more elaborate than this check. `MODULERCFILE` is defined only when a `cdt` module is loaded; it doesn't cover the system default cdt. I think for that you should check the version from the following:

```
cat /opt/cray/pe/cdt/default/modulerc | head -1
```

Note: you should do that Pythonically, not by launching a new shell.

@victorusu Do you agree with the `cray_cdt_version()` function?

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -49,3 +51,13 @@ def __init__(self, linkage):
...
+                self.modules += ['cray-netcdf-hdf5parallel']
```
vkarak, line 59: Why this is needed?

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -49,3 +51,13 @@ def __init__(self, linkage):
...
+                self.prebuild_cmds = [
+                    'ln -s $CRAY_NETCDF_HDF5PARALLEL_PREFIX/lib/pkgconfig/'\
```
vkarak, line 61: Do not use backslash and unindent the following line.

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -49,3 +51,13 @@ def __init__(self, linkage):
...
+                        'netcdf-cxx4_parallel.pc netcdf_c++4_parallel.pc',
+                    'export PKG_CONFIG_PATH=`pwd`:$PKG_CONFIG_PATH']
```
vkarak, line 63: You should set the `variables` instead of using the `export` here. Also, since ReFrame runs always from the stage directory, you can simply export `PKG_CONFIG_PATH=.:$PKG_CONFIG_PATH`.

cscs-checks/libraries/math/trilinos_compile_run.py
vkarak: I think also that we should move that function to a utility package. We are going to need it in different cases.

vkarak: I'm writing a utility function. I will do a PR soon.

vkarak: You can use this utility for retrieving the currently loaded CDT module: https://github.com/eth-cscs/reframe/pull/1416

vkarak: @ChristopherBignamini #1416 is now merged, so you can proceed with this PR.

cscs-checks/libraries/math/trilinos_compile_run.py
ChristopherBignamini: The module load is needed in order to have the CRAY_NETCDF_HDF5PARALLEL_PREFIX variable set.

cscs-checks/libraries/math/trilinos_compile_run.py
ChristopherBignamini: Done.

cscs-checks/libraries/math/trilinos_compile_run.py
ChristopherBignamini: Done.

cscs-checks/libraries/math/trilinos_compile_run.py
victorusu: @vkarak, yes! The `cray_cdt_version` is a much better approach!

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -62,3 +63,12 @@ def cdt2006_workaround_intel(self):
                     'ln -s $CRAY_NETCDF_HDF5PARALLEL_PREFIX/lib/pkgconfig/'
                     'netcdf-cxx4_parallel.pc netcdf_c++4_parallel.pc']
                 self.variables['PKG_CONFIG_PATH'] = '.:$PKG_CONFIG_PATH'
+
+    @rfm.run_before('compile')
+    def cdt2006_workaround_dynamic(self):
+        if (os_ext.cray_cdt_version() == '20.06' and
+            self.linkage == 'dynamic'):
+            self.variables['PATH'] = '/opt/cray/pe/cce/10.0.1/cce-clang/x86_64/bin:$PATH'
+            self.prgenv_flags[self.current_environ.name] += ['-fuse-ld=lld']
```
vkarak, line 72: This workaround is only for `PrgEnv-gnu` and I would suggest to put in another function that would be run before compile also.

cscs-checks/libraries/math/trilinos_compile_run.py
```diff
@@ -62,3 +63,12 @@ def cdt2006_workaround_intel(self):
...
+            if self.current_environ.name == 'PrgEnv-gnu':
+                self.modules += ['gcc/9.3.0']
```
vkarak, line 74: Why is this needed?

vkarak: Now I see it.

PR #1404, opened 2020-07-08
[feat] Add option to disable test stage directory cleanup

### LINKED ISSUE #1366: Add an option to reuse the existing stage directory
For example, `--dont-restage`. If a stage dir exists, then this should be reused, else a a new one will be created as usual.

vkarak:
In the past we removed the stage directory of a test in two places. Inside the `RuntimeContext` and inside and low-level `os_ext.copytree()` function. We have changed `copytree()` and made it completely compatible with Python 3.8, which has added the `dirs_exist_ok` argument. It does not remove the target directory if it exists, as it was the previous version. Responsible for removing the stage directory is only the `RuntimeContext`.

I want some feedback about the name of the command-line option and the naming of the configuration option as well.

Fixes #1366.


vkarak:

vkarak:
@jenkins-cscs retry tsa kesch

PR #1407, opened 2020-07-10
[test] Add account in executable options of RequiredConstraintCheck

lucamar:
`RequiredConstraintCheck` is missing the account in the `srun` command line: see [CSCS RT 39626](https://webrt.cscs.ch/Ticket/Display.html?id=39626).


vkarak:
@jenkins-cscs retry none

PR #1409, opened 2020-07-10
[ci] Set the account dynamically in the configuration file for unit tests 

### LINKED ISSUE #1406: Allow passing job options to `./test_reframe.py`

vkarak:
Adding an option to `./test_reframe.py` for setting the account is not so straightforward as for the standard cli. It would require either to do hacks or to modify all the unit tests that use a temporary runtime. For this reason, I believe the solution proposed here is the best.

Fixes #1406.

### DIFF
--- a/config/cscs-ci.py
+++ b/config/cscs-ci.py
@@ -7,6 +7,9 @@
 # CSCS CI settings
 #
 
+import reframe.utility.os_ext as os_ext
+
+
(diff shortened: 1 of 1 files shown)


vkarak:
Let me know if you want me to follow this approach also on the basic configuration file.

vkarak:
@jenkins-cscs retry all

Issue #1369, opened 2020-06-10
Random failure of the `test_kbd_interrupt_in_wait_with_limited_concurrency` unit test


vkarak:
The following failures happens randomly from time to time. I don't know if it actually corresponds to a bug in the code or is it a problem of this particular unit test.

```
_____________ test_kbd_interrupt_in_wait_with_limited_concurrency ______________
async_runner = (reframe.frontend.executors.Runner(_max_retries=0, _policy=reframe.frontend.executors.policies.AsynchronousExecutionPo...cs.TestStats(...)@0x7fdbd38f2908)@0x7fdbd38f2c18, <unittests.test_policies._TaskEventMonitor object at 0x7fdbd38f2208>)
make_cases = <function make_cases.<locals>._make_cases at 0x7fdbd34edc80>
make_async_exec_ctx = <function make_async_exec_ctx.<locals>._make_async_exec_ctx at 0x7fdbd34f46a8>
    def test_kbd_interrupt_in_wait_with_limited_concurrency(
            async_runner, make_cases, make_async_exec_ctx):
        # The general idea for this test is to allow enough time for all the
        # four checks to be submitted and at the same time we need the
        # KeyboardInterruptCheck to finish first (the corresponding wait should
        # trigger the failure), so as to make the framework kill the remaining
        # three.
        ctx = make_async_exec_ctx(2)
        next(ctx)
    
        runner, _ = async_runner
        with pytest.raises(KeyboardInterrupt):
            runner.runall(make_cases([
                KeyboardInterruptCheck(), SleepCheck(10),
>               SleepCheck(10), SleepCheck(10)
            ]))
E           Failed: DID NOT RAISE <class 'KeyboardInterrupt'>
unittests/test_policies.py:525: Failed
=========================== short test summary info ============================
FAILED unittests/test_policies.py::test_kbd_interrupt_in_wait_with_limited_concurrency
================== 1 failed, 619 passed, 88 skipped in 38.85s ==================
```

vkarak:
It seems that this failure appears only with Python 3.7 and 3.8.

vkarak:
This is a very strange error. It now appears with Python 3.6 and not with 3.7 or 3.8. Also, I couldn't find any scenario that would cause `KeyboardInterrupt` not to be thrown. That would mean that somehow the task associated with the `KeyboardInterruptCheck` is not waited for, which I couldn't find any execution path that would make this happen.

vkarak:
Full dump from failure:

[...]

PR #1410, opened 2020-07-11
[refactor] Add a powerful `repr()` function for debugging

vkarak:
This function replaces the older `debug.repr()` and uses `util.ppretty()` for nicely printing the object attributes. `util.ppretty()` is also extended in order to recursively call the new `repr()` function. Objects are printed only once by this new `repr()` avoiding infinite recursions.

Finally, we use that function to print the whole ReFrame runtime state in case #1369 appears.

This function does not take into account the `__slots__` for the moment. In cases, where an object does not have a `__dict__` attribute, the builtin `repr()` will be used.

An example output for a list of objects with self references is the following (see also the corresponding unit test):

```
[
    _Y({
        'x': 1,
        'y': 'foo',
        'z': _Y(...)@0x102463550,
        'a': None
    })@0x102463550,
    _Y({
        'x': 2,
        'y': 'foo',
        'z': _Y(...)@0x102463510,
        'a': _X({
            '_a': False
        })@0x102463290
    })@0x102463510
]
```


unittests/test_policies.py
```diff
@@ -524,6 +525,8 @@ def test_kbd_interrupt_in_wait_with_limited_concurrency(
             KeyboardInterruptCheck(), SleepCheck(10),
             SleepCheck(10), SleepCheck(10)
         ]))
+        # FIXME: Dump everything in case Github #1369 appears
+        print(util.repr(runner))
```
vkarak, line 529: Unindent this and try running this unit test with `-s` to check what this innocent line prints ;-)

vkarak:
Check comment in issue #1369 as a proof that this PR works wonderfully :-) It served its purpose already!

reframe/utility/__init__.py
```diff
@@ -164,24 +170,55 @@ def ppretty(value, htchar=' ', lfchar='\n', indent=4, basic_offset=0):
...
+        if hasattr(obj, '__dict__'):
+            objects.add(addr)
+
+        return func(obj, **kwargs)
+
+    return _repr
+
+
+@_tracked_repr
+def repr(obj, *, htchar=' ', lfchar='\n', indent=4, basic_offset=0):
```
vkarak, line 211: Actually, here I don't need all the other options... I could update the way I'm trying to call `repr()` in `ppretty()`.

reframe/utility/__init__.py
```diff
@@ -135,26 +137,31 @@ def ppretty(value, htchar=' ', lfchar='\n', indent=4, basic_offset=0):
     :arg indent: Number of htchar characters for every indentation level.
     :arg basic_offset: Basic offset for the representation, any additional
         indentation space is added to the ``basic_offset``.
+    :arg repr: The :func:`repr` to use for printing values. This function may
+        accept also all the arguments of :func:`ppretty` except the ``repr``.
+
     :returns: a formatted string of the ``value``.
     '''
 
+    print(repr)
```
teojgo, line 146: Why do you need this print?

vkarak: No, it's stale...

PR #1417, opened 2020-07-14
[feat] Add utility function to look for available environment modules

### LINKED ISSUE #1159: Parametrizing tests based on module discovery

victorusu:
~~This PR extends the environment modules interface to include the `module avail` command.~~

~~This is the first step to allow the detection of modules to be included in the tests.~~

This PR adds the `find_modules()` utility function, which will retrieve all the available modules given a substring of a module name. Internally, it delegates to `module avail substr` of the current system's modules system, but it does so per system partition and respecting the ReFrame environments defined for each partition. It returns an iterator over a "set" of tuples of `(partition, environment, module)`. This works nicely with hierarchical modules, but it can also work fine with flat modules scheme. In this case, you can pass a dictionary to the `environ_mapping` argument, which maps module names to ReFrame environments. This way, `find_module()` will skip any combinations of `(partition, environment, module)` that are not meaningful. You can use the return value of this function in parameterized tests as follows, in order to generate automatically tests for different modules. Here is an example:

```python
@rfm.parameterized_test(*find_modules('netcdf'))
    def __init__(self, s, e, m):
        self.descr = f'{s}, {e}, {m}'
        self.valid_systems = [s]
        self.valid_prog_environs = [e]
        self.modules = [m]
```

For flat schemes you can use it as follows:

```python
my_find_modules = functools.partial(find_modules, environ_mapping={
    r'.*CrayGNU.*': {'PrgEnv-gnu'},
    r'.*CrayIntel.*': {'PrgEnv-intel'},
    r'.*CrayCCE.*': {'PrgEnv-cray'}
})

@rfm.parameterized_test(*my_find_modules('GROMACS'))
    def __init__(self, s, e, m):
        self.descr = f'{s}, {e}, {m}'
        self.valid_systems = [s]
        self.valid_prog_environs = [e]
        self.modules = [m]
```

Fixes #1159.

Still todo:

- [x] Document the function and add a tutorial example
- [x] Better type checking of the `toolchain_mapping` argument

Future work:

`find_modules()` can be somewhat slow since it calls `module avail` for each partition/environment combination. We could in the future implement a caching mechanism on top of it.


victorusu:
@jenkins-cscs retry dom

reframe/core/modules.py
```diff
@@ -152,6 +152,13 @@ def resolve_module(self, name):
     def backend(self):
         return(self._backend)
 
+    def available_modules(self):
```
vkarak, line 155: The interface should accept a module name.

reframe/core/modules.py
```diff
@@ -470,6 +484,17 @@ def _exec_module_command(self, *args, msg=None):
...
         exec(completed.stdout)
 
+    def available_modules(self):
+        avail = []
+        completed = self._run_module_command(
+            'available', '-t', msg="could not run module available")
+
+        for line in re.finditer(r'\S+[^:]$', completed.stderr, re.MULTILINE):
+            module = re.sub(r"\(default\)", "", line.group(0))
+            avail += [Module(module)]
```
vkarak, line 494:
```suggestion
            module = re.sub(r'\(default\)', '', line)
            avail.append(Module(module))
```

reframe/core/modules.py
```diff
@@ -470,6 +484,17 @@ def _exec_module_command(self, *args, msg=None):
         completed = self._run_module_command(*args, msg=msg)
         exec(completed.stdout)
 
+    def available_modules(self):
+        avail = []
+        completed = self._run_module_command(
+            'available', '-t', msg="could not run module available")
```
vkarak, line 490:
```suggestion
            'avail', '-t', msg="could not run 'module avail'")
```

reframe/core/modules.py
```diff
@@ -681,6 +706,17 @@ def name(self):
     def _module_command_failed(self, completed):
         return completed.stdout.strip() == 'false'
 
+    def available_modules(self):
+        avail = []
+        completed = self._run_module_command(
+            '-t', 'available', msg="could not run module available")
```
vkarak, line 712: Why don't you use spider for Lmod?

victorusu: Because spider can be cached. But that's a great idea too.
Do you want me to use it?

ekouts, line 712:
```suggestion
            '-t', 'available', msg="could not run 'module available'")
```

reframe/core/modules.py
```diff
@@ -681,6 +706,17 @@ def name(self):
...
+
+        for line in re.finditer(r'\S+[^:/]$', completed.stderr, re.MULTILINE):
+            module = re.sub(r"\(\S+\)", "", line.group(0))
```
ekouts, line 715:
```suggestion
            module = re.sub(r'\(\S+\)', '', line.group(0))
```

vkarak:
@mboisson @davidedelvento

This PR should do pretty much what we have been discussing in #1159. Could you test it?

mboisson:
I'll have to put this on the pile. I haven't had time to touch our Reframe tests in months...

vkarak:
@jenkins-cscs retry all

davidedelvento:
I've been very busy with other stuff too, but trying various aspects of ReFrame in NCAR's environment is at the top of my todo list, just after a couple of other things which are due by the end of the month. So I should get to it in early October, and perhaps sooner. 
@vkarak  Do you need/want feedback earlier that this?

vkarak:
@mboisson @davidedelvento I don't want to rush you, I just wanted to put that in your radar and test it whenever you have time. My plan is to merge this in the next weeks and ship it with ReFrame 3.2, but you can test it afterwards and we enhance it further.

vkarak:
@jenkins-cscs retry all

vkarak (approved):

unittests/test_utility.py
```diff
@@ -1293,3 +1297,73 @@ def test_cray_cle_info_missing_parts(tmp_path):
...
+        yield from temp_runtime(config_file, system,
+                                {'systems/modules_system': request.param})
+    except ConfigError as e:
+        pytest.skip(str(e))
+
+
+@pytest.fixture
+def modules_system(user_exec_ctx, monkeypatch):
+    # Pretend to be on a clean modules environment
+    #    monkeypatch.setenv('MODULEPATH', '')
```
vkarak, line 1329: I should remove this.

victorusu:
`lgtm`

PR #1419, opened 2020-07-14
[feat] Add utility function for retrieving Cray Linux Environment (CLE) information

vkarak:
I am not doing any special post processing of the values retrieved, since they are well formed strings that you can nicely compare:

```python
if cray_cle_info().release >= '7.0.UP00' and cray_cle_info().patchset < '09':
    # do sth
```

### DIFF
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -1189,7 +1189,7 @@ def test_cray_cle_info(tmp_path):
     assert cle_info.build == '7.0.1227'
     assert cle_info.date == '20200326'
     assert cle_info.network == 'ari'
-    assert cle_info.patchset == '09-202003261814'
+    assert cle_info.patchset == '09'
 
 
 def test_cray_cle_info_no_such_file(tmp_path):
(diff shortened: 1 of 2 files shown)


unittests/test_utility.py
```diff
@@ -1171,3 +1171,42 @@ def test_cray_cdt_version_no_such_file(tmp_path, monkeypatch):
...
+def test_cray_cle_info(tmp_path):
+    # Mock up a CLE release
+    cle_info_file = tmp_path / 'cle-release'
+    with open(cle_info_file, 'w') as fp:
+        fp.write('RELEASE=7.0.UP01\n'
+                 'BUILD=7.0.1227\n'
+                 'DATE=20200326\n'
+                 'ARCH=noarch\n'
+                 'NETWORK=ari\n'
+                 'PATCHSET=09-202003261814\n')
```
vkarak, line 1185: I could possibly strip off the date from the patchset so as to allow comparisons such as `cray_cle_info().patchset >= '09'`. Now this would compare false.

PR #1424, opened 2020-07-16
[bugfix] Convert Syslog handler port parameter into integer

difey:
In our cluster, we use syslog to collect reframe tests data. We have found SysLogHandler require int type port instead of str type in codes now. It works fine in our cluster after change.

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -222,7 +222,12 @@ def _create_syslog_handler(site_config, config_prefix):
     except ValueError:
         pass
     else:
-        address = (host, int(port))
+        try:
+            address = (host, int(port))
+        except ValueError:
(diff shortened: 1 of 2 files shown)


difey:
> Can I test this patch?

yes

teojgo:
ok to test

reframe/core/logging.py
```diff
@@ -222,7 +222,7 @@ def _create_syslog_handler(site_config, config_prefix):
     except ValueError:
         pass
     else:
-        address = (host, port)
+        address = (host, int(port))
```
vkarak, line 225: We should handle the case when `port` cannot be converted to an `int`.

victorusu: That's true.

PR #1431, opened 2020-07-22
[ci] Change modules system for Dom to TMod4

teojgo:
* Hardcode singularity module to singularity/3.5.3 for dom.


vkarak:
@jenkins-cscs retry all

PR #1433, opened 2020-07-23
[feat] Support multiple tags in `extract*` sanity functions

### LINKED ISSUE #1245: Add a sanity function for extracting multiple capturing groups at a time
This is something that has been requested to me some times, so I guess it would make sense to consider it. The question is how the returned value is meant to be treated. A straightforward return value would be to return  a list of tuples of the extracted values for the different capturing groups or would it make sense to return a list-like data structure that would allow, for example, to extract all the elements of a single capturing group at once? @jgphpc How you would use such a function?

teojgo:
* Support also passing of multiple conversion functions for
  each tag to be extracted.

Fixes #1245

### DIFF
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -538,8 +538,8 @@ def _callable_name(fn):
 
 def _extractiter_singletag(patt, filename, tag, conv, encoding):
     if isinstance(conv, collections.Iterable):
-        raise SanityError('multiple conversion functions given for the single '
-                          f'capturing group {tag!r}')
+        raise SanityError(f'multiple conversion functions given for the '
+                          f'single capturing group {tag!r}')
 
     for m in finditer(patt, filename, encoding):
         try:
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -1,9 +1,10 @@
-# Copyright 2016-2020 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# Copyrigh 2016-2020 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
 # ReFrame Project Developers. See the top-level LICENSE file for details.
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
 import builtins
+import collections
 import glob as pyglob
 import itertools
 import re
@@ -530,28 +531,65 @@ def extractiter(patt, filename, tag=0, conv=None, encoding='utf-8'):
     the extracted values.
     '''
     for m in finditer(patt, filename, encoding):
-        try:
-            val = m.group(tag)
-        except (IndexError, KeyError):
-            raise SanityError(
-                "no such group in pattern `%s': %s" % (patt, tag))
-
-        try:
-            yield conv(val) if callable(conv) else val
-        except ValueError:
-            fn_name = '<unknown>'
-            try:
-                # Assume conv is standard function
-                fn_name = conv.__name__
-            except AttributeError:
+        if isinstance(tag, collections.Iterable) and not isinstance(tag, str):
+            val = []
+            for t in tag:
                 try:
-                    # Assume conv is callable object
-                    fn_name = conv.__class__.__name__
-                except AttributeError:
-                    pass
+                    val.append(m.group(t))
+                except (IndexError, KeyError):
+                    raise SanityError(
+                        "no such group in pattern `%s': %s" % (patt, t))
+        else:
+            try:
+                val = m.group(tag)
+            except (IndexError, KeyError):
+                raise SanityError(
+                    "no such group in pattern `%s': %s" % (patt, tag))
 
-            raise SanityError("could not convert value `%s' using `%s()'" %
-                              (val, fn_name))
+        if isinstance(val, list):
+            converted_vals = []
+            if not isinstance(conv, collections.Iterable):
+                conv = [conv] * len(val)
+
+            # Here we use the last conversion function for the remaining
+            # tags which don't have a corresponding one
+            for v, c in itertools.zip_longest(val, conv, fillvalue=conv[-1]):
+                try:
+                    converted_vals.append(c(v) if callable(c) else v)
+                except ValueError:
+                    fn_name = '<unknown>'
+                    try:
+                        # Assume conv is standard function
+                        fn_name = c.__name__
+                    except AttributeError:
+                        try:
+                            # Assume conv is callable object
+                            fn_name = c.__class__.__name__
+                        except AttributeError:
+                            pass
+
+                    raise SanityError(
+                        "could not convert value `%s' using `%s()'" %
+                        (v, fn_name))
+
+            yield(tuple(converted_vals))
+        else:
+            try:
+                yield conv(val) if callable(conv) else val
+            except ValueError:
+                fn_name = '<unknown>'
+                try:
+                    # Assume conv is standard function
+                    fn_name = conv.__name__
+                except AttributeError:
+                    try:
+                        # Assume conv is callable object
+                        fn_name = conv.__class__.__name__
+                    except AttributeError:
+                        pass
+
+                raise SanityError("could not convert value `%s' using `%s()'" %
+                                  (val, fn_name))
 
 
 @deferrable
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -536,9 +536,11 @@ def _callable_name(fn):
     return fn_name
 
 
-def _extractiter_tag(patt, filename, tag, conv, encoding):
+def _extractiter_singletag(patt, filename, tag, conv, encoding):
     if isinstance(conv, collections.Iterable):
-        conv = conv[0]
+        raise SanityError(
+            f'multiple conversion functions given for single group: {tag}'
+        )
 
     for m in finditer(patt, filename, encoding):
         try:
@@ -551,7 +553,8 @@ def _extractiter_tag(patt, filename, tag, conv, encoding):
         except ValueError:
             fn_name = _callable_name(conv)
             raise SanityError(
-                f'could not convert value {val!r} using {fn_name}()')
+                f'could not convert value {val!r} using {fn_name}()'
+            )
 
 
 def _extractiter_multitag(patt, filename, tags, conv, encoding):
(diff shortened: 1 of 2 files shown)


teojgo:
@vkarak Does this cover our case? If yes I will also add the documentation.

reframe/utility/sanity.py
```diff
@@ -530,28 +531,65 @@ def extractiter(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
-            val = m.group(tag)
-        except (IndexError, KeyError):
-            raise SanityError(
-                "no such group in pattern `%s': %s" % (patt, tag))
-
-        try:
-            yield conv(val) if callable(conv) else val
-        except ValueError:
-            fn_name = '<unknown>'
+        if isinstance(tag, collections.Iterable) and not isinstance(tag, str):
```
vkarak, line 534: Don't do everything inline. Delegate to different functions for each case.

vkarak: It's better if you delegate to separate function completely in case `tag` is an iterable, and do that right at the beginning:

```python
    if isinstance(tag, collections.Iterable) and not isinstance(tag, str):
        return _extractiter_multitag(patt, filename, tag, conv, encoding)

    # And here remains are as before
```

reframe/utility/sanity.py
```diff
@@ -530,28 +531,65 @@ def extractiter(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
+                    fn_name = '<unknown>'
+                    try:
+                        # Assume conv is standard function
+                        fn_name = c.__name__
+                    except AttributeError:
+                        try:
+                            # Assume conv is callable object
+                            fn_name = c.__class__.__name__
+                        except AttributeError:
+                            pass
```
vkarak, line 569: This should be abstracted in a separate function, e.g., `_callable_name()`, and will be used also in the current implementation.

unittests/test_sanity_functions.py
```diff
@@ -650,6 +653,46 @@ def test_extractall(self):
         for expected, v in enumerate(res, start=1):
             assert expected == v
 
+    def test_extractall_multiple_tags(self):
+        # Check multiple numeric groups
+        res = sn.evaluate(sn.extractall(
+            r'Number: (\d+) (\d+)', self.tempfile, (1, 2)))
+        for expected, v in enumerate(res, start=1):
+            assert str(expected) == v[0]
+            assert str(2 * expected) == v[1]
```
vkarak, line 662: You can be more compact here (same later on):

```suggestion
            assert str(2*expected) == v[1]
```

vkarak:
@teojgo We will also need to update the documentation.

teojgo:
> @teojgo We will also need to update the documentation.

Yes I wanted first to see if what I thought about the tag/conv function was correct.

reframe/utility/sanity.py
```diff
@@ -521,6 +521,68 @@ def findall(patt, filename, encoding='utf-8'):
...
+        try:
+            # Assume fn is callable object
+            fn_name = fn.__class__.__name__
+        except AttributeError:
+            pass
+
+    return fn_name
+
+
+def _extractiter_tag(patt, filename, tag, conv, encoding):
```
vkarak, line 539:
```suggestion
def _extractiter_singletag(patt, filename, tag, conv, encoding):
```

reframe/utility/sanity.py
```diff
@@ -521,6 +521,68 @@ def findall(patt, filename, encoding='utf-8'):
...
+    if isinstance(conv, collections.Iterable):
+        conv = conv[0]
```
vkarak, line 541: Should we support that? I don't see the reason. I would rather fail in that case.

reframe/utility/sanity.py
```diff
@@ -521,6 +521,68 @@ def findall(patt, filename, encoding='utf-8'):
...
+            val = m.group(tag)
+        except (IndexError, KeyError):
+            raise SanityError(f'no such group in pattern {patt!r}: {tag}')
+
+        try:
+            yield conv(val) if callable(conv) else val
+        except ValueError:
+            fn_name = _callable_name(conv)
+            raise SanityError(
+                f'could not convert value {val!r} using {fn_name}()')
```
vkarak, line 554: Close the parenthesis in the next line.

reframe/utility/sanity.py
```diff
@@ -521,6 +521,68 @@ def findall(patt, filename, encoding='utf-8'):
...
+
+        converted_vals = []
+        if not isinstance(conv, collections.Iterable):
+            conv = [conv] * builtins.len(val)
+        elif builtins.len(conv) > builtins.len(val):
+            conv = conv[:builtins.len(val)]
+
+        # Here we use the last conversion function for the remaining
+        # tags which don't have a corresponding one, if length of the
+        # conversion function iterable is less that the one of tags
```
vkarak, line 574:
```suggestion
        # Use the last function in case we have less conversion function than tags
```

unittests/test_sanity_functions.py
```diff
@@ -685,6 +685,20 @@ def test_extractall_multiple_tags(self):
             assert 2 * expected == v[1]
             assert isinstance(v[1], float)
 
+        # Check more convert functions than tags
```
vkarak, line 688:
```suggestion
        # Check more conversion functions than tags
```

unittests/test_sanity_functions.py
```diff
@@ -685,6 +685,20 @@ def test_extractall_multiple_tags(self):
...
+        res = sn.evaluate(sn.extractall(r'Number: (?P<no1>\d+) (?P<no2>\d+)',
+                                        self.tempfile, ('no1', 'no2'),
+                                        [int, float, float, float]))
+        for expected, v in enumerate(res, start=1):
+            assert expected == v[0]
+            assert 2 * expected == v[1]
+
+        # Check multiple convert functions and single tag
```
vkarak, line 696:
```suggestion
        # Check multiple conversion functions and a single tag
```

reframe/utility/sanity.py
```diff
@@ -610,11 +616,26 @@ def extractall(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
         returns the whole line that was matched.
-    :arg conv: A callable that takes a single argument and returns a new value.
-        If provided, it will be used to convert the extracted values before
-        returning them.
-    :returns: A list of the extracted values from the matched regex.
+    :arg conv: A callable or iterable of callables taking a single argument
+        and returning a new value.
+        If provided, and is not an iterable it will be used to convert
+        the extracted values for all the capturing groups of ``tag``
+        returning the converted values.
```
vkarak, line 623:
```suggestion
        If not an iterable it will be used to convert
        the extracted values for all the capturing groups specified in ``tag``.
```

reframe/utility/sanity.py
```diff
@@ -610,11 +616,26 @@ def extractall(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
+        If an iterable of callables is provided, each one will be used to
+        convert the corresponding extracted capturing group of `tag`.
```
vkarak, line 625:
```suggestion
        Otherwise each conversion function will be used to
        convert the value extracted from the corresponding capturing group in `tag`.
```

reframe/utility/sanity.py
```diff
@@ -610,11 +616,26 @@ def extractall(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
+        If more callables functions than the corresponding capturing groups of
+        ``tag`` are provided, the last conversion function is used for the
+        remaining capturing groups.
```
vkarak, line 628:
```suggestion
        If more conversion functions are supplied than the corresponding capturing groups in
        ``tag``, the last conversion function will be used for the
        additional capturing groups.
```

reframe/utility/sanity.py
```diff
@@ -610,11 +616,26 @@ def extractall(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
+    :returns: A list of the extracted values from the matched regex if ``tag``
+        converted using the ``conv`` callable if ``tag`` is a single value.
+        In case of multiple capturing groups, a list of tuples where each one
+        contains the extracted capturing converted using the corresponding
+        callable of ``conv``.
```
vkarak, line 633:
```suggestion
    :returns: A list of tuples of converted values extracted from the capturing groups specified in ``tag``, if ``tag`` is an iterable. Otherwise, a list of the converted values extracted from the single capturing group specified in ``tag``.
```

reframe/utility/sanity.py
```diff
@@ -610,11 +616,26 @@ def extractall(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
     :raises reframe.core.exceptions.SanityError: In case of errors.
+
+    .. versionchanged:: 3.1
+        Multiple regex capturing groups are now supporetd via ``tag`` and
+        multiple callables can be used in ``conv``.
```
vkarak, line 638:
```suggestion
        multiple conversion functions can be used in ``conv``.
```

reframe/utility/sanity.py
```diff
@@ -536,9 +536,11 @@ def _callable_name(fn):
     return fn_name
 
 
-def _extractiter_tag(patt, filename, tag, conv, encoding):
+def _extractiter_singletag(patt, filename, tag, conv, encoding):
     if isinstance(conv, collections.Iterable):
-        conv = conv[0]
+        raise SanityError(
+            f'multiple conversion functions given for single group: {tag}'
```
vkarak, line 542:
```suggestion
            f'multiple conversion functions given for the single capturing group {tag!r}'
```

reframe/utility/sanity.py
```diff
@@ -618,24 +621,22 @@ def extractall(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
-        convert the corresponding extracted capturing group of `tag`.
-        If more callables functions than the corresponding capturing groups of
-        ``tag`` are provided, the last conversion function is used for the
-        remaining capturing groups.
-    :returns: A list of the extracted values from the matched regex if ``tag``
-        converted using the ``conv`` callable if ``tag`` is a single value.
-        In case of multiple capturing groups, a list of tuples where each one
-        contains the extracted capturing converted using the corresponding
-        callable of ``conv``.
+        If not an iterable it will be used to convert the extracted values for
```
vkarak, line 624:
```suggestion
        If not an iterable, it will be used to convert the extracted values for
```

PR #1439, opened 2020-07-29
[bugfix] Fix glob pattern to search for both txt and yaml

teojgo:
* Correct also the sanity pattern.

* Use f-strings instead of `%` based string formatting.


teojgo:
@jenkins-cscs retry all

PR #1442, opened 2020-08-04
[test] Update libsci acc symlink test

victorusu:


cscs-checks/compile/libsci_acc_symlink.py
```diff
@@ -4,21 +4,40 @@
...
-                              'dom:login', 'dom:gpu',
-                              'tiger:login', 'tiger:gpu']
+
+        self.valid_systems = ['tiger:login', 'tiger:gpu']
+        regex = (r'libsci_acc_(?P<prgenv>[A-Za-z]+)_((?P<cver>[A-Za-z0-9]+)_)'
+                 r'?(?P<version>\S+)')
+        prgenv = re.match(regex, lib_name).group('prgenv')
+
+        #FIXME: Remove this logic after daint upgrade
+        if prgenv == 'gnu':
```
vkarak, line 32: Why don’t you simply check the Cray OS and CDT version with the utility functions in os_ext?

victorusu: That would also work. 
Should I use it?

victorusu: The implementation that uses `os_ext.cray_cdt_version()` doesn't change the number of lines, neither is simpler.
Because the logic is inverse. The classes are spawned and then we select if we add or not `dom` or `daint` based on the `lib_name`. While in a "normal" scenario you will be taking action about the environment based on the available CDT, but the systems would be already set.

vkarak: That's fine for me. I just made a suggestion in case you haven't considered.

PR #1451, opened 2020-08-11
[test] Remove LINKER workaround from tests

victorusu:


cscs-checks/mch/automatic_arrays_acc.py
```diff
@@ -79,18 +79,10 @@ def setflags(self):
...
-
-    @rfm.run_before('compile')
-    def cdt2006_pgi_workaround(self):
+    def cdt2008_pgi_workaround(self):
         cdt = os_ext.cray_cdt_version()
         if not cdt:
             return
 
-        if (self.current_environ.name == 'PrgEnv-pgi' and cdt == '20.06'):
+        if (self.current_environ.name == 'PrgEnv-pgi' and cdt == '20.08'):
```
vkarak, line 87: `>=` would have worked.

PR #1445, opened 2020-08-06
[test] Update tf2 Horovod check

henrique:

PR #1452, opened 2020-08-11
[test] Update the ipcmagic check to TF2

rsarm:


ekouts:
@jenkins-cscs retry all

cscs-checks/apps/jupyter/check_ipcmagic.py
```diff
@@ -14,7 +14,13 @@ def __init__(self):
         self.descr = 'Distributed training with TensorFlow using ipyparallel'
         self.valid_systems = ['daint:gpu', 'dom:gpu']
         self.valid_prog_environs = ['PrgEnv-gnu']
-        self.modules = ['ipcmagic', 'Horovod']
+        self.modules = ['ipcmagic']
+        # FIXME: Workaround until Dom and Daint are aligned
+        if self.current_system.name == 'daint':
+            self.modules.append('Horovod/0.19.1-CrayGNU-19.10-tf-2.2.0')
+        else:
+            self.modules.append('Horovod/0.19.1-CrayGNU-20.08-tf-2.2.0')
```
vkarak: Please use the trick proposed by @henrique in #1445.

vkarak (changes requested):

cscs-checks/apps/jupyter/check_ipcmagic.py
```diff
@@ -14,15 +15,16 @@ def __init__(self):
         self.descr = 'Distributed training with TensorFlow using ipyparallel'
         self.valid_systems = ['daint:gpu', 'dom:gpu']
         self.valid_prog_environs = ['PrgEnv-gnu']
-        self.modules = ['ipcmagic']
-        self.prerun_cmds = [
-            'module unload dask',
-            'module load Horovod/0.16.4-CrayGNU-19.10-tf-1.14.0'
+        # FIXME: The following will not be needed after the Daint upgrade
+        cray_cdt_version = os_ext.cray_cdt_version()  # or '19.10'
```
vkarak, line 19: Either remove the `or` comment or leave it as part of the statement. Technically, it's not needed, since I don't see the reason that `cray_cdt_version()` would return `None` on Daint or Dom, but it has no meaning as a comment here.

ekouts: I wanted to ask the same, why do we need the or? On daint it failed and it looks like it returns `None`. 
```
Horovod(60):ERROR:105: Unable to locate a modulefile for 'Horovod/0.19.1-CrayGNU-None-tf-2.2.0'
```
https://jenkins.cscs.ch/blue/organizations/jenkins/ReframeCI/detail/ReframeCI/3747/pipeline

rsarm: Yes, sorry. Pushed that by mistake. I was trying to understand why it gives `None` on Daint, as @ekouts pointed out.

vkarak: I confirm it doesn't work on Daint and the reason is because the path it looks for the default system CDT is not present on Daint.

vkarak (approved):

PR #1460, opened 2020-08-27
[bugfix] Add always units in the default reference tuple

### LINKED ISSUE #1447: Obliged to set references in RegressionTest?
Hi -- firstly, I'd like to say thanks for developing ReFrame, a really nice tool!

Today I wanted to update from a 3.1-dev2 commit (c027bc70) to the 3.1 release, but this appears to introduce a non-backwards-compatible change (also the current master gives me this issue).

A somewhat minimal example looks like this, with [this file](https://github.com/eth-cscs/reframe/blob/v3.0-dev6/tutorial/src/example_matrix_vector_multiplication.c) as 'sources/example_matrix_vector_multiplication.c':
```python
import reframe as rfm


@rfm.simple_test
class Example1Test(rfm.RegressionTest):
    def __init__(self):
        self.descr = 'Simple matrix-vector multiplication example'
        self.valid_systems = ['*']
        self.time_limit = '00d00h01m00s'
        self.sourcesdir = 'sources'
        self.sourcepath = 'example_matrix_vector_multiplication.c'
        self.build_system.cflags = ['-O2']
        self.executable_opts = ['1024', '100']
        self.sanity_patterns = sn.assert_found(
            r'time for single matrix vector multiplication', self.stdout)
        self.strict_check = False
        self.perf_patterns = {
            'time': sn.extractsingle(
                r'^time\sfor\ssingle\smatrix\svector\smultiplication\s(?P<perf>\S+)\ss$',
                self.stdout, 'perf', float),
        }

```
The setup/build/run/sanity steps seem to run fine, but the performance step crashes with
```
<path_to_reframe>/bin/reframe: unexpected error: not enough values to unpack (expected 5, got 4)
  File "<path_to_reframe>/reframe/frontend/cli.py", line 772, in main
    run_stats = runner.stats.json()
  File "<path_to_reframe>/reframe/frontend/statistics.py", line 143, in json
    val, ref, lower, upper, unit = ref
ValueError: not enough values to unpack (expected 5, got 4)
```
In 3.1 (or the current master), this can be solved by setting a 'reference' attribute like this:
```
        self.reference = {
            '<system>:<partition>': {
                'time': (0.0004, -0.1, 0.1, 's'),
            }
        }
```
but this was not needed in the prior version I was using. The different behavior is probably related to this commit: https://github.com/eth-cscs/reframe/commit/9705d69355f7e3e9071a5dfd38d06b62868d2dfb. Is this a bug, or are we now "forced" to set references, even if only interested in performance logging and not in performance checking by ReFrame?

teojgo:
Fixes #1447

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -1436,10 +1436,7 @@ class RegressionTest(metaclass=RegressionTestMeta):
 
                 for var in variables:
                     name, unit = var
-                    ref_tuple = (
-                        0, None, None,
-                        unit or '(no unit specified)'
-                    )
+                    ref_tuple = (0, None, None, unit or '(no unit specified)')
                     self.reference.update({'*': {name: ref_tuple}})
 
             # We first evaluate and log all performance values and then we
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -1436,7 +1436,7 @@ class RegressionTest(metaclass=RegressionTestMeta):
 
                 for var in variables:
                     name, unit = var
-                    ref_tuple = (0, None, None, unit or '(no unit specified)')
+                    ref_tuple = (0, None, None, unit)
                     self.reference.update({'*': {name: ref_tuple}})
 
             # We first evaluate and log all performance values and then we
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -1436,10 +1436,10 @@ class RegressionTest(metaclass=RegressionTestMeta):
 
                 for var in variables:
                     name, unit = var
-                    ref_tuple = (0, None, None)
-                    if unit:
-                        ref_tuple += (unit,)
-
+                    ref_tuple = (
+                        0, None, None,
+                        unit or '(no unit specified)'
+                    )
                     self.reference.update({'*': {name: ref_tuple}})
 
             # We first evaluate and log all performance values and then we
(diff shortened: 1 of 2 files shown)


ekouts (approved):

teojgo:
@MaximeVdB can you check if this solves your problem?

reframe/core/pipeline.py
```diff
@@ -1436,10 +1436,7 @@ def check_performance(self):
 
                 for var in variables:
                     name, unit = var
-                    ref_tuple = (0, None, None)
-                    if unit:
-                        ref_tuple += (unit,)
-
+                    ref_tuple = (0, None, None, unit or '(no unit specified)')
```
vkarak, line 1439: I would keep the unit as `None`. The JSON schema allows it to be so.

```suggestion
                    ref_tuple = (0, None, None, unit)
```

reframe/frontend/statistics.py
```diff
@@ -273,11 +273,7 @@ def performance_report(self):
             for key, ref in t.check.perfvalues.items():
                 var = key.split(':')[-1]
                 val = ref[0]
-                try:
-                    unit = ref[4]
-                except IndexError:
-                    unit = '(no unit specified)'
-
+                unit = ref[4]
```
vkarak, line 276: Here leave it as before.

vkarak (changes requested):

vkarak (approved):
lgtm

MaximeVdB:
@teojgo Thanks, I did a quick test and this indeed solved the problem!

PR #1461, opened 2020-08-28
[bugfix] Use new the 2020 pip dependency resolver in the bootstrap script

vkarak:
This solves the following error when installing ReFrame's dependencies with the bootstrap script:

```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.
```

This PR also disallows pip cache in the bootstrap script.


teojgo (approved):

vkarak:
@jenkins-cscs retry all

PR #1470, opened 2020-09-02
[feat] Drop tuple syntax for timer fields

### LINKED ISSUE #1350: Drop support of setting of a timer field from a tuple
As mentioned in #1317 it should be dropped in ReFrame 3.1 or 3.2.

vkarak:
Fixes #1350.

### DIFF
--- a/reframe/core/fields.py
+++ b/reframe/core/fields.py
@@ -114,13 +114,6 @@ class TimerField(TypedField):
 
     def __set__(self, obj, value):
         self._check_type(value)
-        if isinstance(value, tuple):
-            user_deprecation_warning(
-                "setting a timer field from a tuple is deprecated: "
-                "please use a string '<days>d<hours>h<minutes>m<seconds>s'")
-            h, m, s = value
-            value = datetime.timedelta(hours=h, minutes=m, seconds=s)
-
         if isinstance(value, str):
             time_match = re.match(r'^((?P<days>\d+)d)?'
                                   r'((?P<hours>\d+)h)?'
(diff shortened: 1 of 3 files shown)


reframe/core/fields.py
```diff
@@ -110,8 +110,7 @@ class TimerField(TypedField):
     '''Stores a timer in the form of a :class:`datetime.timedelta` object'''
 
     def __init__(self, fieldname, *other_types):
-        super().__init__(fieldname, datetime.timedelta, str,
-                         types.Tuple[int, int, int], *other_types)
+        super().__init__(fieldname, datetime.timedelta, str, *other_types)
 
     def __set__(self, obj, value):
         self._check_type(value)
```
ekouts, line 116: Shouldn't the check that is just after this line be deleted?
```
    if isinstance(value, tuple):
            user_deprecation_warning(
            user_deprecation_warning(
                "setting a timer field from a tuple is deprecated: "
                "setting a timer field from a tuple is deprecated: "
                "please use a string '<days>d<hours>h<minutes>m<seconds>s'")
                "please use a string '<days>d<hours>h<minutes>m<seconds>s'")
            h, m, s = value
            h, m, s = value
            value = datetime.timedelta(hours=h, minutes=m, seconds=s)
            value = datetime.timedelta(hours=h, minutes=m, seconds=s)
```

ekouts (changes requested):

reframe/core/fields.py
vkarak: Oh, yes! I don't know why I forgot that. Perhaps I forgot to push.

teojgo (approved):

ekouts (approved):

PR #1472, opened 2020-09-04
[bugfix] Fix crash when git is not available

### LINKED ISSUE #1464: dependence on git
When I install reframe through `pip` in minimal docker container, it seems it can only work when git is installed; probably it just tries to use git to get the current commit for the version number (?). Is git a hard dependency? If I could do without it would reduce the size of my docker image by 80MB.

It seems a `depends_on('git', type='run')` dep in the spack package is missing too: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/reframe/package.py.

teojgo:
* Catch the exception thrown in case git is not installed

* Add a unittest

Fixes #1464

### DIFF
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -188,13 +188,13 @@ def test_is_url():
 
 
 def test_git_repo_hash(monkeypatch):
-    # a git branch hash consists of 8(short) or 40 characters.
+    # A git branch hash consists of 8(short) or 40 characters.
     assert len(os_ext.git_repo_hash()) == 8
     assert len(os_ext.git_repo_hash(short=False)) == 40
     assert os_ext.git_repo_hash(branch='invalid') is None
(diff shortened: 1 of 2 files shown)


unittests/test_utility.py
```diff
@@ -187,13 +187,17 @@ def test_is_url():
     assert not os_ext.is_url(repo_ssh)
 
 
-def test_git_repo_hash():
-    # A git branch hash consists of 8(short) or 40 characters.
+def test_git_repo_hash(monkeypatch):
+    # a git branch hash consists of 8(short) or 40 characters.
```
vkarak, line 191: I think our style is to start comments with a capital letter unless the comment is inline.

vkarak (changes requested):
Just a minor style comment.

vkarak (approved):
lgtm

Issue #1476, opened 2020-09-08
Add option to disable JSON reporting completely


vkarak:

jgphpc:
We could have it on by default but instead of creating many report files:
```
~/.reframe/reports/run-report-0.json
~/.reframe/reports/run-report-1.json
etc...
```
we could set the default to overwrite:
```
~/.reframe/reports/run-report.json
```
and then add the option to specify new report files (as before) ?

vkarak:
@jgphpc Not a bad idea either.

vkarak:
I think that setting the default as you propose here is the best solution, in which case I don't think we need to have an extra configuration parameter to disable reporting completely, do we?

jgphpc:
If you have hundreds of jobs all writing to the same file (~/.reframe/reports/run-report.json), it will not work, no ?

vkarak:
Good point, but that's a different issue. Even if you had a single file per session as now, the problem would remain the same; you might end up with two reframe instances trying to create the same file. We would need to arbitrate that using a file lock.

PR #1480, opened 2020-09-10
[feat] Change default JSON report file pattern

jgphpc:
Will fix https://github.com/eth-cscs/reframe/issues/1476


reframe/frontend/cli.py
```diff
@@ -423,6 +423,13 @@ def main():
         action='store_true',
         help='Use a login shell for job scripts'
     )
+    argparser.add_argument(
+        dest='save_report_file', default=True,
+        envvar='RFM_SAVE_REPORT_FILE',
+        configvar='general/save_report_file',
+        action='store_true',
```
vkarak, line 430: You don't need the action, since this is not a command-line option.

reframe/schemas/config.json
```diff
@@ -409,6 +410,7 @@
         "general/purge_environment": false,
         "general/report_file": "${HOME}/.reframe/reports/run-report-{sessionid}.json",
         "general/save_log_files": false,
+        "general/save_report_file": true,
```
vkarak, line 413: I prefer this one:

```suggestion
        "general/enable_json_report": false,
```

vkarak (changes requested):
One thing that's not very nice is that the configuration option must be enabled in order for the `--report-file` command-line option to have an effect, which is a bit strange. As a user I would expect to get a report if I pass that option. One solution could be to automatically enable reporting if this option is passed. The downside of this is that it will break somehow the consistency of command line options mapping to configuration options. Another decision to make is what should be the default behaviour: generate or not the report. I kind of tend to not generate, but I'm not sure. This decision will also guide us to what would be the best solution to the problem I just described.

Apart from this, we will need to update the documentation.

reframe/frontend/cli.py
jgphpc: Is `default=...,` needed here ?

vkarak: No, it's not, because we are setting the default in the configuration schema.

jgphpc: https://github.com/eth-cscs/reframe/pull/1480/commits/cf1572eb6a77418045716888dd960846cda4416d

jgphpc:
Tested with:

- pytest-3 unittests/test_cli.py 
- reframe hello1.py -> ~/.reframe/reports/run-report.json
- reframe --report-file=/tmp/0_{sessionid}.json -r -c hello1.py ->  /tmp/0_0.json
- reframe --mode maintenance -C cscs.py -r -c hello1.py -> $APPS/.../run-report-0.json
- RFM_REPORT_FILE=/tmp/0.json reframe  -r -c hello1.py ->  /tmp/0.json

config/cscs.py
```diff
@@ -1027,7 +1027,7 @@
             'check_search_path': [
                 'checks/'
             ],
-            'check_search_recursive': True
+            'check_search_recursive': True,
```
vkarak, line 1030: Please remove the `,` at the end.

docs/manpage.rst
```diff
@@ -241,7 +241,7 @@ Options controlling ReFrame output
 .. option:: --report-file=FILE
 
    The file where ReFrame will store its report.
-   The ``FILE`` argument may contain the special placeholder ``{sessionid}``, in which case ReFrame will generate a new report each time it is run by appending a counter to the report file.
+   The ``FILE`` argument may contain the special placeholder ``{sessionid}``, in which case ReFrame will generate a new report each time it is run by appending a counter to the report file. By default, ReFrame will write to $HOME/.reframe/reports/run-report.json.
```
vkarak, line 244: You don't need the `By default, [...]` part. The defaults are never mentioned here. They are only mentioned in the configuration file reference, so as to avoid stale entries in the documentation.

reframe/frontend/cli.py
```diff
@@ -790,8 +790,8 @@ def print_infoline(param, value):
                     printer.info(runner.stats.performance_report())
 
                 # Generate the report for this session
-                report_file = os.path.normpath(
-                    os_ext.expandvars(rt.get_option('general/0/report_file'))
+                report_file = os.path.normpath(os_ext.expandvars(
+                    rt.get_option('general/0/report_file'))
```
vkarak, line 794: The previous formatting fits in a single line and is better.

reframe/frontend/cli.py
```diff
@@ -813,7 +813,8 @@ def print_infoline(param, value):
                         json.dump(json_report, fp, indent=2)
                 except OSError as e:
                     printer.warning(
-                        f'failed to generate report in {report_file!r}: {e}'
+                        f'failed to generate report in {report_file!r}: '
+                        f'{e}'
```
vkarak, line 817: Same here.

docs/manpage.rst
vkarak: So you should change it in the configuration reference only.

vkarak (changes requested):

docs/manpage.rst
vkarak: You still need to change the default in the documentation here: https://github.com/eth-cscs/reframe/blob/master/docs/config_reference.rst

vkarak (approved):
lgtm now

PR #1481, opened 2020-09-10
[feat] Allow dependencies between partitions

### LINKED ISSUE #1378: Allow dependencies between tests from different partitions

ekouts:
Closes #1378 .


victorusu:
First of all, I want to say that I love this feature! I have been dreaming about this one for quite some time now... 😄 

I am a bit lost with the naming logics of `DEPEND_BY_PARTITION`.
I think we should change the name of the `DEPEND_BY_PARTITION`. 
But I have to say in advance that I have no good naming scheme to propose. 😅 🤦 
Let me try to reason why I think the name `DEPEND_BY_PARTITION` should be changed. 

### Basic idea

Given the tests `A` and `B`, the systems `s0` and `s1`, the partitions `p0` and `p1` and the environments `e0` and `e1`. Using the notation `A->B` to denote that test `B` depends on test `A`, then in order to achieve the following dependency relations one should:

```
1. `A(s0, p0, e0) -> B(s1, p0, e0)` => Error! Inter-system dependencies not supported by ReFrame. Because it either run on `s0` or on `s1`.
2. `A(s0, p0, e0) -> B(s0, p0, e0)` => use `how=DEPEND_BY_ENV`. This is the default behaviour.
3. `A(s0, p0, e0) -> B(s0, p0, e1)` => use `how=DEPEND_FULLY` or `how=DEPEND_EXACT`
4. `A(s0, p0, e0) -> B(s0, p1, e0)` => use `how=DEPEND_BY_PARTITION` (this PR)
```

### DEPEND_BY_ENV meaning
In the context of the ReFrame check, where one writes 
```
B.depends_on('A', how=DEPEND_...)
```

The `DEPEND_BY_ENV` means "when the tests have the SAME environment". 
One can read 
```
B.depends_on('A', how= DEPEND_BY_ENV)
```
as
> The test `B` depends on test `A` when their environments are the same, given that the systems and the partitions of `A` and `B` are the same. 

I think the original mindset we had was the sentence:
> Test `B` depends on `A` [constrained] by the environment.

Using the notation above
`A(s0, p0, e0) -> B(s0, p0, e0)`

### DEPEND_BY_PARTITION meaning
In the context of the ReFrame check, where one writes 
```
B.depends_on('A', how=DEPEND_...)
```

The `DEPEND_BY_PARTITION` means "when the tests have the DIFFERENT partitions"
One can read 
```
B.depends_on('A', how= DEPEND_BY_PARTITION)
```
as
> The test `B` depends on test `A` when their partitions are different, given that the systems and the environment of `A` and `B` are the same.

The `DEPEND_BY_PARTITION` uses the original mindset:
> Test `B` depends on `A` constrained by the partition.
But in this case, the partitions are different.

Using the notation above
`A(s0, p0, e0) -> B(s0, p1, e0)`

### DEPEND_EXACT meaning
In the context of the ReFrame check, where one writes 
```
B.depends_on('A', how=DEPEND_...)
```

The `DEPEND_EXACT ` means "when the tests have exact dependencies"
One can read 
```
B.depends_on('A', how=DEPEND_EXACT, subdeps={'e0': ['e1']})
```
as
> The test `B` with environment `e0` depends on test `A` with the environment `e1`, given that the systems and the partitions of `A` and `B` are the same. 

In the original mindset:
> Test `B` depends on `A` constrained by the exact environments.

Using the notation above
`A(s0, p0, e1) -> B(s0, p0, e0)`



### Conclusions

The original notation for `DEPEND_` does not imply similar environments, rather similar systems and partitions.
In `DEPEND_EXACT` one can have different environments, but similar partitions.
`DEPEND_BY_ENV` implies similar environments and partitions.
And the proposed `DEPEND_BY_PARTITION` implies similar environments but different partitions.

So, I think we should change the name from `DEPEND_BY_PARTITION`.

A non-optimal proposal is to rename it to `DEPEND_DIFF_PARTITION`.
I know, it is not the best name change 😭 , but it does not maintain the same name structure that implies different logic.

victorusu (changes requested):
I am requesting a name change for `DEPEND_BY_PARTITION`, but I am not sure what should be the replacement. We should come up with some logic.

victorusu:
@ekouts and @vkarak, I think we miss an example in the tutorial about this feature.
I am not sure if the tutorial should also come in this PR, or if it should be a different issue.
What do you think?

victorusu:
Should we also add a test for `DEPEND_BY_PARTITION` inside `def test_build_deps_unknown_test(loader, exec_ctx):`?

ekouts:
> The original notation for DEPEND_ does not imply similar environments, rather similar systems and partitions.

@victorusu For me it makes sense as it is now, but of course I am open to change to whatever is more useful. As far as I understand the meaning of `how` implies similarity. Now `DEPEND_BY_STH` means that everything is the same up to `STH`. So `DEPEND_BY_ENV` means partition and env is the same. So the dependency for `B.depends_on('A', how= DEPEND_BY_ENV)` is:
```
A(s0, p0, e0) -> B(s0, p0, e0)

A(s0, p0, e1) -> B(s0, p0, e1)

A(s0, p1, e0) -> B(s0, p1, e0)

A(s0, p1, e1) -> B(s0, p1, e1)
```
Similarly `DEPEND_BY_PARTITION` means a full dependency inside the same partition. This is the old `DEPEND_FULLY`. Now `B.depends_on('A', how= DEPEND_BY_PARTITION)` would be:
```
# partition p0
A(s0, p0, e0) -> B(s0, p0, e0)
A(s0, p0, e1) -> B(s0, p0, e0)
A(s0, p0, e0) -> B(s0, p0, e1)
A(s0, p0, e1) -> B(s0, p0, e1)

# partition p1
A(s0, p1, e0) -> B(s0, p1, e0)
A(s0, p1, e1) -> B(s0, p1, e0)
A(s0, p1, e0) -> B(s0, p1, e1)
A(s0, p1, e1) -> B(s0, p1, e1)
```
The new `B.depends_on('A', how= DEPEND_FULLY)` means:
```
# all to all dependencies
A(s0, p0, e0) -> B(s0, p0, e0)
A(s0, p0, e1) -> B(s0, p0, e0)
A(s0, p1, e0) -> B(s0, p0, e0)
A(s0, p1, e1) -> B(s0, p0, e0)
A(s0, p0, e0) -> B(s0, p0, e1)
A(s0, p0, e1) -> B(s0, p0, e1)
A(s0, p1, e0) -> B(s0, p0, e1)
A(s0, p1, e1) -> B(s0, p0, e1)
A(s0, p0, e0) -> B(s0, p1, e0)
A(s0, p0, e1) -> B(s0, p1, e0)
A(s0, p1, e0) -> B(s0, p1, e0)
A(s0, p1, e1) -> B(s0, p1, e0)
A(s0, p0, e0) -> B(s0, p1, e1)
A(s0, p0, e1) -> B(s0, p1, e1)
A(s0, p1, e0) -> B(s0, p1, e1)
A(s0, p1, e1) -> B(s0, p1, e1)
```

victorusu:
@ekouts, Ah!
I had a different mindset when reviewing this PR!
You are right. In this case, `DEPEND_BY_PARTITION` is the appropriate name!

victorusu (approved):
Since the naming scheme got clarified... 😂  
`lgtm`

vkarak:
The meaning of "by" in the `DEPEND_BY_XXX` is to specify how the test cases of the two tests are connected. It defines a binding or grouping relation based on the `XXX` attribute, until now only the environment. That means that `DEPEND_BY_ENV` means that the test cases of two tests depend on each other only if they refer to the same environment: you can't have cross-environment dependencies. If we extend that logic to partitions we have 2^2 combinations to cover instead of just two (`BY_ENV` and `FULLY`). If we stick to the definition of `BY` as a binding factor, then we can derive the following meaning.

Assume that tests `t0` and `t1` are valid for `{p0, p1}` partitions and for `{e0, e1}` programming environments. Then we have the following cases:

### DEPEND_FULLY

That's the fully connected graph with 16 edges, which I skip from posting it here.

### DEPEND_BY_ENV

That's essentially two graphs, one per each environment, fully connected per partition:

```
(t0, p0, e0) -> (t1, p0, e0)
(t0, p0, e0) -> (t1, p1, e0)
(t0, p1, e0) -> (t1, p0, e0)
(t0, p1, e0) -> (t1, p1, e0)

(t0, p0, e1) -> (t1, p0, e1)
(t0, p0, e1) -> (t1, p1, e1)
(t0, p1, e1) -> (t1, p0, e1)
(t0, p1, e1) -> (t1, p1, e1)
```

### DEPEND_BY_PARTITION

The exact analogous of `DEPEND_BY_ENV`, but the binding is on partition:

```
(t0, p0, e0) -> (t1, p0, e0)
(t0, p0, e0) -> (t1, p0, e1)
(t0, p0, e1) -> (t1, p0, e0)
(t0, p0, e1) -> (t1, p0, e1)

(t0, p1, e0) -> (t1, p1, e0)
(t0, p1, e0) -> (t1, p1, e1)
(t0, p1, e1) -> (t1, p1, e0)
(t0, p1, e1) -> (t1, p1, e1)
```

### DEPEND_BY_ENV and DEPEND_BY_PARTITION

This can be represented as a bitwise OR in the `how` parameter: `DEPEND_BY_ENV | DEPEND_BY_PARTITION`. This will bind on both environment and partition (I don't see a better way to represent this programmatically other than a bitwise OR if we don't want to introduce yet another constant):

```
(t0, p0, e0) -> (t1, p0, e0)
(t0, p0, e1) -> (t1, p0, e1)
(t0, p1, e0) -> (t1, p1, e0)
(t0, p1, e1) -> (t1, p1, e1)
```

### DEPEND_EXACT

Exactly as before. The user can create the exact relationships by herself.

I think the definitions I just presented are consistent and cover all the possibilities. By reading the comments here, I think that the current PR does not follow exactly this principle am I right?

ekouts:
> I think the definitions I just presented are consistent and cover all the possibilities. By reading the comments here, I think that the current PR does not follow exactly this principle am I right?

You are right, the `DEPEND_BY_PARTITION` was supposed to replace the old `DEPEND_FULLY`. I will change the code and the unittests with the new flags.

vkarak:
We discussed this a bit further with @victorusu and came up with a more general and flexible solution. To avoid the confusion that the `DEPEND_BY` variables mean, we could have the following syntax:

```python
self.depends_on('T1', when=always)
self.depends_on('T1', when=part_equal)
self.depends_on('T1', when=env_equal)
```

Then `when` argument is a callable that takes the source and destination nodes of the test case graph and returns true if there is an edge or false otherwise. This is much more flexible compared to the `DEPEND_EXACT` for supporting specific patterns, e.g., partition is not named `x` or partitions are not equal etc. You could even compose a function from others.

These predefined functions are as simple as follows:

```python
def always(src, dst):
    return True

def part_equal(src, dst):
    p0, _ = src
    p1, _  = dst
    return p0 == p1

def env_equal(src, dst):
    _, e0 = src
    _, e1  = dst
    return e0 == e1
```

You could even define a function that would create dependencies only if the target partition has a specific name:

```python
def target_part_is(name):
    def _has_edge(src, dst):
        return dst[0] == name

    return _has_edge
```

vkarak (changes requested):
Check my last comment in the main conversation.

vkarak:
I converted this to a draft.

reframe/frontend/dependency.py
```diff
@@ -85,25 +105,13 @@ def resolve_dep(target, from_map, fallback_map, *args):
...
-                        c.deps.append(
-                            resolve_dep(c, cases_revmap, default_cases_revmap,
-                                        tname, pname, te)
-                        )
+            tname, when = dep
+            for d in resolve_dep(c, all_cases, default_all_cases, tname):
+                dep_cname = d.check.name
+                dep_pname = d.partition.fullname
+                dep_ename = d.environ.name
+                if not when or when((pname, ename), (dep_pname, dep_ename)):
```
vkarak, line 113: You should remove the first check here. If `when` is `None` here, it is a bug and we want to crash.

reframe/frontend/dependency.py
```diff
@@ -85,25 +105,13 @@ def resolve_dep(target, from_map, fallback_map, *args):
...
-                    if env != ename:
-                        continue
-
-                    for te in tenvs:
-                        c.deps.append(
-                            resolve_dep(c, cases_revmap, default_cases_revmap,
-                                        tname, pname, te)
-                        )
+            tname, when = dep
+            for d in resolve_dep(c, all_cases, default_all_cases, tname):
```
vkarak, line 109: I think that the `resolve_dep` implementation can be simplified a lot. You only need to index per test name and this should return you all the test cases of that test (partition and environment). Then you simply add them as dependencies if the `when()` function returns `True`. Dependency errors should be raised only if the target test does not exist.

vkarak (changes requested):

ekouts:
@vkarak and @victorusu since you are using the PR already, what do you think that the predefined exposed `when` functions should be? I had something like this in mind but maybe you can think of better use cases:
```python
# fully connected
def always(src, dst):
    return True

# fully connected in the same partition
def part_equal(src, dst):
    return src[0] == dst[0]

# different partitions but same env
def env_equal(src, dst):
    return src[1] == dst[1]

# same env and part, which is the default also
def part_env_equal(src, dst):
    return src == dst
```

vkarak:
@ekouts I agree with those, we can also add a version that checks for a specific partition and environment, e.g., `when=part_is('login')` and `when=env_is('gnu')`. These functions should return another function that will be used as the actual `when` argument.

vkarak:
I also think these functions should go inside the `utility` package, perhaps `utility.dependency`.

vkarak:
Btw, do you prefer `dependencies` or `dependency` as package names? Whatever we choose, it should be consistent both for the `frontend` and the `utility`.

ekouts:
> @ekouts I agree with those, we can also add a version that checks for a specific partition and environment, e.g., `when=part_is('login')` and `when=env_is('gnu')`. These functions should return another function that will be used as the actual `when` argument.

I like the `env_is` as a name of the function. As @victorusu mentioned, it might be useful to have a list in `when=part_is(['login', 'gpu'])`.

> Btw, do you prefer dependencies or dependency as package names? Whatever we choose, it should be consistent both for the frontend and the utility.

I don't have a strong opinion about this. I would go with `dependencies`.

victorusu:
> @ekouts I agree with those, we can also add a version that checks for a specific partition and environment, e.g., `when=part_is('login')` and `when=env_is('gnu')`. These functions should return another function that will be used as the actual `when` argument.

Does it make sense to implement `part_is` and `env_is` to accept a list instead of a string?

This would allow one to depend on multiple partitions or environments, but at the same being able to select which one.

Or should we have another set of functions that are called `part_in` and `env_in`?

I would vote for the first case

victorusu:
> > @ekouts I agree with those, we can also add a version that checks for a specific partition and environment, e.g., `when=part_is('login')` and `when=env_is('gnu')`. These functions should return another function that will be used as the actual `when` argument.
> 
> Does it make sense to implement `part_is` and `env_is` to accept a list instead of a string?
> 
> This would allow one to depend on multiple partitions or environments, but at the same being able to select which one.
> 
> Or should we have another set of functions that are called `part_in` and `env_in`?
> 
> I would vote for the first case

Sorry, I was too late... 😂

ekouts:
> @ekouts I agree with those, we can also add a version that checks for a specific partition and environment, e.g., `when=part_is('login')` and `when=env_is('gnu')`. These functions should return another function that will be used as the actual `when` argument.

@vkarak @victorusu thinking more about it, this could also create some confusion. Because `when=part_is('login')` could be interpreted as `src` has partition `login` or `dst` has partition `login` or that both have it. I know in @victorusu 's case it should be `src`, but does I am not sure if it helps to predefine them for all the users in the end.

vkarak:
Lots of comments! Let me chime in as well:

> I don't have a strong opinion about this. I would go with dependencies.

If so, then you should also rename the `frontend/dependecy.py` to `frontend/dependencies.py` as well.

> it might be useful to have a list in when=part_is(['login', 'gpu'])

What about something like this instead?

```python
def any(*when_funcs):
    def _any(src, dst):
        return builtins.any(fn(dst, src) for fn in when_funcs)

    return _any

self.depends_on('T0', when=any(part_is('login'), part_is('gpu')))
```

> @vkarak @victorusu thinking more about it, this could also create some confusion. Because when=part_is('login') could be interpreted as src has partition login or dst has partition login or that both have it. I know in @victorusu 's case it should be dst, but I am not sure if it helps to predefine them for all the users in the end. Of course we can make them src_env_in, dst_env_in, both_env_in and src_part_in, dst_part_in, both_part_in but it's very simple for the user to write exactly what he needs. What do you think?

Good point, but I'm not sure about the function names you propose. They sound too verbose. What about a more composable thing like this one:

```python
def always(src, dst):
    return True


def part_is(name):
    def _part_is(src, dst):
        if src and dst:
            return src[0] == name and dst[0] == name

        if src:
            return src[0] == name

        if dst:
            return dst[0] == name

        return False

    return _part_is


def source(fn):
    def _source(src, dst):
        return fn(src, None)

    return _source


def dest(fn):
    def _dest(src, dst):
        return fn(None, dst)

    return _dest


def any(*when_funcs):
    def _any(src, dst):
        return builtins.any(fn(dst, src) for fn in when_funcs)

    return _any


def all(*when_funcs):
    def _all(src, dst):
        return builtins.all(fn(dst, src) for fn in when_funcs)

    return _all


if __name__ == '__main__':
    partitions = ['login', 'gpu']
    environs = ['gnu', 'cray']
    when = always
    when = part_is('login')
    when = source(part_is('login'))
    when = dest(part_is('login'))
    when = any(source(part_is('login')), dest(part_is('login')))

    print('Dependencies:')
    t0_cases = [(p, e) for p in partitions for e in environs]
    t1_cases = [(p, e) for p in partitions for e in environs]
    num_deps = 0
    for t0 in t0_cases:
        for t1 in t1_cases:
            if when(t0, t1):
                num_deps += 1
                print(f'\t{t0} -> {t1}')

    print(f'Total dependencies: {num_deps}')
```

Of course we can shortcuts, but I think the generic filtering done by the `source` and `dest` (we can have more expressive names as well, such as `this` and `other` respectively), as well as the fact that the `part_is` interprets the case of `None` arguments, allows us to compose dependencies without exploding the number of functions. I also think that the more natural implementation for `part_is` without filtering should be that *both* partitions have the same name. What do you think of this proposal?

victorusu:
> Lots of comments! Let me chime in as well:
> 
> > I don't have a strong opinion about this. I would go with dependencies.
> 
> If so, then you should also rename the `frontend/dependecy.py` to `frontend/dependencies.py` as well.
> 
> > it might be useful to have a list in when=part_is(['login', 'gpu'])
> 
> What about something like this instead?
> 
> ```python
> def any(*when_funcs):
>     def _any(src, dst):
>         return builtins.any(fn(dst, src) for fn in when_funcs)
> 
>     return _any
> 
> self.depends_on('T0', when=any(part_is('login'), part_is('gpu')))
> ```
> 
> > @vkarak @victorusu thinking more about it, this could also create some confusion. Because when=part_is('login') could be interpreted as src has partition login or dst has partition login or that both have it. I know in @victorusu 's case it should be dst, but I am not sure if it helps to predefine them for all the users in the end. Of course we can make them src_env_in, dst_env_in, both_env_in and src_part_in, dst_part_in, both_part_in but it's very simple for the user to write exactly what he needs. What do you think?
> 
> Good point, but I'm not sure about the function names you propose. They sound too verbose. What about a more composable thing like this one:
> 
> ```python
> def always(src, dst):
>     return True
> 
> 
> def part_is(name):
>     def _part_is(src, dst):
>         if src and dst:
>             return src[0] == name and dst[0] == name
> 
>         if src:
>             return src[0] == name
> 
>         if dst:
>             return dst[0] == name
> 
>         return False
> 
>     return _part_is
> 
> 
> def source(fn):
>     def _source(src, dst):
>         return fn(src, None)
> 
>     return _source
> 
> 
> def dest(fn):
>     def _dest(src, dst):
>         return fn(None, dst)
> 
>     return _dest
> 
> 
> def any(*when_funcs):
>     def _any(src, dst):
>         return builtins.any(fn(dst, src) for fn in when_funcs)
> 
>     return _any
> 
> 
> def all(*when_funcs):
>     def _all(src, dst):
>         return builtins.all(fn(dst, src) for fn in when_funcs)
> 
>     return _all
> 
> 
> if __name__ == '__main__':
>     partitions = ['login', 'gpu']
>     environs = ['gnu', 'cray']
>     when = always
>     when = part_is('login')
>     when = source(part_is('login'))
>     when = dest(part_is('login'))
>     when = any(source(part_is('login')), dest(part_is('login')))
> 
>     print('Dependencies:')
>     t0_cases = [(p, e) for p in partitions for e in environs]
>     t1_cases = [(p, e) for p in partitions for e in environs]
>     num_deps = 0
>     for t0 in t0_cases:
>         for t1 in t1_cases:
>             if when(t0, t1):
>                 num_deps += 1
>                 print(f'\t{t0} -> {t1}')
> 
>     print(f'Total dependencies: {num_deps}')
> ```
> 
> Of course we can shortcuts, but I think the generic filtering done by the `source` and `dest` (we can have more expressive names as well, such as `this` and `other` respectively), as well as the fact that the `part_is` interprets the case of `None` arguments, allows us to compose dependencies without exploding the number of functions. I also think that the more natural implementation for `part_is` without filtering should be that _both_ partitions have the same name. What do you think of this proposal?
> 
> PS: Comment out the `when` function in the above example to see the different effects.

This is fantastic! I think we should definitely implement this!!!!!
This proposed way clearily shows the intention of the test writer.
The addition of the `any` and `all`... That's the cherry in the cake that make all fit together...

I do have a suggestion, but a very weak one...
I am not sure if we should call `source` and `dest`. Probably the words `this` and `parent` would fit better. 
But as I said, I think that `source` and `dest` can also be used.

vkarak:
`this` and `parent` are also good. I have no strong opinion on that.

ekouts:
> I am not sure if we should call source and dest. Probably the words this and parent would fit better.
But as I said, I think that source and dest can also be used.

I agree, but just to make sure we are on the same page. In the relationship `A depends on B`, `A` is `this` and `B` is the `parent`, right?

vkarak:
@ekouts Yes.

reframe/frontend/dependencies.py
```diff
@@ -24,39 +25,26 @@ def build_deps(cases, default_cases=None):
...
-            ename = c.environ.name
-            ret.setdefault((cname, pname, ename), c)
+            ret.setdefault(cname, [])
+            ret[cname].append(c)
 
         return ret
 
-    def resolve_dep(target, from_map, fallback_map, *args):
-        errmsg = 'could not resolve dependency: %s -> %s' % (target, args)
+    def resolve_dep(target, from_map, fallback_map, key):
```
vkarak, line 40: This function signature is deceiving. The first argument is not the target, but actually the source and the `key` is the target. Also you can build the indexes just above this function and use them without passing them as an argument. This will make this function's signature simpler:

```suggestion
    def resolve_dep(src, dst):
```

I suggest using `src` and `dst` for consistency with the udeps functions.

reframe/frontend/dependencies.py
```diff
@@ -81,29 +67,15 @@ def resolve_dep(target, from_map, fallback_map, *args):
     # partitions and environments
     graph = collections.OrderedDict()
     for c in cases:
-        cname = c.check.name
-        pname = c.partition.fullname
+        pname = c.partition.name
```
vkarak, line 70:
```suggestion
        psrc = c.partition.name
```

reframe/frontend/dependencies.py
```diff
@@ -81,29 +67,15 @@ def resolve_dep(target, from_map, fallback_map, *args):
...
         ename = c.environ.name
```
vkarak, line 71:
```suggestion
        esrc = c.environ.name
```

reframe/frontend/dependencies.py
```diff
@@ -81,29 +67,15 @@ def resolve_dep(target, from_map, fallback_map, *args):
...
-
-                    for te in tenvs:
-                        c.deps.append(
-                            resolve_dep(c, cases_revmap, default_cases_revmap,
-                                        tname, pname, te)
-                        )
+            tname, when = dep
+            for d in resolve_dep(c, all_cases, default_all_cases, tname):
+                dep_pname = d.partition.name
+                dep_ename = d.environ.name
```
vkarak, line 76:
```suggestion
                pdst = d.partition.name
                edst = d.environ.name
```

reframe/utility/dependencies.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2016-2020 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
```
vkarak, line 5: We need to document this module and include it somewhere inside the "ReFrame manuals" section.

Would it also make sense to rename this package to `udeps` for "micro-dependencies." I don't know if you had this in mind when you included as `udeps` elsewhere, but I like it as an idea.

reframe/utility/dependencies.py
```diff
@@ -0,0 +1,86 @@
...
+import builtins
+
+
+# Dependency `when' functions
+def always(src, dst):
```
vkarak, line 10: We need to properly document each of these functions.

unittests/resources/checks_unlisted/deps_simple.py
```diff
@@ -33,8 +48,5 @@ def __init__(self, kind):
         self.sanity_patterns = sn.assert_found(self.name, self.stdout)
         if kind == 'default':
             self.depends_on('Test0')
-        elif kindspec[kind] == rfm.DEPEND_EXACT:
-            self.depends_on('Test0', kindspec[kind],
-                            {'e0': ['e0', 'e1'], 'e1': ['e1']})
         else:
-            self.depends_on('Test0', kindspec[kind])
+            self.depends_on('Test0', kindspec[kind])
```
vkarak, line 52: End the line with a newline character.

unittests/test_dependencies.py
```diff
@@ -112,140 +115,360 @@ def test_eq_hash(loader, exec_ctx):
...
 
     with pytest.raises(DependencyError):
-        check_e0.getdep('Test0', 'eX')
+        check_e0.getdep('Test0', 'eX', 'p0')
 
     with pytest.raises(DependencyError):
-        check_e1.getdep('Test0', 'e0')
-
-
-def test_build_deps_unknown_test(loader, exec_ctx):
```
vkarak, line 202: We still need this test. We still expect to get a dependency error.

vkarak (changes requested):
Looks good overall. I will have a look a the general documentation, but we need to document the `when()` functions properly. I also think that we miss some unit tests:

1. Since we still support the old syntax, we should make sure that we are still recognizing it and it works as expected, and that it issues a deprecation warning.
2. We need a unit test as to test the case that the `when()` function returns always `False`. For example, what should I get if I have this in a test?

```python
self.depends_on('non-existent-test', when=lambda s, d: False)
```

ekouts:
> Since we still support the old syntax, we should make sure that we are still recognizing it and it works as expected, and that it issues a deprecation warning.

You mean something other than `test_build_deps_deprecated_syntax`? I check for deprecation warning and that the right callable is passed. In the case of `part_equal` and `part_env_equal` we already test them in `test_dependecies_when_functions` and I check manually the exact. is this enough?

unittests/test_dependencies.py
```diff
@@ -112,140 +115,367 @@ def test_eq_hash(loader, exec_ctx):
     assert hash(case1) != hash(case0)
 
 
+def test_dependecies_when_functions():
```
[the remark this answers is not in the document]
vkarak: Also change the `when` to `how` everywhere.

unittests/test_dependencies.py
vkarak: What I mean here is that we need to check that if we pass the `how` argument the old way, (a) we get deprecation warning and (b) it creates the correct dependencies. You just need to check in a similar way as in your other unit test, except that you only try the three different parameters for `how`.

unittests/test_dependencies.py
```diff
@@ -112,140 +115,367 @@ def test_eq_hash(loader, exec_ctx):
...
+    with pytest.warns(ReframeDeprecationWarning):
+        t1 = Test1_deprecated('exact')
+        when = t1._userdeps[0][1]
+        t0_cases = [(p, e) for p in ['p0', 'p1']
+                    for e in ['e0', 'e1']]
+        t1_cases = [(p, e) for p in ['p0', 'p1']
+                    for e in ['e0', 'e1']]
+        deps = {(t0, t1) for t0 in t0_cases
+                for t1 in t1_cases
+                if when(t0, t1)}
```
vkarak, line 328: This can be moved to the line above.

vkarak:
> Since we still support the old syntax, we should make sure that we are still recognizing it and it works as expected, and that it issues a deprecation warning.

@ekouts You can ignore this comment. Your current unit test looks fine.

vkarak:
@ekouts There are also some warnings when I'm building the docs:

```
reframe/core/pipeline.py:docstring of reframe.core.pipeline.RegressionTest.depends_on:4: WARNING: Inline interpreted text or phrase reference start-string without end-string.
reframe/core/pipeline.py:docstring of reframe.core.pipeline.RegressionTest.depends_on:4: WARNING: Inline interpreted text or phrase reference start-string without end-string.
reframe/core/pipeline.py:docstring of reframe.core.pipeline.RegressionTest.depends_on:4: WARNING: Inline interpreted text or phrase reference start-string without end-string.
reframe/core/pipeline.py:docstring of reframe.core.pipeline.RegressionTest.depends_on:29: WARNING: Inline interpreted text or phrase reference start-string without end-string.
reframe/core/pipeline.py:docstring of reframe.core.pipeline.RegressionTest.depends_on:29: WARNING: Inline interpreted text or phrase reference start-string without end-string.
reframe/core/pipeline.py:docstring of reframe.core.pipeline.RegressionTest.depends_on:29: WARNING: Inline interpreted text or phrase reference start-string without end-string.
```

vkarak:
I'm now going through the documentation and I think we should find better names for the udeps functions, so that they can be better combined with the `how` argument conceptually. For example, you can't say `how=always`, `how=fully` is better and more accurate.

vkarak:
@ekouts I think that we should also test the behaviour of udeps when valid systems and environments are not symmetric between the two tests.

ekouts:
> I'm now going through the documentation and I think we should find better names for the udeps functions, so that they can be better combined with the how argument conceptually. For example, you can't say how=always, how=fully is better and more accurate.

@vkarak Yes, `always` is not a good name anymore. The only problem I see with `fully` is that the meaning of `DEPEND_FULLY` is different but I don't think it's a big issue, since we support interpartition dependencies.

> @ekouts I think that we should also test the behaviour of udeps when valid systems and environments are not symmetric between the two tests.

In the `test_dependecies_how_functions` the partitions and environments are not symmetric. Do you mean in a different test?

vkarak:
> In the test_dependecies_how_functions the partitions and environments are not symmetric. Do you mean in a different test?

Yes, you are right. All good then.

> @vkarak Yes, always is not a good name anymore. The only problem I see with fully is that the meaning of DEPEND_FULLY is different but I don't think it's a big issue, since we support interpartition dependencies.

I don't think that we should be kept behind by a deprecated name and meaning. Regarding the new names that match better the `how` argument I propose the following (@victorusu please comment as well):

- `always` becomes `fully` (example: `self.depends_on('T1', how=fully)`)
- `part_equal` becomes `by_part` (example: `self.depends_on('T1', how=by_part)`)
- `env_equal` becomes `by_env` (example: `self.depends_on('T1', how=by_env)`)
- `part_env_equal` becomes `by_case` (example: `self.depends_on('T1', how=by_case)`)

For the `source` and `dest` and `part_is` I'm still thinking about it :-) Suggestions are welcome!

vkarak:
Given the work we are doing in #1555 and the suggestion for a `when` that prunes the test parameterization space (including the partition and environments), I doubt whether we need the machinery of `source`, `dest`, `part_is` and `env_is`. This could be served by the future `when` argument (but even now one could implement it as a `how` function, but I suggest that the framework does not provide these functions):

The following will be equivalent in the future:

- `self.depends_on('T1', how=source(part_is('login')))` and `self.depends_on('T1', how=fully, when=self.params['*:login'])`

@victorusu For the `how=dest(part_is('login'))` can you help me understand the use case?

victorusu:
> > In the test_dependecies_how_functions the partitions and environments are not symmetric. Do you mean in a different test?
> 
> Yes, you are right. All good then.
> 
> > @vkarak Yes, always is not a good name anymore. The only problem I see with fully is that the meaning of DEPEND_FULLY is different but I don't think it's a big issue, since we support interpartition dependencies.
> 
> I don't think that we should be kept behind by a deprecated name and meaning. Regarding the new names that match better the `how` argument I propose the following (@victorusu please comment as well):
> 
> * `always` becomes `fully` (example: `self.depends_on('T1', how=fully)`)
> * `part_equal` becomes `by_part` (example: `self.depends_on('T1', how=by_part)`)
> * `env_equal` becomes `by_env` (example: `self.depends_on('T1', how=by_env)`)
> * `part_env_equal` becomes `by_case` (example: `self.depends_on('T1', how=by_case)`)
> 
> For the `source` and `dest` and `part_is` I'm still thinking about it :-) Suggestions are welcome!

> > I'm now going through the documentation and I think we should find better names for the udeps functions, so that they can be better combined with the how argument conceptually. For example, you can't say how=always, how=fully is better and more accurate.
> 
> @vkarak Yes, `always` is not a good name anymore. The only problem I see with `fully` is that the meaning of `DEPEND_FULLY` is different but I don't think it's a big issue, since we support interpartition dependencies.
> 
> > @ekouts I think that we should also test the behaviour of udeps when valid systems and environments are not symmetric between the two tests.
> 
> In the `test_dependecies_how_functions` the partitions and environments are not symmetric. Do you mean in a different test?


I vote the opposite of @vkarak and @ekouts 😄 I prefer much more 
```
self.depends_on('T1', how=part_equal))
```
instead of 
```
self.depends_on('T1', how=by_part))
```
But it is true that I have read the `how` as if it was a `when`. 😬 

For me, the `part_equal` shows the intention much more. And I prefer the intentional programming style much more than the `how` programming style. 

The key difference here is that even if we name the parameter `how`, it actually means `when` we connect the two points in the dependency graph. It is just an unfortunate thing in life that names clash. And we have (wisely IMHO) decided to reserve `when` for function execution (because this is a much more recurrent case/pattern).

So, I personally read the following code
```
self.depends_on('T1', how=part_equal))
```
as How do you connect the dependency graph? Well, when the partitions are equal (`part_equal`) to each other!

If I follow the same logic (if I may claim to be able to follow any - 😁), I would say that 
```
self.depends_on('T1', how=part_is('login'))
```
also fits pretty well. Because I would read
How do you connect the dependency graph? Well, when the partition is (`part_is`) equal to `'login'`!
So, I would vote `part_is` to stay as is, too.

And again, if I follow the same logic (if I may claim to be able to follow any - 😁), I would say that 
```
self.depends_on('T1', how=always)
```
can be read as "How do you connect the dependency graph? Always!"

For the `source` and `dest` names... I would vote for `this` and `parent`.
In the sense, that `this` refers to the current test case. The use of `this` is because we cannot use Python's reserved word `self` (Ah! Another name clash!).
And the word `parent` is because, in the dependency graph, I (the node) do not point to what depends on me, but what I depend on (my parent) that should be executed first (or come before first, before I can exist).

Does it make sense?

victorusu:
And an additional note on the `parent` and `this` for the `part_is`...
The expression
```
self.depends_on('T1', how=parent(part_is('login')))
```
Really means... I depend on `'T1'` and how do I connect the dependency graph? Well, when my parent's (`parent`) partition is (`part_is`) equal to `'login'`. 

Isn't it nice? It really speaks English!

victorusu (changes requested):
I think the time has passed and we cannot approve the PR as is, not due to its quality, but I think we still have some details to discuss. 😄

vkarak:
> I think the time has passed and we cannot approve the PR as is, not due to its quality, but I think we still have some details to discuss. 😄

Poor @ekouts ! 😢

vkarak:
@victorusu I tend to disagree.

> as How do you connect the dependency graph? Well, when the partitions are equal (part_equal) to each other!

I think that you are contradicting yourself, you're asking with "how" and you're replying with "when" :-D I personally read the `self.depends_on('T1', how=by_part)` as "T0" depends on "T1" by grouping the test cases *by partition*. I could accept a `group_by_part` as a name, if that makes it clearer.

> The use of this is because we cannot use Python's reserved word self (Ah! Another name clash!).

`self` is not a keyword. You can use it as you like everywhere! :-P

Also you didn't answer my question about the use case for the `dest(part_is('x'))`.

victorusu:
> @victorusu I tend to disagree.

> I think that you are contradicting yourself, you're asking with "how" and you're replying with "when" :-D I personally read the `self.depends_on('T1', how=by_part)` as "T0 depends on T1 by grouping the test cases _by partition_." I could accept a `group_by_part` as a name, if that makes it clearer.

I agree! That's the reason why I wrote...
> ... if we name the parameter `how`, it actually means `when` we connect the two points in the dependency graph. 

> `self` is not a keyword. You can use it as you like everywhere! :-P

You are right about `self` not being a keyword, but what I mean that you should not be using `self`, because it is confusing. It has clear usage in Python classes. So, it becomes confusing. But you are right that there is a clear distinction.

> Also you didn't answer my question about the use case for the `dest(part_is('x'))`.

I did. I think we should keep it! 😄 

> And an additional note on the `parent` and `this` for the `part_is`...
> The expression
> 
> ```
> self.depends_on('T1', how=parent(part_is('login')))
> ```
> 
> Really means... I depend on `'T1'` and how do I connect the dependency graph? Well, when my parent's (`parent`) partition is (`part_is`) equal to `'login'`.
> 
> Isn't it nice? It really speaks English!


I am not sure if `group_by_xxx` is much easier to read...
But I am biasied, because I do not like the `how`... I prefer when it means `when`... 🤔 
Ok, dropping personal feelings to the side...

If we stick with the `by_part` and the `by_env` naming scheme...

Then `part_is` could be

* `part_is('login')` becomes `by_part_equal('login')` (example: `self.depends_on('T1', how= by_part_equal('login'))`)
* `env_is('PrgEnv-gnu')` becomes `by_env_equal('PrgEnv-gnu')` (example: `self.depends_on('T1', how= by_env_equal('PrgEnv-gnu'))`)

What do you think?

jjotero:
I'm with @vkarak on this one. When you ask a question starting with a "how" must answer with an adverb of manner (e.g. `fully`), but you cannot answer the question with an adverb of time (e.g. `always`). That would be fine though if the question were to start with a "when", but that's not the case here. So when I see something like `self.depends_on('T1', how=fully)`, I'd read it as: _How do I (self) depend on T1? I fully depend on T1_. To me, replacing in there `fully` by `always` sounds like you're not answering what you're being asked. The same applies for all the other ones: _How do I depend on T1? I depend on T1 by partition_.

victorusu:
> I'm with @vkarak on this one. When you ask a question starting with a "how" must answer with an adverb of manner (e.g. `fully`), but you cannot answer the question with an adverb of time (e.g. `always`). That would be fine though if the question were to start with a "when", but that's not the case here. So when I see something like `self.depends_on('T1', how=fully)`, I'd read it as: _How do I (self) depend on T1? I fully depend on T1_. To me, replacing in there `fully` by `always` sounds like you're not answering what you're being asked. The same applies for all the other ones: _How do I depend on T1? I depend on T1 by partition_.

Yeap! I fully agree!

jjotero:
> But I am biasied, because I do not like the `how`... I prefer when it means `when`... 🤔

Yeah, but this is not a time-dependent relation (or is it? 🤯)

vkarak:
In graph theory terms, each of the `by_xxx` functions splits the dependency subgraph between two tests into fully connected components. So, essentially, `self.depends_on('T1', how=by_part)` means that `self` depends on `T1`, but conversely to `fully`, the subgraph of dependencies is split by partition in fully connected components. For me, it is clear and semantically correct.

vkarak:
Thanks @jjotero for your analysis! I see the full syntax in the future as follows:

```python
self.depends_on('T0', how=fully, when=self.params['daint:login'])
```

This example is the exact equivalent of the current `self.depends_on('T1', how=source(part_is('login')))` and given the analysis of @jjotero this does not read correctly as well. Given the replacement syntax above, I would suggest drop this. The open question is when you filter with `dest()`, but I really wonder if there is a practical use case for this. In any case, a user could always define a completely custom connectivity by passing his own callable to `how`.

vkarak:
@victorusu As you said, you want to use the `how` argument beyond its meaning. You can do that by writing your own `how` functions, but I don't think the framework should do this for you. Therefore, I vote against the `source`, `dest`, `part_is`, `env_is`. We could keep them, if you like, but we should leave them undocumented, thus unpublished and use-at-your-own-risk. Instead, what would make sense as a partitioning scheme is an XOR relation between partitions, environments and cases, and I firmly believe that this would cover your case in a proper way.

For example, we can add the following how functions as well:

- `by_xpart` -> group by non-equal partitions
- `by_xenv` -> group by non-equal environments
- `by_xcase -> group by non-equal partitions and non-equal environments.

vkarak:
@ekouts Another thing we will need is to update the tutorial example of dependencies. Since we now support dependencies across partitions, we could have a test that only downloads the OSU benchmarks running on the `login` partitions.

unittests/resources/checks_unlisted/deps_simple.py
```diff
@@ -18,9 +18,10 @@ def __init__(self):
         self.sanity_patterns = sn.assert_found(self.name, self.stdout)
 
 
-@rfm.parameterized_test(*([kind] for kind in ['default', 'always',
-                                              'part_equal', 'part_env_equal',
-                                              'custom', 'any', 'all', 'never']))
+@rfm.parameterized_test(*([kind] for kind in ['default', 'fully',
+                                              'by_part', 'by_case',
+                                              'custom', 'any', 'all',
+                                              'edgeless']))
```
vkarak, line 24:
```suggestion
                                              'nodeps']))
```

unittests/resources/checks_unlisted/deps_simple.py
```diff
@@ -32,15 +33,15 @@ def custom_deps(src, dst):
...
+            'fully': udeps.fully,
+            'by_part': udeps.by_part,
+            'by_case': udeps.by_case,
             'any': udeps.any(udeps.source(udeps.part_is('p0')),
                              udeps.dest(udeps.env_is('e1'))),
             'all': udeps.all(udeps.part_is('p0'),
                              udeps.dest(udeps.env_is('e0'))),
             'custom': custom_deps,
-            'never': (lambda s, d: False),
+            'edgeless': (lambda s, d: False),
```
vkarak, line 44: I don't think you need the parenthesis here.

unittests/test_dependencies.py
```diff
@@ -174,10 +199,23 @@ def test_dependecies_when_functions():
...
+            if how(t0, t1)}
+    assert deps == {
+        (t0, t1)
+        for t0 in t0_cases
+        for t1 in t1_cases
+        if (t0[0] != t1[0] or t0[1] != t1[1])
+    }
+    assert len(deps) == 32
+
+    how = udeps.part_is('p0')
```
vkarak, line 215: From this point onward, everything should go inside a different test function: `test_dependecies_how_functions_undoc`

vkarak: Just to be clear, `env_is`, `part_is`, `source`, `dest`, `all`, `any` will be all undocumented, until we decide about their future.

vkarak (changes requested):
Just minor comments. Regarding the documentation, I will go over it and I will fix myself anything wrong there.

vkarak:
I've updated the documentation. The only thing remaining is to update the tutorial example.

tutorials/config/settings.py
```diff
@@ -33,7 +33,7 @@
                     'descr': 'Login nodes',
                     'scheduler': 'local',
                     'launcher': 'local',
-                    'environs': ['gnu', 'intel', 'pgi', 'cray'],
+                    'environs': ['gnu', 'intel', 'pgi', 'cray', 'builtin'],
```
vkarak, line 36: By adding `builtin` here, we will need to update the explanations in the current tutorial. Do you want to do that, too?

vkarak (changes requested):
The updated tutorial example looks ok. We need to update the text and pay attention to the `builtin` environment that you have added. You may need to update the text of the main tutorial, too.

tutorials/config/settings.py
ekouts: Yes, I can change it. Let me have a look

ekouts: @vkarak Actually, looking at the changes that we have to make in the tutorial maybe it makes more sense to download the benchmarks in one of the `['gnu', 'intel', 'pgi', 'cray']`, what do you think?

vkarak: Haha, I knew that :-) Yes, just pick one!

ekouts: Okay, fixed it 😄 and I also fixed the `.. versionadded::` in the documentation

docs/tutorial_deps.rst
```diff
@@ -2,67 +2,86 @@
...
+   :lines: 13-44
+
+First, since we will have multiple similar benchmarks, we move all the common functionality to the :class:`OSUBenchmarkTestBase` base class.
+Again nothing new here; we are going to use two nodes for the benchmark and we set :attr:`sourcesdir <reframe.core.pipeline.RegressionTest.sourcesdir>` to ``None``, since none of the benchmark tests will use any additional resources.
+Similar to before, we will define the dependencies with the the following line:
+
+.. literalinclude:: ../tutorials/deps/osu_benchmarks.py
+   :lines: 32
+
+Here we tell ReFrame that this test depends on a test named ``OSUBuildTest`` ``by_env``.
```
vkarak, line 68:
```suggestion
Here we tell ReFrame that this test depends on a test named ``OSUBuildTest`` "by environment."
```

docs/tutorial_deps.rst
```diff
@@ -2,67 +2,86 @@
...
+This means that it will depend only on the instances of this test that have the same environment.
```
vkarak, line 69: Don't use the term "instances." Our term is "test cases."

```suggestion
This means that the test cases of this test will depend on the test cases of the ``OSUBuildTest`` that use the same environment.
```

docs/tutorial_deps.rst
```diff
@@ -2,67 +2,86 @@
...
+The partition will be different in this case.
```
vkarak, line 70: This and the sentence above are a bit hard to follow. Can you explain what you want to say?

vkarak (changes requested):

docs/tutorial_deps.rst
ekouts: I just wanted to briefly explain what `by_env` means. They will have same environment but different partition.

vkarak (approved):
All good! Ready to go!

vkarak:
Oh, still some PEP8 issues, I can fix them.

victorusu (approved):
`lgtm` too! Excellent PR. With lots of discussions! Love it!

vkarak:
For the failing tutorial check, see issue #1573.

Issue #1354, opened 2020-06-02
Drop support of scheduler-specific command-line options


ekouts:
Deprecated in #1338.
Should be dropped in 3.1 or 3.2.

vkarak:
Closed by #1483.

PR #1483, opened 2020-09-10
[feat] Drop support of scheduler-specific command-line options

### LINKED ISSUE #1368: Remove `sched_*` internal variables

rsarm:
Closes #1368 and #1354


unittests/test_schedulers.py
```diff
@@ -746,7 +732,7 @@ def test_flex_alloc_sched_access_constraint_partition(make_flexible_job):
 
 
 def test_flex_alloc_sched_access_partition(make_flexible_job):
-    job = make_flexible_job('all', sched_access=['--partition=p1'])
+    job = make_flexible_job('all')
```
ekouts, line 735: Should you pass an extra job option here, like you do for the `test_flex_alloc_partition_idle` unittest?

unittests/test_schedulers.py
```diff
@@ -103,11 +103,6 @@ def minimal_job(make_job):
 @pytest.fixture
 def fake_job(make_job):
     ret = make_job(
-        sched_nodelist='nid000[00-17]',
-        sched_exclude_nodelist='nid00016',
-        sched_partition='foo',
-        sched_reservation='bar',
-        sched_account='spam',
```
ekouts, line 110: Would it make sense to pass all the scheduler variables as job options and keep the same expected slurm output?

unittests/test_schedulers.py
```diff
@@ -158,13 +153,7 @@ def _expected_slurm_directives(job):
         '#SBATCH --ntasks-per-socket=%s' % job.num_tasks_per_socket,
         '#SBATCH --cpus-per-task=%s' % job.num_cpus_per_task,
         '#SBATCH --hint=multithread',
-        '#SBATCH --nodelist=%s' % job.sched_nodelist,
-        '#SBATCH --exclude=%s' % job.sched_exclude_nodelist,
-        '#SBATCH --partition=%s' % job.sched_partition,
-        '#SBATCH --reservation=%s' % job.sched_reservation,
-        '#SBATCH --account=%s' % job.sched_account,
         '#SBATCH --exclusive',
-        # Custom options and directives
```
ekouts, line 167: Why did you delete this comment?

unittests/test_schedulers.py
```diff
@@ -158,13 +153,7 @@ def _expected_slurm_directives(job):
         '#SBATCH --ntasks-per-socket=%s' % job.num_tasks_per_socket,
         '#SBATCH --cpus-per-task=%s' % job.num_cpus_per_task,
         '#SBATCH --hint=multithread',
-        '#SBATCH --nodelist=%s' % job.sched_nodelist,
-        '#SBATCH --exclude=%s' % job.sched_exclude_nodelist,
-        '#SBATCH --partition=%s' % job.sched_partition,
-        '#SBATCH --reservation=%s' % job.sched_reservation,
-        '#SBATCH --account=%s' % job.sched_account,
```
ekouts, line 165: If you pass them as job options when creating the job you will have to hardcode them here, but it doesn't look like a problem to me.

ekouts (changes requested):

unittests/test_schedulers.py
vkarak: I wouldn't say so. You should instead check that we have at least a test that passes options to job and not just a single one, but multiple, in order to check the order.

unittests/test_schedulers.py
```diff
@@ -731,8 +718,7 @@ def test_flex_alloc_sched_access_idle_sequence_view(make_flexible_job):
     from reframe.utility import SequenceView
 
     job = make_flexible_job('idle',
-                            sched_access=SequenceView(['--constraint=f3']),
-                            sched_partition='p3')
+                            sched_access=SequenceView(['--constraint=f3']))
```
vkarak, line 721: Here you need to pass `--partition` as an option, otherwise you break the test.

unittests/test_schedulers.py
vkarak: @rsarm This should not change. Check my comment above.

unittests/test_schedulers.py
```diff
@@ -784,12 +771,6 @@ def test_flex_alloc_valid_multiple_constraints(make_flexible_job):
     assert job.num_tasks == 4
 
 
-def test_flex_alloc_valid_partition_cmd(make_flexible_job):
-    job = make_flexible_job('all', sched_partition='p2')
-    prepare_job(job)
-    assert job.num_tasks == 8
```
vkarak, line 790: You need this test, but pass the partition as a `job.option`. These set of tests exercise a complete different aspect than that the options are passed correctly to the job. So *all* the tests should remain and be adapted.

vkarak (changes requested):

vkarak (approved):

vkarak:
@jenkins-cscs retry all

ekouts (changes requested):
Sorry, but I just realized that there are also parts in the documentation that have to be removed. For example we still have the option `--partition` in the man pages.

victorusu (approved):
`lgtm`

ekouts (approved):

vkarak:
@jenkins-cscs retry tsa

PR #1489, opened 2020-09-18
[bugfix] Fix JSON report generation for deferred expressions

### LINKED ISSUE #1459: Object of type '_DeferredExpression' is not JSON serializable
This code will fail with:

```
version:           3.2-dev0 (rev: 576eb3f1)
PERFORMANCE REPORT
FailTest
- dom:login
   - PrgEnv-gnu
      * num_tasks: 1
      * elapsed_time: 8 s
      * filename: 8 (9.cpp)
reframe: type error: ../../../../../../../../usr/lib64/python3.6/json/encoder.py:180: 
Object of type '_DeferredExpression' is not JSON serializable
                        o.__class__.__name__)
```

when run like this:

```bash
reframe -C cscs.py -p PrgEnv-gnu \
--keep-stage-files --prefix=$SCRATCH \
--report-file eff.json \
--performance-report -v -r -c bug_report.py'
``` 

I believe the issue is in the json reporting.

```python
import os, sys
import reframe as rfm


@rfm.simple_test
class FailTest(rfm.RunOnlyRegressionTest):
    def __init__(self):
        self.descr = ('RunOnlyRegressionTest')
        self.valid_systems = ['dom:login']
        self.executable = 'echo 9.cpp'
        self.sanity_patterns = sn.assert_found('9', self.stdout)
        self.perf_patterns = {
            'elapsed_time': self.elapsed,
            'filename': self.elapsed,
        }
        self.reference = {
            '*': {
                'elapsed_time': (0, None, None, 's'),
                'filename': (0, None, None, self.fnm),
            }
        }

    @property
    @sn.sanity_function
    def elapsed(self):
        return 8

    @property
    @sn.sanity_function
    def fnm(self):
        return '(9.cpp)'
```

teojgo:
* Add a custom 'json' utility package which uses the
`__rfm_json_encode__` method for json encoding if defined.

* Implement the above method for `_Deferred_Expression` objects.

* Add unittests for the custom `json` utility package.

Fixes #1459


unittests/test_utility.py
```diff
@@ -1293,3 +1295,26 @@ def test_cray_cle_info_missing_parts(tmp_path):
...
+    json_dump = tmp_path / 'test.json'
+    json_dump_indent = tmp_path / 'test_indent.json'
+    with open(json_dump, 'w') as f:
+        jsonext.dump({'foo': sn.defer(['bar'])}, f)
+
+    with open(json_dump, 'r') as f:
+        assert '{"foo": ["bar"]}' == f.read()
+
+    with open(json_dump_indent, 'w') as f:
+        jsonext.dump({'foo': sn.defer(['bar'])}, f, indent=2)
```
vkarak, line 1310: Better pass `separators=(',', ':')` instead of `indent=2` so that your assertion below becomes much more compact.

unittests/test_utility.py
```diff
@@ -1293,3 +1295,26 @@ def test_cray_cle_info_missing_parts(tmp_path):
...
+
+    with open(json_dump_indent, 'r') as f:
+        assert '{\n  "foo": [\n    "bar"\n  ]\n}' == f.read()
+
+
+def test_jsonext_dumps():
+    assert '"foo"' == jsonext.dumps('foo')
+    assert '{"foo": ["bar"]}' == jsonext.dumps({'foo': sn.defer(['bar'])})
+    assert ('{\n  "foo": [\n    "bar"\n  ]\n}' ==
+            jsonext.dumps({'foo': sn.defer(['bar'])}, indent=2))
```
vkarak, line 1320: Same comment about the `indent` argument here.

vkarak (changes requested):

vkarak (approved):
lgtm

Issue #1196, opened 2020-02-25
Verify exit code of tests


davidedelvento:
Currently, exit code of tests are ignored.

With all the limitations of the case, namely:
* some scheduler backends
* the job's exit code is essentially the exit code of the last command
* and possibly more

It would be good to have another test performed automatically which could catch some otherwise silent failures

PR #1490, opened 2020-09-18
[feat] Add configuration parameter for automatically capturing command errors and exit code in job scripts

jgphpc:
Will fix https://github.com/eth-cscs/reframe/issues/1196

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1072,7 +1072,7 @@ General Configuration
    :required: No
    :default: ``false``
 
-   Trap command errors and exit in the generated job scripts.
+   Trap command errors in the generated job scripts and let them exit immediately.
 
 
 .. js:attribute:: .general[].keep_stage_files
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1067,6 +1067,14 @@ General Configuration
    Ignore test name conflicts when loading tests.
 
 
+.. js:attribute:: .general[].ignore_exit_code
+
+   :required: No
+   :default: ``true``
+
+   Ignore job exit code
+
+
 .. js:attribute:: .general[].keep_stage_files
 
    :required: No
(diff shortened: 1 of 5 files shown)


docs/config_reference.rst
```diff
@@ -1067,6 +1067,14 @@ General Configuration
    Ignore test name conflicts when loading tests.
 
 
+.. js:attribute:: .general[].ignore_exit_code
```
vkarak, line 1070: I would make the feature a bit more general and call it `trap_job_errors` with default value `false`. When set, apart from what you are doing now, we should generate the job script with `trap_errors=True` here:

https://github.com/eth-cscs/reframe/blob/master/reframe/core/pipeline.py#L1317-L1320

This would cause the script to exit if any command in it fails.

vkarak (changes requested):
We will also need a unit test.

vkarak (changes requested):

victorusu:
Does it make sense to allow the check to bypass this configuration? Something like
```
self. trap_job_errors = False
```
The use case is when one needs to write a check where one expects a command to exit code different than zero, while the default config is set to `trap_job_errors=True`.
In this case, this check cannot be executed together with other checks, it needs a different ReFrame invocation.

docs/config_reference.rst
```diff
@@ -1072,7 +1072,7 @@ General Configuration
    :required: No
    :default: ``false``
 
-   Trap any failing command from the jobscript.
+   Trap command errors and exit in the generated job scripts.
```
vkarak, line 1075: I think this is less confusing.

```suggestion
   Trap command errors in the generated job scripts and let them exit immediately.
```

vkarak (changes requested):

vkarak (approved):
lgtm now

PR #1499, opened 2020-09-24
[bugfix] Load configuration even if the username cannot be retrieved

### LINKED ISSUE #1496: Reframe fails to perform actions when no configuration is passed
The error thrown is:
`failed to load configuration: [Errno 6] No such device or address`

It is easy to reproduce with a docker container using the following Dockerfile:

```
FROM python

RUN git clone https://github.com/eth-cscs/reframe.git && \
    cd reframe && pip install -r requirements.txt

ENTRYPOINT ["/reframe/bin/reframe"]

CMD ["-l"]
```

To build the above run `docker build --tag=reframe .`

And to run: `docker run reframe`

teojgo:
Fixes #1496

### DIFF
--- a/reframe/core/config.py
+++ b/reframe/core/config.py
@@ -5,6 +5,7 @@
 
 import copy
 import fnmatch
+import getpass
 import itertools
 import json
 import jsonschema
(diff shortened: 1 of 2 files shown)


reframe/core/config.py
```diff
@@ -551,7 +552,7 @@ def handler_list(handler_config, basedir=None):
 def _find_config_file():
     # The order of elements is important, since it defines the priority
     prefixes = [
-        os.path.join(os.getlogin(), '.reframe'),
+        os.path.join(getpass.getuser(), '.reframe'),
```
vkarak, line 555: Better use the `os_ext.osuser()` which does this, but captures also the exception that can be possibly thrown and you should also check that the return value is not `None` and add also a unit test to test that the `None` is handled correcty.

vkarak (changes requested):

vkarak (approved):
lgtm

Issue #1362, opened 2020-06-04
monkeypatching in e.g. run_before() methods


sjpb:
I've got a test method decorated with a @run_before() method, like (for example):

```
@rfm.run_before('run')
def no_run(self):
        """ Turn the run phase into a no-op to help debugging. """

        with open(os.path.join(self.stagedir, 'noop.sh'), 'w') as noop:
            noop.write('#!/bin/bash\necho "noop $@"\n')
        self.executable = "./noop.sh"
```

which really I want to pull out into a separate library so I can use it on various classes (it's more complex than the method here so it makes more sense!).

I can't figure out how to monkey-patch this in though - the run_before decorator seems to make a difference. Ideally I'd do something like:

```
@set_no_run
@rfm.simple_test
class MyTest(rfm.RegressionTest):
    <snip>
```

I tried this:

```
def set_no_run(cls):

    def no_run(self):
        """ Turn the run phase into a no-op. """

        with open(os.path.join(self.stagedir, 'noop.sh'), 'w') as noop:
            noop.write('#!/bin/bash\necho "noop $@"\n')
        self.executable = "./noop.sh"

    cls.no_run = no_run
    cls.no_run = rfm.run_before('run')(cls.no_run)
    return cls
```

but (unsurprisingly?) that doesn't work (as in, doesn't change the executable).

Maybe I'm missing something about the way binding works in Py3 here but I'd appreciate any suggestions for how to do something like this, even if it's via another route.

sjpb:
I guess one key motivation not shown in the above is that really the library-provided decorator (or however it's done) is going to add methods at multiple pipeline stages.

vkarak:
The reason that this does not work is that hooks mechanism works essentially through a metaclass, which is responsible for the creation of the `RegressionTest`. All the decorator does is set a function attribute. This attribute will be used by the metaclass when creating the `RegressionTest` class in order to set up the list of pipeline hooks for each pipeline stage. That means that in order of this mechanism to work, the body of the test class should have been executed before. A class decorator, as in your case, acts too late, after the class has been created, so whichever method you decorate it will be simply ignored.

One thing that could potentially help you is that pipeline hooks are inherited, so you could possibly define in a mixin class that your test could inherit from as well:

```python
class DisableRun:
    @rfm.run_before('run')
    def no_run(self):
        """ Turn the run phase into a no-op to help debugging. """

        with open(os.path.join(self.stagedir, 'noop.sh'), 'w') as noop:
[...]

PR #1505, opened 2020-10-02
[test] Reorganise directory structure for benchmarks

jjotero:
- Tests in the CUDA directory have been moved to PrgEnv/cuda.
- Some of the cuda checks in this directory have been updated to fetch the source from the Nvidia repos.
- The microbenchmarks directory has been split between gpu, cpu and mpi tests.


cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -0,0 +1,159 @@
...
+    def __init__(self):
+        super().__init__()
+        self.descr = 'Cuda-aware MPI test from the NVIDIA repo.'
+        self.sourcesdir = ('https://github.com/NVIDIA-developer-blog/'
+                           'code-samples.git')
+        self.valid_systems = ['daint:gpu', 'dom:gpu',
+                              'kesch:cn', 'tiger:gpu',
+                              'arolla:cn', 'tsa:cn',
+                              'ault:amdv100', 'ault:intelv100']
+        if self.current_system.name == 'kesch':
```
victorusu, line 22: Does it make sense to parameterize this test and pass all these combinations as parameters?

victorusu (commented):
I am not sure whether my proposal will make the code more readable. 
But I think that lines 22-34 from cscs-checks/prgenv/cuda/cuda_aware_mpi.py could be replaced by a parameterization set.

victorusu:
Ok to test

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
jjotero: I see what you mean, but this seems to be the way things are implemented in many other cases. From what I understand, parametrised tests are used to change actual test parameters, not stuff related to the environments. Perhaps this could just go in the env settings?

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -0,0 +1,151 @@
...
+        self.num_gpus_per_node = 1
+        self.executable_opts = ['-t %d 1' % self.num_tasks]
+
+    @rfm.require_deps
+    def set_executable(self, CudaAwareMPICheck):
+        self.executable = os.path.join(
+            CudaAwareMPICheck().stagedir,
+            'posts','cuda-aware-mpi-example',
+            'bin', 'jacobi_cuda_aware_mpi'
+        )
```
vkarak, line 150: Since this is always the same you could inherit it from a random mixin class. The decorators should work even for non `RegressionTest`-derived classes. Here was a related discussion: https://github.com/eth-cscs/reframe/issues/1362

cscs-checks/prgenv/cuda/cuda_samples.py
```diff
@@ -0,0 +1,165 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+import os
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+class CudaSamples(rfm.RegressionTest):
+    def __init__(self):
+        super().__init__()
```
vkarak, line 13: No need of this.

cscs-checks/prgenv/cuda/cuda_samples.py
```diff
@@ -0,0 +1,165 @@
...
+        self.build_system = 'Make'
+        if self.current_system.name in ['daint']:
+            self.prebuild_cmds = ['export CUDA_HOME=$CUDATOOLKIT_HOME']
+        else:
+            self.prebuild_cmds = []
+
+        self.build_system.options = [
+            'SMS="%s"' %
+            self.nvidia_sm,
+            'CUDA_PATH=$CUDA_HOME']
```
vkarak, line 48:
```suggestion
        self.build_system.options = [f'SMS="{self.nvidia_sm}"', 
                                     f'CUDA_PATH=$CUDA_HOME']
```

cscs-checks/prgenv/cuda/cuda_samples.py
```diff
@@ -0,0 +1,165 @@
...
+        self.maintainers = ['JO']
+        self.tags = {'production', 'external_resosurces'}
+
+
+@rfm.required_version('>=2.14')
```
vkarak, line 53: You can remove this.

cscs-checks/prgenv/cuda/cuda_samples.py
```diff
@@ -0,0 +1,165 @@
...
+        self.executable = 'Samples/deviceQuery/deviceQuery'
+        self.sanity_patterns = sn.assert_found(
+            r'Result = PASS',
+            self.stdout)
+
+
+class DependentCudaSamples(CudaSamples):
+    def __init__(self):
+        super().__init__()
+        self.depends_on('CudaDeviceQueryCheck')
```
vkarak, line 72: I would add the `depends_on` in every test, cos it makes clearer their dependencies.

cscs-checks/prgenv/cuda/cuda_samples.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+@rfm.required_version('>=2.14')
```
vkarak, line 75: No need of this. Dependencies were also introduced much later :-)

cscs-checks/prgenv/cuda/cuda_samples.py
```diff
@@ -0,0 +1,165 @@
...
+@rfm.simple_test
+class CudaMatrixMultCublasCheck(DependentCudaSamples):
+    def __init__(self):
+        super().__init__()
+        self.descr = 'CUDA simpleCUBLAS test'
+        self.sanity_patterns = sn.assert_found(
+            r'test passed',
+            self.stdout)
```
vkarak, line 106: Doesn't all this fit in a single line?

cscs-checks/prgenv/cuda/cuda_samples.py
```diff
@@ -0,0 +1,165 @@
...
+        self.descr = 'CUDA concurrentKernels test'
+        self.sanity_patterns = sn.assert_found(
+            r'Test passed',
+            self.stdout)
+
+    @rfm.require_deps
+    def set_prebuild_cmds(self, CudaDeviceQueryCheck):
+        self.prebuild_cmds += ['cd %s' % os.path.join(
+            CudaDeviceQueryCheck().stagedir,
+            'Samples', 'concurrentKernels')]
```
vkarak, line 89: It's strange that you have to cd in the other test's stage directory? Couldn't you just use it's executable?

vkarak (changes requested):
Nice that you modernized the tests! The dir structure makes more sense. I have some additional comments here:

1. You can also remove the `cscs-checks/cuda/reframe.log`. It shouldn't be there and it's forgotten :-)
2. Why OSU tests are not under `mpi/`? Technically, they test also the GPU RDMA stuff, so they can't be classified under cpu only.

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -0,0 +1,151 @@
...
+        self.tags={'production', 'external_resources'}
+
+    @ rfm.run_before('compile')
+    def set_compilers(self):
+        if self.current_environ.name == 'PrgEnv-pgi':
+           self.build_system.cflags=['-std=c99', ' -O3']
+
+        self.build_system.options += [
+            'MPICC="%s"' % self.build_system._cc(self.current_environ),
+            'MPILD="%s"' % self.build_system._cxx(self.current_environ)
```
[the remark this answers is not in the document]
jjotero: Agree.

cscs-checks/prgenv/cuda/cuda_samples.py
jjotero: The entire repo is cloned in the CudaDeviceQueryCheck directory, so the "cd" is necessary to compile the code there. Alternatively, one could just copy the source to the directory of the current case. Does that sound better?

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -0,0 +1,151 @@
...
+            self.valid_prog_environs += ['PrgEnv-pgi']
+        elif self.current_system.name in ['ault']:
+            self.valid_prog_environs = ['PrgEnv-gnu']
+
+        if self.current_system.name in ['arolla', 'tsa','ault']:
+            self.valid_prog_environs = ['PrgEnv-gnu']
+        else:
+            self.modules = ['craype-accel-nvidia60']
+
+        self.executable = '../bin/jacobi_cuda_aware_mpi'
```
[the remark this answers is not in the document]
jjotero: Oh, I see! This `self.executable` gets overridden on the child classes. It gets overridden in the same way for both, so it makes sense to move it up here in the parent class.

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -0,0 +1,151 @@
...
+            CudaAwareMPICheck().stagedir,
+            'posts','cuda-aware-mpi-example',
+            'bin', 'jacobi_cuda_aware_mpi'
+        )
+
+
+@rfm.simple_test
+class CudaAwareMPITwoNodesCheck(CudaAwareMPIRuns):
+    def __init__(self):
+        super().__init__()
```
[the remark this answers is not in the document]
jjotero: If I pop this out, the test doesn't run because it won't inherit the `self.valid_systems` and the rest of the settings defined on the parent class `CudaAwareMPIRuns`. This is the same for the `CudaAwareMPIOneNodeCheck`.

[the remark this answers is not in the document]
vkarak: You are right this is needed :-)

cscs-checks/prgenv/cuda/cuda_samples.py
vkarak: I see, but the problem is that the test is no more self contained. It works with the async execution policy just because you clone the repo multiple times, one per each programming environment. If #1481 is merged, you would be able to have a single test running on a login partition that fetches the repo and the rest of the tests to depend on that. In that case, though, your current approach wouldn't work because the tests would start to step on each other during the compilation phase. So for the moment I would rather clone the repo in each test (the effect is practically the same to what you have now I think), but the tests will be self-contained.

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
vkarak: Generally, we can parametrize tests on anything and there are cases that it is useful to parametrize even on valid systems and environments. In this case, I honestly don't what would be the best solution :-)

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -32,79 +32,84 @@ def __init__(self):
...
+        self.postbuild_cmds = ['ls ../bin']
+        self.sanity_patterns = sn.assert_found(r'jacobi_cuda_aware_mpi',
                                                self.stdout)
-        self.maintainers=['JO']
-        self.tags={'production', 'external_resources'}
+        self.maintainers = ['JO']
+        self.tags = {'production', 'external_resources'}
+
+    @rfm.run_before('compile')
+    def set_cuda_home(self):
```
vkarak, line 55: I would call this hook `cdt2008_pgi_workaround` so as to easily grep it and then temporarily disable it with `--disable-hook`.

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -32,79 +32,84 @@ def __init__(self):
...
+        if self.current_system.name in [
```
vkarak, line 56: The problem is only for the PGI compiler, so you should check only against the `current_environ`.

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -32,79 +32,84 @@ def __init__(self):
...
+                'daint', 'dom'] and osx.cray_cdt_version() == '20.08':
+            self.prebuild_cmds += ['export CUDA_HOME=$CUDATOOLKIT_HOME']
```
vkarak, line 58: You could alternatively set it in `self.variables`, in which case it will be set for both the compilation and the run.

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -32,79 +32,84 @@ def __init__(self):
...
 
-        self.postbuild_cmds= ['ls ../bin']
-        self.sanity_patterns= sn.assert_found(r'jacobi_cuda_aware_mpi',
+        self.postbuild_cmds = ['ls ../bin']
+        self.sanity_patterns = sn.assert_found(r'jacobi_cuda_aware_mpi',
                                                self.stdout)
-        self.maintainers=['JO']
-        self.tags={'production', 'external_resources'}
+        self.maintainers = ['JO']
+        self.tags = {'production', 'external_resources'}
```
vkarak, line 52:
```suggestion
        self.tags = {'production', 'scs'}
```

You're not using our external resources anymore.

cscs-checks/prgenv/cuda/cuda_aware_mpi.py
```diff
@@ -0,0 +1,151 @@
...
+        self.depends_on('CudaAwareMPICheck')
+        self.sanity_patterns = sn.assert_found(r'Stopped after 1000 iterations'
+                                               r' with residue 0.00024', 
+                                               self.stdout)
+
+ 
+@rfm.simple_test
+class CudaAwareMPIOneNodeCheck(CudaAwareMPIRuns):
+    def __init__(self):
+        super().__init__()
```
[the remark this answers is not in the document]
vkarak: Indeed, this is not correct.

cscs-checks/prgenv/cuda/cuda_samples.py
jjotero: OK, got it. I just changed all these checks to clone the repo on each test. As expected, this makes the whole check quite a bit slower, so perhaps it's worth revisiting this in the future once #1481 is merged, and we can then try and come up with a way to reduce all these calls to `git clone` in here.
On the coding style side of things, I know that all these checks could just be parametrised, but I think that would take a bit of a dent in terms of readability.

vkarak: I hope to merge #1481 by next week or the week after, so we can then review the structure. Regarding parametrised vs. classes, I don't have a strong feeling. Perhaps it's better as you have it now.

vkarak (approved):
lgtm

vkarak:
@jjotero There is a strange error happening in Tsa? Can you check?

```
  * Reason: spawned process error: command 'git clone https://github.com/NVIDIA/cuda-samples.git /scratch-shared/meteoswiss/scratch/jenscscs/reframe-ci-tsa-e3da1c1-3960/stage/tsa/cn/PrgEnv-gnu-nompi/CudaGraphsCGCheck' failed with exit code 128:
=== STDOUT ===
=== STDERR ===
fatal: destination path '/scratch-shared/meteoswiss/scratch/jenscscs/reframe-ci-tsa-e3da1c1-3960/stage/tsa/cn/PrgEnv-gnu-nompi/CudaGraphsCGCheck' already exists and is not an empty directory.

```

vkarak:
@jenkins-cscs retry tsa

Issue #1502, opened 2020-10-01
Wrong rerun message when a test fails before the setup phase


ekouts:
When a test fails before the setup phase (for example when it fails because of failed dependencies), the rerun message is wrong because the partition hasn't been set yet. `-n T9 -p None --system generic` would filter T9 in this example:
```
$ ./bin/reframe -c unittests/resources/checks_unlisted/deps_complex.py -r
[ReFrame Setup]
  version:           3.2-dev2 (rev: 3c8f324a)
  command:           './bin/reframe -c unittests/resources/checks_unlisted/deps_complex.py -r'

...

------------------------------------------------------------------------------
FAILURE INFO for T9 
  * Test Description: T9
  * System partition: generic
  * Environment: None
  * Stage directory: None
  * Node list: None
  * Job type: batch job (id=None)
  * Maintainers: []
  * Failing phase: startup
  * Rerun with '-n T9 -p None --system generic'
  * Reason: task dependency error: dependencies failed
------------------------------------------------------------------------------
```

PR #1506, opened 2020-10-05
[bugfix] Print current environment and partition correctly even if a test fails before the setup

rsarm:
Fixes https://github.com/eth-cscs/reframe/issues/1502

### DIFF
--- a/reframe/frontend/statistics.py
+++ b/reframe/frontend/statistics.py
@@ -106,9 +106,13 @@ class TestStats:
                     'time_total': t.duration('total')
                 }
                 partition = t.testcase.partition
+                environ = t.testcase.environ
+                # We take partition and environment from the test case and not
+                # from the check, since if the test fails before `setup()`,
+                # these are not set inside the check.
                 entry['system'] = partition.fullname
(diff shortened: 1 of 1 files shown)


reframe/frontend/statistics.py
```diff
@@ -105,15 +105,10 @@ def json(self, force=False):
...
-                partition = check.current_partition
-                environ = check.current_environ
-                if partition:
-                    entry['system'] = partition.fullname
-                    entry['scheduler'] = partition.scheduler.registered_name
-
-                if environ:
-                    entry['environment'] = environ.name
-
+                partition = t.testcase.partition
```
vkarak, line 108: Better write it as follows and add a comment explaining why we are taking these values from the test case and not from the check:

```suggestion
                # We take partition and environment from the test case and not from the check, since if the test fails before `setup()`, these are not set inside the theck.
                partition = t.testcase.partition
                environ = t.testcase.environ
```

vkarak (changes requested):

ekouts (approved):

vkarak (approved):
lgtm

PR #1508, opened 2020-10-06
[bugfix] Do not require a configuration to be resolved always at system level

### LINKED ISSUE #1453: Default environment is required
At the moment, reframe seems to require a default environment. If I do something like this:

```
        {
            'name':'openfoam',
            'target_systems': ['mysys:ib-gcc7-openmpi4-ucx'],
            'modules': ['openfoam-org/7-2ceqb4l']
        },
        {
            'name':'openfoam',
            'target_systems': ['mysys:ib-gcc9-openmpi4-ucx'],
            'modules': ['openfoam-org/7-4zgjbg2']
        },
```

Then I get a message like "environment 'openfoam' not defined for system 'mysys'" (paraphrased, lost the original terminal).

I have to add an empty default environment of the same name for it to work:
```
        {
            'name':'openfoam'
        },
```

Which is fine, because I can use the 'environs' values in the partition to restrict tests to valid partition+environ combinations. *Except* that it is an error trap - for example if I misspelt the 'target_systems' above, then reframe thinks there is a valid combination because of the default one, then I get a cryptic error during module loads.

To me that case with no default seems quite reasonable/normal, but maybe it's not.

vkarak:
Fixes #1453.

### DIFF
--- a/reframe/core/systems.py
+++ b/reframe/core/systems.py
@@ -318,7 +318,9 @@ class System:
                 )
             )
 
-        # Restore configuration
+        # Restore configuration, but ignore unresolved sections or
+        # configuration parameters at the system level; if we came up to this
+        # point, then all is good at the partition level, which is enough.
         site_config.select_subconfig(config_save, ignore_resolve_errors=True)
         return System(
             name=sysname,
--- a/reframe/core/systems.py
+++ b/reframe/core/systems.py
@@ -319,7 +319,7 @@ class System:
             )
 
         # Restore configuration
-        site_config.select_subconfig(config_save)
+        site_config.select_subconfig(config_save, ignore_resolve_errors=True)
         return System(
             name=sysname,
             descr=site_config.get('systems/0/descr'),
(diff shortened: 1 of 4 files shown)


reframe/core/systems.py
```diff
@@ -319,7 +319,7 @@ def create(cls, site_config):
             )
 
         # Restore configuration
-        site_config.select_subconfig(config_save)
+        site_config.select_subconfig(config_save, ignore_resolve_errors=True)
```
vkarak, line 322: I think I should add a comment here, why we do that.

reframe/frontend/cli.py
```diff
@@ -463,7 +463,8 @@ def main():
             site_config = config.load_config(converted)
 
         site_config.validate()
-        site_config.select_subconfig(options.system)
+        site_config.select_subconfig(options.system,
+                                     ignore_resolve_errors=True)
```
vkarak, line 467: I think I should add a comment here, why we do that.

vkarak:
@jenkins-cscs retry daint

teojgo (approved):
Other than the comments that you have posted I have tested and it looks good.

PR #1512, opened 2020-10-07
[test] Add test for parallel HDF5 via h5py

teojgo:


victorusu (approved):
`lgtm`. 
I just wonder @teojgo if it makes sense to take the opportunity and add a performance section to this test.
What do you think?

teojgo:
> `lgtm`.
> I just wonder @teojgo if it makes sense to take the opportunity and add a performance section to this test.
> What do you think?

Well, the purpose of the test is to make sure that h5py is working with parallel HDF5. We could add a performance check either on this but we would need a realistic example.

cscs-checks/libraries/io/h5py_parallel_run.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@rfm.simple_test
+class H5PyTest(rfm.RunOnlyRegressionTest):
+    def __init__(self):
+        self.descr = 'Test that parallel HDF5 can be used by h5py'
+        self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc']
+        self.valid_prog_environs = ['builtin']
+        self.modules = ['h5py']
+        self.num_nodes = 1
```
vkarak, line 17: Do you mean `num_tasks_per_node` here?

cscs-checks/libraries/io/h5py_parallel_run.py
```diff
@@ -0,0 +1,28 @@
...
+        self.num_tasks = 4
+        expected_dataset = ', '.join(str(i) for i in range(self.num_tasks))
+        self.sanity_patterns = sn.assert_found(
+            rf'.*DATA\s+{{\s+\(0\): {expected_dataset}\s*}}',
+            self.stdout
+        )
+
```
vkarak, line 24: Remove this line.

vkarak (changes requested):

vkarak (approved):

PR #1518, opened 2020-10-08
[ci] Remove the unit tests from the coverage report

### LINKED ISSUE #1516: Do not include unit test coverage in the coverage report
Currently, we are including the coverage of the unit tests in the coverage report. I don't think that this is correct. We should only include the coverage of the framework. This will make the reported coverage number to drop a bit, but it's more correct.

teojgo:
Fixes #1516

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ install:
     - pip install codecov
 
 script:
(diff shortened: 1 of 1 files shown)


vkarak (approved):
lgtm

PR #1529, opened 2020-10-12
[feat] Add option to install pygelf in boostrap script

### LINKED ISSUE #1522: Add option to install pygelf when executing the bootstrap.sh

teojgo:
Fixes #1522


bootstrap.sh
```diff
@@ -24,13 +24,15 @@ usage()
     echo "Usage: $0 [-h] [+docs]"
     echo "Bootstrap ReFrame by pulling all its dependencies"
     echo "  -P EXEC  Use EXEC as Python interpreter"
+    echo "  -O       Install optional packages using pip"
```
vkarak, line 27: I would rather prefer `+pygelf` option.

vkarak (changes requested):

bootstrap.sh
teojgo: I did like that in case we need to add more packages in the feature.

vkarak: I know, but I prefer to restrict it and be more explicit. I don't think we will be adding random packages very often.

bootstrap.sh
```diff
@@ -64,11 +63,11 @@ export PYTHONPATH=$(pwd)/external:$(pwd)/external/usr/lib/python$pyver/site-pack
...
+if [ x"$1" == x"+docs" -o x"$2" == x"+docs" ]; then
+    CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q -r docs/requirements.txt --target=external/ --upgrade
+    make -C docs PYTHON=$python
 fi
 
-if [ x"$1" == x"+docs" ]; then
-    CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q -r docs/requirements.txt --target=external/ --upgrade
-    make -C docs
+if [ x"$1" == x"+pygelf" -o x"$2" == x"+pygelf" ]; then
+    CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q pygelf --target=external/ --upgrade
```
vkarak: I'd get in a while loop where I'd be checking the value of `$1` in a `case` statement and act accordingly. At the end of the I'd `shift` the argument list. You need also `shift $OPTIND` before entering the loop. Check the `getopts` documentation in [bash(1)](https://linux.die.net/man/1/bash).

vkarak (changes requested):

bootstrap.sh
```diff
@@ -43,8 +43,15 @@ if [ -z $python ]; then
     python=python3
 fi
 
-pyver=$($python -V | sed -n 's/Python \([0-9]\+\)\.\([0-9]\+\)\..*/\1.\2/p')
+while [ -n "$1" ]; do
+    case "$1" in
+        "+docs") MAKEDOCS="true" && shift ;;
+        "+pygelf") PYGELF="true" && shift ;;
+        *) usage && exit 0 ;;
```
vkarak, line 50: This is an error, you should exit with non-zero:

```suggestion
        *) usage && exit 1 ;;
```

bootstrap.sh
```diff
@@ -63,11 +70,11 @@ export PYTHONPATH=$(pwd)/external:$(pwd)/external/usr/lib/python$pyver/site-pack
 CMD $python -m pip install --no-cache-dir -q --upgrade pip --target=external/
 CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q -r requirements.txt --target=external/ --upgrade
 
-if [ x"$1" == x"+docs" -o x"$2" == x"+docs" ]; then
+if [ "$MAKEDOCS" == "true" ]; then
```
vkarak, line 73: Better like that:

```suggestion
if [ -n "$MAKEDOCS" ]; then
```

vkarak: Similarly for pygelf.

bootstrap.sh
```diff
@@ -63,11 +70,11 @@ export PYTHONPATH=$(pwd)/external:$(pwd)/external/usr/lib/python$pyver/site-pack
...
     CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q -r docs/requirements.txt --target=external/ --upgrade
     make -C docs PYTHON=$python
 fi
 
-if [ x"$1" == x"+pygelf" -o x"$2" == x"+pygelf" ]; then
+if [ "$PYGELF" == "true" ]; then
     CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q pygelf --target=external/ --upgrade
```
vkarak, line 79: We should pin that version, too, but I don't want to pin it here. What if we introduced it in the `requirements.txt` like that?

```python
#.pygelf pygelf==0.3.6
```

So then you create a new temporary `requirements.txt` by uncommenting this line with sed and use that for installing everything.

vkarak (changes requested):

bootstrap.sh
```diff
@@ -68,13 +68,19 @@ export PATH=$(pwd)/external/usr/bin:$PATH
 export PYTHONPATH=$(pwd)/external:$(pwd)/external/usr/lib/python$pyver/site-packages:$PYTHONPATH
 
 CMD $python -m pip install --no-cache-dir -q --upgrade pip --target=external/
-CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q -r requirements.txt --target=external/ --upgrade
 
-if [ "$MAKEDOCS" == "true" ]; then
+if [ -n "$PYGELF" ]; then
+    cp requirements.txt requirements_tmp.txt
```
vkarak, line 73: You should create a real temporary file with  `mktemp`, which you then will delete.

bootstrap.sh
```diff
@@ -68,13 +68,19 @@ export PATH=$(pwd)/external/usr/bin:$PATH
...
+    sed -i -e 's/^#\.pygelf[[:space:]]\+//g' requirements.txt
```
vkarak, line 74: No you need not and should not change the `requirements.txt`, but rather pipe it to `sed` and `sed` will write the temporary file: `cat requirements.txt | sed stuff > $tmp_requirements`

bootstrap.sh
```diff
@@ -68,13 +68,19 @@ export PATH=$(pwd)/external/usr/bin:$PATH
...
+    CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q -r requirements.txt --target=external/ --upgrade
+    mv requirements_tmp.txt requirements.txt
```
vkarak, line 76: No need of this if you follow my comments. Just delete the temporary file.

vkarak (changes requested):

bootstrap.sh
```diff
@@ -70,10 +70,10 @@ export PYTHONPATH=$(pwd)/external:$(pwd)/external/usr/lib/python$pyver/site-pack
 CMD $python -m pip install --no-cache-dir -q --upgrade pip --target=external/
 
 if [ -n "$PYGELF" ]; then
-    cp requirements.txt requirements_tmp.txt
-    sed -i -e 's/^#\.pygelf[[:space:]]\+//g' requirements.txt
-    CMD $python -m pip install --use-feature=2020-resolver --no-cache-dir -q -r requirements.txt --target=external/ --upgrade
-    mv requirements_tmp.txt requirements.txt
+    tmp_requirements=$(mktemp requirements_XXX.txt)
```
vkarak, line 73: Just `mktemp` is the most portable way. I will fix that and merge it.

vkarak (changes requested):

vkarak (approved):
lgtm

PR #1531, opened 2020-10-12
[feat] Set the default location for log files to be the system's temporary directory

### LINKED ISSUE #1177: Change default location of the log file to $TMPDIR
Some users would like to place a unique log file under $TMPDIR, but currently this is not possible by just setting it in the log handler. I'd suggest to change the default log location to be on $TMPDIR and then moved over to output dir with the `--save-log-files`. Besides this use case, I think that this behaviour is more natural and does not leave behind stale files.

vkarak:
This PR renders the `name` configuration parameter of the `file` log handler optional and sets the default location to be in the system's temporary directory. The behaviour of the framework with the current configuration files that set the `name` parameter is not affected.

Documentation and the tutorials are also updated.

This PR allows the parallelization of the unit tests as discussed in #134. 

Fixes #1177.

### DIFF
--- a/config/cscs-ci.py
+++ b/config/cscs-ci.py
@@ -213,7 +213,6 @@ site_configuration = {
             'handlers': [
                 {
                     'type': 'file',
-                    'name': 'reframe.log',
                     'level': 'debug',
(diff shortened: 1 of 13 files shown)


ekouts (approved):
lgtm

teojgo (approved):

PR #1538, opened 2020-10-16
[feat] Add ability to restore a test session and rerun selected tests

### LINKED ISSUE #1222: Add a `--rerun-failed` option that will allow ReFrame to run only the failed test cases in a separate invocation
This request comes from @ZQyou and has been discussed briefly in #1176. 

This option is different from `--max-retries`, which retries the tests in the same run. We want ReFrame to load a result of a previous run (see also #1221) and rerun only the failed test cases.

rsarm:
Closes #1222

EDIT @vkarak 

This PR adds two new command-line options:

1. `--restore-session [REPORT]`, which can restore a previous testing session from a report file and
2. `--failed` to select the failed tests.

To achieve the restoration of tests, it introduces a mechanism for serializing and deserializing arbitrary objects in JSON. Any class the derives from `JSONSerializable` can be encoded in JSON. If an attribute cannot be converted, it will be silently set to `null`. The deserialization recreates the objects from the JSON contents and based on some additional metadata that contains the module file and the class name.

Still todo:

- [x] Documentation
- [x] Improve the design of the JSON dumping of `RegressionTest`
- [x] Expand the dumping of `RegressionTest` attributes
- [x] More unit tests
- [x] Address remaining PR comments.


rsarm:
For a quick test:
```
$> ./bin/reframe -r -c ./unittests/resources/checks_unlisted/deps_complex.py --max-retries=3
$> mv ~/.reframe/reports/run-report.json ~/.reframe/reports/run-report-deps-complex.json
$> ./bin/reframe -r -c ./unittests/resources/checks_unlisted/deps_complex.py --max-retries=3 --retry-failed ~/.reframe/reports/run-report-deps-complex.json 
```

ekouts:
I think you should also add the `load` function in `reframe.utility.json` and the option to have a custom `__rfm_json_decode__`. In our case the `json.load` is enough but a user might want to restore more.

rsarm:
@jenkins-cscs retest dom

reframe/frontend/cli.py
```diff
@@ -537,12 +542,53 @@ def main():
...
+                             f"'{restart_report}'") from e
+
+        failed_checks = set()
+        failed_checks_prefixes = set()
+        # for run in restart_report['runs']:
+        for testcase in restart_report['runs'][-1]['testcases']:
+            if testcase['result'] == 'failure':
+                failed_checks.add(hash(testcase['name']) ^
+                                  hash(testcase['system']) ^
+                                  hash(testcase['environment']))
```
vkarak, line 577: How the hash of the `TestCase` is implemented should be an implementation detail, that's why I don't like this. Why don't you simply keep a tuple of the test name, system and environment?

unittests/test_cli.py
```diff
@@ -152,6 +152,36 @@ def test_check_success(run_reframe, tmp_path, logfile):
...
+    assert 'T0' in stdout
+    assert 'T1' in stdout
+    assert 'T2' in stdout
+    assert 'T3' in stdout
+    assert 'T4' in stdout
+    assert 'T5' in stdout
+    assert 'T6' in stdout
+    assert 'T7' in stdout
+    assert 'T8' in stdout
+    assert 'T9' in stdout
```
vkarak, line 168: This first set of assertions is not needed. It is tested elsewhere.

unittests/test_cli.py
```diff
@@ -152,6 +152,36 @@ def test_check_success(run_reframe, tmp_path, logfile):
...
+    assert 'T0' not in stdout
+    assert 'T1' not in stdout
+    assert 'T2' in stdout
+    assert 'T3' not in stdout
+    assert 'T4' not in stdout
+    assert 'T5' not in stdout
+    assert 'T6' not in stdout
+    assert 'T7' in stdout
+    assert 'T8' in stdout
+    assert 'T9' in stdout
```
vkarak, line 182: Why don't you simply read the produced report file and check if the tests have executed as you expect in more detail?

vkarak (changes requested):
We will also need documentation.

reframe/frontend/cli.py
```diff
@@ -537,12 +542,53 @@ def main():
...
+                failed_checks_prefixes.add(testcase['prefix'])
+
+        loader = RegressionCheckLoader(
+            load_path=site_config.get('general/0/check_search_path'),  #failed_checks_prefixes,
```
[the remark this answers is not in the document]
rsarm: I started doing it like that, but later I found that if there's an incorrect test on that prefix, it is going to be loaded when reframe is discovering tests and it might crash. That happens for instance if trying to restart a test from `resources/checks_unlisted`.

reframe/frontend/statistics.py
```diff
@@ -82,6 +82,7 @@ def json(self, force=False):
                     'build_stderr': None,
                     'build_stdout': None,
                     'description': check.descr,
+                    'prefix': check.prefix,
```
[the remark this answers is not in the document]
rsarm: But this is only the directory, not the filename.

reframe/frontend/executors/policies.py
```diff
@@ -241,7 +241,8 @@ def deps_failed(self, task):
         return any(self._task_index[c].failed for c in task.testcase.deps)
 
     def deps_succeeded(self, task):
-        return all(self._task_index[c].succeeded for c in task.testcase.deps)
+        return all((c not in self._task_index or self._task_index[c].succeeded)
+                   for c in task.testcase.deps)
```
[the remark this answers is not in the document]
rsarm: This is because a restored test is not on the task index since it ran on a previous run.

reframe/frontend/cli.py
rsarm: I did it like that so it can be compared with [the hash of a testcase](https://github.com/eth-cscs/reframe/blob/8abfb270dbb58b4e965d2475219d1e69fb4c8ed9/reframe/frontend/executors/__init__.py#L47-L50) which is defined like that.

rsarm:
Yesterday after adding the options `--retry-failed` with no argument, I noticed that when running twice with the default report file, it get's overwritten, and the successful dependencies are lost. Since we are overwriting by default the report, what could be the behavior in this case?

reframe/frontend/statistics.py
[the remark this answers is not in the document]
ekouts: You can try this if you want to get the file from the check:
```
inspect.getfile(type(check))
```
https://github.com/eth-cscs/reframe/blob/master/reframe/frontend/cli.py#L43

unittests/test_cli.py
```diff
@@ -156,30 +157,25 @@ def test_check_retry_failed(run_reframe, tmp_path, logfile):
...
-    assert 'T4' not in stdout
-    assert 'T5' not in stdout
-    assert 'T6' not in stdout
-    assert 'T7' in stdout
-    assert 'T8' in stdout
-    assert 'T9' in stdout
+    with open(f'{tmp_path}/report.json') as fp:
+        report = json.load(fp)
+
+    report_summary = {t['name']: t['fail_phase']
```
vkarak, line 167:
```suggestion
    failing_phase = {t['name']: t['fail_phase']
```

Also the lines below are not correctly aligned, that's why PEP8 complains I think.

unittests/test_cli.py
```diff
@@ -156,30 +157,25 @@ def test_check_retry_failed(run_reframe, tmp_path, logfile):
...
+        for t in run['testcases']
+    }
+
+    assert report_summary['T2'] == 'sanity'
+    assert report_summary['T7'] == 'startup'
+    assert report_summary['T8'] == 'setup'
+    assert report_summary['T9'] == 'startup'
+
+    assert all(i not in report_summary.keys()
+               for i in ['T0', 'T1', 'T3', 'T4', 'T5', 'T6'])
```
vkarak, line 178:
```suggestion
    assert set(report_summary.keys()) - {'T0', 'T1', 'T3', 'T4', 'T5', 'T6'} == set()
```

reframe/utility/json.py
```diff
@@ -34,8 +34,8 @@ def __init__(self, *args, **kwargs):
                                   *args, **kwargs)
 
     def object_hook(self, obj):
-        if 'rfm_properties' in obj:
-            self.rfm_obj.__rfm_json_restore__(obj['rfm_properties'])
+        if 'modules' in obj:
```
vkarak, line 37: You should look at the "type" attribute instead and you should check that the `__rfm_json_restore__` is defined in the target object.

vkarak (changes requested):
It's on the right track, but still needs some refinement and the dumping must go deeper. I will look closer into the test dumps, and you can work on the other comments.

reframe/frontend/cli.py
```diff
@@ -581,10 +653,26 @@ def main():
     printer.debug(format_env(options.env_vars))
 
     # Setup the check loader
+    if options.retry_failed is not None:
+        if options.retry_failed:
+            filename = options.retry_failed
+        else:
+            filename = os_ext.expandvars(
+                site_config.get('general/report_file'))
```
vkarak, line 661: This is wrong, because you don't take into account the case that the `report_file` has the `{sessionid}` pattern in it. You should use the `generate_report_filename()` function somehow.

reframe/frontend/cli.py
```diff
@@ -581,10 +653,26 @@ def main():
     printer.debug(format_env(options.env_vars))
 
     # Setup the check loader
+    if options.retry_failed is not None:
+        if options.retry_failed:
+            filename = options.retry_failed
+        else:
+            filename = os_ext.expandvars(
```
vkarak, line 660: This is not tested either, because unit tests have passed and this is invalid. It should be `osext`.

reframe/frontend/cli.py
```diff
@@ -95,6 +100,68 @@ def fmt_deps():
...
+            failed_checks.add(hash(testcase['name']) ^
+                              hash(testcase['system']) ^
+                              hash(testcase['environment']))
+
+    return failed_checks
+
+
+def get_filenames_from_report(restart_report):
+    return list({testcase['filename']
+                 for testcase in restart_report['runs'][-1]['testcases']})
```
vkarak, line 144: Why don't you filter out here the cases that succeeded?

reframe/frontend/cli.py
```diff
@@ -633,13 +718,27 @@ def print_infoline(param, value):
...
                 type(c).disable_hook(h)
 
-        testcases = generate_testcases(checks_matched,
-                                       options.skip_system_check,
-                                       options.skip_prgenv_check,
-                                       allowed_environs)
-        testgraph = dependency.build_deps(testcases)
-        dependency.validate_deps(testgraph)
-        testcases = dependency.toposort(testgraph)
+        testcases_unfiltered = generate_testcases(checks_matched,
```
vkarak, line 788: Is this in the right place? The cases are already filtered here.

vkarak (approved):

reframe/frontend/cli.py
```diff
@@ -586,10 +584,44 @@ def main():
...
+                new=False
+            )
+
+        report = runreport.load_report(filename)
+        check_search_path = list(report.slice('filename', unique=True))
+        check_search_recursive = False
+
+        # If `-c` or `-R` are passed explicitly outside the configuration
+        # file, override the values set from the report file
+        if site_config.is_sticky_option('general/check_search_path'):
```
teojgo, line 603: Here is would be good to print that you are overriding the search path of the report with the one passed from the configuration. In my case for example I had the `RFM_CHECK_SEARCH_PATH` because I have loaded the `reframe` module so that search path was always overriding the one from the report.

vkarak: I am planning to print something at "info" level, because it's not necessarily a bad thing or undesirable, so as to give a warning. What do you think?

vkarak: I will make it warning, after all, because otherwise it doesn't stand out...

vkarak:
@jenkins-cscs retry dom

unittests/test_cli.py
```diff
@@ -153,7 +209,7 @@ def test_check_success_force_local(run_reframe, tmp_path):
 
 
 def test_report_file_with_sessionid(run_reframe, tmp_path):
-    returncode, stdout, _ = run_reframe(
+    returncode, _, _ = run_reframe(
```
teojgo, line 212:
```suggestion
    returncode, _* = run_reframe(
```

teojgo (changes requested):
lgtm, I have tested and it works fine. I have a slight comment only if you agree.

teojgo (approved):

PR #1540, opened 2020-10-16
[bugfix] Fix bug when patchlevel in environment modules version is not an integer

jonathanfrawley:
A system I am working on uses version numbers such as "3.2.1+26-ad12ae47" for the module environment. This breaks reframe, but since reframe only looks at the major and minor numbers, it is not required to parse out the bugfix version at all.


reframe/core/modules.py
```diff
@@ -433,7 +433,7 @@ def __init__(self):
 
         version = version_match.group(1)
         try:
-            ver_major, ver_minor, *_ = [int(v) for v in version.split('.')]
+            ver_major, ver_minor = [int(v) for v in version.split('.')[:2]]
```
vkarak, line 436: I'rather do:

```suggestion
            ver_major, ver_minor, *_ = [int(v) for v in version.split('.', maxsplit=2)]
```

vkarak (changes requested):
Hi @jonathanfrawley, thanks for the PR and the fix. We will also need a unit test to reproduce the bug. I can provide one.

vkarak:
> Hi @jonathanfrawley, thanks for the PR and the fix. We will also need a unit test to reproduce the bug. I can provide one.

Ignore my comment about the unit tests. I just realised it was in the way we initialize the modules backend, so it should be fine if you just address my other comment.

vkarak (approved):
`maxsplit=2` won't work, so lgtm as it is now.

vkarak:
ok to test

victorusu (approved):
`lgtm`

vkarak:
@jenkins-cscs retry tsa

PR #1545, opened 2020-10-21
[feat] Add support for logging all performance variables to a single line

### LINKED ISSUE #1543: Allow multiple perf_patterns on one line in handlers_perflog
When specifying multiple entries in perf_patterns (for instance as in cscs-checks/microbenchmarks/mpi/halo_exchange/halo_cell_exchange.py), every performance variable is put on a separate line by the logging handler. All attributes that are not related to the performance variable are repeated on all those lines.

For the HaloCellExchangeTest you could argue that every performance represents a separate aspect of the test, so it makes sense to put them on separate lines. However, when you are for instance looking at walltime, CPU time, communication time, memory, ... in one test, it would make more sense to keep those performance variables tied together. At the moment, it is necessary to join lines on some common attribute (such as jobid) to find out which results belong together.

I think it would be a useful feature to allow the `%(check_perf_*)s` attributes to report multiple values, similar to the `%(check_tags)s` attribute. Before attempting to implement this, I would like the opinion of a developer on this. Or maybe there is a hack to work around this issue?

stevenvdb:
This PR introduces log record attributes such as `%(check_perf_all_values)s` that will be filled in with a comma-separated list of values from all entries in `perf_patterns`. When the existing log record attributes such as `%(check_perf_value)s` are omitted, this changes the existing behavior that a separate line is logged for each entry in `per_patterns`.

Closes #1543

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -699,7 +699,13 @@ All logging handlers share the following set of common attributes:
    - ``%(check_perf_upper_thres)s``: The upper threshold of the performance difference from the reference value expressed as a fractional value.
      See the :attr:`reframe.core.pipeline.RegressionTest.reference` attribute of regression tests for more details.
    - ``%(check_perf_value)s``: The performance value obtained for a certain performance variable.
-   - ``%(check_perf_var)s``: The name of the `performance variable <tutorial_basic.html#writing-a-performance-test>`__ being logged.
+   - ``%(check_perf_var)s``: The name of the `performance variable <tutorial_basics.html#writing-a-performance-test>`__ being logged.
+   - ``%(check_perf_all_lower_thress)s``: Comma-separated list of lower thresholds of the performance differences from the reference values expressed as fractional values.
+   - ``%(check_perf_all_refs)s``: Comma-separated list of reference performance values of all performance variables.
+   - ``%(check_perf_all_units)s``: Comma-separated list of units of measurement for the measured performance variables.
+   - ``%(check_perf_all_upper_thress)s``: Comma-separated list of upper thresholds of the performance differences from the reference values expressed as a fractional values.
+   - ``%(check_perf_all_values)s``: Comma-separated list of performance values obtained for all performance variables.
+   - ``%(check_perf_all_vars)s``: Comma-separated list of names of the `performance variables <tutorial_basics.html#writing-a-performance-test>`__ being logged.
    - ``%(osuser)s``: The name of the OS user running ReFrame.
    - ``%(osgroup)s``: The name of the OS group running ReFrame.
    - ``%(version)s``: The ReFrame version.
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -192,6 +192,29 @@ Finally, there is a special set of handlers for handling performance log message
 These are stored in the ``handlers_perflog`` property.
 The performance handler in this example will create a file per test and per system/partition combination and will append the performance data to it every time the test is run.
 Notice in the ``format`` property how the message to be logged is structured such that it can be easily parsed from post processing tools.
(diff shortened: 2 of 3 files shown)


vkarak (changes requested):
Hi @stevenvdb, thanks for your PR. Instead of having the different `check_perf_all_*` format specifiers, I was wondering wether it would make more sense to have a single one, say `check_perf_data` that would actually print all performance variables and values at once as key value pairs. Something like this:

```
perfX=valX[unitX],perfY=valY[unitY],...
```

We could even have `check_perf_data_fmt` (similarly to `datefmt`) that would allow users to explicitly specify the perf data format that they want based on the existing `check_perf_*` values. This will allow to even log `check_perf_data` as JSON:

```python
'check_perf_data_fmt': '{"perfvar": %(check_perf_var)s, "value": %(check_perf_value)s, ...}'
```

vkarak:
@stevenvdb After merging #1600 and #1644, we will no more need the `check_perf_all` variables you introduce here. You will now be able to use `%(check_perf_patterns)s` and this will log all the performance variables and values at once in JSON format. From this PR, we still need the part about creating a single log entry for the performance. If you are fine with I propose here, I can update the PR and schedule it for ReFrame 3.4.

vkarak:
Also if you need a different formatting of the performance variables, you can define an arbitrary field in your test and log that always. In this case, we probably need to find another way to force logging of performance data to happen only once.

stevenvdb:
> 
> 
> @stevenvdb After merging #1600 and #1644, we will no more need the `check_perf_all` variables you introduce here. You will now be able to use `%(check_perf_patterns)s` and this will log all the performance variables and values at once in JSON format. From this PR, we still need the part about creating a single log entry for the performance. If you are fine with I propose here, I can update the PR and schedule it for ReFrame 3.4.

I think your proposition sounds good. What do you think is the best way to trigger the single log entry on or off? You could set it based on the presence of `%(check_perf_patterns)s` or let the user specify it explicitly in the configuration file.

vkarak:
@stevenvdb Sorry for getting back to you late. I was quite busy and then I can't figure out which is the best way to handle this :-) First of all, I think that the logic of whether to iterate or not over the performance variables should be moved inside the `log_performance()` function in `core/logging.py`. The signature of this function would need to be updated, too, so that it simply takes the evaluated `perf_patterns`. It's then the logger that will decide whether to do log each performance variable separately or not. The question is how this behaviour is controlled. Doing this automatically and having the logger inspecting its handlers for this, hmm, it may be convenient for the user, but implementation-wise breaks encapsulation and separation of concerns and I am a bit skeptical about it. Having a configuration variable that controls this, it's much cleaner from the implementation point of view, but that would require the users to remember to set this configuration parameter if they're using `%(check_perf_patterns)s` in their format string, which can be error-prone. If you have any suggestions, I'm more than happy to hear them.

vkarak:
Hi @stevenvdb, #1816 allows you also to log all the references in a single line, so now you can have `%(check_perf_patterns)s` and `%(check_reference)s` to get everything in a single line. You will still have redundant lines, but there is no need to join them anymore. If you don't mind, I will close this PR, but keep #1543 open and schedule it for the next major release. The reason for this is that it's probably better to drop altogether the `log_performance()` function and the redundant lines, but this is a breaking change, so it should be scheduled for 4.0. Meanwhile, I think that you can use the format specifiers I just mentioned to achieve what you want. Let me know if that fits you.

stevenvdb:
> 
> 
> Hi @stevenvdb, #1816 allows you also to log all the references in a single line, so now you can have `%(check_perf_patterns)s` and `%(check_reference)s` to get everything in a single line. You will still have redundant lines, but there is no need to join them anymore. If you don't mind, I will close this PR, but keep #1543 open and schedule it for the next major release. The reason for this is that it's probably better to drop altogether the `log_performance()` function and the redundant lines, but this is a breaking change, so it should be scheduled for 4.0. Meanwhile, I think that you can use the format specifiers I just mentioned to achieve what you want. Let me know if that fits you.

Hi @vkarak,

That sounds look a good solution to me, thanks. I will close this PR.

PR #1547, opened 2020-10-22
[test] Update GridTools tests

ChristopherBignamini:


ChristopherBignamini:
The way GridTools is tested (build and performances) has changed: instead of multiple executables they have now just one executable (perftests) that takes in input some parameters for domain size/number of iterations and the stencil type. Is there a way to avoid the re-building of the executable (perftests) for each test, if it already exists?

victorusu:
@ChristopherBignamini, yes. You can have a base test that compiles the code and multiple ones that depend on the former that use the compiled executable. The relevant documentation is [this](https://reframe-hpc.readthedocs.io/en/stable/dependencies.html).
Currently, you need to have all tests on the same partition, but after this [PR](https://github.com/eth-cscs/reframe/pull/1481) this won't be the case.

If I were you, I would for now implement on test that is based on `rfm.RegressionTest`, which compiles and runs a very simple test, like `perftest -h`, with partition `daint:gpu`, for instance.
All the other ones, then, will be based on `rfm.RunOnlyRegressionTest` with partition `daint:gpu` too. So they can work with the current setup.

ChristopherBignamini:
Thanks @victorusu , I've tried to split the build and the run steps into 2 different tests and I've added the dependency condition but I think I'm doing something wrong: the build tests seems to be ok, they run and pass, but when I try to run one of the run tests (that depends on the build ones) I get the error message 

/bin/reframe: could not resolve dependency: ('GridToolsGPURunCheck_horizontal_diffusion_gpu_horizontal_double', 'daint:gpu', 'builtin') -> 'GridToolsGPUBuildCheck'

Is there anything obvious I'm doing wrong? 

Also, question about design: I'd like to have a base class for the run only test, as I have for the build ones, and then derived classes for the GPU and CPU cases. Does it make sense wrt reframe guidelines? Thanks!

vkarak:
> Is there anything obvious I'm doing wrong?

@ChristopherBignamini Check this out: #1581. The problem is that test filtering happens now before the dependency resolution and it does not take into consideration the test dependencies. You have to select manually with `-n` the rest of the dependency chain...

ChristopherBignamini:
Thank @vkarak , now it seems to work.

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -9,211 +9,189 @@
...
-                        ['simple_hori_diff_naive'],
-                        ['simple_hori_diff_cpu_ifirst'],
-                        ['simple_hori_diff_cpu_kfirst'],
-                        ['vertical_advection_dycore_gpu'],
-                        ['simple_hori_diff_gpu'])
-class GridToolsCheck(rfm.RegressionTest):
-    def __init__(self, variant):
+class GridToolsBuildCheck(rfm.CompileOnlyRegressionTest):
+    def __init__(self):
         # Check if this is a device check
```
vkarak, line 14: I think this comment is stale.

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -9,211 +9,189 @@
...
+            GridToolsCPUBuildCheck().stagedir,
+            'tests', 'regression', 'perftests')
+
+
+@rfm.parameterized_test(['horizontal_diffusion/gpu_double'],
+                        ['horizontal_diffusion/gpu_horizontal_double'])
+class GridToolsGPURunCheck(GridToolsRunCheck):
+    def __init__(self, variant):
+        super().__init__()
+        # Check if this is a device check
```
vkarak, line 152: I think this comment is stale.

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -9,211 +9,189 @@
...
+
+        self.variant_data = {
+            'horizontal_diffusion/gpu_double': {
+                'executable_opts': [
+                    '256', '256', '80', '1',
+                    '--gtest_filter=horizontal_diffusion/gpu_double*'],
                 'reference': {
                     'daint:gpu': {
-                        'wall_time': (8690, None, 0.1, 'ms')
+                        'timer': (0.0006, None, 0.1, 's')
```
vkarak, line 166: How reliable are these timings, because they are rather small. Can you increase the test size to get higher numbers?

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -9,211 +9,189 @@
...
-                        'wall_time': (2880, None, 0.1, 'ms')
+                        'timer': (9.0, None, 0.1, 's')
                     },
                     'daint:gpu': {
-                        'wall_time': (3020, None, 0.1, 'ms')
+                        'timer': (1.0, None, 0.1, 's')
                     },
                     'dom:mc': {
-                        'wall_time': (2880, None, 0.1, 'ms')
+                        'timer': (9.0, None, 0.1, 's')
```
vkarak, line 126: These numbers seem too round to be real, are they?

vkarak (changes requested):

cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -3,217 +3,194 @@
...
                     'dom:gpu': {
-                        'wall_time': (8690, None, 0.1, 'ms')
+                        'timer': (0.004, None, 0.1, 's')
                     }
                 }
             },
-            'simple_hori_diff_gpu': {
-                'executable_opts': ['150', '150', '150'],
+            'horizontal_diffusion/gpu_horizontal_double': {
+                'executable_opts': [
```
vkarak, line 173: See my comment above about this.

vkarak (changes requested):

vkarak:
@jenkins-cscs retry daint

vkarak (approved):

PR #1548, opened 2020-10-22
[bugfix] Fix `IndexError` crash of `osext.copytree()` when the source directory does not exist 

### LINKED ISSUE #1504: Index out of range error when 'sourcesdir' does not exist
While working on writing an [example helloworld.py test using dependencies](https://gist.github.com/victorusu/78bb0f3b175be274a9d41d67a007bc54), I got the following [error](https://gist.github.com/victorusu/78bb0f3b175be274a9d41d67a007bc54#gistcomment-3473961).

For some reason that I do not understand (yet) only the `pgi` [tests have failed](https://gist.github.com/victorusu/78bb0f3b175be274a9d41d67a007bc54#gistcomment-3473963). 

But even though the failure is strange, I believe that ReFrame should not print the [stack trace in this way](https://gist.github.com/victorusu/78bb0f3b175be274a9d41d67a007bc54#gistcomment-3473961). It should be a bit more descriptive.

vkarak:
This PR also makes `osext.copytree()` errors compliant with the standard `copytree()`.

Fixes #1504.

### DIFF
--- a/reframe/utility/osext.py
+++ b/reframe/utility/osext.py
@@ -156,7 +156,20 @@ def copytree(src, dst, symlinks=False, ignore=None, copy_function=shutil.copy2,
         return shutil.copytree(src, dst, symlinks, ignore, copy_function,
                                ignore_dangling_symlinks)
 
-    # dst exists; manually descend into the subdirectories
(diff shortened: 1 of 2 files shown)


teojgo (approved):
lgtm

PR #1549, opened 2020-10-22
[feat] Color deprecation warnings

### LINKED ISSUE #1267: Allow coloring of deprecation warnings
We should color the output of the `user_deprecation_warning` function, if the user hasn't passed the `--nocolor` option. See also #1263 .

vkarak:
Also:

- Move all the warning related parts into a separate file.
- In order to achieve the coloration of the full message, we had to replace the warning formatting method. Also the message is adjusted.

Fixes #1267.

### DIFF
--- a/reframe/core/warnings.py
+++ b/reframe/core/warnings.py
@@ -17,7 +17,7 @@ def _format_warning(message, category, filename, lineno, line=None):
     import reframe.utility.color as color
 
     if category != ReframeDeprecationWarning:
-        return _format_warning_orig
+        return _format_warning_orig(message, category, filename, lineno, line)
 
     if line is None:
         # Read in the line from the file
--- a/reframe/core/warnings.py
+++ b/reframe/core/warnings.py
@@ -1,0 +1,58 @@
+import inspect
+import warnings
+
+
+class ReframeDeprecationWarning(DeprecationWarning):
+    '''Warning raised for deprecated features of the framework.'''
+
+
+warnings.filterwarnings('default', category=ReframeDeprecationWarning)
+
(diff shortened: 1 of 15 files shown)


vkarak:
I need to add a unit test to check the formatting of the deprecation warning: https://codecov.io/gh/eth-cscs/reframe/pull/1549/diff?src=pr&el=tree#diff-cmVmcmFtZS9jb3JlL3dhcm5pbmdzLnB5

reframe/core/warnings.py
```diff
@@ -0,0 +1,58 @@
...
+    if line is None:
+        # Read in the line from the file
+        with open(filename) as fp:
+            try:
+                line = fp.readlines()[lineno-1]
+            except IndexError:
+                line = '<no line information>'
+
+    message = f'{filename}:{lineno}: WARNING: {message}\n{line}\n'
+    if rt.runtime().get_option('general/0/colorize'):
```
teojgo, line 31: This needs a check whether a runtime is initialized. For example when trying to build the docs now, it fails because no runtime is unitialized.

teojgo (changes requested):

reframe/core/warnings.py
vkarak: Hmm, interesting. I'll have a look!

vkarak:
@jenkins-cscs retry tsa

ekouts (approved):
lgtm

teojgo (approved):
Works fine now.

vkarak:
@jenkins-cscs retry all

PR #1552, opened 2020-10-24
[feat] Control verbosity of check failure info

### LINKED ISSUE #491: Add a command-line option to always print full stack traces in failure report
This is quite useful for debugging.

vkarak:
To achieve this, the way the failure report is generated had to be redesigned:

- The former `failure_report()` and `failure_stats()` are now printing directly through a printer object, thus their output is controlled with the universal verbosity levels of the framework.
- The exception information is kept in detail in the produced JSON report.
- The former `format_exception()` function has been simplified to return only the reason of the failure (no tracebacks).
- The traceback are always stored in the JSON report and are printed either at the INFO or the VERBOSE level depending on the severity of the failure, which can be queried through the `is_severe()` method.

Fixes #491.

### DIFF
--- a/reframe/core/decorators.py
+++ b/reframe/core/decorators.py
@@ -48,7 +48,7 @@ def _register_test(cls, args=None):
             try:
                 ret.append(_instantiate(cls, args))
             except Exception:
-                frame = user_frame(sys.exc_info()[2])
+                frame = user_frame(*sys.exc_info())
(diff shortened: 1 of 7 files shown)


teojgo (approved):
The conflicts due to https://github.com/eth-cscs/reframe/pull/1549 need to be fixed.

teojgo (approved):

vkarak:
@jenkins-cscs retry tsa

PR #1556, opened 2020-10-29
[bugfix] Fix spacing issues when defining options for execution modes

### LINKED ISSUE #1554: Inconsistency in mode option passing
If one runs reframe using the `-J` option inside the 

```python
    'modes': [
        {
          ...
                '-J reservation=maintenance',
          ...
            ]
        },
```

One gets the following job script

```bash
#!/bin/bash                                                                        
...                                                             
#SBATCH -- reservation=maintenance                                                 
...
```

However, if one submits thru the command line interface, then one gets the correct job script.

```bash
reframe  -n GromacsGPUCheck_small_maint -r -J reservation=maintenance
```

```
#!/bin/bash                                                                        
...
#SBATCH --reservation= maintenance                                                          
...
```

rsarm:
Fixes #1554


reframe/frontend/cli.py
```diff
@@ -475,6 +477,8 @@ def main():
             mode_args = site_config.get(f'modes/@{options.mode}/options')
 
             # Parse the mode's options and reparse the command-line
+            mode_args = list(itertools.chain(*[shlex.split(m)
+                                               for m in mode_args]))
```
ekouts, line 481:
```suggestion
            mode_args = list(itertools.chain(*(shlex.split(m)
                                               for m in mode_args)))
```
I am not sure whether it makes a different, @vkarak is this better than the temporary list?

ekouts:
```suggestion
            mode_args = list(itertools.chain.from_iterable(shlex.split(m) for m in mode_args))
```
Or this? 🤔

vkarak: The second one is the most pythonic.

reframe/frontend/cli.py
```diff
@@ -475,6 +477,8 @@ def main():
             mode_args = site_config.get(f'modes/@{options.mode}/options')
 
             # Parse the mode's options and reparse the command-line
+            mode_args = list(itertools.chain(*[shlex.split(m)
```
vkarak, line 480: We will need a small comment here as well as of why we have to lexically split each option.

vkarak (changes requested):
Now that I see the fix, I don't think that this was actually a bug, but rather the configuration was wrong. Nonetheless, this will prevent common user errors in the future. Apart from the rest of the comments, we will need to adjust the unit test configuration to exercise this. Simply change the following two lines:

https://github.com/eth-cscs/reframe/blob/master/unittests/resources/settings.py#L134-L135

to

```python
                '-c unittests/resources/checks/hellocheck.py',
                '-p builtin-gcc',
```

ekouts (approved):

vkarak (approved):
lgtm now

PR #1570, opened 2020-11-03
[feat] Add possibility to inherit the prefix from the parent class

### LINKED ISSUE #1569: Allow derived tests to inherit the prefix of their parent tests
The idea is discussed in #1555 and it should be easy to implement. We want that a test inherits the prefix of its parent instead of setting its own. This will allow a library base test to carry the required test resources and let the derived tests specialize only on system-specific stuff.

jjotero:
Closes #1569


vkarak (changes requested):
We will need a unit test for this functionality.

jjotero:
> We will need a unit test for this functionality.

Done.

victorusu (approved):
`lgtm`

reframe/core/pipeline.py
```diff
@@ -734,10 +738,16 @@ def __init__(self):
         pass
 
     @classmethod
-    def __init_subclass__(cls, *, special=False, **kwargs):
+    def __init_subclass__(cls, *, special=False, base_test=False, **kwargs):
```
vkarak, line 741: I don't like very much `base_test` as a name. It can be confusing when writing a test. I  kind of prefer something like `template`, although it is not ideal, too.

reframe/core/pipeline.py
```diff
@@ -734,10 +738,16 @@ def __init__(self):
...
         super().__init_subclass__(**kwargs)
         cls._rfm_special_test = special
 
+        # Insert the base test path as the prefix if the class is marked
+        # as a base test.
+        if base_test:
+            cls._rfm_base_test_prefix = os.path.abspath(
+                os.path.dirname(inspect.getfile(cls)))
```
vkarak, line 749: Here you need also the `is_interactive()` check, so I suggest moving the setting of the prefix into a separate function.

reframe/core/pipeline.py
```diff
@@ -722,10 +722,14 @@ def __new__(cls, *args, **kwargs):
...
-            else:
-                prefix = os.path.abspath(os.path.dirname(inspect.getfile(cls)))
+            try:
+                prefix = cls._rfm_base_test_prefix
+            except AttributeError:
+                if osext.is_interactive():
+                    prefix = os.getcwd()
+                else:
+                    prefix = os.path.abspath(
+                        os.path.dirname(inspect.getfile(cls)))
```
vkarak, line 732: Instead of all these nested `try/except`, I would suggest this, which is more compact:

```python
prefix = None
with contextlib.suppress(AttributeError):
    for p in ['_rfm_base_test_prefix', '_rfm_custom_prefix']:
        prefix = getattr(cls, p)

if prefix is None:
    prefix = cls._set_prefix()
```

unittests/resources/checks/hellocheck_base.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2016-2020 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+class HelloBaseTest(rfm.RunOnlyRegressionTest, base_test=True):
```
vkarak, line 10: Since we only need to check that the prefix is correctly set, I would go for a minimal test:

```python
class TemplateTest(rfm.RegressionTest, template=True):
    pass
```

vkarak: And I would put that in a separate file under `unittests/resources/checks/`

unittests/test_pipeline.py
```diff
@@ -246,6 +247,13 @@ def __init__(self):
     _run(MyTest(), *local_exec_ctx)
 
 
+def test_base_test(local_exec_ctx):
+    class MyTest(HelloBaseTest):
+        pass
+
+    _run(MyTest(), *local_exec_ctx)
```
vkarak, line 254: You don't need to run the test, just check that its prefix is correct.

vkarak (changes requested):

reframe/core/pipeline.py
jjotero: I've changed (for now 😂) the `base_test` argument to `pin_prefix`.

reframe/core/pipeline.py
jjotero: What about just adding this attribute in any case, and then using the `is_interactive` check from the comment below?

reframe/core/pipeline.py
jjotero: I feel that this changes the behaviour of `_rfm_custom_prefix`. Rewriting the try/excepts as above would mean that `_rfm_custom_prefix` is no longer a 'master override' kind of thing and it would have the same functionality as `_rfm_base_test_prefix`. I've changed this a bit to use the `is_interactive` check also for the `_rfm_base_test_prefix`.

vkarak: I don't understand why what I propose here does not give the "master override" power to `_rfm_custom_prefix`. It will also be the last prefix to try, so it will override `_rfm_pinned_prefix`.

vkarak: Hmm, you mean in respect with the `is_interactive()`?

vkarak: You are right. I am resolving this. Also, the `try/except` is also more pythonic (that's the reason I had it like that initially), although I don't like it very much, but on the other hand if you go with `hasattr()`, then you type twice the attributes. So I'm leaving it like now.

vkarak (approved):
Lgtm. I'm going to push a couple of minor changes.

Issue #795, opened 2019-05-21
More informative messages when checks are undetected


shoshijak:
I'm trying out reframe by running a simple hello world, like this: 

```
@rfm.simple_test
class HelloWorldTest(rfm.RegressionTest):
    def __init__(self):
        super().__init__()
        self.descr = 'C++ Hello World Test'
        self.name="hello-world"
        self.valid_systems = ['daint:mc']
        self.valig_prog_environs = ['*']
        self.sourcesdir = os.path.join(self.prefix, '..')
        self.build_system = 'SingleSource'
        self.sanity_patterns = sn.assert_found(r'Hello world!', self.stdout)
        self.tags = {'hello-world'}
```

When I run reframe however, no checks are detected: 
```
Command line: /apps/common/UES/jenkins/SLES12/easybuild/software/reframe/2.17/bin/reframe -C /users/alicej/reframe/settings.py -c reframe/dummy.py -l
Reframe version: 2.17
Launched by user: alicej
Launched on host: daint103
Reframe paths
=============
    Check prefix      :
    Check search path : 'reframe/dummy.py'
    Stage dir prefix     : /users/alicej/dbcsr/stage/
    Output dir prefix    : /users/alicej/dbcsr/output/
    Perf. logging prefix : /users/alicej/dbcsr/perflogs
List of matched checks
======================
Found 0 check(s).
``` 

It took me a while to figure out that this happened because I had simply misspelled "valid": 
```
                 |
                 V
        self.valig_prog_environs = ['*']
```

Since no information was printed on the reason why no checks were found, finding the spelling error was not straightforward. 

Just as a suggestion for user-friendliness, would it be possible, for example:
- to add a warning or message if non-supported attributes are added to the class (e.g. in this case `valig_prog_environs`), 
- or to add a warning message if a test was instantiated, but no valid programming environments were found/specified for it, as is also the case here?

vkarak:
Yeah, we've gone through it before and it's a bit annoying. Good that you opened an issue for that! Regarding your suggestions:

> to add a warning or message if non-supported attributes are added to the class (e.g. in this case valig_prog_environs),

This is not possible, cos there's no such thing as "non-supported" attributes. We want users to define their own attributes.

> or to add a warning message if a test was instantiated, but no valid programming environments were found/specified for it, as is also the case here?

[...]

PR #1571, opened 2020-11-03
[feat] Better logging of framework activities, especially test loading

### LINKED ISSUE #1046: Improve debug information that is being logged
Currently, it is not very helpful in spotting problems.

vkarak:
- The initialization of the framework and the loading of checks is much better logged now. The test loading and filtering is logged at the verbose level.
- A new debug level, DEBUG2, was introduced. All messages from the execution policies and the actual OS commands that are being executed are logged in this level.
- The default level for the logger objects is now set to "undefined." This allows for the seamless addition of new debug levels in the future.
- The stdout and stderr markers in errors for spawned processes have changed.

Fixes #1046.

It also addresses #795, since by increasing verbosity you can now see what files are being loaded and how many tests.


reframe/frontend/loader.py
```diff
@@ -96,10 +103,14 @@ def load_from_module(self, module):
             )
 
         if not hasattr(module, '_rfm_gettests'):
+            getlogger().debug('no tests registered')
```
ekouts, line 106: Usually in `getlogger().debug` you start with a capital letter the message. I don't have a preference, I am just commenting in case this was by accident
```suggestion
            getlogger().debug('No tests registered')
```
I am not sure at what point this message is printed but should we add the module in the message? Unless something else was printed above it and the message is not alone

ekouts (approved):
Lgtm

vkarak:
@jenkins-cscs retry daint

PR #1572, opened 2020-11-04
[feat] Add support to generate dynamic Gitlab pipelines

### LINKED ISSUE #1523: Create dynamic Gitlab pipelines from ReFrame tests
The idea is very similar to `spack ci generate`. ReFrame will generate a Gitlab pipeline dynamically that would run in each stage different set of tests as pipeline jobs. For the moment, we don't treat dependencies a proper test restore mechanism is need, which will be addressed in #1222. Here is the basic idea:

```yaml
stages: [generate, build]

generate-pipeline: 
  stage: generate
  tags: 
    - <custom-tag>
  script: 
    - reframe -c /path/to/repo/checks --ci-generate="${CI_PROJECT_DIR}/jobs_scratch_dir/pipeline.yml" --ci-tags=unittest,verification,production
  artifacts: 
    paths: 
      - "${CI_PROJECT_DIR}/jobs_scratch_dir/pipeline.yml"

test-jobs: 
  stage: build
  trigger: 
    include: 
      - artifact: "jobs_scratch_dir/pipeline.yml"
        job: generate-pipeline
    strategy: depend
```

The `pipeline.yaml` would then define a Gitlab pipeline with different stages, one per CI tag (in this example, `unittest`, `verification` and `production`) and each pipeline job will execute a single ReFrame test for that set of tags: `reframe -c /path/to/repo/checks -t unittest -n test_0`, `reframe -c /path/to/repo/checks -t unittest -n test_1` etc.

victorusu:
Fixes #1523.

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -111,12 +111,7 @@ def generate_report_filename(filepatt):
     return filepatt.format(sessionid=new_id)
 
 
-def locate_and_load_tests(options, loader, ci_tag=[]):
-    # Locate and load checks
-    try:
-        checks_found = loader.load_all()
-    except OSError as e:
-        raise errors.ReframeError from e
+def filter_checks(checks_found, options, rt, loader, ci_tag=None):
 
     # Filter checks by name
     checks_matched = checks_found
@@ -134,8 +129,8 @@ def locate_and_load_tests(options, loader, ci_tag=[]):
         checks_matched = filter(filters.have_tag(tag), checks_matched)
 
     # Filter checks by ci tags
-    for tag in ci_tag:
-        checks_matched = filter(filters.have_tag(tag), checks_matched)
+    if ci_tag:
+        checks_matched = filter(filters.have_tag(ci_tag), checks_matched)
 
     # Filter checks by prgenv
     if not options.skip_prgenv_check:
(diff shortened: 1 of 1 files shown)


reframe/frontend/cli.py
```diff
@@ -111,6 +111,124 @@ def generate_report_filename(filepatt):
...
+
+# TODO place this function in a proper module
+# TODO define mechanism on how to propagate command line options, currently
+#      we are passing site_config, but that's clearly a bad design
+#      read below the options we need to propagate or make available to each
+#      ci invokation
+#      An alternative design would be to have environment variables exported
+#      during the trigger phase of reframe that would influence the tests
+#      individually
+def generate_ci_pipeline_file(ci_pipeline_file, ci_tags, checks_found, options, rt, loader, site_config):
```
vkarak, line 170: All the CI integration functionality must go into a separate module.

vkarak: And the interface seems a bit hacky, too.

vkarak: The reason is because this function mixes everything: selection of tests and the generation of the pipeline. These must be separated and, especially, the selection of the tests must be viewed in the light of the work done in #1594.

reframe/frontend/cli.py
```diff
@@ -111,6 +111,124 @@ def generate_report_filename(filepatt):
...
+        # + --skip-system-check
+        # + --skip-prgenv-check
+        # + --mode
+        # + --max-retries
+        # + -M
+        # + -m
+        # + --module-mappings
+        # + --failure-stats
+        # + --performance-report
+        for test in checks_matched:
```
vkarak, line 213: Everything must happen at the test case level. Now the test cases are generated immediately and everything in the frontend operates on them.

reframe/frontend/cli.py
```diff
@@ -111,6 +111,124 @@ def generate_report_filename(filepatt):
...
+    for tag in ci_tags:
+        pipeline = {}
+        checks_matched = filter_checks(checks_found, options, rt, loader, ci_tag=tag)
+
+        # TODO there are some options that need  to be revisited
+        # - about the test folder: reframe's -c option
+        # - about the configuration file to use: reframe's -C option
+        # - about the artifacts artifacts: {paths: [jobs_scratch_dir], when: always}
+
+        # TODO there are some missing options:
```
vkarak, line 191: It's not clear to me how you want to treat the different flags. Can you explain?

vkarak (changes requested):
I see two parts of this PR: (a) selecting the tests for which to generate a pipeline and (b) the generation of the pipeline. Currently these parts are mixed up and they should be separated. Part (a) must be reconsidered completely, since it will be obsolete as soon as #1594 is merged. The second one should be independent: generate a pipeline file given a set of test cases. This would be also testable, whereas it's not as it is written now.

Finally, I don't understand how we plan the `--ci-tags` and `--ci-generate` to interoperate with the rest of the test selection options.

victorusu:
I am closing this PR, because another one (#1641) replaces it.

PR #1576, opened 2020-11-04
[feat] Print more information about dependencies in detailed test listing

### LINKED ISSUE #1521: More detailed information about dependencies
In one of my tests, reframe reports the `-L` reports
```
- BTest:
    Dependencies:
      ATest
...
    Valid environments:
      builtin, PrgEnv-gnu

    Valid systems:
      daint:gpu, dom:gpu, daint:mc, dom:mc
```

But my `BTest` only depends on `ATest` when
* The `ATest` has partition `login` and programming environment `PrgEnv-gnu`; and
* The `BTest` has partition `gpu` and programming environment `builtin` or `PrgEnv-gnu`

I have the following seam in my `BTest` class
```
        self.valid_systems = ['daint:gpu','dom:gpu', 'daint:mc','dom:mc']
        self.valid_prog_environs = ['builtin', 'PrgEnv-gnu', 'PrgEnv-intel']
...
        subdeps = {('gpu', 'builtin'): [('login', environment)]}
        subdeps[('gpu', 'PrgEnv-gnu')] = [('login', environment)]
        self.depends_on('ATest', how=rfm.DEPEND_EXACT, subdeps=subdeps)
...
```
> Note: My test is based on PR #1481.

 I wanted to know in which conditions `BTest` depends on the `ATest` test. I would love to see something like the following (or better)
```
- BTest:
    Dependencies:
      ATest('login', 'PrgEnv-gnu') when ('gpu', 'builtin')
      ATest('login', 'PrgEnv-gnu') when ('gpu', 'PrgEnv-gnu')
...
    Valid environments:
      builtin, PrgEnv-gnu, PrgEnv-intel

    Valid systems:
      daint:gpu, dom:gpu, daint:mc, dom:mc
```

ekouts:
Fixes #1521.

Some output examples:

```
- OSUAllreduceTest_4:
    Description:
      OSU Allreduce test

    Environment modules:
      <none>

    Location:
      /users/karakasv/Devel/reframe/tutorials/deps/osu_benchmarks.py

    Maintainers:
      <none>

    Node allocation:
      standard (4 task(s))

    Pipeline hooks:
      - post_setup: set_executable

    Tags:
      <none>

    Valid environments:
      gnu, pgi, intel

    Valid systems:
      daint:gpu

    Dependencies (conceptual):
      OSUBuildTest

    Dependencies (actual):
      - ('OSUAllreduceTest_4', 'daint:gpu', 'gnu') -> ('OSUBuildTest', 'daint:login', 'gnu')
      - ('OSUAllreduceTest_4', 'daint:gpu', 'intel') -> ('OSUBuildTest', 'daint:login', 'intel')
      - ('OSUAllreduceTest_4', 'daint:gpu', 'pgi') -> ('OSUBuildTest', 'daint:login', 'pgi')
```

```
- OSUDownloadTest:
    Description:
      OSU benchmarks download sources

    Environment modules:
      <none>

    Location:
      /Users/karakasv/Repositories/reframe/tutorials/deps/osu_benchmarks.py

    Maintainers:
      <none>

    Node allocation:
      standard (1 task(s))

    Pipeline hooks:

    Tags:
      <none>

    Valid environments:
      gnu

    Valid systems:
      daint:login

    Dependencies (conceptual):
      <none>

    Dependencies (actual):
      <none>
```

I kept the old `Dependencies` because the check itself might have dependencies to more checks than will appear on this specific system.


ekouts:
@jenkins-cscs retry all

vkarak:
What I can't understand from the output samples you posted is why the "Dependencies for current system" are empty for the second set of examples, when you have for example test dependencies.

teojgo:
> What I can't understand from the output samples you posted is why the "Dependencies for current system" are empty for the second set of examples, when you have for example test dependencies.

Executing this:
`./bin/reframe -C unittests/resources/settings.py  --system=sys0 -c unittests/resources/checks_unlisted/deps_simple.py -L`

looks good for me.

ekouts:
> What I can't understand from the output samples you posted is why the "Dependencies for current system" are empty for the second set of examples, when you have for example test dependencies.

@vkarak The environments `e0` and `e1` are not in the configuration, so no test is going to run in the end

teojgo (approved):

vkarak:
> @vkarak The environments e0 and e1 are not in the configuration, so no test is going to run in the end

This is a bit problematic and comes from the fact that the printing is based on tests not on the test cases. I was about to comment something related to that for your PR. Perhaps we can address also this problem. If the tests are not going to run, they should not be listed.

ekouts:
> This is a bit problematic and comes from the fact that the printing is based on tests not on the test cases. I was about to comment something related to that for your PR. Perhaps we can address also this problem. If the tests are not going to run, they should not be listed.

Okay, I can easily print only the checks that are included in the `index_of_deps`. Or do you prefer a different way to solve this issue?

vkarak:
I don't like the `index_of_deps` concept being computed outside and then just being passed along all the way down to `format_checks()`. The interface doesn't make sense. The `list_checks()` should take the test cases instead and derive the actual tests and dependencies from there. You build the dependency index privately inside that function or wherever it is actually needed.

ekouts:
Okay, I think I can pass only the testcases and infer the checks from there. This way we will only have the checks that will run

reframe/frontend/cli.py
```diff
@@ -77,6 +77,20 @@ def fmt_list(x):
 
         lines.append('')
 
+    lines.append('    Dependencies for current system:')
```
vkarak, line 80: Couldn't you add this as an element in `check_info`? And similarly to `fmt_list()`, have something like `fmt_deps()`.

vkarak (changes requested):

vkarak (approved):
I did some small changes. Lgtm now.

PR #1578, opened 2020-11-04
[ci] Disable parallel unit testing on tsa

teojgo:
Also:

* Remove kesch specific `module load`

EDIT: This is due to the random hangs of the unit tests on Tsa.


vkarak (approved):
lgtm

PR #1579, opened 2020-11-05
[test] OSU Benchmarks for Tsa extension

kraushm:


vkarak:
This PR starts by rewriting the OSU tests for Tsa, but the plan is to expand these changes to the other systems and remove the old tests.

vkarak:
These are now outdated.

PR #1583, opened 2020-11-06
[bugfix] Fix unit test error due the PID 1 problem when running inside a Docker container

vkarak:
The `assert_process_died()` function was failing for the local scheduler when the unit tests were run inside a container. The reason for that is that the Python process was run with pid=1, thus inheriting the children that were killed in some of the unit tests. Since these children were never waited for, the remained zombies, thus having still a valid PID.


victorusu (approved):
`lgtm`

victorusu:
Funny thing though. The fact that the unit tests fail but the framework doesn't! 😄

PR #1589, opened 2020-11-09
[ci] Skip unit tests requiring git if git is not available

### LINKED ISSUE #1582: Skip unit tests using `git` when `git` is not available

teojgo:
Fixes #1582

### DIFF
--- a/ci-scripts/dockerfiles/Lmod.dockerfile
+++ b/ci-scripts/dockerfiles/Lmod.dockerfile
@@ -7,7 +7,9 @@ FROM ubuntu:20.04
 ENV TZ=Europe/Zurich
 ENV DEBIAN_FRONTEND=noninteractive
 ENV _LMOD_VER=8.4.12
-WORKDIR /root
+
+# ReFrame user
+RUN useradd -ms /bin/bash rfmuser
 
 # ReFrame requirements
 RUN \
@@ -28,12 +30,15 @@ RUN \
   cd Lmod-${_LMOD_VER} && \
   ./configure && make install
 
-
 ENV BASH_ENV=/usr/local/lmod/lmod/init/profile
 
-# Install ReFrame from the current directory
-COPY . /root/reframe/
-RUN cd reframe && ./bootstrap.sh
+USER rfmuser
 
-WORKDIR /root/reframe
+# Install ReFrame from the current directory
+COPY --chown=rfmuser . /home/rfmuser/reframe/
+
+WORKDIR /home/rfmuser/reframe
+
+RUN ./bootstrap.sh
+
 CMD ["/bin/bash", "-c", "./test_reframe.py --rfm-user-config=ci-scripts/configs/lmod.py -v"]
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -243,8 +243,6 @@ def git_only():
         pytest.skip('no git installation found on system')
 
 
-
-
 def test_git_repo_hash(monkeypatch, git_only):
     try:
         completed = osext.run_command('git status', check=True, log=False)
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -236,18 +236,20 @@ def test_is_url():
 
 
 @pytest.fixture
-def no_git_installed():
+def git_only():
     try:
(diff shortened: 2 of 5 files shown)


ekouts (approved):

unittests/test_utility.py
```diff
@@ -235,7 +235,20 @@ def test_is_url():
     assert not osext.is_url(repo_ssh)
 
 
-def test_git_repo_hash(monkeypatch):
+@pytest.fixture
+def no_git_installed():
```
vkarak, line 239: I would inverse the check and call the fixture `git_only`.

unittests/test_utility.py
```diff
@@ -235,7 +235,20 @@ def test_is_url():
...
+    try:
+        completed = osext.run_command(f'git --version', check=True, log=False)
```
vkarak, line 241:
```suggestion
        completed = osext.run_command('git --version', check=True, log=False)
```

unittests/test_utility.py
```diff
@@ -235,7 +235,20 @@ def test_is_url():
...
+    except (SpawnedProcessError, FileNotFoundError):
+        return True
+
+    return False
+
+
+def test_git_repo_hash(monkeypatch, no_git_installed):
+    if no_git_installed:
+        pytest.skip('no git installation found on system')
```
vkarak, line 250: The skipping can happen in the fixture.

vkarak (changes requested):

unittests/test_utility.py
vkarak: If you skip the test inside the fixture, you don't need to return anything.

vkarak:
Can you also update the Dockerfiles for the modules systems?

ci-scripts/dockerfiles/Lmod.dockerfile
```diff
@@ -7,7 +7,9 @@ FROM ubuntu:20.04
 ENV TZ=Europe/Zurich
 ENV DEBIAN_FRONTEND=noninteractive
 ENV _LMOD_VER=8.4.12
-WORKDIR /root
+
+# ReFrame user
```
vkarak, line 11: Can you expand the comment there why we need a special user? If it's a standard practice, there is no need.

unittests/test_utility.py
```diff
@@ -236,18 +236,18 @@ def test_is_url():
...
 
 
-def test_git_repo_hash(monkeypatch, no_git_installed):
-    if no_git_installed:
-        pytest.skip('no git installation found on system')
+def test_git_repo_hash(monkeypatch, git_only):
+    try:
+        completed = osext.run_command('git status', check=True, log=False)
+    except (SpawnedProcessError, FileNotFoundError):
```
vkarak, line 249: Why don't you simply check if `.git` exists?

vkarak (changes requested):

ci-scripts/dockerfiles/Lmod.dockerfile
teojgo: It's just better to have a normal user without root privileges to be as close as possible to the typical use case.

unittests/test_utility.py
teojgo: I think it's better to rely on git to check if it's a valid repo.

vkarak: My only problem is whether this is worth it. This four lines of code and much more verbose instead of just two.

teojgo: `.git` can exist but in case it's not a proper repo you cannot get the hash.

vkarak: Ok, I see the scenario that sb downloads the source (without cloning), has git and runs the unit tests, so this check has a meaning. I'm not gonna argue further about the way for writing it.

unittests/test_utility.py
```diff
@@ -247,7 +260,7 @@ def test_git_repo_hash(monkeypatch):
     assert osext.git_repo_hash() is None
```
vkarak, line 260: I think here we should add a case that monkeypatches the `GIT_DIR`, so as to check the case that the `rev-parse` returns an error.

vkarak (changes requested):

vkarak (approved):

PR #1590, opened 2020-11-10
[test] Update memory bandwidth checks references

jjotero:
This PR extends the support of the memory bandwidth tests for all the current partitions in `ault` with NVIDIA cards.
The most relevant change to these tests is that now the `nvidia_sm` parameter is set in a hook due to the different device architectures present in this system.


vkarak (approved):
lgtm

PR #1591, opened 2020-11-10
[feat] Add support for building tests remotely

### LINKED ISSUE #1175: Add support for submitting build step as a job rather than build locally
This feature is would help us a lot because we have a heterogenous cluster, so the microarchitecture of the login node is not necessarily the same as that of the compute nodes.

ekouts:
Support remote building for tests by adding the attribute `build_locally`. By default it is `True` but it can be set to `False` and the job will be built in the partition of the test.

Fixes #1175.

In this PR the exitcode of `squeue` is also set `None` instead of `0`.


reframe/core/pipeline.py
```diff
@@ -708,6 +708,14 @@ def disable_hook(cls, hook_name):
     extra_resources = fields.TypedField('extra_resources',
                                         typ.Dict[str, typ.Dict[str, object]])
 
+    #: .. versionadded:: 3.3
```
vkarak, line 711: This goes at the end.

reframe/core/pipeline.py
```diff
@@ -777,6 +785,9 @@ def _rfm_init(self, name=None, prefix=None):
         # True only if check is to be run locally
         self.local = False
 
+        # True only if check is to be built locally
```
vkarak, line 788: I don't think that this comment is needed. For `local` it has a bit more meaning, but most probably it's there before even documenting the fields.

reframe/core/pipeline.py
```diff
@@ -1037,29 +1048,29 @@ def _setup_paths(self):
         except OSError as e:
             raise PipelineError('failed to set up paths') from e
 
-    def _setup_job(self, **job_opts):
+    def _setup_job(self, local, name, **job_opts):
```
vkarak, line 1051:
```suggestion
    def _setup_job(self, name, force_local=False, **job_opts):
```

reframe/core/pipeline.py
```diff
@@ -1037,29 +1048,29 @@ def _setup_paths(self):
...
+        if local:
             scheduler = getscheduler('local')()
             launcher = getlauncher('local')()
         else:
             scheduler = self._current_partition.scheduler
             launcher = self._current_partition.launcher_type()
 
         self.logger.debug(
-            f'Setting up run job descriptor '
+            f'Setting up {name} job descriptor '
```
vkarak, line 1062:
```suggestion
            f'Setting up job {name!r} '
```

reframe/core/pipeline.py
```diff
@@ -1750,6 +1762,20 @@ class RunOnlyRegressionTest(RegressionTest, special=True):
     module.
     '''
 
+    @_run_hooks()
+    def setup(self, partition, environ, **job_opts):
+        '''The setup stage of the regression test pipeline.
+
+        Similar to the :func:`RegressionTest.setup`, except that no job
+        descriptor is set up for the build of this test.
```
vkarak, line 1770:
```suggestion
        Similar to the :func:`RegressionTest.setup`, except that no build job is created for this test.
```

vkarak: Can you update in a similar way also the documentation of the compile only test?

vkarak (changes requested):
As discussed, we will also need a unit test.

reframe/core/pipeline.py
ekouts: Are you sure? Because in the filelds `max_pending_time`, `postrun_cmds`, `prerun_cmds` and all the others that were added at some point, it's the first thing. Should I change this everywhere?

vkarak: 🤔  I thought it was the other way around. Let me check.

unittests/test_pipeline.py
```diff
@@ -170,6 +171,21 @@ def test_hellocheck_local(hellotest, local_exec_ctx):
...
+        def __init__(self):
+            super().__init__()
+            self.build_locally = False
+
+    test = MyTest()
+    partition = fixtures.partition_by_name('gpu')
+    environ = partition.environment('builtin-gcc')
+    test.setup(partition, environ)
+
+    assert isinstance(test._build_job, _SlurmJob)
```
vkarak, line 186: I think it's better if you run the test:

```suggestion
def test_hellocheck_remote_building(hellotest, remote_exec_ctx):
    hellotest.build_locally = False
    _run(hellotest, *remote_exec_ctx)    
```

vkarak (changes requested):

reframe/core/pipeline.py
vkarak: You are right. Nonetheless we will need to expand the documentation of this field. For example, what does it mean when this is set to `False`

vkarak (approved):

vkarak:
@jenkins-cscs retry all

vkarak:
The reason the unit tests are failing for the PBS/Torque backends is here:

https://github.com/eth-cscs/reframe/blob/7666a2becc1df28719ab6c3786c72043c0d72fb2/reframe/core/pipeline.py#L1227-L1229

The `FIXME` says it all :-) PBS/Torque backends do not return the exit code (I think it's `None`), that's why the tests is failing. We should think how this can be fixed.

teojgo (approved):

vkarak (changes requested):
We still need to figure out how we deal with schedulers that don't give us the exit code.

ekouts:
@jenkins-cscs retry all

reframe/core/schedulers/slurm.py
```diff
@@ -562,9 +562,6 @@ def poll(self, *jobs):
                 job_match = jobinfo[job.jobid]
             except KeyError:
                 job._state = 'CANCELLED' if job.is_cancelling else 'COMPLETED'
-                if job.exitcode is None:
```
ekouts, line 565: @vkarak I removed the `exitcode` from the squeue scheduler since we cannot get it.

vkarak (approved):

PR #1598, opened 2020-11-13
[test] Add test for Cray module prefix environment variables

ekouts:
This is a test to ensure that some cray modules include the variables `<module_name>_PREFIX` and `<module_name>_VERSION`. It will be more useful as soon as we get a new cdt so that we catch the missing variables early. Moreover, it would be nicer to skip the tests that we know are already failing, instead of making them succeed with the workaround.


lucamar (approved):
Thanks @ekouts, that's really useful!

cscs-checks/prgenv/environ_check.py
```diff
@@ -41,3 +42,44 @@ def __init__(self):
...
+
+@rfm.parameterized_test(['cray-fftw'], ['cray-hdf5'], ['cray-hdf5-parallel'],
+                        ['cray-libsci'], ['cray-netcdf'],
+                        ['cray-netcdf-hdf5parallel'], ['cray-petsc'],
+                        ['cray-petsc-64'], ['cray-petsc-complex'],
+                        ['cray-petsc-complex-64'], ['cray-python'],
+                        ['cray-R'], ['cray-tpsl'], ['cray-tpsl-64'],
+                        ['cudatoolkit'], ['gcc'], ['papi'], ['pmi'])
+class CrayVariablesCheck(rfm.RunOnlyRegressionTest):
+    def __init__(self, variant):
```
vkarak, line 55:
```suggestion
    def __init__(self, module_name):
```

cscs-checks/prgenv/environ_check.py
```diff
@@ -41,3 +42,44 @@ def __init__(self):
...
+        self.variant = variant
+        self.descr = 'Check for standard Cray variables'
+        self.valid_systems = ['daint:login', 'dom:login']
+        self.valid_prog_environs = ['builtin']
+
+        self.executable = 'module'
+        self.executable_opts = ['show', variant]
+
+        mod_name = variant.upper().replace('-', '_')
```
vkarak, line 64:
```suggestion
        envvar_prefix = variant.upper().replace('-', '_')
```

cscs-checks/prgenv/environ_check.py
```diff
@@ -41,3 +42,44 @@ def __init__(self):
...
+    def cdt_workaround(self):
+        cdt = osext.cray_cdt_version()
+        if not cdt:
+            return
+
+        # these modules should be fixed in later releases
+        if (cdt <= '20.10' and
+            self.variant in ['cray-petsc-complex',
+                             'cray-petsc-complex-64',
+                             'cudatoolkit', 'gcc']):
```
vkarak, line 84: You could explicitly empty the `valid_systems` for these modules and for CDT <= 20.10 directly in the `__init__()`. This way these tests will not even be run if the cdt is not the right one.

vkarak (changes requested):

vkarak (approved):
lgtm

PR #1599, opened 2020-11-13
[feat] Validate tests after loading them

### LINKED ISSUE #55: Deep copying regression tests imposes unnecessary restrictions
We should rethink the fact that we deep copy the whole regression test instances for each test case (environment, system, partition). It is definitely necessary for isolating them and unties our hands in the asynchronous execution policy, but it imposes some restrictions on how we should write regression tests. Anything that cannot be deep-copied (e.g., I/O streams, generators etc.) it cannot be used in a regression test. For example, the following although it's pretty much good valid Python, we do not support it:

```python
        self.sanity_patterns = sn.all(sn.assert_found(r'sth', fname)
                                      for fname in sn.iglob('file*.txt'))
```

The problem is that the argument to `sn.all()` is a generator and cannot be deep-copied. Instead, we should write this using a list comprehension:

```python
        self.sanity_patterns = sn.all([sn.assert_found(r'sth', fname)
                                       for fname in sn.iglob('file*.txt')])
```

Internal issue: https://madra.cscs.ch/scs/reframe/issues/524

### LINKED ISSUE #701: Is a validation stage needed, to make sure a test contains a set of required parameters?
During the creation of the Fieldextra performance test, Reframe submitted all 3 tests (2x accuracy + 1x perf) through SLURM, although the class FieldextraPerformance(FieldextraCheck) (and the base class FieldextraCheck) did not contain any sanity patterns. (see output below)

```
Command line: ./reframe.py -C config/cscs.py -c cscs-checks/mch/fieldextra_check.py --run --keep-stage-files --prefix=/scratch/kraushm/.
Reframe version: 2.17-dev0
Launched by user: kraushm
Launched on host: keschln-0003
Reframe paths
=============
    Check prefix      :
    Check search path : 'cscs-checks/mch/fieldextra_check.py'
    Stage dir prefix     : /scratch/kraushm/stage/
    Output dir prefix    : /scratch/kraushm/output/
    Perf. logging prefix : /scratch/kraushm/perflogs
[==========] Running 3 check(s)
[==========] Started on Fri Feb 22 10:43:33 2019[----------] started processing FieldextraAccuracy_opt (Fieldextra validation test (accuracy); MCH)
[ RUN      ] FieldextraAccuracy_opt on kesch:cn using PrgEnv-gnu-nompi
[       OK ] FieldextraAccuracy_opt on kesch:cn using PrgEnv-gnu-nompi
[----------] finished processing FieldextraAccuracy_opt (Fieldextra validation test (accuracy); MCH)[----------] started processing FieldextraAccuracy_opt_omp (Fieldextra validation test (accuracy); MCH)
[ RUN      ] FieldextraAccuracy_opt_omp on kesch:cn using PrgEnv-gnu-nompi
[       OK ] FieldextraAccuracy_opt_omp on kesch:cn using PrgEnv-gnu-nompi
[----------] finished processing FieldextraAccuracy_opt_omp (Fieldextra validation test (accuracy); MCH)[----------] started processing FieldextraPerformance (Fieldextra validation test (performance); MCH)
[ RUN      ] FieldextraPerformance on kesch:cn using PrgEnv-gnu-nompi
[     FAIL ] FieldextraPerformance on kesch:cn using PrgEnv-gnu-nompi
[----------] finished processing FieldextraPerformance (Fieldextra validation test (performance); MCH)[  FAILED  ] Ran 3 test case(s) from 3 check(s) (1 failure(s))
[==========] Finished on Fri Feb 22 10:53:11 2019==============================================================================
SUMMARY OF FAILURES
FAILURE INFO for FieldextraPerformance
  * System partition: kesch:cn
  * Environment: PrgEnv-gnu-nompi
  * Stage directory: /scratch/kraushm/stage/kesch/cn/PrgEnv-gnu-nompi/FieldextraPerformance
  * Job type: batch job (id=1009420)
  * Maintainers: ['Mkr']
  * Failing phase: sanity
  * Reason: sanity error: sanity_patterns not set
```

The error message is correct, but the test wasted compute resources, since it does not check anything. Should there be an additional stage in the pipeline, which would filter out "incomplete" tests, e. g. by checking for certain (required) keywords/parameters?

### LINKED ISSUE #795: More informative messages when checks are undetected

vkarak:
This PR adds a post-init validation step to the check loader. A test must fulfil the following requirements in order to be loaded:

1. Define `valid_systems`.
2. Define `valid_prog_environs`.
3. Be (deep) copyable.

Implementing (3) without copying the object was quite tricky. I had to implement a validator that recursively descends into the object and checks that every attribute fulfils a condition. The validator examines also any built-in container types element-by-element. The second tricky part was to implement the `is_copyable()` condition, i.e., understanding that an object can be copied without doing the actual copy. For this, I had to sneak in the `copy.deepcopy()`'s implementation, because, essentially, we want to return `True` iff the object can be deep-copied.

The validation step does not check whether `sanity_patterns` is set, because this can be set in a pipeline hook.

Fixes #701. 
Fixes #795.
Fixes #55.


reframe/utility/__init__.py
```diff
@@ -270,6 +271,199 @@ def repr(obj, htchar=' ', lfchar='\n', indent=4, basic_offset=0):
...
+    return (isinstance(cls, types.BuiltinFunctionType) or
+            isinstance(cls, types.FunctionType))
+
+
+def attr_validator(validate_fn):
+    '''Validate object attributes recursively.
+
+    This returns a function which you can call with the object to check. It
+    will return :class:`True` if the :func:`validate_fn` returns :class:`True`
+    for all object attributes recursively. If the object to be validate is an
```
ekouts, line 297:
```suggestion
    for all object attributes recursively. If the object to be validated is an
```

ekouts (approved):

PR #1600, opened 2020-11-16
[feat] Allow any test attribute to be logged

### LINKED ISSUE #1068: Allow any test field to be logged as well as allow tests to log arbitrary data
This is a custom change introduced to @ZQyou's fork.

### LINKED ISSUE #1432: Allow non-numbers as values to performance variables
Users would like to use a performance variable to simply include an additional information to the performance log, so the tend to define the following pattern and reference:

```python
self.perf_patterns = {
    'non-numeric-var': sn.extractsingle(...)
}
self.reference = {
    '*': {
        'non-numeric-var': (None, None, None, 'n/a')
    }
}
```

We could simply avoid the `assert_reference()`check if the reference value is `None` or we could even go further by skipping the test if the extracted value is not a number. This would not require to define a special reference for such variables at all.

This request has emerged in #1146 and #1430.

rsarm:
EDIT: @vkarak 

This PR adds the ability to log any test attribute. If the attribute cannot be found, `<undefined>` will be logged instead. If an attribute is deferrable it will be evaluated and if this is not possible `<error>` will be logged.

Fixes #1068.
Fixes #1432.


reframe/core/logging.py
```diff
@@ -405,12 +403,25 @@ def __init__(self, logger=None, check=None):
...
                 'check_tags': None,
                 'version': osext.reframe_version(),
             }
-        )
+
+        self.user_log_vars = []
+        if logger:
+            log_vars = []
+            for h in logger.handlers:
+                fmt = re.findall(r'\%\((check_\S+?)\)s', h.formatter._fmt)
```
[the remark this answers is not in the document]
vkarak: But I think the design should be different. The fact that the logger has to deal with the formatter is a bit hacky and the `LoggerAdapter` is created at least every time a check-specific log call is done, so this is not so efficient. We should find a proper design.

reframe/core/pipeline.py
```diff
@@ -180,6 +180,11 @@ def disable_hook(cls, hook_name):
     #: :default: ``[]``
     valid_systems = fields.TypedField('valid_systems', typ.List[str])
 
+    #: Logging of arbitrary information in per test basis
+    #: :type: :class:`typ.Dict[str, object]`
+    #: :default: ``{}``
+    log_extra = fields.TypedField('log_extra', typ.Dict[str, object])
```
vkarak, line 186: As discussed, this should be a string.

vkarak (changes requested):

reframe/core/logging.py
```diff
@@ -165,7 +165,40 @@ def _format_time_rfc3339(timestamp, datefmt):
...
 
-class RFC3339Formatter(logging.Formatter):
+class DynamicFieldFormatter(logging.Formatter):
+    def __init__(self, fmt=None, datefmt=None, style='%'):
+        super().__init__(fmt, datefmt, style)
+        self.__extras = {}
+        fmt = re.findall(r'\%\((check_\S+?)\)s', fmt)
+        for v in set(fmt):
+            if v not in self.__extras:
+                self.__extras[v] = None
```
ekouts, line 175:
```suggestion
            self.__extras.setdefault(v, None)
```
I think this will also work.

reframe/core/logging.py
```diff
@@ -165,7 +165,39 @@ def _format_time_rfc3339(timestamp, datefmt):
     return re.sub(r'(%)?\:z', tz_rfc3339, time.strftime(datefmt, timestamp))
 
 
-class RFC3339Formatter(logging.Formatter):
+class DynamicFieldFormatter(logging.Formatter):
+    def __init__(self, fmt=None, datefmt=None, style='%'):
+        super().__init__(fmt, datefmt, style)
+        self.__extras = {}
+        log_vars = re.findall(r'\%\((check_\S+?)\)s', fmt)
+        for v in set(log_vars):
```
vkarak, line 173:
```suggestion
        fmt_specs = re.findall(r'\%\((check_\S+?)\)s', fmt)
        for spec in set(fmt_specs):
```

reframe/core/logging.py
```diff
@@ -165,7 +165,39 @@ def _format_time_rfc3339(timestamp, datefmt):
...
+        if isinstance(val, list) or isinstance(val, tuple):
+            return ','.join(val)
+        else:
+            return val
+
+    def format(self, record):
+        if record.check:
+            for spec in self.__extras.keys():
+                _, *attr = spec.split('_')
+                if spec in record.__dict__.keys():
```
vkarak, line 186:
```suggestion
                if hasattr(record, spec):
```

reframe/core/logging.py
```diff
@@ -385,7 +417,9 @@ def debug2(self, message, *args, **kwargs):
 
 class LoggerAdapter(logging.LoggerAdapter):
     def __init__(self, logger=None, check=None):
-        extra_vars = {
+        super().__init__(
+            logger,
+            {
                 'check_name': 'reframe',
```
vkarak: You still need to remove from here all the extras that correspond to an actual test attribute.

reframe/core/logging.py
```diff
@@ -165,7 +165,39 @@ def _format_time_rfc3339(timestamp, datefmt):
...
+            return val
+
+    def format(self, record):
+        if record.check:
+            for spec in self.__extras.keys():
+                _, *attr = spec.split('_')
+                if spec in record.__dict__.keys():
+                    continue
+
+                record.__dict__[spec] = self.__extras[spec]
```
vkarak, line 189: The record should have the `extras` even if `record.check` is `None`, otherwise formatting will crash when something is logged outside the context of a check.

reframe/core/logging.py
```diff
@@ -165,7 +165,39 @@ def _format_time_rfc3339(timestamp, datefmt):
...
+                if attr[0] == 'job' and record.check.job:
+                    self.__extras[spec] = self._fmt_check_attr(
+                        getattr(record.check.job, '_'.join(attr[1:])))
```
vkarak, line 192: We won't support `check_job_xxx` to be translated to `check.job.xxx`. For these cases, we keep the extras as we have them.

reframe/core/logging.py
```diff
@@ -408,20 +442,8 @@ def __init__(self, logger=None, check=None):
...
-                log_vars.extend(fmt)
-
-            for v in set(log_vars):
-                if v not in extra_vars:
-                    extra_vars[v] = None
-                    self.user_log_vars.append(v)
-
-        super().__init__(logger, extra_vars)
+        )
+        self.extra['check'] = check
```
vkarak, line 446: You can pass this when creating the logger above. No need to do it here.

reframe/core/pipeline.py
```diff
@@ -181,9 +181,9 @@ def disable_hook(cls, hook_name):
     valid_systems = fields.TypedField('valid_systems', typ.List[str])
 
     #: Logging of arbitrary information in per test basis
```
vkarak, line 183: Leave a new line after here.

vkarak (changes requested):
The documentation must also be updated.

reframe/core/logging.py
vkarak: These two are not equivalent. The second one will only set the field if it's not present. I don't think that this can ever be the case at this point. You just fill in the `__extras` that you have just created. So @rsarm's original solution is correct.

reframe/core/pipeline.py
vkarak: I removed completely the `log_extra` attribute. It's not needed. If the `foo` attribute in `%(check_foo)s` cannot be found, it will be simply logged as `None`.

vkarak (approved):

vkarak:
I have finalized this PR. @rsarm @ekouts Have a look in case something is missing.

ekouts (approved):
lgtm

rsarm:
lgtm

PR #1603, opened 2020-11-16
[bugfix] Fix regression when filtering with multiple criteria

vkarak:
This was introduced with #1594 and it is a regression. I've added also a unit test to catch this in the future.


unittests/test_cli.py
```diff
@@ -481,6 +481,20 @@ def test_list_with_details(run_reframe):
...
+def test_filtering_multiple_criteria(run_reframe):
+    returncode, stdout, stderr = run_reframe(
+        checkpath=['unittests/resources/checks'],
+        action='list',
+        more_options=['-t', 'foo', '-n', 'hellocheck',
+                      '--ignore-check-conflicts']
+    )
+    assert 'Traceback' not in stdout
+    assert 'Traceback' not in stderr
+    print(stdout)
```
ekouts, line 493: you should remove this `print`

ekouts (approved):

teojgo (changes requested):
Remove the print that @ekouts mentioned.

teojgo (approved):

PR #1605, opened 2020-11-17
[doc] Update ReFrame's requirements in the documentation

### LINKED ISSUE #1585: Update requirements for running ReFrame
We need at least these for the unit tests:

```Dockerfile
RUN \
  apt-get -y update && \
  apt-get -y install gcc && \
  apt-get -y install make && \
  apt-get -y install python3

# Erroneously required by ReFrame unit tests; see GH #1582
RUN apt-get -y install git
```

`gcc` per se is not a requirement, but any C compiler available through `cc` (we need to verify that for the unit tests though).

And we should check what are the minimal requirements for running the framework, but not the unit test.

See also #1582.

teojgo:
Fixes #1585

### DIFF
--- a/docs/started.rst
+++ b/docs/started.rst
@@ -15,8 +15,7 @@ Requirements
 Optional Requirements
 ---------------------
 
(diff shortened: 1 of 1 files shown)


vkarak (approved):
I just rephrased it a bit. Lgtm.

PR #1606, opened 2020-11-17
[feat] Add command-line option to manipulate the module path before executing any tests

### LINKED ISSUE #1395: Add a command line option to use or unuse module paths
This is useful for quick tests sometimes that you don't want to manually do `module [un]use` and the absolute module paths do not always work correctly.

teojgo:
* Add the corresponding documentation.

* Add unittests.

Fixes #1395


docs/manpage.rst
```diff
@@ -387,6 +387,26 @@ It does so by leveraging the selected system's environment modules system.
    This option can also be set using the :envvar:`RFM_UNLOAD_MODULES` environment variable or the :js:attr:`unload_modules` general configuration parameter.
 
 
+.. option:: --unuse-module-path=PATH
```
vkarak, line 390: Why not use a single option and use `+/-` to control addition and deletion of paths?

- `--module-path=+/path/to/add`
- `--module-path=-/path/to/remove`

vkarak: If somebody passes it without `+/-` we could completely override the module path.

reframe/frontend/cli.py
```diff
@@ -733,6 +746,28 @@ def print_infoline(param, value):
...
+        printer.debug('Unusing module paths from command line')
+        for d in site_config.get('general/0/unuse_module_paths'):
+            try:
+                rt.modules_system.searchpath_remove(d)
+            except errors.EnvironError as e:
+                printer.warning(
+                    f'could not unuse module path {d["name"]!r} correctly; '
+                    f'skipping...'
+                )
+                printer.debug(str(e))
```
vkarak, line 758: I would rather print that in `verbose()`.

unittests/test_cli.py
```diff
@@ -591,6 +591,42 @@ def test_unload_module(run_reframe, user_exec_ctx):
     assert returncode == 0
 
 
+def test_unuse_module_path(run_reframe, user_exec_ctx, monkeypatch):
+    ms = rt.runtime().modules_system
+    if ms.name == 'nomod':
+        pytest.skip('no modules system found')
+
+    # We need to have MODULEPATH set to unuse a module
```
vkarak, line 599: This comment is a bit confusing. What do you mean?

unittests/test_cli.py
```diff
@@ -591,6 +591,42 @@ def test_unload_module(run_reframe, user_exec_ctx):
...
+    monkeypatch.setenv('MODULEPATH', '')
+    module_path = os.path.abspath('unittests/modules')
```
vkarak, line 601: Making this absolute doesn't offer something.

reframe/frontend/cli.py
```diff
@@ -733,6 +746,28 @@ def print_infoline(param, value):
             printer.debug(str(e))
             raise
 
+        printer.debug('Unusing module paths from command line')
+        for d in site_config.get('general/0/unuse_module_paths'):
+            try:
+                rt.modules_system.searchpath_remove(d)
+            except errors.EnvironError as e:
```
vkarak, line 753: I don't think that this is needed. Adding/removing paths does not throw any errors. If paths do not exist, it's harmless.

unittests/test_cli.py
```diff
@@ -591,6 +591,42 @@ def test_unload_module(run_reframe, user_exec_ctx):
...
+    returncode, stdout, stderr = run_reframe(
+        config_file=fixtures.USER_CONFIG_FILE,
+        more_options=[f'--unuse-module-path={module_path}'], action='run'
+    )
+
+    assert stdout != ''
```
vkarak, line 607: Remove this, it's not so useful.

unittests/test_cli.py
```diff
@@ -591,6 +591,42 @@ def test_unload_module(run_reframe, user_exec_ctx):
...
+
+def test_use_module_path(run_reframe, user_exec_ctx):
+    ms = rt.runtime().modules_system
+    if ms.name == 'nomod':
+        pytest.skip('no modules system found')
+
+    module_path = os.path.abspath('unittests/modules')
+    returncode, stdout, stderr = run_reframe(
+        config_file=fixtures.USER_CONFIG_FILE,
+        more_options=[f'--use-module-path={module_path}'], action='run'
```
vkarak, line 621: You should also try to load a module as well, otherwise you don't know that the module was loaded. You need to pass `-m testmod_foo`.

vkarak: And then you should check that there is not warning in the output (check the code the warning printeed when the requested module could not be loaded).

unittests/test_cli.py
vkarak: I think in order to test the unuse, you should add a path to the `MODULEPATH`, then invoke reframe with the option to unuse it and then check that the expected warning was printed in the output.

vkarak (changes requested):

docs/manpage.rst
teojgo: I think that it's better to have only the `-` convension for unusing otherwise the module path should be used.

reframe/frontend/cli.py
teojgo: If a module path is not defined yet removing a path returns an error.

docs/manpage.rst
```diff
@@ -387,12 +387,13 @@ It does so by leveraging the selected system's environment modules system.
    This option can also be set using the :envvar:`RFM_UNLOAD_MODULES` environment variable or the :js:attr:`unload_modules` general configuration parameter.
 
 
-.. option:: --unuse-module-path=PATH
+.. option:: --module-path=PATH
 
-   Unuse module path ``PATH`` before acting on any tests.
-   This option may be specified multiple times, in which case all specified modules paths will be unused in order.
+   Use module path ``PATH`` before acting on any tests.
```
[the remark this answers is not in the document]
vkarak: My problem with the `--module-path=/path/to/modules` is that it is a bit confusing. One could think that this would set the whole `MODULEPATH`. That's why I suggest to have the following three options:

1. `+/path/to/modules` -> `module use`
2. `-/path/to/modules` -> `module unuse`
3. `/path/to/modules` -> override completely `MODULEPATH`.

Since, as I suggest in the following, you need not provide a corresponding configuration parameter and/or environment variable, it should be quite straightforward to implement this logic.

vkarak (changes requested):

vkarak:
Also I would expect the actions using the different modules systems to fail 🤔

docs/manpage.rst
```diff
@@ -389,21 +389,10 @@ It does so by leveraging the selected system's environment modules system.
...
-   .. versionadded:: 3.3
-
-
-.. option:: --use-module-path=PATH
-
-   Use module path ``PATH`` before acting on any tests.
-   This option may be specified multiple times, in which case all specified module paths will be used in order.
-
-   This option can also be set using the :envvar:`RFM_USE_MODULE_PATHS` environment variable or the :js:attr:`use_module_paths` general configuration parameter.
+   Use ``PATH`` to manipulate the ``MODULEPATH`` before acting on any tests.
```
vkarak, line 392:
```suggestion
   Manipulate the ``MODULEPATH`` environment variable before acting on any tests.
```

docs/manpage.rst
```diff
@@ -389,21 +389,10 @@ It does so by leveraging the selected system's environment modules system.
...
+   Paths starting with the `-` character are going to be removed from the ``MODULEPATH``, while the ones starting with the `+` character are going to be added to it.
```
vkarak, line 393:
```suggestion
   If ``PATH`` starts with the `-` character, it will be removed from the ``MODULEPATH``, whereas if it starts with the `+` character, it will be added to it.
```

reframe/frontend/cli.py
```diff
@@ -751,7 +749,19 @@ def print_infoline(param, value):
                         f'skipping...'
                     )
                     printer.verbose(str(e))
+            elif d.startswith('+'):
+                try:
+                    rt.modules_system.searchpath_add(d[1:])
+                except errors.EnvironError as e:
+                    printer.warning(
+                        f'could not use module path {d} correctly; '
```
vkarak, line 757:
```suggestion
                        f'could not add module path {d} correctly; '
```

reframe/frontend/cli.py
```diff
@@ -751,7 +749,19 @@ def print_infoline(param, value):
...
+                        f'skipping...'
+                    )
+                    printer.verbose(str(e))
             else:
+                rt.modules_system.searchpath_remove(
+                    *rt.modules_system.searchpath
+                )
```
vkarak, line 764: I don't understand this very well. Wouldn't you simply set the `MODULEPATH` here?

vkarak (changes requested):

ekouts (approved):
lgtm

reframe/frontend/cli.py
teojgo: I wanted to manipulate the module path without setting `MODULEPATH` directly but through the modules api.

vkarak: As discussed offline, this is fine.

docs/manpage.rst
```diff
@@ -389,8 +389,8 @@ It does so by leveraging the selected system's environment modules system.
 
 .. option:: --module-path=PATH
 
-   Use ``PATH`` to manipulate the ``MODULEPATH`` before acting on any tests.
-   Paths starting with the `-` character are going to be removed from the ``MODULEPATH``, while the ones starting with the `+` character are going to be added to it.
+   Manipulate the ``MODULEPATH`` environment variable before acting on any tests.
+   If ``PATH`` starts with the `-` character it will be removed from the ``MODULEPATH``, whereas if it starts with the `+` character, it will be added to it.
```
vkarak, line 393:
```suggestion
   If ``PATH`` starts with the `-` character, it will be removed from the ``MODULEPATH``, whereas if it starts with the `+` character, it will be added to it.
```

unittests/test_cli.py
```diff
@@ -627,6 +627,23 @@ def test_use_module_path(run_reframe, user_exec_ctx):
...
+
+    module_path = 'unittests/modules'
+    returncode, stdout, stderr = run_reframe(
+        more_options=[f'--module-path={module_path}', '-m testmod_foo'],
+        config_file=fixtures.USER_CONFIG_FILE, action='run',
+        system=rt.runtime().system.name
+    )
+
+    assert 'Traceback' not in stdout
+    assert 'Traceback' not in stderr
```
vkarak, line 643: I think you should also add an assertion here that you didn't get the warning that `testmod_foo` could not be loaded. See here:

https://github.com/eth-cscs/reframe/blob/4aa8ac9029e9d55bc4e2283c6c16fcb4ba6c54dd/reframe/frontend/cli.py#L741-L744

unittests/test_cli.py
```diff
@@ -596,32 +596,30 @@ def test_unuse_module_path(run_reframe, user_exec_ctx, monkeypatch):
...
     )
 
-    assert stdout != ''
-    assert 'Traceback' not in stdout
-    assert 'Traceback' not in stderr
-    assert returncode == 0
+    # Here we test that ReFrame fails to run because 'testmod_foo' cannot
+    # be found and therefore the module name is included in the given error
+    assert 'testmod_foo' in stdout
+    assert returncode == 1
```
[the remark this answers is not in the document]
vkarak: I don't think it's correct to exit with 1 in such a case.

unittests/test_cli.py
```diff
@@ -591,6 +591,59 @@ def test_unload_module(run_reframe, user_exec_ctx):
...
+    monkeypatch.setenv('MODULEPATH', module_path)
+    returncode, stdout, stderr = run_reframe(
+        more_options=[f'--module-path=-{module_path}', '-m testmod_foo'],
+        config_file=fixtures.USER_CONFIG_FILE, action='run',
+        system=rt.runtime().system.name
+    )
+
+    # Here we test that ReFrame fails to run because 'testmod_foo' cannot
+    # be found and therefore the module name is included in the given error
+    assert 'testmod_foo' in stdout
```
vkarak, line 609: Be a bit more precise here regarding the message you are looking for.

vkarak (changes requested):
I have minor comments except that I don't understand why it is correct to exit with 1 in the `test_unuse_module_path()` test.

unittests/test_cli.py
```diff
@@ -599,15 +599,13 @@ def test_unuse_module_path(run_reframe, user_exec_ctx, monkeypatch):
     module_path = 'unittests/modules'
     monkeypatch.setenv('MODULEPATH', module_path)
     returncode, stdout, stderr = run_reframe(
-        more_options=[f'--module-path=-{module_path}', '-m testmod_foo'],
+        more_options=[f'--module-path=-{module_path}', '--module=testmod_foo'],
```
vkarak, line 602: Oh yes! You would have to pass them as separate arguments otherwise.

vkarak (approved):

PR #1608, opened 2020-11-18
[bugfix] Properly define default values for the 'modules' and 'variables' configuration parameters of the 'container_platforms'

### LINKED ISSUE #1607: ReFrame crashes if `container_platforms` is defined without modules
Using the following configuration:

```python
        {
            'name': 'tresa',
            'hostnames': ['.*'],
            'modules_system': 'lmod',
            'partitions': [
                {
                    'name': 'default',
                    'scheduler': 'local',
                    'launcher': 'local',
                    'environs': ['builtin'],
                    'container_platforms': [{'type': 'Docker'}],
                    'max_jobs': 8
                }
            ]
        },
```

ReFrame crashes as follows:

```
  File "./bin/reframe", line 21, in <module>
    cli.main()
  File "/Users/karakasv/Repositories/reframe/reframe/frontend/cli.py", line 536, in main
    runtime.init_runtime(site_config)
  File "/Users/karakasv/Repositories/reframe/reframe/core/runtime.py", line 183, in init_runtime
    _runtime_context = RuntimeContext(site_config)
  File "/Users/karakasv/Repositories/reframe/reframe/core/runtime.py", line 33, in __init__
    self._system = System.create(site_config)
  File "/Users/karakasv/Repositories/reframe/reframe/core/systems.py", line 282, in create
    f'{partid}/container_platforms/{i}/modules'
  File "/Users/karakasv/Repositories/reframe/reframe/core/config.py", line 57, in _get
    ret = norm_fn(ret)
  File "/Users/karakasv/Repositories/reframe/reframe/core/environments.py", line 20, in normalize_module_list
    for m in modules:
TypeError: 'NoneType' object is not iterable
```
If we add `'modules': []`, the test works.

This is a regression introduced by #1586.

teojgo:
Fixes #1607


reframe/core/environments.py
```diff
@@ -16,6 +16,9 @@ def normalize_module_list(modules):
 
     :meta private:
     '''
+    if modules is None:
+        return None
```
vkarak, line 20: I'm not sure whether returning `None` is correct here or `[]`. What I cannot understand how it came and this was called with `None`. It is called here:

```
  File "/Users/karakasv/Repositories/reframe/reframe/core/systems.py", line 282, in create
```

And there, the `site_config.get()` should return the default value, which it should be `[]`.

vkarak: Indeed querying the variable from the command line, it fails instead of returning the default.

```
./bin/reframe -C config/tresa.py --show-config=systems/0/partitions/@default/container_platforms/0/modules
```
```
./bin/reframe: no such configuration parameter found: systems/0/partitions/@default/container_platforms/0/modules
```

vkarak (changes requested):
I think the problem is deeper. Check my comment.

vkarak:
The problem existed long ago, but it wasn't appearing... Since the `container_platform` uses the `type` attribute, this is treated specially when `get()` constructs the default key: it uses `systems/partitions/container_platforms/Docker_modules` instead of `systems/partitions/container_platforms/modules`, which is what we set as default key. The fix is simple, but in a completely different place :-)

```diff
diff --git a/reframe/schemas/config.json b/reframe/schemas/config.json
index 6fea2162..99fabceb 100644
--- a/reframe/schemas/config.json
+++ b/reframe/schemas/config.json
@@ -460,8 +460,8 @@
         "systems/partitions/access": [],
         "systems/partitions/environs": [],
         "systems/partitions/container_platforms": [],
-        "systems/partitions/container_platforms/modules": [],
-        "systems/partitions/container_platforms/variables": [],
+        "systems/partitions/container_platforms/*modules": [],
+        "systems/partitions/container_platforms/*variables": [],
         "systems/partitions/resources": [],
         "systems/partitions/resources/options": [],
         "systems/partitions/modules": [],
```

vkarak (approved):

PR #1611, opened 2020-11-19
[bugfix] Treat conflicts in module mappings properly

### LINKED ISSUE #1274: Module conflicts are ignored in module mappings
In the following example `PrgEnv-gnu` loads `gcc/9.2.0`. Using the following mapping `-M 'PrgEnv-gnu: PrgEnv-gnu gcc/8.3.0'` it will cause the error when loading `gcc/8.3.0`:

```
gcc/8.3.0(3):ERROR:150: Module 'gcc/8.3.0' conflicts with the currently loaded module(s) 'gcc/9.2.0'
gcc/8.3.0(3):ERROR:102: Tcl command execution failed: conflict gcc
```

The problem is that the mapping module mechanism is not taking into account module conflicts in mappings. In this case the rhs of the mapping will `PrgEnv-gnu` with `gcc/9.2.0` and then it will try to load `gcc/8.3.0`, which will cause the error.

I don't know if this is easily fixable, but I am creating this issue to be listed as known issue at least. If we decide not to fix it, we should document it.

vkarak:
When modules are mapped with the `-M` option and there are conflicts in the replacement sequence of modules, emit the the module load and unload commands in the correct order.

Fixes #1274.


reframe/core/modules.py
```diff
@@ -222,15 +222,21 @@ def load_module(self, name, force=False, collection=False):
             conflicting modules currently loaded. If module ``name`` refers to
             multiple real modules, all of the target modules will be loaded.
         :arg collection: The module is a "module collection" (TMod4 only)
-        :returns: the list of unloaded modules as strings.
+
```
teojgo, line 225: Remove this line

teojgo (changes requested):
Maybe also change the documentation for `emit_load_commands` and `emit_unload_commands`

vkarak:
> Maybe also change the documentation for `emit_load_commands` and `emit_unload_commands`

You mean the fact that they do not take into account anymore the module mapping?

teojgo:
> > Maybe also change the documentation for `emit_load_commands` and `emit_unload_commands`
> 
> You mean the fact that they do not take into account anymore the module mapping?

No, I mean you should use plural here since you get a list of commands:
https://github.com/eth-cscs/reframe/blob/76adeea3a17fc268374d270cc2e044e1a19e6214/reframe/core/modules.py#L351

reframe/core/modules.py
```diff
@@ -349,37 +354,41 @@ def searchpath_remove(self, *dirs):
...
 
-        return ret
+        # We don't consider module mappings here, because we cannot treat
+        # correctly possible conflicts
+        return [self._backend.emit_load_instr(Module(name, collection))]
 
     def emit_unload_commands(self, name, collection=False):
-        '''Return the appropriate shell command for unloading module
-        ``name``.
+        '''Return the appropriate shell command for unloading a module.
```
teojgo, line 376:
```suggestion
        '''Return the appropriate shell commands for unloading a module.
```

reframe/core/modules.py
```diff
@@ -349,37 +354,41 @@ def searchpath_remove(self, *dirs):
         return self._backend.searchpath_remove(*dirs)
 
     def emit_load_commands(self, name, collection=False):
-        '''Return the appropriate shell command for loading module ``name``.
+        '''Return the appropriate shell command for loading a module.
```
teojgo, line 357:
```suggestion
        '''Return the appropriate shell commands for loading a module.
```

teojgo (changes requested):

teojgo (approved):
Lgtm

PR #1612, opened 2020-11-20
[refactor] Align implementations of the PBS and the Torque scheduler regarding the use of `qstat`

### LINKED ISSUE #1473: Implement max_pending_time for the PBS scheduler

### LINKED ISSUE #1563: Align implementations of the PBS and the Torque scheduler regarding the use of `qstat`
Also for the PBS scheduler we can safely assume the `qstat` will be available.

ekouts:
Closes #1563
Closes #1473


reframe/core/schedulers/torque.py
```diff
@@ -18,115 +18,6 @@
...
-    'T': 'MOVED',
-    'W': 'WAITING',
-    'S': 'SUSPENDED',
-    'C': 'COMPLETED',
-}
-
-
 @register_scheduler('torque')
 class TorqueJobScheduler(PbsJobScheduler):
     TASKS_OPT = '-l nodes={num_nodes}:ppn={num_cpus_per_node}'
```
vkarak, line 23: Since the class is now so minimal, I suggest moving this over to `pbs.py`.

vkarak (changes requested):
Just a minor comment.

vkarak (approved):

PR #1613, opened 2020-11-20
[test] Add MPI version of the eat memory check

jgphpc:
thanks @ekouts


ekouts (dismissed):
Even though none of the tests fail there is an error printed in the end:
```
[  PASSED  ] Ran 15 test case(s) from 10 check(s) (0 failure(s))
[==========] Finished on Fri Nov 20 15:11:26 2020 
./bin/reframe: rfm_MemoryMpiCheck_job.out: No such file or directory
```
It is coming from these lines, but I am not sure how you should write the test to avoid it
```
        regex_mem = r'^Currently avail memory: (\d+)'
        self.reference_meminfo = \
            sn.extractsingle(regex_mem, self.stdout, 1,
                             conv=lambda x: int(int(x) / 1024**3))
```
Also this part doesn't need to be in a hook, since you use the deferable, right?

jgphpc:
:+1: I moved it out of the hook.
I still get:

cscs-checks/system/slurm/src/eatmemory_mpi.c
```diff
@@ -0,0 +1,173 @@
...
+        else {
+          printf("Invalid size format\n");
+          exit(0);
+        }
+      } else {
+        size = atoi(arg);
+      }
+
+      if (rank == 0) {
+        cscs_read_proc_meminfo(i);
```
vkarak, line 147: If you have edited this file, you should mention it at the beginning, that you are also an author.

vkarak (changes requested):
What I cannot understand very well is the value of this test. Why running the single node eat memory test on multiple nodes wouldn't do the same?

jgphpc:
>  value of this test

It's an extension of the MemoryOverconsumptionCheck: it tests a slurm flag with a fixed value, I am testing the same but I am trying test values closer to the max  with a limit

cscs-checks/system/slurm/src/eatmemory_mpi.c
jgphpc: It came from the problem we had with: https://github.com/eth-cscs/reframe/pull/1613#issuecomment-731994574
Do you have an example of what you want ?

vkarak: Why did you have to change the code? I think the problem in that comment was due to how you were computing the sanity function. As I proposed, you go without a hook.

jgphpc: refers to https://github.com/eth-cscs/reframe/pull/1613#discussion_r533725483

vkarak:
@jgphpc I did some minor enhancements to your test, but on Dom it fails to meet some references. Can you check? And then we can merge.

cscs-checks/system/slurm/slurm.py
```diff
@@ -243,10 +243,10 @@ def set_tasks(self):
...
     def reference_meminfo(self):
         reference_meminfo = {
-            'dom:gpu': 64,
-            'dom:mc': 64,
-            'daint:gpu': 64,
-            'daint:mc': 64,  # this will pass with 64 GB and above memory sizes
+            'dom:gpu': 62,
+            'dom:mc': 62,
+            'daint:gpu': 62,
+            'daint:mc': 62,  # this will pass with 64 GB and above memory sizes
```
vkarak, line 249: Is this comment still relevant?

jgphpc: I think yes because some daint:mc nodes have RealMemory=126000

vkarak (approved):

Issue #1568, opened 2020-11-02
Syntactic elements for creating test libraries and managing efficiently large parameterization spaces


vkarak:
The idea is to introduce new syntax for declaring typed test variables along with additional attributes, such as whether the variable is required. This will allow users to create extensible base library tests setting the requirements for their derived tests. The following examples shows relevant excerpts from this [gist](https://gist.github.com/vkarak/be9e88fa3c86abd50bfaac04a97d5fa5), which is how our current GROMACS test is envisioned with this new syntax:

```python
import os

import reframe as rfm
import reframe.utility.sanity as sn


class GromacsCheck(rfm.RunOnlyRegressionTest):
    '''This will be part of a library.'''

    parameterize('variant' ['cpu', 'gpu'])
    perfvar('throughput', 'ns/day')

    # inherit=True, redefines the parameter but keeps the original type
    # NOTE: Where do we validate that a required field is provided?
    var('modules', inherit=True, required=True)
    var('num_tasks', inherit=True, required=True)
    var('num_tasks_per_node', inherit=True, required=True)

    # NOTE: I'm not sure whether the variant should be passed here or set in
    # the test in the __new__() method perhaps
    def __init__(self):
        ...
        self.executable = 'gmx_mpi'
        self.keep_files = ['md.log']
        energy = sn.extractsingle(r'\s+Potential\s+Kinetic En\.\s+Total Energy'
                                  r'\s+Conserved En\.\s+Temperature\n'
                                  r'(\s+\S+){2}\s+(?P<energy>\S+)(\s+\S+){2}\n'
                                  r'\s+Pressure \(bar\)\s+Constr\. rmsd',
                                  output_file, 'energy', float, item=-1)
        energy_reference = -3270799.9
        self.sanity_patterns = sn.all([
            sn.assert_found('Finished mdrun', output_file),
            sn.assert_reference(energy, energy_reference, -0.001, 0.001)
        ])
        self.perf_patterns = {
            'throughput': sn.extractsingle(r'Performance:\s+(?P<perf>\S+)',
                                           'md.log', 'perf', float)
        }


@rfm.register_test
class CSCSGromacsCheck(rfm.RunOnlyRegressionTest):
    parameterize('scale', ['small', 'large'])
    parameterize('run_type', ['production', 'maintenance'])

    def __init__(self):
        self.valid_prog_environs = ['builtin']
        if self.variant == 'cpu':
            self.valid_systems = ['daint:mc']
[...]

PR #1614, opened 2020-11-23
[feat] Introduce new syntax for allowing the creation of extensible parametrised tests

### LINKED ISSUE #1567: Parameterized base tests and extensible parameterization of tests
This was discussed in #1555, but I'm putting it as a separate issue, since it can be implemented as a stand-alone feature. The idea is to extend the notion of the parameterized test to base tests (not necessarily registered for running) , from which other concrete tests derive. To achieve this, we propose the use a `parameterize` directive at the class level that will act similarly to the current `@parameterized_test` decorator. Here is an example:

```python
class MyBaseTest(rfm.RegressionTest):
    parameterize('variant', str, ['cpu', 'gpu'])
    def __init__(self):
        ...

class MyTest(MyBaseTest):
    parameterize('scale', str, ['small', 'large'])
    def __init__(self):
        ...
        if self.variant == 'cpu' and self.scale == 'large':
            do_sth()
```

`MyTest` will be parametrized in two dimensions ('variant' and 'scale') leading to four actual tests. More functionality as the one suggested in #1555 can be incorporated to this directive.

jjotero:
Implementation of a new way to write more flexible parametrised tests (see #1567). This new approach does not rely on the test parameters to be passed as arguments to the `__init__` method. Instead, the parameters are defined through class attribute directives that append these parameters to the namespace of the class. This allows extending the parameter space directly through class inheritance, which was not possible before. This method to write parametrised test introduces the concept of abstract parametrised tests. This would be tests with at least one parameter that does not have any values assigned to it. See this examples below.
```python
class A(rfm.RegressionTest):                                                                                                          
    '''                                                                                                                                       
    This is a regular non-parametrised test                                                                                                           
    '''                                                                                                                                       
    def __init__(self):                                                                                                                       
        self.valid_systems = ['*']                                                                                                            
        self.sourcepath = 'hello.cpp'                                                                                                         
        self.sanity_patterns = sn.assert_found(r'Hello, World\!', self.stdout)                                                                
                                                                                                                                              
                                                                                                                                              
@rfm.test                                                                                                                                     
class B(A):                                                                                                                
    '''                                                                                                                                       
    This is now a parametrised test. The @rfm.test decorator registers the test.                                                              
    This will result into an all to all combination of the parameter space.                                                                   
    '''                                                                                                                                       
    rfm_parameter('P0', ['a', 'b', 'c'])                                                                                                      
    rfm_parameter('P1', ['1', '2'])                                                                                                           
    def __init__(self):                                                                                                                       
        super().__init__()                                                                                                                    
        self.build_system.cxxflags = [r'-DFLAG1="\"%s\""' % self.P0,                                                                          
                                      r'-DFLAG2="\"%s\""' % self.P1,                                                                          
                                     ]                                                                                                        
                                                                                                                                              
                                                                                                                                              
class C(B):                                                                                                             
    '''                                                                                                                                       
    Make the test an abstract test - this will raise an error if decorated with @rfm.test                                                     
    '''                                                                                                                                       
    # P0 is an abstract parameter.                                                                                                            
    rfm_parameter('P0', [])                                                                                                                   
                                                                                                                                              
    # Previous P1 values are wiped.                                                                                                           
    rfm_parameter('P1', ['3', '4', '5'])                                                                                                      
                                                                                                                                              
                                                                                                                                              
@rfm.test                                                                                                                                     
class D(C):                                                                                                             
    '''                                                                                                                                       
    Change and extend the parameter space.                                                                                                    
    '''                                                                                                                                       
    # P0 no longer abstract                                                                                                                   
    rfm_parameter('P0', ['d'])                                                                                                                
                                                                                                                                              
    # P1 now has ['3', '4', '6']                                                                                                              
    rfm_parameter('P1', ['6'],                                                                                                                
                  inherit_params=True, # Extend the existing values for P1                                                                    
                  filt_params=lambda x: x[:-1] # Pop the last value of the inherited P1 values                                                
                 )                                                                                                                            
                                                                                                                                              
                                                                                                                                              
class E(D):                                                                                                             
    '''                                                                                                                                       
    Make P0 and P1 abstract parameters.                                                                                                      
    '''                                                                                                                                       
    rfm_purge_parameters('P0', 'P1')                                                                                                                
                                                                                                                                              
                                                                                                                                              
class F(D):                                                                                                             
    '''                                                                                                                                       
    Alternatively, one can also purge the whole parameter space without listing all the available parameters.
    '''                                                                                                                                       
    rfm_purge_all_parameters()   
```
Closes #1567


vkarak:
Some initial comments by reading the high-level description only:

1. I would prefer the directive to be named `parameter()` instead of `rfm_parameter()` and be exposed through the `reframe` module, so that users would write `rfm.parameter()` as they now write other `rfm.*` stuff.
2. What is the difference between `rfm.test` and `rfm.simple_test`? I find confusing to have both. Couldn't the `rfm.simple_test` take also on the role your `rfm.test` here?
3. In class `C`, it's not clear to me what makes it an abstract parameterized class. The fact that the parameter `P0` is initialized as an empty list? That can be a bit counter intuitive, because somebody could simply set a parameter to an empty list for testing or a function that the user calls may return an empty list. I would vote for a more explicit way (if necessary) for marking that a parameter is abstract.
4. I would probably prefer the `purge_parameter()` to be an option of the `parameter()` directive instead.
5. Regarding the extension of the parameter space, now done with the `inherit_params`, I don't know if it would be possible or nicer to be achieved using language elements directly, e.g., `rfm_parameter('P1', C.P1 + ['6'])`.

jjotero:
I think it's better if we just focus on the high-level stuff for now :)
 1. `parameter()` sounds good to me. Not sure there is the need to add the `rfm.*` to it before though - that's already done when you inherit from the `rfm.RegressionTest`. 
 2. With this new syntax, if we go for `simple_test` doing the test registration, one could write a parametrised test and decorate it with `simple_test`, which I would find a bit confusing. Moving forward, if the current way of writing parametrised tests is deprecated, there should be no need for two different decorators, and having something like `test` would be just enough.
3. `P0` is not initialised as an empty list. For that you'd need `parameter('P0', [[]])`.
4. That is what something like `parameter('P0', [])` would do. The purpose of `purge_parameters` is just to give the user a shortcut to make multiple parameters abstract in a shorter way (`purge_parameters('P0', 'P1', ..., 'Pn')`).
5. I guess that could be done. But if you do this on, say `class D` in the example above and you do `parameter('P0', A.P0 + ['4'])`, if one were to change `class B` and make it inherit from another class other than `A`, you'd be breaking all the classes upstream.

vkarak:
1. I just mean that it should be exported at the `reframe` module, as we do with the `RegressionTest` etc., although they are defined in more esoteric modules.
2. If we go with `test` now, then practically we would need to either deprecate both `simple_test` and `parameterized_test` or have three decorators. Neither is very good I think.
3. Why do you need a list of lists? Also, wouldn't `parameter('P0')` be more natural as an "abstract parameter?"
4. Oh, so `purge_parameters()` purges the parameter space and makes them abstract. I thought it was just removing them. What if you were redefining them without any values, e.g.,`parameter('P0')`. Then you loose the convenience function, but that's not a big deal for the moment, I think.
5. I think we can leave it as of now.

jjotero:
1. I'm not sure I follow 😅. In order to write `rfm.parameter` and so on, we'd simply need to wrap all these directives into a class and instantiate it as `rfm` in the `RegressionTest` namespace. Is that what you mean? 
2. There isn't really a need to deprecate `parameterized_test`. That does not interfere at all with the new syntax and users could keep using that decorator if they want. This would be more of a renaming of `simple_test` into just `test`. But I guess that could stay as `simple test` for now.
3. I don't need lists (good point 😂). Yeah, agreed.
4. Yes, if we go for `parameter('P0')` to make a parameter abstract, then `purge_parameters()` would be a very minor shortcut to make multiple parameters abstract. I think it makes more sense to start without `purge_parameters` and we could include it later if we see it could be useful.
5. 👍

vkarak:
1. No, I mean something simpler. Check how we even expose the `RegressionTest` under the `reframe` module. It will not require you to prefix the directive with `rfm.`, as you are not required to write `rfm.RegressionTest` if you do `from reframe import RegressionTest`. I am just saying that we should do the same here. I need to look into the code, so ignore this for the moment. It's really just cosmetic.
2. Yes, let's stay with the not-so-`simple_test`-anymore for the moment :-)
3. Nice.
4. Let's go with `parameter('P0')` syntax and leave the purge shortcut for the future.

I will start looking into the implementation now for more details.

jjotero:
Sounds good to me. I'll push some of these changes today.

jjotero:
I have now addressed the initial PR comments and the syntax to define a reframe test is 
```python
@rfm.simple_test
class test(rfm.RegressionTest):
    parameter('P0', 1, 2, 3, 4, 5, inherit_params=True, filter_params=lambda x: x[:-1])
    def __init__(self):
        ...
```
All the other directives have been dropped for now. 

I've also reviewed the implementation and I've done some changes for clarity/simplicity in the code structure.

vkarak (changes requested):
I haven't gone through the whole PR, but I think it's worth if I leave a first set of comments now. It's a bit hard to follow especially how the parameter space is built. I would suggest the following (it's mostly a slight refactoring):

- Rename the `RegressionTestParameter` to `_TestParam`. The rest should stay the same, but I would prefer if it had private fields accessed through properties.
- I think the `ParameterStagingArea` should be changed a bit. First, I would call it `LocalParamSpace`, since the parameter space local to the class that is being defined. Then I would make it simply a wrapper to your current `staging_area` dict and I would make it to behave transparently as a dictionary. This is quite easy to do with the following trick:

```python
class LocalParamSpace:
    def __init__(self):
        self._param_space = {}

    def __getattr__(self, name):
        # Delegate any unknown attribute access to the actual parameter space
        return getattr(self._param_space, name)

    def __setitem__(self, key, value):
        if name not in self._param_space:
            self._param_space[key] = value
        else:
            raise ValueError(
                'Cannot double-define a parameter in the same class.'
            )

    def add_param(self, name, *values, **kwargs):
        self[name] = _TestParam(name, *values, **kwargs)
```
- Now this local parameter space should be stored in the namespace instead of the `_rfm_build_param_space`, say `_rfm_local_param_space`.
- Now the `build_parameter_space()` and the helpers ("inherit" and "extend") can be free functions operating on the classes only. They will access the local parameter space from the special attribute we just defined above. Also you don't need to pass the bases, you can get them from the class' `mro()`. So then the interface becomes very simple: `build_param_space(cls)`!

I will come later with more comments, but I think these are more fundamental and will make it more readable. The problem was that `build_param_space(...)` does not make too much sense to be defined as a method of the local param space, because it is just using it in order to build the param space of a class, so that was confusing.

vkarak:
We will also need unit tests. I would suggest in a completely separate file, e.g., `test_parameterization.py`, where you can define your classes inside that, or even at the test level.

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+
+#
+# attribute extension to the RegressionTest classes.
+#
+
+
+class RegressionTestParameter:
+    '''
+    Regression test paramter class.
+    Stores the attributes of a regression test parameter needed to build
```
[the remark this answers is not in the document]
jjotero: 👍 Updated.

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+        '''
+        name: parameter name
+        values: parameter values. If no values are passed, the parameter is
+            considered as declared but not defined (i.e. an abstract param).
+        inherit_params: If false, this parameter is marked to not inherit any
+            values that might have been defined in a parent class.
+        filter_params: Function to filter/modify the inherited parameter values
+             from the parent classes. This only has an effect if used with
+             inherit_params=True.
+        '''
```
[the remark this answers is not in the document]
jjotero: I've updated the docstrings of these new functions.

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+#
+# attribute extension to the RegressionTest classes.
+#
+
+
+class RegressionTestParameter:
```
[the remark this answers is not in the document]
jjotero: Agree. It makes a lot more sense.

reframe/core/meta.py
```diff
@@ -8,12 +8,37 @@
 #
 
 from reframe.core.exceptions import ReframeSyntaxError
+import reframe.core.directives as ReframeDirectives
```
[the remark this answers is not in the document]
jjotero: I moved the parameter logic to `reframe/core/parameters.py`. The reason why I've gone for parameters is because all the other files in `reframe/core` are also nouns.

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+        See the RegressionTestParameter class for further information on the
+        function arguments.
+        '''
+        if name not in self.staging_area:
+            self.staging_area[name] = RegressionTestParameter(
+                name, *values, **kwargs
+            )
+        else:
+            raise ValueError(
+                'Cannot double-define a parameter in the same class.'
```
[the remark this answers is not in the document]
jjotero: 👍

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+    add_regression_test_parameter.
+
+    The regression test's parameter space is assembled during the class object
+    initialization (i.e. the __init__ method of the metaclass), where a call
+    to the member function build_parameter_space is made. This member function
+    handles the parameter inheritance from the parent classes.
+    '''
+
+    def __init__(self):
+        self.staging_area = {}
```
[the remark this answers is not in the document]
jjotero: Done.

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+        param_space_key: class attribute name under which the parameter space
+            is stored.
+        '''
+        # Temporary dict where we build the parameter space from the base
+        # clases
+        temp_parameter_space = {}
+
+        # Iterate over the base classes and inherit the parameter space
+        for b in bases:
+            base_params = b.__dict__.get(param_space_key, ())
```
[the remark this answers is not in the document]
jjotero: Done.

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+            def filter_params(x):
+                return x
+
+        self.name = name
+        self.values = values
+        self.inherit_params = inherit_params
+        self.filt_params = filter_params
+
+
+class ParameterStagingArea:
```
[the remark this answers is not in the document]
jjotero: Agreed and updated.

reframe/core/meta.py
```diff
@@ -8,12 +8,37 @@
...
+        # Directive to add a regression test parameter directly in the
+        # class body as: `parameter('P0', 0,1,2,3)`.
+        namespace['parameter'] = param_stage.add_regression_test_parameter
+
+        @classmethod
+        def build_param_space(cls, *args, **kwargs):
+            return param_stage.build_parameter_space(cls, *args, **kwargs)
+
+        # Method to build the final parameter space
+        namespace['_rfm_build_param_space'] = build_param_space
```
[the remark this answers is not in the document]
jjotero: I've changed the `build_parameter_space` is now a free function in the `parameters.py` module. So no need for the above.

reframe/core/pipeline.py
```diff
@@ -706,9 +706,19 @@ def pipeline_hooks(cls):
     def __new__(cls, *args, **kwargs):
         obj = super().__new__(cls)
 
+        # Abort the class instantiation if the test is an abstract test.
+        if cls.is_abstract_test():
+            raise ValueError('Cannot instantiate an abstract test.') from None
```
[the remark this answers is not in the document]
jjotero: 👍

reframe/core/pipeline.py
```diff
@@ -734,6 +744,81 @@ def __new__(cls, *args, **kwargs):
...
+
+        # If the param space is not present anymore, it's because we
+        # have already instantiated the full parameter space. However
+        # this will get called if the instance is copied.
+        for key in cls._rfm_params:
+            obj.__dict__[key] = None
+
+    @classmethod
+    def is_abstract_test(cls):
+        ''' Checks if the test is an abstract test '''
```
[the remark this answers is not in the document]
jjotero: 👍 Done.

reframe/core/directives.py
```diff
@@ -0,0 +1,169 @@
...
+        if name not in self.staging_area:
+            self.staging_area[name] = RegressionTestParameter(
+                name, *values, **kwargs
+            )
+        else:
+            raise ValueError(
+                'Cannot double-define a parameter in the same class.'
+            ) from None
+
+    def _inherit_parameter_space(self, bases, param_space_key):
```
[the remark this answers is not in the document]
jjotero: I just did that to keep all the parameter logic fully agnostic to what happens in the metaclass. With the updated implementation of `build_parameter_space`, `cls._rfm_params` is set by the method itself.

jjotero:
I have the impression that after these changes, the implementation looks more coherent and logical. I'm still not sure if the new methods in `pipeline.py` and how the `simple_test` decorator handles the parameter space is clear enough. In any case, it looks like we won't be changing the interface (for now 😂) so I'll go ahead and get the unit tests done.

jjotero:
I've now added the unit tests that check all the new functionalities in this PR. Though, some of the unit tests in `test_parameters.py` deal with functions implemented in `pipeline.py`. Not sure if some of these tests should be moved to `test_pipeline.py` instead.

reframe/core/pipeline.py
```diff
@@ -762,55 +766,70 @@ def param_space_len(cls):
 
     @classmethod
     def prepare_param_space(cls):
```
vkarak, line 768: Actually you don't need this. Here is the trick to create a generator from another one:

```python
    @classmethod
    def walk_param_space(cls):
        yield from itertools.product(*(p for p in cls._rfm_params.values()))
```

When you want to chain generators, you need the `yield from`, not just `yield`. So I think this will simplify the design in this part. You will only need to call `walk_param_space()` whenever you need to iterate over the parameter space.

vkarak (changes requested):
The design of the param space creation is now very clean! I really like it. For the machinery in `RegressionTest` I think I have a solution to avoid the iterator tricks that you do. I will skip for the moment the unit tests and come back later with comments there.

unittests/resources/checks/paramcheck.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2016-2020 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
```
vkarak, line 4: I wouldn't use a separate file for the tests but rather define them directly in the unit test file.

unittests/resources/checks/paramcheck.py
```diff
@@ -0,0 +1,35 @@
...
+        self.name = 'noParams'
+        self.descr = 'Hello World test'
+
+        # All available systems are supported
+        self.valid_systems = ['*']
+        self.valid_prog_environs = ['*']
+        self.executable = 'echo "Hello, World!"'
+        self.tags = {'foo', 'bar'}
+        self.sanity_patterns = sn.assert_found(r'Hello, World', self.stdout)
+        self.maintainers = ['JO']
```
vkarak, line 21:
```suggestion
class NoParams(rfm.RunOnlyRegressionTest):
    pass
```

unittests/resources/checks/paramcheck.py
vkarak: The tests can be really minimal and since we don't plan to run them, they can go in the unit test file directly.

unittests/test_parameters.py
```diff
@@ -0,0 +1,150 @@
...
+
+    assert MyTest._rfm_params == {}
+
+
+def test_params_are_present():
+    class MyTest(TwoParams):
+        pass
+
+    assert MyTest._rfm_params['P0'] == ('a',)
+    assert MyTest._rfm_params['P1'] == ('b',)
```
vkarak, line 22: Here I wouldn't look into the `_rfm_params`, since this is an implementation detail. I would rather register normally the tests that I'm interested in and then check that the `P0`, `P1` are defined correctly. This way, you will also check that the test registration happens correctly.

vkarak (changes requested):
I finished reviewing also the unit tests.

reframe/core/parameters.py
```diff
@@ -0,0 +1,200 @@
...
+    for b in bases:
+        base_params = getattr(b, '_rfm_params', ())
+        for key in base_params:
+            # With multiple inheritance, a single parameter
+            # could be doubly defined and lead to repeated
+            # values.
+            if key in param_space:
+                if not (param_space[key] == () or
+                        base_params[key] == ()):
+                    raise NameConflictError(f'parameter space conflict '
```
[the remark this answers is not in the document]
jjotero: I've changed it to `ReframeSyntaxError`

reframe/core/parameters.py
```diff
@@ -0,0 +1,200 @@
...
+    # Iterate over the base classes and inherit the parameter space
+    for b in bases:
+        base_params = getattr(b, '_rfm_params', ())
+        for key in base_params:
+            # With multiple inheritance, a single parameter
+            # could be doubly defined and lead to repeated
+            # values.
+            if key in param_space:
+                if not (param_space[key] == () or
+                        base_params[key] == ()):
```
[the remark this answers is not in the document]
jjotero: Makes sense. Done.

reframe/core/parameters.py
```diff
@@ -0,0 +1,200 @@
...
+                    raise NameConflictError(f'parameter space conflict '
+                                            f'(on {key}) due to '
+                                            f'multiple inheritance.'
+                                           ) from None
+
+            param_space[key] = base_params.get(
+                key, ()) + param_space.get(key, ())
```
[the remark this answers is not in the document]
jjotero: The order does not matter here. These two tuples only get merged if one of them is empty. If that were not the case, an exception would be raised just before this.

reframe/core/parameters.py
```diff
@@ -0,0 +1,200 @@
...
+        :class:`reframe.core.parameters.LocalParamSpace`.
+    :param param_space: an existing parameter space. This **must** have been
+        generated with
+        :meth:`reframe.core.parameters._merge_parameter_spaces`.
+
+    '''
+    # Ensure that the local parameter space is an instance of LocalParamSpace
+    if not isinstance(local_param_space, LocalParamSpace):
+        raise ReframeSyntaxError(f'the local_param_space must be an '
+                                 f'instance of the LocalParamSpace class')
```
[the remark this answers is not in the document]
jjotero: No, it can't. Changed this to an `assert`.

reframe/core/parameters.py
```diff
@@ -0,0 +1,200 @@
...
+
+    :returns: dictionary containing the full parameter space. The keys are the
+        parameter names and the values are tuples containing all the values
+        for each of the parameters.
+    '''
+    # Inherit the parameter space from the base classes
+    param_space = _merge_parameter_spaces(cls.__bases__)
+
+    # Extend the parameter space with the local parameter space
+    _extend_parameter_space(cls._rfm_local_param_space, param_space)
```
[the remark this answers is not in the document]
jjotero: Using `cls.mro()` wouldn't work here. `cls.mro()` returns the full hierarchy of parent classes, and that would trigger the `parameter space conflict (on {key}) due to multiple inheritance` error. There is no need to pass the full inheritance tree in here because the direct parent has already it's own `cls._rfm_params` built in there. So just passing the direct parents with `cls.__bases__` is what we want here.

reframe/core/pipeline.py
```diff
@@ -762,55 +766,70 @@ def param_space_len(cls):
...
                 *(p for p in cls._rfm_params.values())
             )
 
-        return cls.param_space_len()
-
     @classmethod
-    def _set_parameter_space(cls, obj):
-        '''
-        Adds to obj.__dict__ the keys corresponding to the test parameters.
+    def _set_param_space(cls, obj):
```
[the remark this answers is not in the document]
jjotero: `_construct_params`?

reframe/core/pipeline.py
```diff
@@ -762,55 +766,70 @@ def param_space_len(cls):
...
+
+        If any of the parameters declared in the test parameter space
+        (cls._rfm_params) has no defined values, the parameter is considered
+        an abstract parameter. Therefore, a regression test with at least one
+        abstract parameter is considered an abstract test.
+
+        :return: bool indicating wheteher the test is abstract or not
+        '''
+        if cls.param_space_len() == 0:
+            return True
```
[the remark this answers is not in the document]
jjotero: 😅

reframe/core/pipeline.py
jjotero: As discussed before, this is still needed by the `simple_test` decorator. A call to `prepare_param_space` will create the iterator that feeds the values to the parameters as the instances of the class get created. If this `prepare_param_space` method is not called before class instantiation, or the class is instantiated after the iterator has been exhausted, the test parameters will be set to `None`.

unittests/resources/checks/paramcheck.py
jjotero: Makes sense. Moved theses classes to `unittests/test_parameters.py`

unittests/test_parameters.py
jjotero: Such test is also further down. My intention with this test was to check only the logic related to the parameter space build, and make sure that the parameter overriding, inheritance and filtering works. 
Building `cls._rfm_params` and populating the actual parameters in the object are two different steps and I think they should be tested separately. The build of `cls._rfm_params` is dealt with by the logic implemented into the metaclass, but populating the actual parameters into the object is done by the pipeline.

How about adding to `RegressionTest` a method such as 
```python
@classmethod
def get_parameter_space(cls):
    return cls._rfm_params
```
This would allow us to test that the parameter space is built correctly without the need of using any implementation details.

jjotero:
I think it would make more sense to offload some of the `@classmethods` from the `pipeline.py` into `meta.py`. I'm thinking of `is_abstract`, `param_space_len` and `walk_param_space` (this last one maybe changed to `param_space_iter`).
On the other hand, `prepare_param_space` and `_construct_params` should remain in the pipeline, as these methods are specific for the instantiation of the `RegressionTest` class.

jjotero:
The `pipeline.py` and the `simple_test` decorator look pretty clean now. I've wrapped most of the stuff into a new class in `parameters.py` named `ParamSpace`. This new class has a similar role to `LocalParamSpace`, but it deals with the whole process of building the full parameter space for the RegressionTest class. This `ParamSpace` class has the `__len__` and `__next__` methods implemented, which means that the class is iterable, and to know the length of the parameter space one must simply do `len(cls.param_space)`. 

The parameter space is now available to the users through `self.param_space` and the `is_abstract` method is now a property.

reframe/core/pipeline.py
```diff
@@ -707,7 +707,7 @@ def __new__(cls, *args, **kwargs):
         obj = super().__new__(cls)
 
         # Set the test parameters in the object
-        cls._set_param_space(obj)
+        cls._construct_params(obj)
```
vkarak, line 710: I think after all `_init_params()` is a better name, because we are kind of supplementing the constructor here and in Python that's the `__init__()`.

unittests/test_parameters.py
```diff
@@ -3,120 +3,109 @@
...
 
-def test_params_are_none():
+def test_param_len_is_zero():
     class MyTest(Abstract):
         pass
 
-    test = MyTest()
-    assert test.P0 == None
-    assert test.P1 == None
+    assert len(MyTest.param_space) == 0;
```
vkarak, line 86: I think it's good if you also do `assert MyTest.param_space == ??`

unittests/test_parameters.py
```diff
@@ -136,15 +125,13 @@ def test_extended_params_are_set():
     class MyTest(ExtendParams):
         pass
 
-    MyTest.prepare_param_space()
-    for i in range(MyTest.param_space_len()):
+    for i in range(len(MyTest.param_space)):
```
vkarak, line 128: I think this is equivalent and simpler:

```suggestion
    for _ in MyTest.param_space:
```

vkarak (changes requested):
Looks great now! Apart from my really minor comments, there are still two things we need to address:

1. Setting the name of the test based on the parameter values. This is something the `parameterized_test` does, but it does not happen here. I'm not saying that we should solve the naming problem now, we could still use the naive scheme from the `parameterized_test` for the moment. If we don't do, we will have test name clashes and ReFrame will not be very happy.
2. We need also unit tests to test how the decorators work together with the new parameterization infrastructure.

reframe/core/parameters.py
```diff
@@ -99,102 +104,141 @@ def add_param(self, name, *values, **kwargs):
...
+                               target_cls.__bases__):
+                self.join(base._rfm_param_space)
+
+            # Extend the parameter space with the local parameter space
+            try:
+                for name, p in target_cls._rfm_local_param_space.items():
+                    self._params[name] = (
+                        p.filter_params(self._params.get(name, ())) + p.values
+                    )
+            except AttributeError:
```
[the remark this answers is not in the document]
jjotero: Good point. I'm removing the try/except. Though, I think it makes sense to
```python
assert hasattr(target_cls, '_rfm_local_param_space')
assert isinstance(target_cls._rfm_local_param_space, LocalParamSpace)
```
inside the `if target_cls` to catch any potential misuse of this method.

reframe/core/parameters.py
```diff
@@ -99,102 +104,141 @@ def add_param(self, name, *values, **kwargs):
...
+    the extension of the inherited parameter space with the local parameter
+    space (this must be an instance of
+    :class `reframe.core.parameters.LocalParamSpace`), and lastly, a check to
+    ensure that none of the parameter names clashes with any of the class
+    attributes existing in the target class. If no target class is provided,
+    the parameter space is initialized as empty. After the parameter space is
+    set, a parameter space iterator is created, which allows traversing the
+    full parameter space walking though all posible parameter combinations.
+    Since this class is iterable, this may be used by the RegressionTest
+    constructor to assing the values to the test parameters. Note that the
```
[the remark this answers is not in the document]
jjotero: Good catch! :)

reframe/core/pipeline.py
jjotero: Agree. That name makes a lot more sense.

unittests/test_parameters.py
jjotero: I've added a new test that does the `assert MyTest.param_space == ...` in a separate test. I think it's cleaner to test the build of the param space and the param initialisation in different tests.

unittests/test_parameters.py
jjotero: With this implementation, doing that loop as `for _ in MyTest.param_space` doesn't work, because the internal iterator in the `ParamSpace` class is consumed in the wrong place and then the parameters are not initialised correctly (a `StopIteration` exception is raised during the regression test instantiation and the parameters become `None` instead of getting their right values).  This shows that the implementation for setting the parameter space is not yet very clean (or at least not very safe). I've got an idea on how to get this done in a much cleaner way. I'll try it out and then summarise it in here if successful.

jjotero:
1. If I understand correctly, I think this is already implemented in the `pipeline.py` with the method `_append_parameters_to_name`. This method is called in `__new__` and it gives exactly the same name as if the parameterisation were to happen through the `@rfm.parameterized_test` decorator.
2. I'm still not 100% happy the way the test registering and parameter initialisation is done. Let me give it another twist and see if I come up with something better. Then I'll get these unit tests done for that.

victorusu:
Will this PR allow me to

+ Extend the test to also use the parameterized_test syntax? (check `SpackInstallSupportedGPUApps` class)
+ Depend on a long inheritance chain, without having to redeclare the parameters? (check `SpackInstallDaintGPUApps` class)
+ Depend on a previously defined parameter set, like when we have the `parameterized_test` syntax? (check `SpackInstallSupportedGPUApps` class)

```python

import reframe as rfm

class SpackInstallSpecs(rfm.RegressionTest):
    rfm_parameter('specs', [])
    rfm_parameter('options', [])
    rfm_parameter('arch', [])
    def __init__(self):
      ...

class SpackInstallDaintGPUApps(SpackInstallSpecs):
    # Do I have to set the parameters empty or is it ok to skip them?
    # rfm_parameter('specs', [])
    # rfm_parameter('options', [])
    rfm_parameter('arch', ['haswell'])

    def __init__(self):
      ...

class SpackInstallDaintMCApps(SpackInstallSpecs):
    # Do I have to set the parameters empty or is it ok to skip them?
    # rfm_parameter('specs', [])
    # rfm_parameter('options', [])
    rfm_parameter('arch', ['broadwell'])

    def __init__(self):
      ...


@rfm.parameterized_test(['', '+plumed', '+craypat'])
class SpackInstallSupportedGPUApps(SpackInstallDaintGPUApps):
    rfm_parameter('specs', ['gromacs@02020.4', 'gromacs@02020.2'])
    rfm_parameter('options', ['+cuda'])
    # Do I have to reset the arch parameter?
    # rfm_parameter('arch', ['haswell'])

    def __init__(self, variant):
      ...
      self.valid_systems = ['daint:gpu','dom:gpu']
      self.variant = variant
      self.dep_name = re.sub(f'{util.toalphanum(self.variant)}', r'_', self.name)
      # make the +plumed and +craypat variants depend on the '' one
      # it is a waste of compute node hours to test those if the basic one fails
      self.depends_on(self.dep_name, when=parent_part_env('login', 'builtin'))

@rfm.parameterized_test(['', '+plumed', '+craypat'])
class SpackInstallSupportedMCApps(SpackInstallDaintMCApps):
    rfm_parameter('specs', ['gromacs@02020.4', 'gromacs@02020.2'])
    rfm_parameter('options', ['~cuda'])
    # Do I have to reset the arch parameter?
    # rfm_parameter('arch', ['haswell'])

    def __init__(self, variant):
      ...
      self.valid_systems = ['daint:mc','dom:mc']
      self.variant = variant
      self.dep_name = re.sub(f'{util.toalphanum(self.variant)}', r'_', self.name)
      # make the +plumed and +craypat variants depend on the '' one
      # it is a waste of compute node hours to test those if the basic one fails
      self.depends_on(self.dep_name, when=parent_part_env('login', 'builtin'))


```

jjotero:
@victorusu short answer: can be easily implemented, yes and yes.

Longer answer:
1. I'm still working on the final details as in how the reframe test consumes the parameter space and how that is dealt with the decorators and the `__new__` method in the `RegressionTest` class. But with the current form of the code, it should be trivial to extend this to work with the `parameterized_test` decorator. These two approaches are orthogonal.
2. A parameter defined in a parent class is inherited as is from the parent class unless redefined in the current class. For example, in your snippet above, if you define a parameter as  
```python
    parameter('arch', 'haswell')
    def __init__(self, ...):
        ...
``` 
the parameter `arch` will always `self.arch == 'haswell'` unless any of the child classes that inherit from this base class  override/modify this parameter.
3. The full parameter space is available through `cls.param_space`, so you can pretty much do anything with it.

jjotero:
IMO, the way the decorators interact with the test class is much much cleaner now. Before the default behaviour was that, if the parameter space had `N` points, the first `N` instances of the test class would get these values from the parameter space, and the subsequent instantiations of the class (or deep-copies fo the object) would init these parameters as `None`. Even though the instantiation order is unlikely to change, this method provided no control over which instances do feed from the parameter space and which are inited as `None`.

This last method goes along the lines of what @vkarak once suggested, which is using the decorator to feed the test with the parameters. Though, instead of doing this, we just use both decorators (`simple_test` and `parameterized_test`) pass a single additional argument `_rfm_use_params=True` to the `__new__` method of the test class, and this triggers the object to get the values from the parameter space. This `_rfm_use_params` argument is then intercepted and is not forwarded to the `__init__` method. If this argument is not passed to the `__new__` method, the parameters are simply initialised as `None`. Now, on the error-catching side, if one were to register the class a number of times greater than length of the parameter space, there would be a `StopIteration` exception raised during the `__new__` call, as the iterator of the parameter space would simply run out.

With the code on its current state, all three scenarios raised by @victorusu work.

Since this changes things a bit, I'll wait until we all agree on this and then I'll move on to finish the unit tests.

reframe/core/meta.py
```diff
@@ -79,12 +79,21 @@ def __init__(cls, name, bases, namespace, **kwargs):
...
 
+    def __call__(cls, *args, **kwargs):
+        obj = cls.__new__(cls, *args, **kwargs)
+
+        # Intercept constructor arguments
+        if '_rfm_use_params' in kwargs:
+            del kwargs['_rfm_use_params']
+
+        obj.__init__(*args, **kwargs)
+        return obj
```
vkarak, line 90: Nice trick! I haven't thought about it. I think you can write more simply though:

```suggestion
    def __call__(cls, *args, **kwargs):
        kwargs.pop('_rfm_use_params', None)
        return super().__call__(*args, **kwargs)
```

reframe/core/pipeline.py
```diff
@@ -703,11 +703,11 @@ def pipeline_hooks(cls):
     #: :type: boolean : :default: :class:`True`
     build_locally = fields.TypedField('build_locally', bool)
 
-    def __new__(cls, *args, **kwargs):
+    def __new__(cls, *args, _rfm_use_params=False, **kwargs):
```
vkarak, line 706: Are you sure that this can ever be passed as true? Since it's consumed by the metaclass.

reframe/core/pipeline.py
```diff
@@ -741,29 +741,35 @@ def __init__(self):
...
-        # Try to set the values of the test parameters from the param iterator.
-        try:
-            param_values = next(cls._rfm_param_space)
+        # Set the values of the test parameters (if any)
+        if use_params and cls._rfm_param_space.params:
+            param_values = next(cls._rfm_param_space.unique_iter)
             for index, key in enumerate(cls._rfm_param_space.params):
                 setattr(obj, key, param_values[index])
 
             return
```
```suggestion
        else:
```

vkarak (changes requested):

reframe/core/meta.py
vkarak: Ok, now I see why you are doing as such. You want to call `__new__()` and `__init__()` with different keyword arguments. So ignore my comment here and the other below! Perhaps you need to comment this further so as to help others understand the trick here.

reframe/core/pipeline.py
vkarak: Ignore this, I haven't seen closely the metaclass' `__call__()` implementation.

reframe/core/decorators.py
```diff
@@ -27,13 +27,22 @@
 
 
 def _register_test(cls, args=None):
+    '''Register the test
+
+    Register the test with _rfm_use_params=True. This additional argument flags
+    this case to consume the parameter space. Otherwise, the regression test
+    parameters would simply be initialized to None.
+
+    :meta private:
```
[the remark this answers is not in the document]
jjotero: 👍 I took this out.

reframe/core/meta.py
jjotero: I've still changed that `if '_rfm_use_params' in kwargs` by `kwargs.pop('_rfm_use_params', None)` since it reads much better. I've also added some comments to the `__call__` method to explain what it does.

jjotero:
The unit tests that check the parameter space consumption with the decorators are still to do.

vkarak:
Looks great! I haven't pressed the "Approve" button yet, because we need to think a bit about some forward-looking parts of the API and the finalization of the unit tests:

1. Perhaps we need to allow a type argument for the parameter, to be in sync with the future `var()` directive.
2. The public access to the parameter space could be done from something called `params` instead of `param_space`. Again, this is forward looking. Also, technically, the access is not read-only, since the users can still mess with the dictionary behind the scenes, since only the handle is read-only. We can consider returning a `reframe.utility.MappingView` there instead.

jjotero:
1. What does `var()` do and how is it different from a parameter?
2. I think renaming `param_space` to `params` would add some confusion to the naming we use. When we register the tests we iterate over the parameter space and not over the parameters, therefore, the decorators do
```python
def simple_test(cls):
    ...
    for _ in cls.parameter_space:
        _register_test(cls)
```
Here `cls` could have 2 parameter with 2 different values for each parameter. Therefore, your parameter space would have a length of 4. Then, the above loop would do 4 iterations if you do `for _ in cls.params` even though you've got only 2 parameters. That doesn't sound right to me. The users can access the parameters directly with `cls.param_space.params` in any case.

I'm adding a third item here :)
3. When writing the unit tests, I just noticed that the examples provided above by @victorusu would not work as things stand right now. The `ParamSpace` class has a `unique_iter` implemented which ensures that you only walk through the full parameter space once. Thus, when you decorate the class with `parameterized_test`, you would need to iterate over the full parameter space as many times as parameter sets are provided by the decorator. I see two solutions here:
  - Allow for walking through the parameter space more than once (which I don't think it makes much sense in the context of how the parameters are defined and I also believe this opens a way for errors)
  - Or we simply make incompatible to use the `parameterized_test` decorator with the new parameters. There cases shown above by @victorusu don't need to be decorated with the `parameterized_test` and they could simply add these parameters with the new method.
```python
@rfm.simple_test
class SpackInstallSupportedMCApps(SpackInstallDaintMCApps):
    parameter('specs', 'gromacs@02020.4', 'gromacs@02020.2')
    parameter('options', '~cuda')
    parameter(variant, '', '+plumed', '+craypat')

    def __init__(self):
      ...
      self.valid_systems = ['daint:mc','dom:mc']
      ...
```

vkarak:
Hi @jjotero,

> What does var() do and how is it different from a parameter?

This is to help users define new variables in their library tests and attach attributes to them (for example, types, whether it is required or not etc.) Check an older suggestion here: https://gist.github.com/vkarak/be9e88fa3c86abd50bfaac04a97d5fa5#file-gromacs-py-L20-L22

> I think renaming param_space to params would add some confusion to the naming we use. When we register the tests we iterate over the parameter space and not over the parameters, therefore, the decorators do

For what you do, what you say makes perfectly sense. I'm trying to correlate this to the bigger picture of a test in the future. Again from that [gist](https://gist.github.com/vkarak/be9e88fa3c86abd50bfaac04a97d5fa5#file-gromacs-py-L96-L99) the idea was that the users could use `setvar()` with a `when` parameter to quickly slice the parameter space. When writing this as a user, I would like to write something compact, not `parameter_space`.

For your third point, I think the best idea is to make `parameter()` and `@parameterized_test` incompatible. I don't see a case that you would need both or that you couldn't achieve something with `parameter()` that you could with `@paramterized_test`, do you?

jjotero:
OK, I now understand why there is a need for a different mechanism to set test variables that may depend on the parameter space and other factors. There are still a few things that confuse me on that gist, but let's not get into that yet, since it's independent from what's being done on this PR (perhaps move the discussion over to #1568?).

What would be needed for the `setvar` method mentioned on the previous comment, is a different structure that holds only the actual values that the parameters take on each instance of the test. No need for the full parameter space there, which is what `self.param_space` is. Therefore, that `self.params` could be added together with the `setvar` method at a later stage. Though, I'm not very convinced by the slicing of `self.params` suggested on the gist. That would fix the position of each parameter and that could become unmanageable for large parameter spaces. I have the feeling that using a dict there instead would simplify things.

Agree on the third point. I'll make them incompatible.

vkarak:
The gist is certainly under discussion and we can revise it. It was a very first idea of how the tests might look like in the future. I just brought it in order to keep this in context, too.

reframe/core/decorators.py
```diff
@@ -119,10 +119,13 @@ def parameterized_test(*inst):
     '''
     def _do_register(cls):
         _validate_test(cls)
+        if not cls.param_space.is_empty():
+            raise ValueError(
+                'the decorated test already is a parameterized test'
```
vkarak, line 124:
```suggestion
                f'{cls.__qualname__!r} is already a parameterized test'
```

unittests/test_parameters.py
```diff
@@ -100,45 +104,68 @@ class MyTest(ExtendParams):
...
     assert test.P0 is None
     assert test.P1 is None
 
 
-def test_extended_params():
+def test_consume_param_space():
     class MyTest(ExtendParams):
-        pass
+        def __init__(self):
+            pass
```
vkarak, line 128: Do you need this?

vkarak (changes requested):
Just a minor comment.

reframe/core/decorators.py
jjotero: 👍

unittests/test_parameters.py
jjotero: Nope. Cleaned that up.

vkarak (approved):
Latm. I will open a separate issue for updating the documentation and perhaps adding a tutorial about it ;-)

vkarak:
I have not merged this yet, because I need to see how it shows up the documentation that you have put already.

jjotero:
I've extended the directives section a bit and expanded the example. Even though this example uses the `parameter` directive, everything mentioned in there should also apply to the `var` directive.  Also, the description of the `parameter` function is now more oriented towards its functionality rather than its implementation.

The `parameter` directive also shows as a member function of the `RegressionTest` class in the docs. I've included a link to the `Directives` section in there, and made clear that such function should only be called directly in the class body.

docs/regression_test_api.rst
```diff
@@ -36,6 +36,59 @@ Pipeline Hooks
...
+   Instead, if no parameter with a matching name exist in any of the parent parameter spaces, a new regression test parameter is created.
+
+   :param name: parameter name.
+   :param values: parameter values.
+       If no values are passed when creating a new parameter, the parameter is considered as declared but not defined (i.e. an abstract parameter).
+       Instead, for an existing parameter, this depends on the parameter's inheritance behaviour and on whether any values where provided in any of the parent parameter spaces.
+   :param inherit_params: If false, no parameter values that may have been defined in any of the parent parameter spaces will be inherited.
+   :param filter_params: Function to filter/modify the inherited parameter values that may have been provided in any of the parent parameter spaces.
+       This function must only expect a tuple containing the inherited parameter values as its only argument.
```
vkarak, line 86: Something like this perhaps?


```suggestion
       This function must accept a single argument, which will be passed an iterable containing the inherited parameter values.
```

vkarak (changes requested):
Looks good, just minor comments and corrections.

docs/regression_test_api.rst
jjotero: I think there is an `as` missing after `passed`: `This function must accept a single argument, which will be passed as an iterable containing the inherited parameter values.` Does this read better?

docs/regression_test_api.rst
```diff
@@ -36,6 +36,59 @@ Pipeline Hooks
...
+                override_this()
+            else:
+                override_other()
+
+
+.. py:function:: reframe.core.pipeline.RegressionTest.parameter(name, *values, inherit_params=False, filter_params=None)
+
+   Inserts or modifies a regression test parameter.
+   If a parameter with a matching name is already present in the parameter space of a parent class, the existing parameter values will be combined with those provided by this method following the inheritance behaviour set by the arguments ``inherit_params`` and ``filt_params``.
+   Instead, if no parameter with a matching name exist in any of the parent parameter spaces, a new regression test parameter is created.
```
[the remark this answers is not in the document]
jjotero: I've lost hope in ever being able to spot these kinds of typos 😭😅

jjotero:
@jenkins-cscs retry all

vkarak (approved):
lgtm

vkarak:
Goes in finally! 🍾  🎉

PR #1619, opened 2020-11-24
[bugfix] Fail properly when the source code language cannot be guessed

### LINKED ISSUE #1546: Compilation phase passes if a non-existent path is passed in `sourcepath`
```python
@rfm.simple_test
class CompileOnlyHelloTest(rfm.CompileOnlyRegressionTest):
    def __init__(self):
        self.descr = 'Compile-only C Hello World test'

        self.valid_systems = ['*']
        self.sourcepath = 'does-not-exist'
        self.sanity_patterns = sn.assert_not_found(r'(?i)error', self.stdout)
```

```
[ReFrame Setup]
  version:           3.3-dev0 (rev: 6d531e0a)
  command:           './bin/reframe -c ../misc/reframe-tests/issue-1504/hellocheck.py -n CompileOnlyHelloTest --keep-stage-files -r'
  settings file:     '<builtin>'
  check search path: '/users/karakasv/Devel/misc/reframe-tests/issue-1504/hellocheck.py'

[==========] Started on Wed Oct 21 23:00:44 2020

[----------] started processing CompileOnlyHelloTest (Compile-only C Hello World test)
[ RUN      ] CompileOnlyHelloTest on generic:default using builtin
[----------] finished processing CompileOnlyHelloTest (Compile-only C Hello World test)

[       OK ] (1/1) CompileOnlyHelloTest on generic:default using builtin [compile: 0.018s run: 0.021s total: 0.052s]

[  PASSED  ] Ran 1 test case(s) from 1 check(s) (0 failure(s))
[==========] Finished on Wed Oct 21 23:00:44 2020
```

teojgo:
* The SingleSource build system now raises an exception when
it fails to guess the programming language of the source file.

* Add unittests.

Fixes #1546

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -375,7 +375,6 @@ class SingleSource(BuildSystem):
             raise BuildSystemError(
                 'a source file is required when using the %s build system' %
                 type(self).__name__)
-
         cc = self._cc(environ)
         cxx = self._cxx(environ)
         ftn = self._ftn(environ)
@@ -426,8 +425,9 @@ class SingleSource(BuildSystem):
             cmd_parts += [nvcc, *cppflags, *cxxflags, self.srcfile,
                           '-o', executable, *ldflags]
         else:
-            BuildSystemError('could not guess language of file: %s' %
-                             self.srcfile)
+            raise BuildSystemError(
+                f'could not guess language of file: {self.srcfile}'
+            )
 
         return [' '.join(cmd_parts)]
 
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -375,6 +375,7 @@ class SingleSource(BuildSystem):
             raise BuildSystemError(
                 'a source file is required when using the %s build system' %
                 type(self).__name__)
+
         cc = self._cc(environ)
         cxx = self._cxx(environ)
         ftn = self._ftn(environ)
--- a/unittests/test_buildsystems.py
+++ b/unittests/test_buildsystems.py
@@ -239,16 +239,6 @@ def test_compiler_pick(lang):
 
 def test_singlesource_unknown_language():
     build_system = bs.SingleSource()
-    build_system.cc = 'cc'
-    build_system.cxx = 'CC'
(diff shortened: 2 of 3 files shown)


reframe/core/buildsystems.py
```diff
@@ -426,8 +426,9 @@ def emit_build_commands(self, environ):
             cmd_parts += [nvcc, *cppflags, *cxxflags, self.srcfile,
                           '-o', executable, *ldflags]
         else:
-            BuildSystemError('could not guess language of file: %s' %
-                             self.srcfile)
+            raise BuildSystemError(
```
vkarak, line 429: That was a very silly bug!

unittests/test_buildsystems.py
```diff
@@ -235,3 +235,20 @@ def test_compiler_pick(lang):
...
     assert ([f'{compilers[lang]} {build_system.srcfile} -o foo.x'] ==
             build_system.emit_build_commands(ProgEnvironment('testenv')))
+
+
+def test_singlesource_unknown_language():
+    build_system = bs.SingleSource()
+    build_system.cc = 'cc'
+    build_system.cxx = 'CC'
+    build_system.ftn = 'ftn'
+    build_system.nvcc = 'nvcc'
```
vkarak, line 245: Why do you need to set all these?

vkarak (commented):
What is not so clear to me is how this was triggering the bug that we observed. Can you explain a bit?

ekouts (approved):

unittests/test_buildsystems.py
teojgo: I will remove. I put there to have the compilers in place.

teojgo:
> What is not so clear to me is how this was triggering the bug that we observed. Can you explain a bit?

So in our case, even though the sourcepath was non existent, the build system did not raise the exception during autodetection. Therefore, no actual compilation/build command was generated and the compilation script did nothing so it was not failing.

vkarak (approved):
lgtm

vkarak:
@jenkins-cscs retry dom

PR #1620, opened 2020-11-26
[ci] Add configuration entries for Eiger

jgphpc:


config/cscs.py
```diff
@@ -627,6 +627,101 @@
...
+            'hostnames': [
+                'eiger'
+            ],
+            'modules_system': 'lmod',
+            'resourcesdir': '/apps/common/UES/reframe/resources',
+            'partitions': [
+                {
+                    'name': 'login',
+                    'scheduler': 'local',
+                    'modules': [],
```
vkarak, line 642: Remove this.

config/cscs.py
```diff
@@ -627,6 +627,101 @@
...
+                    'descr': 'Login nodes',
+                    'max_jobs': 4,
+                    'launcher': 'local'
+                },
+                {
+                    'name': 'mc_lowmem',
+                    'descr': 'Multicore nodes (AMD EPYC 7742, 256GB/cn)',
+                    'scheduler': 'slurm',
+                    'access': [
+                        '--exclude=nid00[1512-1703,1768-1927]'
```
vkarak, line 658: Can you add a `FIXME`  to remember updating this line?

config/cscs.py
```diff
@@ -627,6 +627,101 @@
...
+                        # f'--constraint=mc',
```
vkarak, line 659: Remove this.

config/cscs.py
```diff
@@ -627,6 +627,101 @@
...
+                        'builtin',
+                        'PrgEnv-cray',
+                        'PrgEnv-gnu',
+                        'PrgEnv-aocc'
+                    ],
+                    'max_jobs': 100,
+                    'resources': [
+                        {
+                            'name': 'bcast',
+                            'options': ['--bcast={path}']
```
vkarak, line 671: Will we often need this?

config/cscs.py
```diff
@@ -627,6 +627,101 @@
...
+                        },
+                        {
+                            'name': 'switches',
+                            'options': [
+                                '--switches={num_switches}'
+                            ]
+                        },
+                        {
+                            'name': 'mem-per-cpu',
```
vkarak, line 680: Will we often need this?

config/cscs.py
```diff
@@ -627,6 +627,101 @@
...
+                        },
+                    ],
+                    'launcher': 'srun'
+                },
+                {
+                    'name': 'mc_himem',
+                    'descr': 'Multicore nodes (AMD EPYC 7742, 512GB/cn)',
+                    'scheduler': 'slurm',
+                    'access': [
+                        '--exclude=nid00[1000-1003,1006-1191,1256-1415]'
```
vkarak, line 693: Same comment as above here.

vkarak (changes requested):

jgphpc:
You're right, I can't comment inline if I am not a reviewer 
My comment was:
for large job, we point users to: https://user.cscs.ch/access/running/#large-jobs

jgphpc:
We have: `eiger:mc_lowmem` and `eiger:mc_himem`
I could also add the full partition: `eiger:mc` ?

jgphpc:
Looks like the constraint will not be set. The users will have to use the `--mem=` slurm flag.

vkarak:
> Looks like the constraint will not be set. The users will have to use the `--mem=` slurm flag.

So the way we have set up the configuration now is not correct. In such case, I would suggest to have a single partition and add the memory as a resource.

config/cscs.py
```diff
@@ -627,6 +627,108 @@
...
+                    'max_jobs': 100,
+                    'resources': [
+                        {
+                            'name': 'switches',
+                            'options': [
+                                '--switches={num_switches}'
+                            ]
+                        },
+                        {
+                            'name': 'mem-per-node',
```
vkarak, line 671:
```suggestion
                            'name': 'memory',
```

config/cscs.py
```diff
@@ -627,6 +627,108 @@
...
```
vkarak, line 722:
```suggestion
                            'name': 'memory',
```

config/cscs.py
```diff
@@ -773,6 +875,33 @@
             'cxx': 'g++',
             'ftn': 'gfortran'
         },
+        {
+            'name': 'PrgEnv-cray',
+            'target_systems': [
+                'alps', 'eiger', 'rigi', 'pilatus'
```
vkarak, line 881: I would only put the systems that we have defined here.

```suggestion
                'eiger', 'pilatus'
```

config/cscs.py
```diff
@@ -773,6 +875,33 @@
...
+            ],
+            'modules': [
+                {'name': 'PrgEnv-cray', 'collection': True}
+            ]
+        },
+        {
+            'name': 'PrgEnv-gnu',
+            'target_systems': [
+                'alps', 'eiger', 'rigi', 'pilatus'
```
vkarak, line 890: Similarly here.

vkarak (changes requested):
I have only a couple of minor comments that I could fix myself.

jgphpc:
I want to add `mem-per-cpu` to the config and here is why we need it:

> scontrol show conf

`DefMemPerCPU = 1000`

ok ?

config/cscs.py
```diff
@@ -668,9 +668,9 @@
                             ]
                         },
                         {
-                            'name': 'mem-per-node',
+                            'name': 'memory',
                             'options': [
-                                '--mem={mem_per_node}'
```
vkarak: This you can still keep it as `mem_per_node`. The name of the resource can have multiple options, that's they should not necessarily be the same.

vkarak (approved):
lgtm

PR #1621, opened 2020-11-26
[test] Add HIP support for GPU microbenchmarks

jjotero:
The following tests have been modified to support both HIP and CUDA:

- kernel latency
- memory bandwidth
- shared memory bandwidth
- gpu_burn

These implementations use the Xdevice library that lives under `memory_bandwidth/src`. The other tests have symlinks to that directory.

Found a workaround to #1639 by including the symlinks as read only.


cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
```diff
@@ -99,21 +79,57 @@ def __init__(self):
...
-    @rfm.run_before('compile')
-    def set_nvidia_sm_arch(self):
-        nvidia_sm = '60'
-        if self.current_system.name in ['arolla', 'tsa', 'ault']:
-            nvidia_sm = '70'
+    @rfm.run_after('setup')
+    def select_makefile(self):
+        cp = self.current_partition.fullname
+        if cp == 'ault:amdvega':
+            self.prebuild_cmds = ['cp makefile_memoryBandwidth.hip Makefile']
```
vkarak, line 86: Check my comment above.

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
```diff
@@ -99,21 +79,57 @@ def __init__(self):
...
+        else:
+            self.prebuild_cmds = ['cp makefile_memoryBandwidth.cuda Makefile']
+
+    @rfm.run_after('setup')
+    def set_gpu_arch(self):
```
vkarak, line 91: I guess these functions will be moved to a "library test" as soon as we have have this functionality. How do you see that?

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
```diff
@@ -13,40 +13,20 @@ def __init__(self):
         self.valid_systems = ['daint:gpu', 'dom:gpu',
                               'arolla:cn', 'tsa:cn',
                               'ault:amdv100', 'ault:intelv100',
-                              'ault:amda100']
+                              'ault:amda100', 'ault:amdvega']
```
vkarak, line 16: I think that you are also missing a reference, don't you?

cscs-checks/microbenchmarks/gpu/shmem/shmem.py
```diff
@@ -48,3 +59,40 @@ def __init__(self):
...
+            self.build_system.cxxflags += [f'-arch=sm_{nvidia_sm}']
+            if cp in {'dom:gpu', 'daint:gpu'}:
+                self.modules += ['craype-accel-nvidia60']
+            else:
+                self.modules += ['cuda']
+
+        # Deal with the AMD options
+        amd_trgt = None
+        if cp == 'ault:amdvega':
+            amd_trgt = 'gfx906,gfx908'
```
vkarak, line 94: Question out of my ignorance: why do you need two targets?

vkarak (changes requested):

cscs-checks/microbenchmarks/gpu/kernel_latency/kernel_latency.py
```diff
@@ -70,6 +48,18 @@ def __init__(self, kernel_version):
...
+                'ault:intelv100': {
+                    'latency': (7.15, None, 0.10, 'us')
+                },
+                'ault:amdv100': {
+                    'latency': (7.15, None, 0.10, 'us')
+                },
+                'ault:amda100': {
+                    'latency': (9.65, None, 0.10, 'us')
+                },
+                'ault:amdv100': {
```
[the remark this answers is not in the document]
jjotero: Yup, that's what I meant.

cscs-checks/microbenchmarks/gpu/shmem/shmem.py
jjotero: Oh, there used to be a node in the `amdvega` partition with different AMD cards than the rest of the nodes. So that's just the product of me messing about with the target and forgetting to cleanup after Mark took that node to other partition.

cscs-checks/microbenchmarks/gpu/memory_bandwidth/src/Xdevice/common/smi.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+  /*
+   * Get the total memory size present in a device.
+   *
+   * Parameters:
+   *   + id: Device id.
+   *   + mSize: Size (in bytes) of the total memory.
+   *
+   */
+  void getDeviceMemorySize(int id, size_t * mSize);
```
vkarak, line 79: I would probably mark these as `const`. Btw, why don't you return the size and pass it as a pointer instead?

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
vkarak: No, sorry, I was confused. I thought you haven't put references for all the systems, but I see they are there.

vkarak (changes requested):

cscs-checks/microbenchmarks/gpu/memory_bandwidth/src/Xdevice/common/smi.hpp
jjotero: All these functions are void just to be consistent with both ROCm-SMI and NVML implementations. The functions in these libraries return a status, not the requested value. Actually, I should probably also change the getGpuTemp to be consistent with the others.

jjotero:
- Added the `gpu_burn` case to the PR.
-  Implemented a workaround to the symlink issue (see above) by marking the symlinks as `self.readonly_files` in the reframe test.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
```diff
@@ -39,553 +42,369 @@
...
+      XMalloc((void**)&d_A, d_resultSize);
+      XMalloc((void**)&d_B, d_resultSize);
+      XMalloc((void**)&d_numberOfErrors, sizeof(int));
+
+      // Populating matrices A and B
+      XMemcpy(d_A, h_A, d_resultSize, XMemcpyHostToDevice);
+      XMemcpy(d_B, h_B, d_resultSize, XMemcpyHostToDevice);
+  }
+
+  void compute()
```
vkarak, line 163: I would only define this in the specializations and let it fail in compilation, or you could use a static assert.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
```diff
@@ -39,553 +42,369 @@
...
+            return -1;
 
-// Returns the number of devices
-int initCuda() {
-    checkError(cuInit(0));
-    int deviceCount = 0;
-    checkError(cuDeviceGetCount(&deviceCount));
+        // Get the time difference and return the flops
+        std::chrono::duration<double> diff = end-start;
+        double Gflops = (double)(iters*reps) * ((double)(OPS_PER_MUL)/diff.count()/1000.0/1000.0/1000.0);
```
vkarak, line 273: The following, I think, should be enough. The operations will be promoted to double due to `diff.count()`.

```suggestion
        double Gflops = 1e-9 * iters * reps * OPS_PER_MUL * diff.count();
```

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
```diff
@@ -39,553 +42,369 @@
...
 
-    // kill(tempPid, 15);
-    // close(tempHandle);
+    // Print device count
+    printf("[%s] Found %d device(s).\n", hostname, devCount);
 
-    while (wait(NULL) != -1);
-    printf("Node %s:\n", hostname);
+    // All the burn info is stored in instances of the BurnTracker class.
+    BurnTracker * trackThreads = new BurnTracker[devCount];
```
vkarak, line 360: I think you are leaking this, because you are constructing again `BurnTracker`s in the loop below with the placement new. Either make this an array of pointers to `BurnTracker`s or dummy construct them here and initialize them with an explicit method that you would call below.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
```diff
@@ -39,553 +42,369 @@
...
-    T *A = (T*) malloc(sizeof(T)*SIZE*SIZE);
-    T *B = (T*) malloc(sizeof(T)*SIZE*SIZE);
-    srand(10);
-    for (size_t i = 0; i < SIZE*SIZE; ++i) {
-        A[i] = (T)((double)(rand()%1000000)/100000.0);
-        B[i] = (T)((double)(rand()%1000000)/100000.0);
+    // Hold until all the threads are done with the init.
+    {
+        std::unique_lock<std::mutex> lk(cv_m);
+        cv.wait(lk, []{return startUpCounter == devCount;});
```
vkarak, line 377: Perhaps you can replace those with barriers in C++20.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
```diff
@@ -39,553 +42,369 @@
...
-                    clientPids.push_back(slavePid);
-                    close(slavePipe[1]);
-                }
-            }
-            listenClients(clientPipes, clientPids, runLength);
-        }
+    // Burn-time.
+    burn = true;
+    cv.notify_all();
+    std::this_thread::sleep_for( std::chrono::seconds(duration) );
```
vkarak, line 383: What's the purpose of this sleep?

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
```diff
@@ -39,553 +42,369 @@
...
-        clientErrors.push_back(0);
-        clientCalcs.push_back(0);
-        struct timespec thisTime;
-        clock_gettime(CLOCK_REALTIME, &thisTime);
-        clientUpdateTime.push_back(thisTime);
-        clientGflops.push_back(0.0f);
-        clientFaulty.push_back(false);
-    }
+    // Initialise the SMI
+    Smi * smi_handle = new Smi();
```
vkarak, line 348: Couldn't you simply have `Smi smi_handle` here?

cscs-checks/microbenchmarks/gpu/shmem/shmem.py
```diff
@@ -48,3 +60,63 @@ def __init__(self):
     @sn.sanity_function
     def num_tasks_assigned(self):
         return self.job.num_tasks
+
+    @sn.sanity_function
+    def assert_num_gpus(self):
```
vkarak, line 65: We should be consistent with these names. In other places we call it `assert_count_gpus()`.

vkarak (changes requested):

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
jjotero: Yes - forgot to clean that up.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
jjotero: That is what sets the duration of the burn time. After that sleep, the main thread sets `burn=false` which triggers the work threads to finish up their work.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
jjotero: Wouldn't this be a bit early for systems like Daint where GCC there is at 8.2?

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
jjotero: Fixed. I've gone for the array of pointers approach.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
jjotero: I've gone for `void compute() = delete;`

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -10,39 +10,24 @@
...
-        self.build_system.ldflags = ['-lcuda', '-lcublas', '-lnvidia-ml']
-        self.sanity_patterns = sn.assert_eq(
-            sn.count(sn.findall('OK', self.stdout)), self.num_tasks_assigned)
-
-        patt = r'GPU\s+\d+\(\S*\): (?P<perf>\S*) GF\/s  (?P<temp>\S*) Celsius'
+        self.exclusive_access = True
+        self.executable_opts = ['-d', '40']
+        self.build_system = 'Make'
+        self.executable = './gpu_burn.x'
+        self.readonly_files = ['Xdevice']
```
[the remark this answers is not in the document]
jjotero: Added that to the `shmem` and `kernel_latency` tests too.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
vkarak: Yes, leave it as it is now. It's just a comment for the future.

cscs-checks/microbenchmarks/gpu/gpu_burn/src/gpu_burn.cu
vkarak: Yes, better.

vkarak (approved):
lgtm

PR #1623, opened 2020-11-26
[feat] Allow using colon separated paths with the `--module-path` option

vkarak:
This is only the case when neither the `+` nor the `-` prefixes are used. The reason for this enhancement is to allow an easier manipulation of the path through. The unit tests are also adapted.

This blocks #1538, since the unit tests are broken in master on Dom.

### DIFF
--- a/unittests/test_cli.py
+++ b/unittests/test_cli.py
@@ -3,14 +3,14 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+import contextlib
+import io
 import itertools
 import os
 import pathlib
 import pytest
 import re
 import sys
-from contextlib import redirect_stdout, redirect_stderr, suppress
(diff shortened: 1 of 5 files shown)


unittests/test_cli.py
```diff
@@ -632,6 +632,11 @@ def test_overwrite_module_path(run_reframe, user_exec_ctx):
         pytest.skip('no modules system found')
 
     module_path = 'unittests/modules'
+    try:
+        module_path += f':{os.environ["MODULEPATH"]}'
```
teojgo, line 636: Do this with the `suppress`

teojgo (changes requested):

teojgo (approved):

vkarak:
@jenkins-cscs retry none

PR #1624, opened 2020-11-27
[feat] Support `spack load` for loading software packages

### LINKED ISSUE #1584: Support `spack load` for loading software packages

rsarm:
Closes #1584


reframe/core/modules.py
```diff
@@ -109,6 +109,8 @@ def create(cls, modules_kind=None):
             return ModulesSystem(TMod4Impl())
         elif modules_kind == 'lmod':
             return ModulesSystem(LModImpl())
+        elif modules_kind == 'slack':
```
vkarak, line 112: I guess you mean "spack" here. :-)

reframe/core/modules.py
```diff
@@ -939,3 +941,111 @@ def emit_load_instr(self, module):
...
+        # try:
+        #     # Try the Python bindings now
+        #     completed = osext.run_command(self.modulecmd())
+        # except OSError as e:
+        #     raise ConfigError(
+        #         'could not get the Python bindings for TMod: ' % e) from e
+
+        # if re.search(r'Unknown shell type', completed.stderr):
+        #     raise ConfigError(
+        #         'Python is not supported by this TMod installation')
```
vkarak, line 969: @victorusu If there are no Python bindings for `spack load`, then the only thing we can do for this backend is to emit the right commands and list the loaded modules.

reframe/core/modules.py
```diff
@@ -939,3 +941,111 @@ def emit_load_instr(self, module):
...
+        return ' '.join(['spack', *args])
+
+    def _execute(self, cmd, *args):
+        modulecmd = self.modulecmd(cmd, *args)
+        completed = osext.run_command(modulecmd)
+        if re.search(r'Error', completed.stderr) is not None:
+            raise SpawnedProcessError(modulecmd,
+                                      completed.stdout,
+                                      completed.stderr,
+                                      completed.returncode)
```
vkarak, line 987: You should simply check the error code. Tmod32 is just misbehaving not returning a proper exit code when it fails, that's why we do this hack. In your case, you should simply execute `run_command()` with `check=True`.

reframe/core/modules.py
```diff
@@ -939,3 +941,111 @@ def emit_load_instr(self, module):
...
+    def available_modules(self, substr):
+        output = self.execute('find', '--format', '{name}-{version}-{hash}',
+                              substr)
+        ret = []
+        for line in output.split('\n'):
+            if not line or line[-1] == ':':
+                # Ignore empty lines and path entries
+                continue
+
+            ret.append(Module(line))
```
vkarak, line 1000: Are you should that this is going to be parsed correctly by `Module`? I think that `Module` expects module names as `name/version`.

reframe/core/modules.py
```diff
@@ -939,3 +941,111 @@ def emit_load_instr(self, module):
...
+                output = self.execute('find', '--loaded', '--format',
+                                      '{name}-{version}-{hash}', f'/{hashes}')
+                return [Module(m) for m in output.split('\n') if m]
+            else:
+                return []
+        except KeyError:
+            return []
+
+    def conflicted_modules(self, module):
+        pass
```
vkarak, line 1022: This should return `[]`.

vkarak: Generally, don't simply do `pass` in any of these functions. Check what return type the callers expect.

reframe/core/modules.py
```diff
@@ -939,3 +941,111 @@ def emit_load_instr(self, module):
...
+        try:
+            hashes = '/'.join(
+                [m for m in os.environ['SPACK_LOADED_HASHES'].split(':') if m])
+            if hashes:
+                output = self.execute('find', '--loaded', '--format',
+                                      '{name}-{version}-{hash}', f'/{hashes}')
+                return [Module(m) for m in output.split('\n') if m]
+            else:
+                return []
+        except KeyError:
```
vkarak, line 1018: `KeyError` for what? If it's for the environment variable use `getenv` and do the following:

```python
hashes = '/'.join(m for m in os.getenv('SPACK_LOADED_HASHES', '').split(':') if m)
if not hashes:
    return []

# Now do your stuff
```

reframe/core/modules.py
```diff
@@ -939,3 +941,111 @@ def emit_load_instr(self, module):
...
+                # Ignore empty lines and path entries
+                continue
+
+            ret.append(Module(line))
+
+        return ret
+
+    def _module_full_name(self, name):
+        return self.execute('find', '--format', '{name}-{version}-{hash}',
+                            name)
```
vkarak, line 1006: I'm not sure about this format's interoperability with `Module`.

vkarak (changes requested):

reframe/core/modules.py
victorusu: yes

ci-scripts/dockerfiles/Spack.dockerfile
```diff
@@ -0,0 +1,42 @@
+#
+# Execute this from the top-level ReFrame source directory
+#
+
+FROM ubuntu:20.04
+
+ENV TZ=Europe/Zurich
+ENV DEBIAN_FRONTEND=noninteractive
+ENV _TMOD_VER=4.6.0
```
vkarak, line 9: I don't think you need this.

ci-scripts/dockerfiles/Spack.dockerfile
```diff
@@ -0,0 +1,42 @@
...
+  apt-get -y update && \
+  apt-get -y install ca-certificates && \
+  update-ca-certificates && \
+  apt-get -y install gcc && \
+  apt-get -y install make && \
+  apt-get -y install git && \
+  apt-get -y install python3 python3-pip
+
+# Required utilities
+RUN apt-get -y install wget
```
vkarak, line 25: Neither this one.

ci-scripts/dockerfiles/Spack.dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Install ReFrame from the current directory
+COPY --chown=rfmuser . /home/rfmuser/reframe/
+
+USER rfmuser
+
+# Install Spack
+RUN git clone https://github.com/spack/spack ~/spack && \
+    cd ~/spack && \
+    git checkout releases/v0.16 && \
```
vkarak, line 35: Don't hard code the version here, but use an environment variable as `_TMOD_VER`.

ci-scripts/dockerfiles/Spack.dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    echo 'source ~/spack/share/spack/setup-env.sh' >> ~/.bashrc
```
vkarak, line 36: No, use `BASH_ENV`. Check the other Dockerfiles. `.bashrc` is not sourced if the container is not executed interactively.

reframe/core/modules.py
vkarak: @rsarm You need to remove these commented lines.

reframe/core/modules.py
```diff
@@ -1002,41 +997,38 @@ def available_modules(self, substr):
         return ret
 
     def _module_full_name(self, name):
-        return self.execute('find', '--format', '{name}-{version}-{hash}',
+        return self.execute('find', '--format', self._name_format,
                             name)
```
vkarak: Everything fits in a single line.

reframe/core/modules.py
```diff
@@ -1002,41 +997,38 @@ def available_modules(self, substr):
...
-                [m for m in os.environ['SPACK_LOADED_HASHES'].split(':') if m])
-            if hashes:
-                output = self.execute('find', '--loaded', '--format',
-                                      '{name}-{version}-{hash}', f'/{hashes}')
-                return [Module(m) for m in output.split('\n') if m]
-            else:
-                return []
-        except KeyError:
+        hashes = '/'.join(
+            [m for m in os.environ['SPACK_LOADED_HASHES'].split(':') if m])
```
vkarak, line 1005:
```suggestion
        hashes = '/'.join(
            m for m in os.environ['SPACK_LOADED_HASHES'].split(':') if m
        )
```

reframe/core/modules.py
```diff
@@ -1002,41 +997,38 @@ def available_modules(self, substr):
...
+        if hashes:
+            output = self.execute('find', '--loaded', '--format',
+                                  self._name_format, f'/{hashes}')
```
vkarak, line 1008: This produces this output `/hash1/hash2/.../hashn`. Is this valid? @victorusu ?

reframe/core/modules.py
```diff
@@ -1002,41 +997,38 @@ def available_modules(self, substr):
...
+        return []
 
     def is_module_loaded(self, module):
         module = self._module_full_name(module.name)
         module = Module(module)
         return module in self.loaded_modules()
 
     def load_module(self, module):
-        pass
+        self.execute('load', '--sh', str(module))
```
vkarak, line 1022: This has no effect. This will produce the shell commands to load the module. This is not what we want by a module system. We would want the Python bindings, but Spack does not offer any, so we should simply return the same as the `nomod` backend here.

reframe/core/modules.py
```diff
@@ -1002,41 +997,38 @@ def available_modules(self, substr):
...
 
     def unload_module(self, module):
-        pass
+        self.execute('unload', '--sh', str(module))
```
vkarak, line 1025: See my comment above.

reframe/core/modules.py
```diff
@@ -1002,41 +997,38 @@ def available_modules(self, substr):
...
 
     def unload_all(self):
-        pass
+        self.execute('unload', '--sh', '--all', str(module))
```
vkarak, line 1028: See my comment above.

vkarak (changes requested):
We will also need unit tests and documentation.

reframe/core/modules.py
rsarm: Yes.  Module names can be passed to Spack as hashes, the syntax is `/<hash>`. For instance, this converts from hashes, which is how they are stored on the `SPACK_LOADED_HASHES` variable, to module names:
```
$> spack find --loaded --format '{name}/{version}' /vcsrq3vgchevv2weu3cfwe37jclspuuv/injxlcmccya6nslhqf7chwsh4e5hpgcp/zeyd2m5zugdxbnx5mepvbmfnht4tjrqr/oxpxbrpqqca5hjh6blb4qv5chqqv3ykv/sl7m27mzkbejtkrajigj3a3m37ygv4u2/4kcnj3oypwyyr3o46ipejwuk3x5gzrar/af5tdk6ilv6mah2ntgb5odryvlosijnz/fueeuyvb37exbhwxkgte4kurdjyxli5u
hwloc/1.11.11  libpciaccess/0.16  numactl/2.0.14  xz/5.2.5
libiconv/1.16  libxml2/2.9.10     openmpi/3.1.6   zlib/1.2.11
```
It doesn't need spaces between hashes.

rsarm: I just realized that the env variable is no needed, I think that was the initial way I was doing it. With `--loaded` it gives the loaded modules directly.

docs/config_reference.rst
```diff
@@ -115,6 +115,7 @@ System Configuration
    - ``tmod32``: A synonym of ``tmod``.
    - ``tmod4``: The `new environment modules <http://modules.sourceforge.net/>`__ implementation (versions older than 4.1 are not supported).
    - ``lmod``: The `Lua implementation <https://lmod.readthedocs.io/en/latest/>`__ of the environment modules.
+   - ``Spack``: `Spack <https://spack.readthedocs.io/en/latest/>`'s built-in mechanism for managing modules.
```
vkarak, line 118:
```suggestion
   - ``spack``: `Spack <https://spack.readthedocs.io/en/latest/>`'s built-in mechanism for managing modules.
```

vkarak: We should also mention that the Spack support is new in 3.4.

unittests/test_modules.py
```diff
@@ -35,7 +35,7 @@ def modules_system(request, monkeypatch):
 
 
 def test_searchpath(modules_system):
-    if modules_system.name == 'nomod':
+    if modules_system.name in ['nomod', 'spack']:
```
vkarak, line 38: @victorusu Are we sure that Spack does not support any sort of module path concept?

reframe/core/modules.py
```diff
@@ -997,18 +985,12 @@ def available_modules(self, substr):
         return ret
 
     def _module_full_name(self, name):
-        return self.execute('find', '--format', self._name_format,
-                            name)
+        return self.execute('find', '--format', self._name_format, name)
 
     def loaded_modules(self):
```
vkarak, line 990: This function is untested for the Spack backend...

vkarak (changes requested):
Can you also fix the unused import error?

reframe/core/modules.py
rsarm: Initially `_module_full_name` was needed several times but it ended up being used only once. Maybe we get rid of it?

vkarak: Yes, we can get rid of it.

vkarak (approved):

PR #1635, opened 2020-12-02
[doc] Remove superlinter badge from README

teojgo:


vkarak (approved):
Lgtm, but did you understand why it doesn't work?

teojgo:
> Lgtm, but did you understand why it doesn't work?

Yes it would need to be in a separate workflow to work. But I don't think it offers anything since the `ReFrame CI` badge now includes also the linting.

vkarak:
@jenkins-cscs retry dom

PR #1636, opened 2020-12-02
[feat] Drop support of pipeline stage method override

### LINKED ISSUE #1352: Drop support of test syntax that overrides `RegressionTest` methods

teojgo:
Fixes #1352

### DIFF
--- a/docs/migration_2_to_3.rst
+++ b/docs/migration_2_to_3.rst
@@ -65,8 +65,7 @@ Pipeline methods and hooks
 
 ReFrame 2.20 introduced a new powerful mechanism for attaching arbitrary functions hooks at the different pipeline stages.
 This mechanism provides an easy way to configure and extend the functionality of a test, eliminating essentially the need to override pipeline stages for this purpose.
-ReFrame 3.0 deprecates the old practice of overriding pipeline stage methods in favor of using pipeline hooks.
-From ReFrame 3.4, overriding pipeline stage methods is no longer allowed.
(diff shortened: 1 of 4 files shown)


docs/migration_2_to_3.rst
```diff
@@ -66,6 +66,7 @@ Pipeline methods and hooks
 ReFrame 2.20 introduced a new powerful mechanism for attaching arbitrary functions hooks at the different pipeline stages.
 This mechanism provides an easy way to configure and extend the functionality of a test, eliminating essentially the need to override pipeline stages for this purpose.
 ReFrame 3.0 deprecates the old practice of overriding pipeline stage methods in favor of using pipeline hooks.
+From ReFrame 3.4, overriding pipeline stage methods is no longer allowed.
```
vkarak, line 69:
```suggestion
ReFrame 3.0 deprecates the old practice of overriding pipeline stage methods in favor of using pipeline hooks and ReFrame 3.4 disables that by default.
```

reframe/core/pipeline.py
```diff
@@ -1092,6 +1092,12 @@ def setup(self, partition, environ, **job_opts):
               <migration_2_to_3.html#force-override-a-pipeline-method>`__ for
               more details.
 
+           .. versionchanged:: 3.4
+              Overriding this method directly unless you are in
+              special test. See `here
+              <migration_2_to_3.html#force-override-a-pipeline-method>`__ for
+              more details.
```
vkarak, line 1099: I don't think that this adds anything. I would suggest to remove. Just add an "a" here: "unless you are in a special test."

vkarak (approved):
lgtm

PR #1637, opened 2020-12-02
[bugfix] Fix detailed listing message for node allocation scheme

### LINKED ISSUE #1520: Wrong node allocation description
When using the `-L/--list-detailed` in order to get detailed information about one of my self-backed tests, ReFrame reports the following:
```bash
...
    Node allocation:
      standard (1 task(s))
... 
```
This is misleading information because I have the following seam in my code
```
...
    @rfm.run_after('setup')
    def set_num_tasks(self):
        if self.current_partition.fullname in ['daint:gpu', 'dom:gpu']:
            self.num_tasks = 72
            self.num_tasks_per_node = 12
            self.num_cpus_per_task = 1
...
```

I do not know what to do here, because I understand that the information is taken from after `init`.
And if ReFrame executes all steps up to `compile` (excluding `compile`) to display the information, I could still write:
```
...
    @rfm.run_after('compile')
    def set_num_tasks(self):
        if self.current_partition.fullname in ['daint:gpu', 'dom:gpu']:
            self.num_tasks = 72
            self.num_tasks_per_node = 12
            self.num_cpus_per_task = 1
...
```

One cannot execute compile to give me the `-L` flag because in this particular case, for instance, it takes 20 min to compile the test... 😄 

My proposal is to either remove this entry from `-L` or change the wording, in case, there is a hook for any other phase after `init`. An example would be
```
...
    Node allocation (assumed):
      standard (1 task(s))
... 
```

The word *assumed* has to be replaced by a better word, but you get the idea.

vkarak:
Fixes #1520.

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -56,7 +56,8 @@ def format_check(check, check_deps, detailed=False):
         return f'- {check.name} (found in {location!r})'
 
     if check.num_tasks > 0:
-        node_alloc_scheme = f'standard ({check.num_tasks} task(s))'
+        node_alloc_scheme = (f'standard ({check.num_tasks} task(s) -- '
+                             f'may be set differently in hooks)')
     elif check.num_tasks == 0:
         node_alloc_scheme = 'flexible'
     else:


victorusu (approved):
`lgtm`

PR #1644, opened 2020-12-05
[bugfix] Send test attributes correctly with the Graylog handler

### LINKED ISSUE #1640: Sending data to Graylog is broken
Currently, when using a Graylog handler, no data is transferred. This has been introduced with #1600 and specifically with commit: b30b2509527dd06404c7aa17186d72ed5d1bb5a7

vkarak:
The implementation of the dynamic log record based on arbitrary test attributes is revised completely. It does not happen anymore in the formatter, because the formatter is not used by the Graylog handler and this was the reason for #1640. All public test attributes as well as some interesting properties are put in the logger's `extra` dict. This will be transferred to the log record by Python's logging mechanism. At this point, we do NOT format anything, except the `check_job_completion_time`, in order to be sent correctly to Graylog. The formatting of the record is an exclusive duty of the formatter object, which formats the final log message through its `formatMessage()` method. We don't format the record attributes directly, but we use a proxy dictionary, which holds the formatted values, in order to produce the final message from the format string. All attributes, except lists, tuples, sets and strings are formatted using our `jsonext.dumps()` function. For compatibility with the current behaviour, lists, tuples and sets are formatted as comma-separated lists, whereas strings are passed as-is (NB: if they were JSON formatted, they would be quoted). This allows users to use the `%(check_perf_patterns)s` format specifier in order to get all the performance variables and values at once as a JSON object. If a requested test attribute cannot be found, it will be logged as `null`.

The Graylog handler sends to the remote end the full log record converted to JSON. Again here, we use our JSON encoding method. There are some limitations to this. If a field is logged as `null` it will not be indexed and it will not show up in the Kibana logs. Also, for some reason that I cannot understand, the same happens for boolean fields. It would be rather complicated to format these fields as strings and transfer them, without messing with the log record, so I chose not to do it, because in principle it's not a handler's responsibility to format a log record.

Fixes #1640.

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import abc
-import collections.abc
 import logging
 import logging.handlers
 import numbers
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -5,6 +5,7 @@
 
 import abc
 import collections.abc
+import functools
 import logging
 import logging.handlers
 import numbers
@@ -16,6 +17,7 @@ import socket
 import time
 
 import reframe.utility.color as color
+import reframe.utility.jsonext as jsonext
 import reframe.utility.osext as osext
 from reframe.core.exceptions import ConfigError, LoggingError
 
@@ -144,7 +146,8 @@ class MultiFileHandler(logging.FileHandler):
         except OSError as e:
             raise LoggingError('logging failed') from e
 
-        self.baseFilename = os.path.join(dirname, record.check.name + '.log')
+        self.baseFilename = os.path.join(
+            dirname, record.__rfm_check__.name + '.log')
         self.stream = self._streams.get(self.baseFilename, None)
         super().emit(record)
         self._streams[self.baseFilename] = self.stream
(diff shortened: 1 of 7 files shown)


jgphpc (approved):
lgtm

reframe/core/logging.py
```diff
@@ -144,7 +144,8 @@ def emit(self, record):
         except OSError as e:
             raise LoggingError('logging failed') from e
 
-        self.baseFilename = os.path.join(dirname, record.check.name + '.log')
+        self.baseFilename = os.path.join(dirname,
```
teojgo, line 147:
```suggestion
        self.baseFilename = os.path.join(
            dirname, f'{record.__rfm_check__.name}.log'
        )
```

reframe/core/pipeline.py
```diff
@@ -1842,12 +1856,18 @@ def setup(self, partition, environ, **job_opts):
     @property
     @sn.sanity_function
     def stdout(self):
-        return self._build_job.stdout
+        if self.build_job is None:
+            return None
+
+        return self.build_job.stdout
```
teojgo, line 1862: Do this in one line

reframe/core/pipeline.py
```diff
@@ -1842,12 +1856,18 @@ def setup(self, partition, environ, **job_opts):
...
 
     @property
     @sn.sanity_function
     def stderr(self):
-        return self._build_job.stderr
+        if self.build_job is None:
+            return None
+
+        return self.build_job.stderr
```
teojgo, line 1870: Do this in one line

reframe/core/pipeline.py
```diff
@@ -964,12 +972,18 @@ def build_job(self):
...
+        return self.build_job.stdout
 
     @property
     @sn.sanity_function
     def build_stderr(self):
-        return self._build_job.stderr
+        if self.build_job is None:
+            return None
+
+        return self.build_job.stderr
```
teojgo, line 986: Do this in on line

reframe/core/pipeline.py
```diff
@@ -964,12 +972,18 @@ def build_job(self):
     @property
     @sn.sanity_function
     def build_stdout(self):
-        return self._build_job.stdout
+        if self.build_job is None:
+            return None
+
+        return self.build_job.stdout
```
teojgo, line 978: Do this in one line

reframe/core/pipeline.py
```diff
@@ -953,9 +957,13 @@ def stderr(self):
...
 
-        :type: :class:`str`.
+        :type: :class:`str` or :class:`None` if a run job has not yet been
+            created.
         '''
-        return self._job.stderr
+        if self.job is None:
+            return None
+
+        return self.job.stderr
```
teojgo, line 966:
```suggestion
       return self.job.stderr if self.job else None
```

reframe/core/pipeline.py
```diff
@@ -939,9 +939,13 @@ def stdout(self):
...
 
-        :type: :class:`str`.
+        :type: :class:`str` or :class:`None` if a run job has not yet been
+            created.
         '''
-        return self._job.stdout
+        if self.job is None:
+            return None
+
+        return self.job.stdout
```
teojgo, line 948:
```suggestion
    return self.job.stdout if self.job else None
```

teojgo (changes requested):

reframe/core/pipeline.py
victorusu: I agree with @teojgo

reframe/core/logging.py
vkarak: The way I have it is equivalent and still fits in a single line.

teojgo (approved):
lgtm now

PR #1653, opened 2020-12-12
[test] Extend GPU burn test to report GPU node with smallest flops

jgphpc:
GpuBurnTest
- dom:gpu
   - PrgEnv-gnu
      * num_tasks: 22
      * perf: 4091.0 Gflop/s
      * smallest_flops: 4091.0 GF/s (nid00008)

Goal is to report nidname when the test fails to meet perf. reference 
`'perf': (4115, -0.10, None, 'Gflop/s')`
but I am not sure how to do that... Feel free to push 
For now, replacing `perf:` with the last line gives some info.

### DIFF
--- a/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
+++ b/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import os
-import sys
 
 import reframe as rfm
 import reframe.utility.sanity as sn
--- a/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
+++ b/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import os
+import sys
 
 import reframe as rfm
 import reframe.utility.sanity as sn
@@ -142,20 +143,19 @@ class GpuBurnTest(rfm.RegressionTest):
         rptf = os.path.join(self.stagedir, sn.evaluate(self.stdout))
         self.nids = sn.extractall(regex, rptf, 1)
         self.flops = sn.extractall(regex, rptf, 2, float)
-        index = -1
-        flops_min = sn.min(self.flops)
-        for ii in range(len(sn.evaluate(self.flops))):
-            if self.flops[ii] == flops_min:
-                index = ii
-                break
+        flops_min, index = sys.float_info.max, None
+        for i, flops in enumerate(self.flops):
+            if flops < flops_min:
+                flops_min, index = flops, i
 
         self.unit = f'GF/s ({self.nids[index]})'
+        unit = (0, None, None, self.unit)
         self.perf_patterns['smallest_flops'] = flops_min
-        self.reference['dom:gpu:smallest_flops'] = (0, None, None, self.unit)
-        self.reference['daint:gpu'] = (0, None, None, self.unit)
-        self.reference['arolla:cn'] = (0, None, None, self.unit)
-        self.reference['tsa:cn'] = (0, None, None, self.unit)
-        self.reference['ault:amda100'] = (0, None, None, self.unit)
-        self.reference['ault:amdv100'] = (0, None, None, self.unit)
-        self.reference['ault:intelv100'] = (0, None, None, self.unit)
-        self.reference['ault:amdvega'] = (0, None, None, self.unit)
+        self.reference['dom:gpu:smallest_flops'] = unit
+        self.reference['daint:gpu:smallest_flops'] = unit
+        self.reference['arolla:cn:smallest_flops'] = unit
+        self.reference['tsa:cn:smallest_flops'] = unit
+        self.reference['ault:amda100:smallest_flops'] = unit
+        self.reference['ault:amdv100:smallest_flops'] = unit
+        self.reference['ault:intelv100:smallest_flops'] = unit
+        self.reference['ault:amdvega:smallest_flops'] = unit
--- a/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
+++ b/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+import os
+
 import reframe as rfm
 import reframe.utility.sanity as sn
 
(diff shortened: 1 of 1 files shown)


jjotero (commented):
How about adding an optional step after `performance` with something like `self.perf_error_patterns`? This would process the output and collect all the information on how and where the test failed. In that way, this additional step would only run if the performance fails, and you would avoid having redundant output when the test passes performance.

vkarak:
@jjotero Is this a general feature request?

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -133,3 +135,27 @@ def set_gpus_per_node(self):
...
+        self.unit = f'GF/s ({self.nids[index]})'
+        self.perf_patterns['smallest_flops'] = flops_min
+        self.reference['dom:gpu:smallest_flops'] = (0, None, None, self.unit)
+        self.reference['daint:gpu'] = (0, None, None, self.unit)
+        self.reference['arolla:cn'] = (0, None, None, self.unit)
+        self.reference['tsa:cn'] = (0, None, None, self.unit)
+        self.reference['ault:amda100'] = (0, None, None, self.unit)
+        self.reference['ault:amdv100'] = (0, None, None, self.unit)
+        self.reference['ault:intelv100'] = (0, None, None, self.unit)
+        self.reference['ault:amdvega'] = (0, None, None, self.unit)
```
vkarak, line 161: I don't think these are needed.

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -133,3 +135,27 @@ def set_gpus_per_node(self):
...
+        regex = r'\[(\S+)\] GPU\s+\d\(OK\): (\d+) GF/s'
+        rptf = os.path.join(self.stagedir, sn.evaluate(self.stdout))
+        self.nids = sn.extractall(regex, rptf, 1)
+        self.flops = sn.extractall(regex, rptf, 2, float)
+        index = -1
+        flops_min = sn.min(self.flops)
+        for ii in range(len(sn.evaluate(self.flops))):
+            if self.flops[ii] == flops_min:
+                index = ii
+                break
```
vkarak, line 150:
```suggestion
        flops_min, index = sys.float_info.max, None
        for i, flops in enumerate(self.flops):
            if flops < flops_min:
                flops_min, index = flops, i
```

vkarak (changes requested):

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
jgphpc: :+1: 
I can't manage to call `self.flops.index(min(self.flops))`.
What's the right way to call a deferred function ?

vkarak: `index` is no method of `self.flops`, is it?

jgphpc:
> you would avoid having redundant output when the test passes performance.
the test always report the smallest flops, I only want to always add the nidname, so the report will not be much different from what we have now:
`perf: 4091.0 GF/s (nid00008)` 
However having `self.perf_error_patterns` sounds nice.

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
vkarak: If you want to call `index()`, you would have to do it like this: `self.flops.evaluate().index(min(self.flops))`.

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -142,20 +143,19 @@ def report_smallest_node(self):
...
         self.perf_patterns['smallest_flops'] = flops_min
-        self.reference['dom:gpu:smallest_flops'] = (0, None, None, self.unit)
-        self.reference['daint:gpu'] = (0, None, None, self.unit)
-        self.reference['arolla:cn'] = (0, None, None, self.unit)
-        self.reference['tsa:cn'] = (0, None, None, self.unit)
-        self.reference['ault:amda100'] = (0, None, None, self.unit)
-        self.reference['ault:amdv100'] = (0, None, None, self.unit)
-        self.reference['ault:intelv100'] = (0, None, None, self.unit)
-        self.reference['ault:amdvega'] = (0, None, None, self.unit)
+        self.reference['dom:gpu:smallest_flops'] = unit
```
vkarak, line 154: I think if you only put a single entry, it will print out the unit for all systems.

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
jgphpc: pushed

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -37,35 +37,35 @@ def __init__(self):
 
         self.reference = {
             'dom:gpu': {
-                'perf': (4115, -0.10, None, 'Gflop/s'),
+                # 'perf': (4115, -0.10, None, 'Gflop/s'),
```
vkarak, line 40: I prefer this here.

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -137,25 +137,19 @@ def set_gpus_per_node(self):
...
-        self.perf_patterns['smallest_flops'] = flops_min
-        self.reference['dom:gpu:smallest_flops'] = (0, None, None, self.unit)
-        self.reference['daint:gpu'] = (0, None, None, self.unit)
-        self.reference['arolla:cn'] = (0, None, None, self.unit)
-        self.reference['tsa:cn'] = (0, None, None, self.unit)
-        self.reference['ault:amda100'] = (0, None, None, self.unit)
-        self.reference['ault:amdv100'] = (0, None, None, self.unit)
-        self.reference['ault:intelv100'] = (0, None, None, self.unit)
-        self.reference['ault:amdvega'] = (0, None, None, self.unit)
+        self.reference['dom:gpu:perf'] = (4115, -0.10, None, self.unit)
```
vkarak, line 148: Now I'm really confused with this PR!

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
vkarak: All I meant is that this is enough. You don't need to set the reference here and no need to touch the references in the `__init__()`.

jjotero:
> @jjotero Is this a general feature request?

I guess it could be. Should I open an issue and move the discussion over there?

vkarak:
> I guess it could be. Should I open an issue and move the discussion over there?

Yes, it would make sense.

vkarak (approved):
@jgphpc I fixed the PR.

PR #1655, opened 2020-12-14
[ci] Ensure that module paths are restored correctly in unit tests

teojgo:


unittests/test_cli.py
```diff
@@ -673,7 +674,7 @@ def test_use_module_path(run_reframe, user_exec_ctx):
         config_file=fixtures.USER_CONFIG_FILE, action='run',
         system=rt.runtime().system.name
     )
-
+    ms.searchpath_remove(module_path)
```
vkarak, line 677: Here we need a comment explaining why we are doing that, because it's counter-intuitive.

vkarak (changes requested):
I have a minor comment, which I can fix.

unittests/test_cli.py
vkarak: I am having second thoughts about this. I don't think it's needed, because `run_reframe()` restores the environment to after it finishes. Can you check if this is needed? Similarly, for the `test_overwrite_module_path()`. In the `test_unuse_module_path()`, it is needed, however.

vkarak (approved):

PR #1656, opened 2020-12-14
[bugfix] Restore module paths specified in command line when restoring module collections

### LINKED ISSUE #1631: module-path=+<path> conflict with module collections
I believe there is a conflict between `--module-path` and lmod `module restore`.
This check:

```python
import reframe as rfm


@rfm.simple_test
class HelloTest(rfm.CompileOnlyRegressionTest):
    def __init__(self):
        self.valid_systems = ['pilatus:login']
        self.sourcepath = 'hello.c'
        self.modules = ['ddt']
```

when run with:
> ~/reframe.git/bin/reframe -C ~/reframe.git/config/cscs.py -c jg.py -r -p PrgEnv-cray \
`--module-path=+/apps/eiger/UES/tools/easybuild/modules/all`

fails with (attached log):
```
  * Reason: environ error: could not execute module operation: command '/usr/share/lmod/lmod/libexec/lmod python show ddt' failed with exit code 1:
```

teojgo:
Fixes #1631

### DIFF
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -729,8 +729,7 @@ class TMod4Impl(TModImpl):
                 (version, self.MIN_VERSION))
 
         self._version = version
-        self._module_paths_added = []
-        self._module_paths_removed = []
+        self._extra_module_paths = []
 
     def name(self):
         return 'tmod4'
@@ -760,8 +759,12 @@ class TMod4Impl(TModImpl):
 
             # Here the module search path removal/addition is repeated since
             # 'restore' discards previous module path manipulations
-            super().searchpath_remove(*self._module_paths_removed)
-            super().searchpath_add(*self._module_paths_added)
+            for op, mp in self._extra_module_paths:
+                 if op == '+':
+                     super().searchpath_add(mp)
+                 else:
+                     super().searchpath_remove(mp)
+
             return []
         else:
             return super().load_module(module)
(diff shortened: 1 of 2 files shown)


teojgo:
@jgphpc can you check if this works for you?

reframe/core/modules.py
```diff
@@ -729,6 +729,8 @@ def __init__(self):
                 (version, self.MIN_VERSION))
 
         self._version = version
+        self._module_paths_added = []
+        self._module_paths_removed = []
```
vkarak, line 733: I would have a single list of tuples here of the form: `[('+', 'path/added'), ('-', '/path/removed')]`. This way you keep the sequence of additions and deletions and be completely sure that you preserve the same behaviour.

vkarak (changes requested):
As for testing, you can create a temporary module collection as we do in the module collection unit tests and play with the module paths.

reframe/core/modules.py
```diff
@@ -803,11 +806,9 @@ def emit_load_instr(self, module):
...
             # Here we append module searchpath removal/addition commands
             # since 'restore' discards previous module path manipulations
-            for mp in self._module_paths_removed:
-                instructions += [f'module unuse {mp}']
-
-            for mp in self._module_paths_added:
-                instructions += [f'module use {mp}']
+            for op, mp in self._extra_module_paths:
+                operation = 'use' if op == '+' else 'unuse'
+                instructions += [f'module {operation} {mp}']
```
vkarak, line 811:
```suggestion
                cmds += [f'module {operation} {mp}']
```

vkarak (approved):
lgtm. I only have a minor cosmetic comment.

PR #1660, opened 2020-12-16
[test] Adapt CP2K checks for Eiger

lucamar:
I have re-defined `self.executable = '--cpu-bind=cores cp2k.psmp'` as suggested by @jjotero, since I could not find a simpler way to add the `srun` flag `--cpu-bind=cores` to the command line. The apparently weird setting of `num_cpus_per_task=16` and `OMP_NUM_THREADS=8` yields a better performance with respect to `num_cpus_per_task=8` and `'OMP_NUM_THREADS': str(self.num_cpus_per_task)` (see the related issue https://jira.cscs.ch/projects/SD/queues/custom/48/SD-51051).

In order to improve a bit more the performance of the test, I would also like to add the following Slurm directive:
```
#SBATCH --distribution=block:block
```
I found in the documentation the `job_opts` parameter, but the warning says that in version 3.0 you may not override this method directly unless you are in special test (https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html?highlight=job_opts#reframe.core.pipeline.RegressionTest.setup).

Are there other options to add custom Slurm directives to the check without creating a check using `special=True`?

### DIFF
--- a/cscs-checks/apps/cp2k/cp2k_check.py
+++ b/cscs-checks/apps/cp2k/cp2k_check.py
@@ -3,7 +3,6 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
-import os
 import reframe as rfm
 import reframe.utility.sanity as sn
 
--- a/cscs-checks/apps/cp2k/cp2k_check.py
+++ b/cscs-checks/apps/cp2k/cp2k_check.py
@@ -11,11 +11,7 @@ class Cp2kCheck(rfm.RunOnlyRegressionTest):
     def __init__(self):
         self.valid_prog_environs = ['builtin']
         self.modules = ['CP2K']
-        if (self.current_system.name in ['eiger']):
-            self.executable = '--cpu-bind=cores cp2k.psmp'
-        else:
-            self.executable = 'cp2k.psmp'
-
+        self.executable = 'cp2k.psmp'
         self.executable_opts = ['H2O-256.inp']
 
         energy = sn.extractsingle(r'\s+ENERGY\| Total FORCE_EVAL \( QS \) '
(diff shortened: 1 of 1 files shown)


ekouts (approved):

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -50,31 +53,63 @@ class Cp2kCpuCheck(Cp2kCheck):
     def __init__(self, scale, variant):
         super().__init__()
         self.descr = 'CP2K CPU check (version: %s, %s)' % (scale, variant)
-        self.valid_systems = ['daint:mc']
+        self.valid_systems = ['daint:mc', 'eiger:mc']
         if scale == 'small':
             self.valid_systems += ['dom:mc']
-            self.num_tasks = 216
+            if (self.current_system.name in ['daint', 'dom']):
```
vkarak, line 59:
```suggestion
            if self.current_system.name in ['daint', 'dom']:
```

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -50,31 +53,63 @@ class Cp2kCpuCheck(Cp2kCheck):
...
+                self.num_tasks = 216
+                self.num_tasks_per_node = 36
+            elif (self.current_system.name in ['eiger']):
```
vkarak, line 62:
```suggestion
            elif self.current_system.name == 'eiger':
```

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -50,31 +53,63 @@ class Cp2kCpuCheck(Cp2kCheck):
...
+                self.variables = {
+                    'MPICH_OFI_STARTUP_CONNECT': '1',
+                    'OMP_NUM_THREADS': '8',
+                    'OMP_PLACES': 'cores',
+                    'OMP_PROC_BIND': 'close'
+                }
+
         else:
-            self.num_tasks = 576
+            if (self.current_system.name in ['daint', 'dom']):
```
vkarak, line 76:
```suggestion
            if self.current_system.name in ['daint', 'dom']:
```

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -50,31 +53,63 @@ class Cp2kCpuCheck(Cp2kCheck):
...
+                self.num_tasks = 576
+                self.num_tasks_per_node = 36
+            elif (self.current_system.name in ['eiger']):
```
vkarak, line 79:
```suggestion
            elif self.current_system.name in ['eiger']:
```

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -12,7 +11,11 @@ class Cp2kCheck(rfm.RunOnlyRegressionTest):
     def __init__(self):
         self.valid_prog_environs = ['builtin']
         self.modules = ['CP2K']
-        self.executable = 'cp2k.psmp'
+        if (self.current_system.name in ['eiger']):
+            self.executable = '--cpu-bind=cores cp2k.psmp'
```
vkarak, line 15: This is not the best way to add this. You should do it with pipeline hooks:

```python
@rfm.run_before('run')
def set_cpu_binding(self):
    self.job.launcher.options = ['--cpu-bind=cores']
```

vkarak (changes requested):
Also to answer your other question about the sbatch option, you should again do it with a pipeline hook:

```python
@rfm.run_before('run')
def set_task_distribution(self):
    self.job.options = ['--distribution=block:block']
```

cscs-checks/apps/cp2k/cp2k_check.py
lucamar: Thanks @vkarak, I have implemented the change that you have requested with the pipeline hook

vkarak (approved):
lgtm

vkarak:
@jenkins-cscs retry dom

PR #1662, opened 2020-12-17
[bugfix] Pass the job id when raising job-related errors

### LINKED ISSUE #1632: Revisit how `JobError` is constructed and raised
Currently, we simply raise `JobError('message')`, but job errors take also the job id and print that in the error message. Since we are not passing the job id, we are getting this type of errors in the output, which can be a bit confusing:

```
  * Job type: batch job (id=27350600)
...
  * Reason: job error: [jobid=None] maximum pending time exceeded
```

I think we should either revisit the way we are throwing `JobError`s or change the way we format the error.

teojgo:
Fixes #1632


vkarak (approved):
lgtm

vkarak:
@jenkins-cscs retry daint

PR #1580, opened 2020-11-05
[feat] Preserve symbolic links in sources directory

teojgo:
* Add unittest.

Fixes #1189

PR #1670, opened 2021-01-04
[bugfix] Revert feature to preserve symbolic links in sources directory since it can lead to dangling links in the stage directory; users should use `readonly_files` instead

### LINKED ISSUE #1639: Preserving symlinks in sources directories creates dangling symlinks in the staging directories
#1580 introduced in reframe the default behaviour of copying symlinks onto the stage directory as symlinks rather than copying the content of the directory that the symlink points to. This works fine when the symlink points to another file that is under the same directory tree. However, this will create dangling symlinks if the original link pointed to a file/directory outside this directory tree.

In reframe, this could happen when a test case needs resources defined in another test.

teojgo:
This PR reverts #1580.

* Remove temporary workarounds from affected regression tests.

Fixes #1639


jjotero (approved):
lgtm

vkarak (approved):
lgtm

PR #1672, opened 2021-01-05
[feat] Support multiple job submissions when slurm's job submit limit is reached

### LINKED ISSUE #1595: job submit limits don't count other user jobs that may already be in the queue
We ran into an issue where it looks like the configurable "max_jobs" setting for a resource only looks at the jobs being submitted by the current instance of reframe. However, the user running reframe can have other jobs in the queue that could count against this limit. If they do, this will cause numerous test to fail with "QOSMaxSubmitJobPerUserLimit". It would be useful to have a flag that would expand this limit to all jobs in the queue that belong to a user.

ekouts:
Will fix #1595 .

This PR still has several issues:
- It should work with remote building but this complicates the PR a bit.
- Currently it works only in the asynchronous policy. In the serial policy I would have to create another loop with some sleep interval to be able to submit the compilation job multiple times.
- There are no unittests since it is a bit hard to test. Maybe one way would be to overload the slurm submission to return the exception and see how the policies handle it?
- I think the logic of the policies file is getting a bit too complicated and we should rethink how to name the buckets of the tasks (now called "ready", "running", "compiled" etc) and how we move the tasks through them. I think it's reasonable to also have a simpler `runcase` which puts jobs just in the "waiting" list and handle everything in `exit`. This would require a bigger restructure and more thorough testing, but I made this draft PR anyway, mostly to get some feedback.
- It also needs documentation and some configuration parameter, so that the users can decide whether to treat `QOSMaxSubmitJobPerUserLimit` as an error or submit the job later.

### DIFF
--- a/reframe/frontend/executors/__init__.py
+++ b/reframe/frontend/executors/__init__.py
@@ -16,6 +16,7 @@ import reframe.core.runtime as runtime
 import reframe.frontend.dependencies as dependencies
 import reframe.utility.jsonext as jsonext
 from reframe.core.exceptions import (AbortTaskError, JobNotStartedError,
+                                     JobQOSMaxSubmitJobPerUserLimitError,
                                      ReframeForceExitError, TaskExit)
 from reframe.core.schedulers.local import LocalJobScheduler
 from reframe.frontend.printer import PrettyPrinter
@@ -252,6 +253,8 @@ class RegressionTask:
         except ABORT_REASONS:
             self.fail()
             raise
+        except JobQOSMaxSubmitJobPerUserLimitError as e:
+            raise e
         except BaseException as e:
             self.fail()
             raise TaskExit from e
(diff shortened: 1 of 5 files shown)


reframe/frontend/executors/policies.py
```diff
@@ -351,13 +389,15 @@ def runcase(self, case):
             if len(self._running_tasks[partname]) < partition.max_jobs:
                 # Task was put in _ready_tasks during setup
                 self._ready_tasks[partname].pop()
-                self._reschedule(task)
+                self._reschedule_compile(task)
+                self._reschedule_run(task)
             else:
                 self.printer.status('HOLD', task.check.info(), just='right')
-        except TaskExit:
+        except TaskExit: ## WHY?
```
ekouts, line 396: I can't find why we need this here so I am not sure if I am handling the exception correctly

reframe/frontend/executors/__init__.py
```diff
@@ -261,14 +264,22 @@ def setup(self, *args, **kwargs):
         self._notify_listeners('on_task_setup')
 
     def compile(self):
-        self._safe_call(self.check.compile)
+        try:
+            self._safe_call(self.check.compile)
+            self._notify_listeners('on_task_compile')
+        except JobQOSMaxSubmitJobPerUserLimitError as e:
+            raise e
```
ekouts, line 271: It is not nice that we treat `JobQOSMaxSubmitJobPerUserLimitError` differently here and in the `run` method below, but I couldn't find a better way without changing the structure of policies and how we poll for compiled jobs.

reframe/frontend/executors/policies.py
vkarak: It is needed and it acts effectively as a "break" statement for a non-critical condition that that has happened in currently executing task. Check also #1331.

PR #1676, opened 2021-01-07
[feat] Add `--maxfail` option to stop execution after a certain number of failures

### LINKED ISSUE #941: Stop execution after a certain number of failures

ekouts:
Fixes #941 .


reframe/frontend/executors/policies.py
```diff
@@ -92,6 +93,11 @@ def runcase(self, case):
         self._task_index[case] = task
         self.stats.add_task(task)
         try:
+            if (self.max_failures and
+                len(self.stats.failures()) >= self.max_failures):
+                raise MaxFailError('the maximum number of failures has been '
```
vkarak, line 98: Simply raise it as a `ReframeError`.

reframe/frontend/executors/policies.py
```diff
@@ -499,5 +504,11 @@ def exit(self):
                 self._failall(e)
                 raise
 
+            if len(self.stats.failures()) >= self.max_failures:
+                exc = MaxFailError('the maximum number of failures has been '
+                                   'reached')
```
vkarak, line 509: Instead of duplicating the logic here, you could put it in the `on_task_failure()`, couldn't you?

reframe/core/exceptions.py
```diff
@@ -98,6 +98,12 @@ class TaskDependencyError(ReframeError):
     '''
 
 
+class MaxFailError(ReframeError):
```
vkarak, line 101: Just raise a `ReframeError` directly.

vkarak (changes requested):
We also need a unit test + documentation.

reframe/core/exceptions.py
ekouts: Then the output of async policy, in the summary of failures, will be simply ```* Reason: aborted due to ReframeError``` instead of ```* Reason: aborted due to MaxFailError``` that it is now.
Should we change more in the output?

reframe/core/exceptions.py
```diff
@@ -98,6 +98,12 @@ class TaskDependencyError(ReframeError):
...
+    '''Raised inside a regression task by the runtime when the maximum allowed
+    when the maximum number of failures has been exceeded.
```
teojgo, line 103:
```suggestion
    number of failures has been exceeded.
```

reframe/frontend/executors/policies.py
ekouts: I don't know if it is a good idea, because it creates a loop now, since `fail` will call `_failall` , which will call `fail` for each task and since the maxfail will be reached it will try to fail all every time. And even when I try to abort them only once I still get some issues with the other lists. If you think it is much better like this, I can look into where the errors are coming from.

vkarak:
@ekouts Is this still a draft?

vkarak:
Can you also fix the unused import errors (see the CI checks).

reframe/frontend/cli.py
```diff
@@ -923,13 +923,18 @@ def module_unuse(*paths):
             ) from None
 
         try:
-            max_fail = int(options.max_fail)
+            max_failures = int(options.maxfail)
+            if max_failures < 0:
+                raise errors.ConfigError(
+                    f'--maxfail should be a non-negative integer: '
+                    f'{options.max_fail!r}'
+                ) from None
```
vkarak, line 931: You don't need the `from None` here.

reframe/frontend/executors/__init__.py
```diff
@@ -480,7 +480,7 @@ def __init__(self):
         self.stats = None
 
     def enter(self):
-        pass
+        self._failed_tests = 0
```
vkarak, line 483:
```suggestion
        self._num_failed_tasks = 0
```

reframe/frontend/executors/policies.py
```diff
@@ -94,7 +94,7 @@ def runcase(self, case):
         self.stats.add_task(task)
         try:
             if (self.max_failures and
```
vkarak, line 96: In order to keep things simple, I suggest initializing `max_failures` to `sys.maxsize` and skip the first check here.

reframe/frontend/executors/policies.py
vkarak: No, leave it as it is. I have made another suggestion, which would make the `if` simpler.

reframe/core/exceptions.py
vkarak: I don't like that it fails like that. Let me play a bit with this to get the feeling.

ekouts:
> @ekouts Is this still a draft?

Yes, because I still need to write documentation + unittests. I will fix them and change it.

reframe/frontend/cli.py
```diff
@@ -923,13 +923,18 @@ def module_unuse(*paths):
...
+            max_failures = int(options.maxfail)
+            if max_failures < 0:
+                raise errors.ConfigError(
+                    f'--maxfail should be a non-negative integer: '
+                    f'{options.max_fail!r}'
+                )
         except ValueError:
             raise errors.ConfigError(
-                f'--max-fail is not a valid integer: {max_fail}'
+                f'--maxfail is not a valid integer: {options.max_fail!r}'
```
vkarak, line 934:
```suggestion
                f'--maxfail is not a valid integer: {options.maxfail!r}'
```

reframe/frontend/cli.py
```diff
@@ -923,13 +923,18 @@ def module_unuse(*paths):
             ) from None
 
         try:
-            max_fail = int(options.max_fail)
+            max_failures = int(options.maxfail)
+            if max_failures < 0:
+                raise errors.ConfigError(
+                    f'--maxfail should be a non-negative integer: '
+                    f'{options.max_fail!r}'
```
vkarak, line 930:
```suggestion
                    f'{options.maxfail!r}'
```

vkarak: Which btw means that this is not tested :-) We should add a unit test.

vkarak (changes requested):
We also need unit tests for the policies + documentation.

vkarak (changes requested):
I played a bit with the feature and it's not exactly as I have imagined it. I've run `./bin/reframe -c unittests/resources/checks/frontend_checks.py -r --maxfail=1` and the output is not intuitive. Let's start with the baseline:

```
[  FAILED  ] Ran 6 test case(s) from 6 check(s) (5 failure(s))
```

When running with `--maxfail=1`, I get

```
[  FAILED  ] Ran 6 test case(s) from 6 check(s) (6 failure(s))
```

This is wrong, because you mark as failures all other tests, like sending a keyboard interrupt. Additionally, in the output I see 6 failures instead of just one, which does not serve the purpose of this option, which is basically to avoid the "fail clutter." When the `--maxfail=N` is passed I expect to see

```
[  FAILED  ] Ran X test case(s) from Y check(s) (N failure(s))
```

And the summary must contain exactly N failure infos.

The obvious thing to ask is why we treat `KeyboardInterrupt` differently. The answer is that it was just like that, but now I think that we should probably handle the `KeyboardInterrupt` and perhaps all the abort reasons in the same way and avoid "fail clutter" all the way. If I'm running 1000 test cases, and I type Ctrl-C, I don't want to see 1000 failure infos!

ekouts:
> The obvious thing to ask is why we treat `KeyboardInterrupt` differently. The answer is that it was just like that, but now I think that we should probably handle the `KeyboardInterrupt` and perhaps all the abort reasons in the same way and avoid "fail clutter" all the way. If I'm running 1000 test cases, and I type Ctrl-C, I don't want to see 1000 failure infos!

I agree with but I have some more questions:
1. When `maxfail` is reached should tests be retried? For `KeyboardInterrupt` I assume not.
2. Do I include the aborted tests in the json report?
> And the summary must contain exactly N failure infos.
3. In the asynchronous policy is it okay if it is not always exact? We can do it but this might not include tests that have actually finished in the meantime and failed for other reasons (eg failed dependency)

ekouts:
@vkarak Some more questions:
> When the --maxfail=N is passed I expect to see
```[  FAILED  ] Ran X test case(s) from Y check(s) (N failure(s))```

I am not sure what the X number should be in this case, the ones that actually finished? Or the total number that has at least been setup? In the asynchronous policy all tests are trying to do the setup before the first failure prints the message of failure which can lead to confusing output this way.

Let's say we have `--maxfai=1` and a tests fails in setup, then how should the output continue? As usual with `[ RUN      ] Testcase on partition using env` of the rest of the testcases? Or move one directly to `[----------] waiting for spawned checks to finish` and print the successful + failed tests so far? In the first case you might get extra failures and not have exactly N failures.

**Comment by vkarak** [comment]
> When maxfail is reached should tests be retried? For KeyboardInterrupt I assume not.

No need to retry them, I agree.

> Do I include the aborted tests in the json report?

Good point. I would say not, but perhaps we need a way to identify directly or indirectly how many were aborted (e.g., we might need a `num_cases_run` property or similar).

> In the asynchronous policy is it okay if it is not always exact? We can do it but this might not include tests that have actually finished in the meantime and failed for other reasons (eg failed dependency)

I'm fine of not being exact with the async policy.

> I am not sure what the X number should be in this case, the ones that actually finished? Or the total number of the ones that have at least been setup? In the asynchronous policy all tests are trying to do the setup before the first failure prints the message of failure which can lead to confusing output this way.

Those that have actually finished, I would say.

> Let's say we have --maxfai=1 and a tests fails in setup, then how should the output continue? As usual with [ RUN ] Testcase on partition using env of the rest of the testcases? Or move on directly to [----------] waiting for spawned checks to finish and print the successful + failed tests so far? In the first case you might get extra failures and not have exactly N failures.

None of this. As soon as it reached the maximum failure count, we abort everything (kill those pending) and report everything that has been finalized up until this point. This should be exactly as with keyboard interrupt. So I think this answers also the exactness question, right? If you abort immediately, you don't finalize any pending test, so the number of failures should be exact, shouldn't?

File: reframe/frontend/statistics.py [hunk 10]
```diff
@@ -187,6 +188,10 @@ def print_failure_report(self, printer):
             if r['result'] == 'success':
                 continue
 
+            if (r['fail_reason'] == 'aborted due to MaxFailError' or
+                r['fail_reason'] == 'aborted due to KeyboardInterrupt'):
```
vkarak, line 192: This should somehow come directly from the exception.

reframe/frontend/statistics.py
```diff
@@ -33,7 +33,8 @@ def tasks(self, run=-1):
             raise errors.StatisticsError('no such run: %s' % run) from None
 
     def failures(self, run=-1):
-        return [t for t in self.tasks(run) if t.failed]
+        return [t for t in self.tasks(run) if (t.failed and
+                                               not t.aborted)]
```
vkarak, line 37: Would it make sense that `t.failed` was true only if the test has failed and be aborted?

vkarak (changes requested):
The output is ok, I had a minor comment on how you log it. I have some issues with the implementation, which I have commented.

reframe/frontend/executors/__init__.py
```diff
@@ -392,15 +397,35 @@ def runall(self, testcases, restored_cases=None):
...
-
+        except AssertionError:
+            abort_reason = "assertion error"
+            raise
+        except MaxFailError:
+            abort_reason = "maximum number of failures reached"
+        except KeyboardInterrupt:
+            abort_reason = "keyboard interrupt"
+            raise
+        except ReframeForceExitError:
```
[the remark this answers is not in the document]
ekouts: We don't always raise the exception. In MaxFailError we want the failure summary to be printed.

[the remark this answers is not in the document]
vkarak: 1. Can you rename the `ReFrameForceExitError` to simply `ForceExitError`?
2. What is the exact difference? Also I don't like that you have to explicitly set the `abort_reason`. ReFrame automatically does that in the exception message.

vkarak (changes requested):

reframe/frontend/executors/__init__.py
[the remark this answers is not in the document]
vkarak: I'm going to push the change for (2).

reframe/frontend/executors/__init__.py
```diff
@@ -386,20 +404,38 @@ def runall(self, testcases, restored_cases=None):
...
+            abort_reason = "reframe error"
+            raise
         finally:
             # Print the summary line
-            num_failures = len(self._stats.failures())
+            num_failures = len(self._stats.failed())
+            num_completed = len(self._stats.completed())
+            if abort_reason or num_failures:
```
vkarak, line 419: I don't understand this `if`. I'm trying to simplify the exception handling and I have a design that wouldn't even the `except` above nor the `abort_reason`. So I'm wondering why not simply `if num_failures` here?

vkarak: I removed completely the `abort_reason`. Let me know if you see something that I can't figure out.

vkarak (approved):
lgtm now

vkarak:
The Tmod32 unit tests fail for some. We need to have a look. Other than, that the PR is fine.

reframe/frontend/executors/__init__.py
ekouts: No, I think it is okay.

PR #1678, opened 2021-01-11
[test] Add basic hostname check for the `xfer` partition

ekouts:


rsarm (approved):

config/cscs.py
```diff
@@ -289,6 +289,24 @@
...
+                },
+                {
+                    'name': 'xfer',
+                    'scheduler': 'slurm',
+                    'environs': [
+                        'builtin',
+                        'PrgEnv-cray',
+                        'PrgEnv-gnu',
+                        'PrgEnv-intel',
+                        'PrgEnv-pgi'
```
vkarak, line 301: Do we really need all these environments?

vkarak (changes requested):

config/cscs.py
ekouts: No, actually now that you mention it only `builtin` is available, so I think it's better if the hostname check has the `builtin` programming environment.

vkarak (approved):
lgtm

PR #1682, opened 2021-01-13
[test] Add Eiger system to NAMD check

ChristopherBignamini:


cscs-checks/apps/namd/namd_check.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, arch, scale, variant):
 
         self.maintainers = ['CB', 'LM']
         self.tags = {'scs', 'external-resources'}
-        self.strict_check = False
+        self.strict_check = True
```
vkarak, line 52: You can remove this completely, since `True` is the default.

vkarak:
@jenkins-cscs retry eiger

vkarak (changes requested):
I have just a minor comment + is the reference on Eiger up to date?

lucamar (approved):
Thanks @ChristopherBignamini , I think that the reference values need to be updated according to the last successful build of NAMD SMP on Eiger, which showed a better performance

cscs-checks/apps/namd/namd_check.py
ChristopherBignamini: Reference on Eiger is updated for the current module, which is for the SMP build of NAMD. According to the comments reported here https://jira.cscs.ch/browse/SD-51014, I need to investigate a bit more the NAMD affinity options in order to get better result (currently a pure MPI build provides better performances).

vkarak:
I agree with @lucamar. Performance numbers need to be updated and take care as well of the PEP8 complaints.

ChristopherBignamini:
The reference numbers are best I have been able to obtain so far for the smp build (current module on Eiger). I have tried to use the cpuaffinity option of NAMD but I don't see any change in the performances. I will have a look to the slurm affinity options/job configuration on Eiger, hopefully I will get better results. In the meantime I will also try with a pure MPI build that, according to the performances reported in the jira issue SD-51014, is faster than the smp one.

ChristopherBignamini:
@jenkins-cscs retry all

vkarak (approved):
lgtm

PR #1683, opened 2021-01-13
[test] Fix tests on dom requiring cdt-cuda

teojgo:


vkarak (commented):
Is this a temporary workaround? I don't like that we are testing explicitly if we are on Dom, because we will have to re-adapt the tests when moving on Daint. Is there a better way?

teojgo:
> Is this a temporary workaround? I don't like that we are testing explicitly if we are on Dom, because we will have to re-adapt the tests when moving on Daint. Is there a better way?

For the moment it's only for dom. When we have on daint we will have to also load the `cdt-cuda` to compile cuda applications.

jjotero (commented):
lgtm

vkarak (approved):

PR #1687, opened 2021-01-15
[test] GPU dgemm and pointer chasing tests

jjotero:
New `dgemm` and `pointer_chase` regression tests. Like all the other microbenchmark tests, these run with both `cuda` and `hip`.

Pointer chase checks:
 - Check the latency of calling the `clock` function.
 - Average latency on a single device. This test uses several stride values.
 - Average P2P latency. The stride is fixed to 1 here.
 - L1 latency, L1 miss rate, and miss latency on a single device.
 - L1 latency, L1 miss rate and miss latency when executing the pointer chase through P2P


jjotero:
@jenkins-cscs retry all

jjotero:
The clockLatency test is ver very far from meeting the performance in Daint. However, it passes in Dom without any problems.

sebkelle1:
Great work, seeing a working hip/cuda abstraction really makes me want to try it out for my projects too!
Also impressive level of detail with the timers.
I'm sorry I just saw the review request now, will add my review shortly.

cscs-checks/microbenchmarks/gpu/pointer_chase/src/pointer_chase.cu
```diff
@@ -0,0 +1,367 @@
...
+
+
+int main(int argc, char ** argv)
+{
+  // Set program defaults before parsing the command line args.
+  int list_init_mode = 0;
+  size_t stride = 1;
+  size_t buffSize = NODES*stride;
+  int multiGPU = 0;
+  int print_mode = 0;
```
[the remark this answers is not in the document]
vkarak: I would vote for underscore naming, but I leave it up to you. :-)

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,383 @@
...
+__global__ void initialize_list(Node * buffer, int stride = 1)
+{
+  /* List serial initializer.
+   * - buffer: where the list is to be placed.
+   * - stride: argument controls the number of empty node spaces
+   *   in between two consecutive list nodes.
+   */
+
+  // Set the head
+  Node * prev = new (&(buffer[0])) Node();
```
sebkelle1, line 57: You can just say `Node* prev = buffer;`

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,383 @@
...
+
+  // Init the rest of the list
+  for (int n = 1; n < NODES; n++)
+  {
+    Node * temp = new (&(buffer[n*stride])) Node();
```
sebkelle1, line 62: `Node* temp = buffer + n*stride;`

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,383 @@
...
+__global__ void initialize_random_list(Node * buffer, uint32_t *indices)
+{
+  /* List random initializer
+   * - buffer: where the list is to be placed.
+   * - indices: array containing the node ordering indices as offsets in
+   *     the buffer.
+   */
+
+  // Set the head
+  Node * prev = new (&(buffer[indices[0]])) Node();
```
sebkelle1, line 78: `Node* prev = buffer;`

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,383 @@
...
+
+  // Init the rest of the list
+  for (int n = 1; n < NODES; n++)
+  {
+    Node * temp = new (&(buffer[indices[n]])) Node();
```
sebkelle1, line 83: `Node* temp = buffer + n*stride;`

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,383 @@
...
+        {
+          if (currentIndex < NODES-1)
+          {
+            currentIndex++;
+          }
+          else
+          {
+            currentIndex = 0;
+          }
+        }
```
sebkelle1, line 296: To generate a random permutation, you usually start from the sorted sequence 0...NODES, then apply a random transposition to each element. `std::iota` plus `std::shuffle` does quite nicely:
```suggestion
            std::random_device rd;
            std::mt19937_64 gen(rd());
            uint32_t * nodeIndices = (uint32_t*)malloc(sizeof(uint32_t)*NODES);
            std::shuffle(nodeIndices, nodeIndices + NODES, gen);
```

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,383 @@
...
+  // Set the head
+  Node * prev = new (&(buffer[indices[0]])) Node();
+
+  // Init the rest of the list
+  for (int n = 1; n < NODES; n++)
+  {
+    Node * temp = new (&(buffer[indices[n]])) Node();
+    prev->next = temp;
+    prev = temp;
+  }
```
sebkelle1, line 86: Close the list with `prev->next = buffer + indices[0];`

sebkelle1 (changes requested):
A short description of what I understand you are doing is this:
You place 64 nodes containing a single pointer in memory, either next to each other, or spaced  `strides` apart. Nodes are then either linked sequentially or in random order. You then read out the clock counter and traverse this list a single time with one GPU thread, followed by reading out the clock counter again and report the latency as the clock difference divided by 64.

While your assembly timer is definitely the most advanced I can imagine or I've ever seen, traversing only 64 nodes, taking only a few microseconds, is definitely not enough to get accurate measurements. From my experience with pointer chasing on the CPU, I would imagine that you need at least a couple dozen milliseconds. This means that you will need your linked list to be a closed loop such that you can tune the number of hops to ensure a sufficient runtime. Don't bother with unrolling the whole traversal, either the compiler will unroll it a bit for you or you can do it manually. Either way, after the first couple unrolls, there's no difference.

In terms of the size of the linked list, you'd expect the memory footprint of your list to be significantly larger than the caches below the level you're measuring. On a GPU you have a couple KB of L1 and a couple MB of L2. If you also want to measure DRAM latency, your pointer chain needs to touch probably around 100MB, otherwise all your accesses will stay in cache. Empty holes between nodes don't count towards the footprint, because the holes are never accessed (except where part of a cacheline).

cscs-checks/microbenchmarks/gpu/pointer_chase/src/pointer_chase.cu
```diff
@@ -0,0 +1,367 @@
...
+      std::cout << "--stride #  : Sets the stride between the nodes in the list (in number of nodes)." << std::endl;
+      std::cout << "              If --rand is used, this parameter just changes the buffer size." << std::endl;
+      std::cout << "--buffer #  : Sets the size of the buffer where the linked list is allocated on. " << std::endl;
+      std::cout << "              The number indicates the size of the buffer in list nodes." << std::endl;
+      std::cout << "--multiGPU  : Runs the pointer chase algo using all device-pair combinations." << std::endl;
+      std::cout << "              This measures the device-to-device memory latency." << std::endl;
+      std::cout << "--summary   : When timing each node jump individually and used alongside --multiGPU, " << std::endl;
+      std::cout << "              this collapses the output into two tables with the min and max latencies." << std::endl;
+      std::cout << "--clock     : Skip all the above and just print the clock latency for all devices." << std::endl;
+      std::cout << "--help (-h) : I guess you figured what this does already ;)" << std::endl;
```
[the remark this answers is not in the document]
jjotero: Makes sense. I've taken out the `--buffer` option and renamed `--stride` to `--sparsity`. This simplifies things and now the buffer size is guaranteed to have the valid size.

cscs-checks/microbenchmarks/gpu/pointer_chase/src/pointer_chase.cu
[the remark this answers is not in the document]
jjotero: 👍 I've opted for the underscore naming.

cscs-checks/microbenchmarks/gpu/pointer_chase/src/pointer_chase.cu
```diff
@@ -0,0 +1,367 @@
...
+  }
+# endif
+
+}
+
+
+int main(int argc, char ** argv)
+{
+  // Set program defaults before parsing the command line args.
+  int list_init_mode = 0;
```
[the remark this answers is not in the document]
jjotero: All the test I've written so far use the serial init, and I don't think we need anything beyond that for benchmarking purposes. The random init is just in there for experimental purposes, so I don't think we'll be adding more methods. 

Changed `list_init_mode` to `init_list_random`.

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,383 @@
...
+  // Set the head
+  Node * prev = new (&(buffer[0])) Node();
+
+  // Init the rest of the list
+  for (int n = 1; n < NODES; n++)
+  {
+    Node * temp = new (&(buffer[n*stride])) Node();
+    prev->next = temp;
+    prev = temp;
+  }
```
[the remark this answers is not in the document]
jjotero: I've skipped this for now. See comment below.

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
jjotero: Same here.

jjotero:
Hmm, not sure there is a need for more node jumps. All these reframe tests do not use the random init, and despite having only 64 nodes, the timings are pretty consistent. 

There are two types of tests in here. The first type measures the overall traversal time, and the second one measures each jump independently. For the second type, the reframe tests work out the L1 hits and misses, and compute the average timing only for these two categories. As it can be seen in the figure below (data from a P100), the timings for an L1 hit are pretty consistent at 158 cycles. If we were to capture the latency of L2 (which sits somewhere below 400 cycles), then I'd agree that we might need more nodes. But in that case, extracting the data that only belongs to an L2 hit is a challenge on its own. 

Though, as you say, if we make it circular we could easily tune the list to fit in any memory level. Let me experiment a bit.

P100_latency

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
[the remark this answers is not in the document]
jjotero: The lists are now circular.

jjotero:
The linked lists are now circular and the number of node jumps is fixed to 256. The number of nodes is now an input parameter from the user, which means that we can do something like a list with 2 nodes (resident in L1) and jump around 256 times to collect all the data we need.

At the moment, there is some inconsistent behaviour with the clock latency on the P100s. This needs further testing in the other cards present in Ault.

cscs-checks/microbenchmarks/gpu/pointer_chase/src/linked_list.hpp
```diff
@@ -0,0 +1,356 @@
...
+  int jump = 0;
+
+#ifndef TIME_EACH_STEP
+  // start timer
+  XClocks64 clocks;
+  clocks.start();
+#endif
+
+  // Traverse the list
+  REPEAT_JUMPS(next_node(&ptr, s_timer, ptrs, jump);)
```
sebkelle1, line 177: I would really strongly discourage you from using macros for essential code-logic. The list traversal is THE core essential part of the entire benchmark and this part is implemented here with the macro `REPEAT_JUMPS` that's not even defined in the same file. It's perfectly adequate to use a plain loop here:
```suggestion
 while(num_jumps--)
 {
     buffer = buffer->next;
 }
```
where `num_jumps` is the number of jumps that I've added as a command line parameter in my tests.

The same goes for `JUMPS`. Together with the number of nodes, the number of jumps is the key parameter of the benchmark and there is simply no way around making it dynamically controllable for achieving accurate results.

In general, if you really must use a macro, my advice is to only define/reference them within the same file.

sebkelle1 (changes requested):

sebkelle1:
The timings don't seem converged. As I mentioned previously, setting the number of hops to a fixed number is not a viable approach and also 256 jumps are not enough by a long shot to get converged results.
I took the code and made a small modification: I added `num_jumps` as a command line parameter, changed the traversal to a `while` loop as described above and increased `num_jumps` for each list size until the time per jump converged.

Here are the results on V100:
[V100.pdf](https://github.com/eth-cscs/reframe/files/5971335/V100.pdf)

and P100:
[P100.pdf](https://github.com/eth-cscs/reframe/files/5971336/P100.pdf)

List linkage was random, with no additional node-padding.

On the V100, the L1 latency now is 28 cycles which agrees with other reports that I found online.
The reason why 256 jumps are not nearly enough is in fact not due to timing overhead. Like CPUs, GPUs nowadays have multiple (opaque hardware-controlled) levels of cache and GPUs also use virtual memory, which means that page table lookup also happens. Despite a warm-up pass through the list, where exactly an address resides and whether or not there's a page-fault is beyond your control for individual accesses. You can only hope to iron out these effects by averaging over multiple repeat passes. The faster the device and the faster the cache, the more pronounced these effects become.
In this context, I would say that trying to time individual jumps is an interesting experiment, but in practice actually just yields random linear combinations of the different cache/DRAM values. Again, not because the timer is not accurate enough, but due to the mentioned background effects that you cannot control.

jjotero (commented):
I've now stripped all the code related to measuring latencies of single node jumps and done the cleanup.

The plot below shows a parametric sweep for all the cards in ault, for a large range of list sizes, and different strides; all using 400k node jumps for the timing. The nodes are placed in sequential order in the buffer, so the `--stride` parameter here was used to control the filling of the cache lines.

plot
A100: solid; V100: dashed; P100: dash-dotted; Vega20: dotted

Now I just need to add the actual tests. For now, I'll fix the stride to 32 in all of them and check the L1, L2 and DRAM latencies.

jjotero:
The tests are in now.

cscs-checks/microbenchmarks/gpu/dgemm/dgmemm.py
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }
+
+        self.maintainers = ['JO']
+        self.tags = {'benchmark'}
+
+    @property
+    @sn.sanity_function
+    def num_tasks_assigned(self):
+        return self.job.num_tasks
```
vkarak, line 54: You don't need that anymore. You can use `sn.getattr` in your sanity patterns and get the `num_tasks` of the test. It's updated to the actual value when you run a flexible test.

cscs-checks/microbenchmarks/gpu/dgemm/dgmemm.py
```diff
@@ -0,0 +1,100 @@
...
+
+    @sn.sanity_function
+    def assert_num_gpus(self):
+        return sn.assert_eq(
+            sn.count(sn.findall(r'^\s*\[[^\]]*\]\s*Test passed', self.stdout)),
+            self.num_tasks_assigned)
+
+    @rfm.run_after('setup')
```
vkarak, line 62: Just because conceptually makes more sense. Similarly below:

```suggestion
    @rfm.run_before('compile')
```

cscs-checks/microbenchmarks/gpu/pointer_chase/pointer_chase.py
```diff
@@ -0,0 +1,345 @@
...
+import reframe as rfm
+
+import os
+
+
+class PchaseGlobal(rfm.RegressionMixin):
+    '''Handy class to store common test settings.
+    '''
+    single_device_systems = variable(
+        list,
```
vkarak, line 16:
```suggestion
        typ.List[str],
```

cscs-checks/microbenchmarks/gpu/pointer_chase/pointer_chase.py
```diff
@@ -0,0 +1,345 @@
...
+        value=['daint:gpu', 'dom:gpu']
+    )
+    multi_device_systems = variable(
+        list,
```
vkarak, line 20: Same here.

cscs-checks/microbenchmarks/gpu/pointer_chase/pointer_chase.py
```diff
@@ -0,0 +1,345 @@
...
+            self.single_device_systems + self.multi_device_systems
+        )
+        self.valid_prog_environs = self.global_prog_environs
+        self.exclusive_access = True
+        self.build_system = 'Make'
+        self.num_tasks = 0
+        self.num_tasks_per_node = 1
+        self.postbuild_cmds = ['ls .']
+        self.sanity_patterns = sn.assert_found(r'pChase.x', self.stdout)
+        self.maintainers = ['JO']
```
vkarak, line 42: Add `SKa  for back up. In both tests.

vkarak (commented):
I didn't go through the C++ code, but I've gone through your discussions and the results and I think that we are good to go. Great work of both of you! I've only some minor ReFrame test comments.

jjotero (commented):
PR comments addressed.

vkarak (approved):

PR #1688, opened 2021-01-15
[ci] Use Github Action for uploading ReFrame releases on PyPI

### LINKED ISSUE #1559: Deploy new versions to PyPI using Github Actions

teojgo:
Fixes #1559

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: Deploy and Publish ReFrame to PyPI
 
 on:
   release:
-    types: [released]
+    types: [prereleased, released]
 
 jobs:
   publish:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,29 @@
+name: Deploy and Publish ReFrame to PyPI
+
+on:
+  release:
+    types: [released]
+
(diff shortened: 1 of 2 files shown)


ekouts (approved):

.github/workflows/publish.yml
```diff
@@ -0,0 +1,29 @@
+name: Deploy and Publish ReFrame to PyPI
+
+on:
+  release:
+    types: [released]
```
vkarak, line 5: Does this cover pre-releases too?

vkarak (approved):
lgtm

.github/workflows/publish.yml
teojgo: No, the various alternatives are given here: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release

teojgo: No, the alternatives types are given here: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#release

vkarak: You should then add `prereleased` here, too.

vkarak (changes requested):

vkarak (approved):
lgtm

PR #1694, opened 2021-01-19
[test] Load cdt-cuda in dom for GPU checks

jjotero:
Extended hook for `dom:gpu` so that the tests under `microbenchmarks/gpu` load the `cdt-cuda` module.


teojgo (changes requested):
Better use `craype-accel-nvidia60` for dom/daint.

vkarak (changes requested):
I agree with @teojgo's comment.

jjotero:
Done. I'll update also the new tests in PR #1687.

teojgo (approved):

vkarak (approved):

PR #1695, opened 2021-01-19
[feat] Add sanity function equivalents of some of `os.path` functions

### LINKED ISSUE #1634: Add sanity function versions of some basic `os.path` functions
Such as `os.path.exists()`, `os.path.isdir()` etc. so that they can be readily used in sanity patterns expressions.

teojgo:
Fixes #1634

### DIFF
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -781,31 +781,35 @@ def iglob(pathname, recursive=False):
 
 @deferrable
 def path_exists(path):
-    '''Replacement for the :func:`os.path.exists() <python:os.path.exists>`
-       function.
+    '''Replacement for the :func:`os.path.exists` function.
+
+    .. versionadded:: 3.4
     '''
     return os.path.exists(path)
 
 
 @deferrable
 def path_isdir(path):
-    '''Replacement for the :func:`os.path.isdir() <python:os.path.isdir>`
-       function.
+    '''Replacement for the :func:`os.path.isdir` function.
+
+    .. versionadded:: 3.4
     '''
     return os.path.isdir(path)
 
 
 @deferrable
 def path_isfile(path):
-    '''Replacement for the :func:`os.path.isfile() <python:os.path.isfile>`
-       function.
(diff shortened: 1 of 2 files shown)


reframe/utility/sanity.py
```diff
@@ -776,3 +777,35 @@ def iglob(pathname, recursive=False):
     '''Replacement for the :func:`glob.iglob() <python:glob.iglob>`
     function.'''
     return pyglob.iglob(pathname, recursive=recursive)
+
+
+@deferrable
+def path_exists(path):
+    '''Replacement for the :func:`os.path.exists() <python:os.path.exists>`
+       function.
```
vkarak, line 785: Add a `.. versionadded:: 3.4` in all the new functions.

reframe/utility/sanity.py
```diff
@@ -776,3 +777,35 @@ def iglob(pathname, recursive=False):
...
+    '''
+    return os.path.exists(path)
+
+
+@deferrable
+def path_isdir(path):
+    '''Replacement for the :func:`os.path.isdir() <python:os.path.isdir>`
+       function.
```
vkarak, line 793:
```suggestion
    '''Replacement for the :func:`os.path.exists` function.
```

vkarak: And similarly for the rest. Can you also do the same for the `glob` function?

vkarak (changes requested):

reframe/utility/sanity.py
teojgo: Do you mean the `glob.glob` function?

teojgo: Because those are already there.

vkarak: Yes, but I mean to fix the link in the docstring. Because you copied-pasted the format from there and it was wrong. It didn't create a hyperlink. You have to write as I propose here.

vkarak (approved):

vkarak:
@jenkins-cscs retry dom

ekouts (approved):

PR #1697, opened 2021-01-20
[test] Put back import that registers magic command

rsarm:
`import ipcmagic` registers the magic command `%ipcluster`. The module is not used on the rest the code but without it script crashes.


cscs-checks/apps/jupyter/src/tf-hvd-sgd-ipc-tf2.py
```diff
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+import ipcmagic
```
vkarak, line 6: If this is not used, add a `# noqa: F401` annotation comment in this line so as to be ignored by flake.

teojgo (approved):

PR #1699, opened 2021-01-20
[feat] Add built-ins to manage test variables and parameters

### LINKED ISSUE #1568: Syntactic elements for creating test libraries and managing efficiently large parameterization spaces

jjotero:
This PR introduces some changes suggested in #1568. Building on the same concept as used to implement the extensible parameterized tests, we introduce the `var` directive, which injects a new variable in the `RegressionTest` class. 
```python
class Spam(rfm.RegressionTest):
    var('myTestVariable', int, float, value=3.141592)
    
    def __init__(self):
        do_something_with_my_own_var(self.myTestVariable)
```
This variable is injected into `Spam` as a data descriptor (`TypedField`) though the `__new__` method. If an alternative `Field` were needed, the `var` directive accepts the`field` argument. The value passed as this argument must be a subclass of the `Field` class located in `reframe/core/fields.py`. 

On this `var` directive, the `value` argument is entirely optional, and one might choose to leave the variable undefined. Logically, if that is the case and the variable is referenced in the class, an exception will be raised. However, If one inherits from a class with an undefined variable and need to set its value, or simply overwriting an old value, one can use the `set_var` directive.
```python
class Ham(Spam):
    set_var('myTestVariable', 1.5)
```
Note that, when using the `set_var` directive, the variable `myTestVariable` must have been declared through the `var` directive in a parent class. If that were not the case, another error would be raised.

Now with the converse. Suppose you're writing a library extending the work of someone else, but that "base" library has set a value for a variable. Instead of using such value, you can force the users to set the value for themselves by using the `require_var` directive. This directive will effectively undefine the variable, which means that an error will be raised if it hasn't been set.
```python
class Eggs(Ham):
    require_var('myTestVariable')

    def __init__(self):
        self.myTestVariable = 20 # We set the variable first ...
        super().__init__()       # and now we can call Spam's __init__ method.
```
Same as with `set_var`, the variable referred to by `require_var` must have been declared with the `var` directive in any of the parent classes. __Flagging a variable that is not present in the test as *required* is implemented as a no-op__ (a variable that is not present in the test has no meaning to it).

There are still a few minor things to be addressed:
- [x] Fix pipeline unit tests
- [x] Add var unit tests
- [x] Refactor the extensible parameter spaces to use the same base classes as the variables
- [x] Add the docs
- [x] Implement a namespace management to track all attributes introduced through directives

### This PR will also:
- Make illegal any name clashes between variables, parameters and other existing class attributes in the regression test.
- Make illegal to call more than one directive on the same var in the same class (the three var directives have excluding actions).
- Make illegal to redeclare a variable. 
- Make illegal for a test to inherit from multiple classes derived from the `RegressionTest` class (direct consequence of the previous item).
- Introduce the `RegressionMixin` class. This class supports reframe hooks and directives, solving the limitation introduced by the previous item.

Closes #1568.


jjotero:
@jenkins-cscs retry all

vkarak:
In the example with the `Eggs` class, I guess that the users wouldn't even have to call `super.__init__()` if the base class only declared or (re)set some variables, right?

reframe/core/variables.py
```diff
@@ -0,0 +1,217 @@
...
+    def join(self, other):
+        '''Join an existing VarSpace into the current one.'''
+        for key, var in other.items():
+
+            # Make doubly declared vars illegal. Note that this will be
+            # triggered when inheriting from multiple RegressionTest classes.
+            if key in self.vars:
+                raise ValueError(
+                    f'attribute {key!r} is declared in more than one of the '
+                    f'parent classes'
```
vkarak, line 151: "of class XX" Is it possible? Perhaps we could store this information.

reframe/core/variables.py
```diff
@@ -0,0 +1,217 @@
...
+        :param obj: The test object.
+        :param cls: The test class.
+        '''
+
+        for name, var in self.items():
+            setattr(cls, name, var.field(*var.types))
+            getattr(cls, name).__set_name__(obj, name)
+
+            # If the var is defined, set its value
+            if not var.is_undef():
```
vkarak, line 209: I probably prefer this.

```suggestion
            if var.is_defined():
```

reframe/core/variables.py
```diff
@@ -0,0 +1,217 @@
...
+                setattr(obj, name, var.value)
```
vkarak, line 210: Should we call the `value`, `default_value`, because this is what it is, isn't it?

reframe/core/variables.py
```diff
@@ -0,0 +1,217 @@
...
+
+    @property
+    def vars(self):
+        return self._attr
+
+    def undefined_vars(self):
+        return list(filter(lambda x: self.vars[x].is_undef(), self.vars))
```
vkarak, line 217: Better write this as a list comprehension.

vkarak (changes requested):
I have mostly suggestions for better names. Overall the implementation is pretty good! I will look at the unit tests and the documentation in second round.

reframe/core/variables.py
jjotero: It is the value that gets assigned to the variable when the class gets instantiated, so I guess we could call it `default_value`. Though, I think it still makes sense to keep the API as
```python
    var('my_var', int, value=10)
    ...
```

reframe/core/variables.py
jjotero: This is not used anywhere, so I'm taking it out.

jjotero (commented):
> In the example with the `Eggs` class, I guess that the users wouldn't even have to call `super.__init__()` if the base class only declared or (re)set some variables, right?

Correct.

I've updated the PR with your suggestions:

- The `attributes.py` along all the base classes in this file have been renamed to `namespace`. Also, the word `attribute` has been replaced by either `variable` or `parameter` when raising an exception.

- ~~Then only main issue that I see left in here is the docs from the `RegerssionTestClass`.~~(see below)

jjotero:
The docs are now working (🎉). The trick has been pulled off by updating the `_TestVar` class to have a `__set_name__` method. Then, the class is made available in the `RegressionTest` class and we can now simply create a new variable doing
```python
    my_var = _var('/', int, value=3)
    def __init__(self):
        pass
```
where the `'/'` argument is just a placeholder for the variable name, which gets later replaced by the `__set_name__` method anyways. This is equivalent to 
```python
class Bar(rfm.RegressionTest):
    var('my_var', int, value=3)
    def __init__(self):
        pass
```
Now, I do like a lot more the first option. So if we eliminate the `var('my_var', ...)` syntax and stick to the one in the top, there's no need for the name placeholder and the whole thing would look as follows
```python
    my_var = var(int, value=3)
    p0 = parameter(1,2,3,5,8,13)

    def __init__(self):
        pass
```
For the same money, we can also change the `parameter` directive as shown in the snippet above. This would be only for the directives that declare a parameter or a variable; the `set_var` and `require_var` would remain the same. Thoughts?

vkarak:
I was about to finish my review and I was right at the moment that was wondering what the heck is this new `_var` 😂 Nice!

vkarak:
I don't a problem with this syntax, except I would still prefer the `parameter()` to get an iterable of values.

vkarak:
One obvious downside is that it breaks completely the already introduced syntax for parameters, but since it's so fresh and hasn't yet become the way to go for parameters, I can accept it.

unittests/test_variables.py
```diff
@@ -0,0 +1,152 @@
...
+    try:
+        inst = MyTest()
+    except Exception:
+        pytest.fail('class instantiation failed')
+
+
+def test_require_undef(novars):
+    with pytest.raises(ValueError):
+        class MyTest(novars):
+            require_var('foo')
```
vkarak, line 143: Hmm, if a variable is required, does it have to be set in the class? Couldn't it be set in the `__init__()`?

unittests/test_variables.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def test_required_var_not_present(onevar):
+    class MyTest(onevar):
+        require_var('foo')
+
+        def __init__(self):
+            pass
+
+    try:
```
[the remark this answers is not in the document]
vkarak: I also don't understand how this should pass. Isn't `foo` required here?

docs/regression_test_api.rst
```diff
@@ -87,6 +87,40 @@ For instance, continuing with the example above, one could override the :func:`_
...
+   If the ``value`` argument is not provided, the variable is considered *declared* but not *defined*.
+   Thus, a variable may not be declared more than once. However, it is possible to alter a variable's value after it was declared by using the :func:`set_var` and :func:`require_var` directives.
+   Note that a variable must be defined before is referenced in the regression test. Otherwise, an :py:exc:`AttributeError` will be raised.
+
+   :param name: the variable name.
+   :param types: the supported types for the variable.
+   :param value: the default value assigned to the variable.
+   :param field: the field validator to be used for this variable.
+        If no field argument is provided, it defaults to
+        :class:`reframe.core.fields.TypedField`.
```
vkarak, line 102: This interface is not exposed to the users, so we will need to do so. I see it relevant for test library developers, but we should expose only a subset of the fields.

docs/regression_test_api.rst
```diff
@@ -87,6 +87,40 @@ For instance, continuing with the example above, one could override the :func:`_
...
+        Note that the field validator provided by this argument must derive from
+        :class:`reframe.core.fields.Field`.
+
+
+.. py:function:: reframe.core.pipeline.RegressionTest.set_var(name, value)
+
+   Assign a value to a regression test variable. The variable must have been defined in a parent class using the :func:`var` directive.
```
vkarak, line 109:
```suggestion
   Assign a default value to a regression test variable. The variable must have been defined in a parent class using the :func:`var` directive.
```

vkarak: Which why not call it `set_default()`? `set_var()` might confuse people thinking that it is the only way to assign values to these variables.

docs/regression_test_api.rst
```diff
@@ -87,6 +87,40 @@ For instance, continuing with the example above, one could override the :func:`_
...
+
+   :param name: the variable name.
+   :param value: the value assigned to the variable.
+
+
+.. py:function:: reframe.core.pipeline.RegressionTest.require_var(name)
+
+   Undefine a regression test variable. The variable must have been defined in a parent class using the :func:`var` directive.
```
vkarak, line 117:
```suggestion
   Make variable `name` required for this class and any descendants.
```

Then you can talk about the "undefining" etc., which I think is an implementation detail.

docs/regression_test_api.rst
```diff
@@ -87,6 +87,40 @@ For instance, continuing with the example above, one could override the :func:`_
        This only has an effect if used with ``inherit_params=True``.
 
 
+.. py:function:: reframe.core.pipeline.RegressionTest.var(name, *types, value=None, field=None)
+
+   Inserts a new regression test variable.
+   If the ``value`` argument is not provided, the variable is considered *declared* but not *defined*.
```
vkarak, line 93: This might confuse the readers. I would probably omit it. And we should say that the `value` is the default value assigned to it.

docs/regression_test_api.rst
```diff
@@ -87,6 +87,40 @@ For instance, continuing with the example above, one could override the :func:`_
...
+   Thus, a variable may not be declared more than once. However, it is possible to alter a variable's value after it was declared by using the :func:`set_var` and :func:`require_var` directives.
+   Note that a variable must be defined before is referenced in the regression test. Otherwise, an :py:exc:`AttributeError` will be raised.
```
vkarak, line 95: I see a problem with this, which I think I realised when I was going through the unit tests.  Is the following valid? If not, I think we are making the "required" attribute impractical.

```python
class MyLibTest(rfm.RegressionTest):
    required_var('num_tasks')

class MyTest(MyLibTest):
    def __init__(self):
        self.num_tasks = 10  # and generally take a decision here
```

If I understand correctly, I would have to write my test as follows:

```python
class MyTest(MyLibTest):
    set_var('num_tasks', 1) # new default
    def __init__(self):
        self.num_tasks = 10  # and generally take a decision here
```

vkarak (changes requested):

unittests/test_variables.py
```diff
@@ -0,0 +1,152 @@
...
+    class NoVars(rfm.RegressionTest):
+        pass
+    yield NoVars
+
+
+@pytest.fixture
+def onevar(novars):
+    class OneVar(novars):
+        var('foo', int, value=10)
+    yield OneVar
```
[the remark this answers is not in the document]
jjotero: > Do these have to be fixtures? Why not like this:
> 
> ```python
> class OneVarTest(NoVarTest):
>     var('foo', int, value=10)
> ```

Same reason as stated in the first comment in the pipeline unit tests. 

> It's also a bit misleading as a styling convention to use full lowercase for something that is eventually a class.

Changed it to `OneVarTest` and so on.

unittests/test_variables.py
```diff
@@ -0,0 +1,152 @@
...
+    assert isinstance(onevar.foo, Field)
+    assert hasattr(inst, 'foo')
+    assert inst.foo == 10
+
+
+def test_instantiate_and_inherit(novars):
+    inst = novars()
+    with pytest.raises(ValueError):
+        class MyTest(novars):
+            '''Error from name clashing'''
```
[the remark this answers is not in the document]
jjotero: See above my first reply in the pipeline unit tests.

unittests/test_variables.py
```diff
@@ -0,0 +1,152 @@
...
+    with pytest.raises(ValueError):
+        class MyTest(rfm.RegressionTest):
+            var('v0', int, value=1)
+            var('v0', float, value=0.5)
+
+
+def test_double_action_on_var():
+    with pytest.raises(ValueError):
+        class MyTest(rfm.RegressionTest):
+            set_var('v0', 2)
```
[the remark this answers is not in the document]
jjotero: No, that's checked below by the test `test_set_undef`. This just tests that a double action on a var is not allowed in the same class body.

unittests/test_variables.py
```diff
@@ -0,0 +1,152 @@
...
+    inst = novars()
+    with pytest.raises(ValueError):
+        class MyTest(novars):
+            '''Error from name clashing'''
+
+
+def test_redeclare_var_clash(novars):
+    with pytest.raises(ValueError):
+        class MyTest(novars):
+            var('name', str)
```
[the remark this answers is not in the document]
jjotero: I see. I've renamed this test to `test_redeclare_builtin_var_clash` and created another one more explicit for `test_redeclare_var_clash`.

unittests/test_variables.py
jjotero: Yeah. But this test checks that you can't call a `require_var` directive on an undeclared var. I see how the name of the test is not the best one. I've now changed that to `test_require_undeclared_var`.

unittests/test_variables.py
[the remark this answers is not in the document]
jjotero: > You don't need the `try/except` here if this is expected to work. If it throws an exception, the test will fail in any case.

Removed the `try/except` block.

> I also don't understand how this should pass. Isn't `foo` required here?

`foo` is required, but it is not referenced in the entire test, so this is all good. If a variable is required, the exception is thrown by the descriptor itself, because otherwise a required variable could be set in a hook at some point and then the value is already in there when the required var's value is accessed.
The only way I see we could be a bit more strict with this is to check at the very end of the test if the variable has been set or not and throw the error there if needed. Though, I'm not sure what the point of that would be, because that means that the test made it all the way to the end without any issues.

reframe/core/pipeline.py
```diff
@@ -189,13 +205,13 @@ def pipeline_hooks(cls):
...
     #: :default: ``[]``
-    valid_systems = fields.TypedField(typ.List[str], type(None))
+    var('valid_systems', typ.List[str], type(None), value=None)
 
     #: A detailed description of the test.
     #:
     #: :type: :class:`str`
     #: :default: ``self.name``
-    descr = fields.TypedField(str)
+    var('descr', str)
```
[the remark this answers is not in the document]
jjotero: I guess this is sorted out now?

jjotero (commented):
The main points standing in from these comments are:
- The constructor arguments to the `_TestVar` class and how they interact with the different fields.
- Variable space name clashing check.

I vote for polishing up the final syntax for all these directives once we get these points are done.

unittests/test_variables.py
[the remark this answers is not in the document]
vkarak: Yes, but my question is if `set_var()` cannot be before `var()`, then you are not sure the error is set because of the double action. Shouldn't it be `var()` first and `set_var()` afterwards?

unittests/test_pipeline.py
```diff
@@ -30,9 +28,25 @@ def _run(test, partition, prgenv):
     test.cleanup(remove_files=True)
 
 
-def load_test(testfile):
-    loader = RegressionCheckLoader(['unittests/resources/checks'])
-    return loader.load_from_file(testfile)
+@pytest.fixture
+def hellotest():
+    from unittests.resources.checks.hellocheck import HelloTest
+    yield HelloTest
```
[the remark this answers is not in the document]
vkarak: > because the vars are inserted as descriptors, they become class attributes after the first class instantiation.

Actually, it's because they are inserted as descriptors too late, at the final object instantiation, which is also something you wouldn't expect for a descriptor. Couldn't they be inserted earlier, during the class creation? Is it just for uniformity with parameters? Parameters needed to be injected at object creation, but why the variable descriptors? 

This behaviour is *really* counter-intuitive and I'm afraid we will have to face it the hard way sometime in the future.

I've also seen the following "inconsistency". If I do the following in the `HelloTest`

```python
@rfm.simple_test
class HelloTest(rfm.RegressionTest):
    var('name', int)

```

I correctly get the `re-declaration` error, even if I disable the sanity checking. However, this is not case for other variables, e.g., `prefix`, `current_environ` etc. (which is the expected behaviour with the current implementation if sanity checking is disabled).

unittests/test_variables.py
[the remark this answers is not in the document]
vkarak: > If a variable is required, the exception is thrown by the descriptor itself

You mean that the descriptor is not defined, so you will get an `AttributeError` or something, right? Which brings us to the point of the counter-intuitive inheritance behaviour discussed above, because in order to get this behaviour here, we need to define the descriptor too late, at the object creation. Perhaps, your other solution with the `__set__` counter inside the descriptor, for which I was not too much in favor at the beginning (😬) might be a better solution, because it could solve both problems.

Although the pattern of instantiating and then inheriting is not something that is done in any test, it's really scary that this implementation does not allow it! Even if you accidentally do it, you would be really wondering what's wrong with my test! Just adding this in the `hellocheck.py`, is enough to block everything:

```python
t = rfm.RegressionTest()
```

And the error message doesn't help.. 

```
unittests/resources/checks/hellocheck.py:13: 'name' already defined in class 'HelloTest'
```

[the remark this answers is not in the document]
vkarak: Also playing a bit with the `require_var`, I'm further less in favour of getting an `AttributeError`. Look at this:

```python
@rfm.simple_test
class HelloTest(rfm.RegressionTest):
    require_var('modules')

    def __init__(self):
        self.x = self.modules
        ...
```

```
./bin/reframe -c unittests/resources/checks/hellocheck.py -l
```

```
./bin/reframe: skipping test due to errors: HelloTest: use `-v' for more information
  FILE: /Users/karakasv/Repositories/reframe/unittests/resources/checks/hellocheck.py:25
[List of matched checks]
- CompileOnlyHelloTest (found in '/Users/karakasv/Repositories/reframe/unittests/resources/checks/hellocheck.py')
Found 1 check(s)
```

If I do `-v` I get an `AttributeError` for `modules`, which is not the best. I would prefer something like `you must set the 'modules' attribute`.

unittests/test_pipeline.py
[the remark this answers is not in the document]
jjotero: > Couldn't they be inserted earlier, during the class creation? 

This would bring us back to square 1. 
The reason why the variables get inserted at the very end boils down to getting more control on the vars and their interaction with the user. This is unrelated with whether they're descriptors or not. If you insert the vars during class creation, a user could do
```python
    num_tasks = 'bananas'
    def __init__(self):
        ...
```
and that's enough to break everything. If you insert the variables at the end, you know that the user has messed up and then you can throw the right error. There are now unit tests in `test_variables.py` to test for an error when overriding a builtin variable (`name`) and builtin attributes (`current_environ`).

The "instantiate and inherit" issue you pointed out above has now been fixed by making the variable space a bit smarter. When the variables get injected into the class, the variable space now tracks which variables have been injected, and it knows there can be no conflict with those variables at any point onwards.

unittests/test_variables.py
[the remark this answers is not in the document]
jjotero: I see your point. I've changed the order.

However, in this case the outcome would be the same regardless of the order. But I do see how can it be confusing from the unit test point of view.

jjotero (commented):
Regarding your comment (which for some reason GitHub did not show in my review)

> Also playing a bit with the require_var, I'm further less in favour of getting an AttributeError. Look at this:
> ```python
> @rfm.simple_test
> class HelloTest(rfm.RegressionTest):
>     require_var('modules')
> 
>     def __init__(self):
>         self.x = self.modules
>         ...
>```
>```
> ./bin/reframe -c unittests/resources/checks/hellocheck.py -l
> ./bin/reframe: skipping test due to errors: HelloTest: use `-v' for more information
>   FILE: /Users/karakasv/Repositories/reframe/unittests/resources/checks/hellocheck.py:25
> ```
> If I do -v I get an AttributeError for modules, which is not the best. I would prefer something like you must set the 'modules' attribute.

The issue here is that reframe can set an attribute in the test at any stage of the test execution. So in the case of the variables, the test must insert the descriptor into the class from the beginning, passing on all controls to the descriptor itself. But in any case, isn't an `AttributeError` the way python tells you that you must set the attribute?

reframe/core/pipeline.py
```diff
@@ -126,13 +126,31 @@ def _wrapped(*args, **kwargs):
...
+
 class RegressionTest(jsonext.JSONSerializable, metaclass=RegressionTestMeta):
     '''Base class for regression tests.
 
     All regression tests must eventually inherit from this class.
     This class provides the implementation of the pipeline phases that the
     regression test goes through during its lifetime.
 
+    .. warning::
+        .. versionchanged:: 3.5
```
jjotero, line 151: Just marking this one to make sure we also update the version here.

reframe/core/pipeline.py
```diff
@@ -569,24 +596,24 @@ def pipeline_hooks(cls):
...
+    modules = _var('/', typ.List[str], value=[])
 
     #: Environment variables to be set before running this test.
     #:
     #: These variables will be set during the :func:`setup` phase.
     #:
     #: :type: :class:`Dict[str, str]`
     #: :default: ``{}``
-    variables = fields.TypedField(typ.Dict[str, str])
+    variables = _var('/', typ.Dict[str, str], value={})
```
jjotero: We need to deprecate this and change it to `env_variables` or something like that.

unittests/test_variables.py
```diff
@@ -22,12 +22,12 @@ class NoVarsTest(rfm.RegressionTest):
 @pytest.fixture
 def OneVarTest(NoVarsTest):
```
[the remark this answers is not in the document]
vkarak: Apparently not, because I tried it and it passes, so I suggest reverting that part there as well.

vkarak (changes requested):
This is great now! 🎉  My comments are mostly about polishing the implementation. Before merging, though, we need to check again the documentation and do an extensive run with the full production test suite, so that we are sure that nothing is broken.

reframe/core/pipeline.py
vkarak: That's for 4.0 😄 It will be sooo disruptive. Nonetheless, I suggest opening an issue about it. We can collect such disruptive changes to bundle them for the next major bump.

docs/regression_test_api.rst
vkarak: This is awesome!

docs/regression_test_api.rst
vkarak: Agreed for both.

reframe/core/pipeline.py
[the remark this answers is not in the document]
vkarak: Yep, new syntax is pretty nice.

jjotero (commented):
The main part of the changes are in the metaclass:

-  Made the `MetaNamespace` a subclass of `RegressionTestMeta`
-  I've replaced the `dir` call with something more robust
-  Implemented a `__getattribute__` method.

I'll push some updates to the docs in a bit.

unittests/test_pipeline.py
```diff
@@ -615,18 +615,6 @@ class MyTest(rfm.RunOnlyRegressionTest, HelloTest):
...
-def test_extend_after_instantiation(HelloTest):
-    '''Instantiation will inject the vars as class attributes.
-
-    Therefore, inheriting from this class after the instantiation will create
-    a namespace clash with the vars.
-    '''
-    hellotest = HelloTest()
-    with pytest.raises(ValueError):
-        class MyTest(HelloTest):
-            pass
```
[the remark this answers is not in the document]
jjotero: I've included a test in the `test_variables.py` to make sure this always works.

reframe/core/pipeline.py
jjotero: Issue created: https://github.com/eth-cscs/reframe/issues/1774

reframe/core/meta.py
```diff
@@ -58,7 +58,12 @@ def __init__(cls, name, bases, namespace, **kwargs):
         super().__init__(name, bases, namespace, **kwargs)
 
         # Create a set with the attribute names already in use.
-        used_attribute_names = set(dir(cls)) - set(cls.__dict__)
+        cls._rfm_dir = set()
+        for base in bases:
+            if hasattr(base, '_rfm_dir'):
+                cls._rfm_dir.update(base._rfm_dir)
+
+        used_attribute_names = copy.deepcopy(cls._rfm_dir)
```
vkarak, line 66: This should be enough, shouldn't be? `_rfm_dir` stores only strings, which are immutable, so I wouldn't go through the machinery of `copy.deepcopy()`.

```suggestion
        used_attribute_names = set(cls._rfm_dir)
```

reframe/core/pipeline.py
```diff
@@ -760,6 +760,22 @@ def __new__(cls, *args, _rfm_use_params=False, **kwargs):
...
+            # Intercept the AttributeError if the name corresponds to a
+            # required variable.
+            if (name in self._rfm_var_space.vars and
+                not self._rfm_var_space.vars[name].is_defined()):
+                raise AttributeError(
+                    f'required variable {name!r} has not been defined'
+                ) from None
+
+            else:
+                super().__getattr__(name)
```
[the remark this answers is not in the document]
vkarak: You are right, because this way you respect the `__getattr__` semantics. My comment was because you were already handling the `AttributeError` from `__getattribute__`, so I was like how could `__getattr__` raise it again. But you are right, the superclass might define it, so we should call it.

reframe/core/variables.py
```diff
@@ -206,32 +138,59 @@ def extend(self, cls):
...
+                if key in self.vars:
+                    raise ValueError(
+                        f'cannot redeclare a variable ({key})'
+                    )
+
+                # Add a new var
+                self.vars[key] = var
+
+            elif isinstance(var, VarDirective):
+                # Modify the value of a previously declarated var
```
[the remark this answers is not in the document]
vkarak: I prefer the `my_var = 5` syntax. Imho, it's not a problem if the variable is created when it doesn't exist. Let's not overdo it. It's Python after all :-)

docs/regression_test_api.rst
```diff
@@ -36,71 +36,97 @@ Pipeline Hooks
 .. autodecorator:: reframe.core.decorators.require_deps
 
 
-.. _directives:
 
-Directives
-----------
+Builtins
+--------
```
vkarak, line 42: Add a `.. versionadded:: 3.5` here.

docs/regression_test_api.rst
```diff
@@ -36,71 +36,97 @@ Pipeline Hooks
...
-        :class:`reframe.core.fields.TypedField`.
-        Note that the field validator provided by this argument must derive from
-        :class:`reframe.core.fields.Field`.
+    class Bar(Foo):
+        my_var = 4
+
+        def __init__(self):
+            print(self.my_var) # prints 4.
+
+  A variable may be set as required by simply assigning a variable as `required`.
```
vkarak, line 112:
```suggestion
  A variable may be set as required by simply assigning a variable as ``required``.
```

docs/regression_test_api.rst
```diff
@@ -36,71 +36,97 @@ Pipeline Hooks
...
+  Note that a variable must be defined before is referenced in the regression test. Otherwise, an :py:exc:`AttributeError` will be raised.
```
vkarak, line 113: This can be a bit confusing and have users declaring every variable that they would normally define with a `self.my_var` in their test. What I find missing in this section of the documentation is when somebody should use the `variable()` built-in and what does it offer?

docs/regression_test_api.rst
```diff
@@ -36,71 +36,97 @@ Pipeline Hooks
...
-For example, a test can be parameterized using the :func:`parameter` directive as follows:
+ReFrame provides built-in functions that facilitate the creation of extensible tests (i.e. a test library).
+These *builtins* are intended to be used directly in the class body of the test, allowing the ReFrame internals to *pre-process* their input before the actual test creation takes place.
+This provides the ReFrame internals with further control over the user's input, making the process of writing regression tests less error-prone thanks to a better error checking.
+In essence, these builtins exert control over the test creation, and they allow adding and/or modifying certain attributes of the regression test.
 
-.. code:: python
+.. py:function:: reframe.core.pipeline.RegressionTest.parameter(values=None, inherit_params=False, filter_params=None)
+
+  Inserts or modifies a regression test parameter.
```
vkarak, line 50: In the documentation, I am missing a bit -as a user- the connection with the `@parameterized_test`. What are the differences and when should I use this syntax over the other. We recently came across an example with @ekouts that the `@parameterized_test` would be at least as useful:

```python
@rfm.parameterized_test([1, 2], [3, 4])
class X(rfm.RegressionTest):
    def __init__(self, a, b):
       pass
```

With the parameter syntax, I think that you have to combine the two parameters into one, since we are only looking for two points in the parameter space:

```python
class X(rfm.RegressionTest):
    ab = parameter([1, 2], [3, 4])
    def __init__(self):
        self.a, self.b = self.ab
```

Is a better way to write this with the `parameter()` built-in ?

vkarak (changes requested):
I have only very minor comments and some suggestions about the documentation. Overall, it's almost ready to 🚀 

> on every test (which is longer than the fixtures 😢). Otherwise the class attributes from one test might end up in the next tests. This may or may not affect the test outcome, but I think it's a good practice if every test gets a fresh class.

Indeed, it's better to get a fresh class, but it's better if we comment it in the fixture somehow. My original suggestion would be to have the classes for the lifetime of the module.

> The way I see it, the same reasoning applies here. I think that a fixture should always clean after itself, and these would not do so unless the module is deleted. Otherwise, I think we should change the scope of the fixture to module and not function.

Agreed.

> This didn't pass because the variables and the parameters go straight into the cls._rfm_local_var_space and cls._rfm_local_param_space. It's only when the class gets instantiated that they get injected in there. So to fix this counter-intuitive behaviour, I've added the __getattribute__ method to the metaclass, which would also look into these local param and variable spaces.

Correct, because of the `MetaNamespace`. That also explains now the `__getattribute__` in the metaclass, so I would suggest to add a comment there.

vkarak:
@jjotero I updated the branch and I'm gonna trigger a full production run.

vkarak:
There seems to be a strange side effect:

```
./bin/reframe -C config/cscs.py -c cscs-checks/ -n MemoryOverconsumptionCheck -r
```

```
[  FAILED  ] Ran 2/2 test case(s) from 1 check(s) (2 failure(s))
```

And this generates this in the build script:

```bash
module load daint-gpu
module load PrgEnv-cray
cd posts/cuda-aware-mpi-example/src
cc eatmemory.c -o ./MemoryOverconsumptionCheck
```

Which is apparently very strange :-D

reframe/core/meta.py
jjotero: Yes, good point. I'll change that.

docs/regression_test_api.rst
jjotero: Agree. I've added quite a bit more of text in this section.

jjotero (commented):
I've added some more comments into the metaclass and also extended the docs.

Before we move forward with this, I think we can already address the naming issue in this PR for the parameterized tests with the built-in `parameter` (see comment).

jjotero:
> There seems to be a strange side effect:
> 
> ```
> ./bin/reframe -C config/cscs.py -c cscs-checks/ -n MemoryOverconsumptionCheck -r
> ```
> 
> ```
> [  FAILED  ] Ran 2/2 test case(s) from 1 check(s) (2 failure(s))
> ```
> 
> And this generates this in the build script:
> 
> ```shell
> module load daint-gpu
> module load PrgEnv-cray
> cd posts/cuda-aware-mpi-example/src
> cc eatmemory.c -o ./MemoryOverconsumptionCheck
> ```
> 
> Which is apparently very strange :-D

😱😱😱😱😱😱😱😱😱😱
I just did `./bin/reframe -C config/cscs.py -c cscs-checks/system/slurm/slurm.py -r` and it ran fine 🤔🤯

jjotero:
The issue with the `MemoryOverconsumptionCheck ` had its origin in the `CudaAwareMPICheck`, which had a `+=` on the `prebuild_cmds` variable.
```python
self.prebuild_cmds += ['cd posts/cuda-aware-mpi-example/src']
```
This changed default value of `prebuild_cmds`, which is a reference to the very default value set in the `RegressionTest` class. This means that test that run after this is executed, will also see this "modified" `prebuild_cmds` variable. 

All this would not be a problem if all tests do the first assignment as a new list, but that is simply too dangerous. So to fix this, the values that get injected into the variables are now deep-copies of the original class-default values, which totally removes this cross-contamination in the tests. The same could have happened for the parameters, so another fix has been implemented in there. To make sure these changes stick, I've added two new unit tests that check this for the variables and parameters.

I've also improved the docs on the built-in section.

reframe/core/namespaces.py
```diff
@@ -0,0 +1,177 @@
...
+        return f'{type(self).__name__}({self._namespace!r})'
+
+    def _raise_namespace_clash(self, name):
+        '''Raise an error if there is a namespace clash.'''
+        raise ValueError(
+            f'{name!r} is already present in the current namespace'
+        )
+
+
+class Namespace:
```
jjotero, line 67: I'm adding the `metaclass=abc.ABCMeta` here. Otherwise the abstract methods below give a very strange error if they're not overridden.

reframe/core/parameters.py
```diff
@@ -145,76 +78,121 @@ class ParamSpace:
...
         '''Create a generator object to iterate over the parameter space
 
+        The parameters must be deep-copied to prevent an instance from
+        modifying the class parameter space.
+
         :return: generator object to iterate over the parameter space.
         '''
-        yield from itertools.product(*(p for p in self._params.values()))
+        yield from itertools.product(
+            *(copy.deepcopy(p) for p in self.params.values())
```
jjotero, line 190: Added the deep-copy to the parameters.

reframe/core/variables.py
```diff
@@ -0,0 +1,216 @@
...
+    def undefine(self):
+        self._default_value = _Undefined
+
+    def define(self, value):
+        self._default_value = value
+
+    def __set_name__(self, owner, name):
+        self.name = name
+
+    @property
```
jjotero, line 63: Adding the deep-copy to the variables

unittests/test_parameters.py
```diff
@@ -168,3 +169,62 @@ def test_parameterized_test_is_incompatible():
...
+
+
+def test_overwrite_param():
+    with pytest.raises(ValueError):
+        class MyTest(TwoParams):
+            P0 = [1, 2, 3]
+
+
+def test_param_deepcopy():
+    '''Test that there is no cross-class pollution.
```
jjotero, line 209: New test to check the deep-copy

unittests/test_variables.py
```diff
@@ -0,0 +1,186 @@
...
+def test_invalid_field():
+    class Foo:
+        '''An invalid descriptor'''
+
+    with pytest.raises(ValueError):
+        class MyTest(rfm.RegressionTest):
+            a = variable(int, value=4, field=Foo)
+
+
+def test_var_deepcopy():
```
jjotero, line 166: Added new test to check the deep-copy on the variables

docs/regression_test_api.rst
vkarak: The naming should certainly change, but let's not do in this PR. The current naming scheme will work even for lists, but it will produce a very ugly output, something like this for this example: `X_1__2_` when `ab = [1, 2]` 😬

vkarak: I suggest opening an issue not forget about updating the naming scheme. I believe that all these changes will bring us to 4.0 :-)

docs/regression_test_api.rst
```diff
@@ -86,42 +111,86 @@ In essence, these builtins exert control over the test creation, and they allow
...
 
         def __init__(self):
             print(self.my_var) # prints 8.
+            # self.my_var = 'override' # Error: my_var must be an int!
+            self.not_a_var = 'override' # However, this would work. Dangerous!
 
-  The argument ``value`` sets the default value for the variable.
-  A variable may not be declared more than once. However, it is possible to alter a variable's value after it was declared by simply assigning it a new value directly in the class body.
+  The argument ``value`` in the :func:`variable` built-in sets the default value for the variable.
+  As mentioned above, a variable may not be declared more than once, but its value can be updated by simply assigning it a new value directly in the class body.
```
vkarak, line 134:
```suggestion
  As mentioned above, a variable may not be declared more than once, but its default value can be updated by simply assigning it a new value directly in the class body.
```

Users might think that this is the only way to update a variable, whereas they can do that in their tests as well. I think we should show that, too. Perhaps by simply adding another assignment to the example above:

```python
self.my_var = 10 # tests may also assign values the standard way
```

docs/regression_test_api.rst
```diff
@@ -86,42 +111,86 @@ In essence, these builtins exert control over the test creation, and they allow
...
         def __init__(self):
             print(self.my_var) # prints 4.
 
-  A variable may be set as required by simply assigning a variable as `required`.
-  Note that a variable must be defined before is referenced in the regression test. Otherwise, an :py:exc:`AttributeError` will be raised.
+  Here, the class :class:`Bar` inherits the variables from :class:`Foo` and can see that ``my_var`` has already been declared in the parent class. Therefore, the value of ``my_var`` is updated ensuring that the new value complies to the original variable declaration.
+
+  These examples above assumed that a default value can be provided to the variables in the bases tests, but that might not always be the case.
+  For example, when writing a test library, one might want to leave some variables undefined and force the user to set these when using the test.
+  As shown in the example below, imposing such requirement is as simple as not passing any ``value`` to the :func:`variable` built-in, which marks the given variable as ``required``.
```
vkarak, line 149: Here "required" is not the keyword, so just emphasize it.

```suggestion
  As shown in the example below, imposing such requirement is as simple as not passing any ``value`` to the :func:`variable` built-in, which marks the given variable as *required*.
```

docs/regression_test_api.rst
vkarak: I see that you mention it later, but it wouldn't hurt here, too. I leave it up to you.

vkarak (changes requested):
I've only cosmetic suggestions and it's ready to go. I will also give it a try with a full production run.

unittests/test_variables.py
```diff
@@ -157,3 +161,26 @@ class Foo:
...
+    '''Test that there is no cross-class pollution.
+
+    Each instance must have its own copies of each variable.
+    '''
+    class MyType:
+        def __init__(self, val):
+            self.val = val
```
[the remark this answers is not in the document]
jjotero: Yeah, this was just a modified copy from the equivalent test in `test_parameters.py`. Changed `MyType` to a list.

jjotero (commented):
Did the full production run pass?

reframe/core/parameters.py
```diff
@@ -107,8 +108,8 @@ def join(self, other, cls):
             # could be doubly defined and lead to repeated
             # values
             if (key in self.params and
-                self.params[key] != () and
-                other.params[key] != ()):
+                    self.params[key] != () and
+                    other.params[key] != ()):
```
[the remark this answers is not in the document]
vkarak: We ignore this ;-)

Check the [coding style guide](https://github.com/eth-cscs/reframe/wiki/coding-style-guide) for a complete list of the PEP8 errors that we ignore in order to set up correctly your style checker.

vkarak (approved):
Perfect!  🎉  Just fix the conflicts with the master and change all the `versionadded` annotations to 3.4.2, because it's going in today!

vkarak:
For the conflicts, just keep your version. I have simply fixed the styling in a previous PR.

reframe/core/pipeline.py
```diff
@@ -135,6 +135,8 @@ class RegressionMixin(metaclass=RegressionTestMeta):
     the metaclass magic implemented in
     :class:`RegressionTestMeta`. Using this metaclass allows mixin classes to
     use powerful ReFrame features, such as hooks, parameters or variables.
+
+    .. versionadded:: 3.5
```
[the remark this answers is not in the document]
jjotero: Updated to `3.4.2`

reframe/core/pipeline.py
jjotero: Updated to `3.4.2`

jjotero (commented):
Updated version notes to `3.4.2`

vkarak:
> Did the full production run pass?

Yes. The failures were just ordinary test failures, not related to framework problems.

victorusu (approved):
`lgtm`

vkarak:
@jenkins-cscs retry daint

PR #1701, opened 2021-01-21
[bugfix] Ensure that job options passed from the command-line are always the last in the generated job script

### LINKED ISSUE #1700: `self.job.options` override any job options passed from the command-line
Currently, if I have a test that does:

```python
@rfm.run_before('run')
def set_job_options(self):
    self.job.options = ['--foo']
```

And then run reframe with `-J bar=X`, the `--bar=X` will never be passed to the job script.

rsarm:
Closes #1700

### DIFF
--- a/reframe/core/schedulers/__init__.py
+++ b/reframe/core/schedulers/__init__.py
@@ -180,7 +180,7 @@ class Job(jsonext.JSONSerializable):
         self.num_cpus_per_task = None
         self.use_smt = None
         self.time_limit = None
-        self.sched_options = list(sched_options) if sched_options else []
+        self.cli_options = list(sched_options) if sched_options else []
         self.options = []
 
         self._name = name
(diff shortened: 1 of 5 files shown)


teojgo (approved):
lgtm

reframe/core/schedulers/__init__.py
```diff
@@ -180,7 +180,8 @@ def __init__(self,
         self.num_cpus_per_task = None
         self.use_smt = None
         self.time_limit = None
-        self.options = list(sched_options) if sched_options else []
+        self.sched_options = list(sched_options) if sched_options else []
```
vkarak, line 183: `sched_options` is a bit confusing as a name here. What about `cli_options`?

vkarak (changes requested):
Lgtm except that we should probably use a better name than `sched_options` inside the `Job`.

vkarak (approved):
lgtm

PR #1704, opened 2021-01-22
[feat] Support multiple job submissions when slurm's job submit limit is reached

### LINKED ISSUE #1595: job submit limits don't count other user jobs that may already be in the queue

ekouts:
Second, much simpler implementation of #1672. In this PR the submission blocks until slurm stops giving the `QOSMaxSubmitJobPerUserLimit` error. This could cause significant slowdown of the jobs.
Closes #1595.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1006,7 +1006,7 @@ Common scheduler options
    In such cases, you may set this parameter to ``true`` to avoid this.
 
 
-.. js:attribute:: .schedulers[].resubmit_on_qos_errors
+.. js:attribute:: .schedulers[].resubmit_on_errors
 
    :required: No
    :default: ``[]``
@@ -1014,10 +1014,11 @@ Common scheduler options
    This option is relevant to the Slurm backends only.
 
    When a job is submitted certain errors can be ignored and the framework will try to submit again the job after some seconds.
-   ReFrame is checking with a regular expression if any of the given expressions is contained in the ``stderr`` of the submission command.
-   Keep in mind that the submission is blocking and ReFrame will not continue until the submission is successful or has a different error.
+   An example of this could be QOS errors like ``QOSMaxSubmitJobPerUserLimit``, in which case you would have to set the option to ``["QOSMaxSubmitJobPerUserLimit"]``. You can ignore multiple errors at the same time if you include all the error strings in the list.
+   Job submission is a synchronous operation in ReFrame.
+   If this option is set, it will block the whole execution until the error conditions specified in this list are raised.
 
-   .. versionadded:: 3.4
+   .. versionadded:: 3.5
 
 
 Execution Mode Configuration
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1006,6 +1006,20 @@ Common scheduler options
    In such cases, you may set this parameter to ``true`` to avoid this.
 
 
+.. js:attribute:: .schedulers[].resubmit_on_qos_errors
+
+   :required: No
+   :default: ``[]``
+
+   This option is relevant to the Slurm backends only.
+
+   When a job is submitted certain errors can be ignored and the framework will try to submit again the job after some seconds.
+   ReFrame is checking with a regular expression if any of the given expressions is contained in the ``stderr`` of the submission command.
+   Keep in mind that the submission is blocking and ReFrame will not continue until the submission is successful or has a different error.
+
+   .. versionadded:: 3.4
+
+
 Execution Mode Configuration
 ----------------------------
 
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1013,12 +1013,16 @@ Common scheduler options
 
    This option is relevant to the Slurm backends only.
 
-   When a job is submitted certain errors can be ignored and the framework will try to submit again the job after some seconds.
-   An example of this could be QOS errors like ``QOSMaxSubmitJobPerUserLimit``, in which case you would have to set the option to ``["QOSMaxSubmitJobPerUserLimit"]``. You can ignore multiple errors at the same time if you include all the error strings in the list.
-   Job submission is a synchronous operation in ReFrame.
-   If this option is set, it will block the whole execution until the error conditions specified in this list are raised.
+   If any of the listed errors occur, ReFrame will try to resubmit the job after some seconds.
+   As an example, you could have ReFrame trying to resubmit a job in case that the maximum submission limit per user is reached by setting this field to ``["QOSMaxSubmitJobPerUserLimit"]``.
+   You can ignore multiple errors at the same time if you add more error strings in the list.
 
    .. versionadded:: 3.5
+
+   .. warning::
+      Job submission is a synchronous operation in ReFrame.
+      If this option is set, ReFrame's execution will block until the error conditions specified in this list are resolved.
+      No other test would be able to proceed.
 
 
 Execution Mode Configuration
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -244,7 +244,7 @@ class SlurmJobScheduler(sched.JobScheduler):
 
                 t = next(intervals)
                 self.log(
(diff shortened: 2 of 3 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -227,7 +230,22 @@ def emit_preamble(self, job):
...
+                completed = _run_strict(cmd, timeout=self._submit_timeout)
+                break
+            except SpawnedProcessError as e:
+                sbatch_error = 'sbatch: error: QOSMaxSubmitJobPerUserLimit'
+                if (
+                    not self._block_submission or
+                    e.exitcode != 1 or
+                    sbatch_error not in e.stderr
+                ):
+                    raise e
```
vkarak, line 245:
```suggestion
                sbatch_error = 'sbatch: error: QOSMaxSubmitJobPerUserLimit'
                if (not self._block_submission or
                    sbatch_error not in e.stderr):
                    raise
```

- The `SpawnedProcessError` will be raised only if the exit code is non-zero, so this check is superfluous.
- Perhaps a regex matching instead of the `in` check would be better and less strict.

reframe/schemas/config.json
```diff
@@ -293,6 +293,7 @@
                         "enum": ["local", "pbs", "slurm", "squeue", "torque"]
                     },
                     "ignore_reqnodenotavail": {"type": "boolean"},
+                    "block_submission": {"type": "boolean"},
```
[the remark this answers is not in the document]
vkarak: The default should be `none` or the empty list.

vkarak (changes requested):
We will also need documentation.

docs/config_reference.rst
```diff
@@ -1006,6 +1006,20 @@ Common scheduler options
...
+   :required: No
+   :default: ``[]``
+
+   This option is relevant to the Slurm backends only.
+
+   When a job is submitted certain errors can be ignored and the framework will try to submit again the job after some seconds.
+   ReFrame is checking with a regular expression if any of the given expressions is contained in the ``stderr`` of the submission command.
+   Keep in mind that the submission is blocking and ReFrame will not continue until the submission is successful or has a different error.
+
+   .. versionadded:: 3.4
```
vkarak, line 1020:
```suggestion
   .. versionadded:: 3.5
```

reframe/core/schedulers/slurm.py
```diff
@@ -227,7 +230,21 @@ def emit_preamble(self, job):
 
     def submit(self, job):
         cmd = f'sbatch {job.script_filename}'
-        completed = _run_strict(cmd, timeout=self._submit_timeout)
+        intervals = itertools.cycle([1, 2, 3])
+        while True:
+            try:
+                completed = _run_strict(cmd, timeout=self._submit_timeout)
+                break
+            except SpawnedProcessError as e:
```
vkarak, line 238:
```suggestion
            except SpawnedProcessError:
```

reframe/core/schedulers/slurm.py
```diff
@@ -227,7 +230,21 @@ def emit_preamble(self, job):
...
+                sbatch_error = '|'.join(self._resubmit_on_qos_errors)
+                if (
+                    not self._resubmit_on_qos_errors or
+                    not re.search(sbatch_error, e.stderr)
+                ):
+                    raise e
```
vkarak, line 244:
```suggestion
                if (not self._resubmit_on_qos_errors or
                    not re.search(sbatch_error, e.stderr)):
                    raise
```

reframe/schemas/config.json
```diff
@@ -445,6 +449,7 @@
         "modes/options": [],
         "modes/target_systems": ["*"],
         "schedulers/ignore_reqnodenotavail": false,
+        "resubmit_on_qos_errors": [],
```
vkarak, line 452:
```suggestion
        "schedulers/resubmit_on_qos_errors": [],
```

reframe/core/schedulers/slurm.py
```diff
@@ -227,7 +230,21 @@ def emit_preamble(self, job):
...
+        intervals = itertools.cycle([1, 2, 3])
+        while True:
+            try:
+                completed = _run_strict(cmd, timeout=self._submit_timeout)
+                break
+            except SpawnedProcessError as e:
+                sbatch_error = '|'.join(self._resubmit_on_qos_errors)
+                if (
+                    not self._resubmit_on_qos_errors or
+                    not re.search(sbatch_error, e.stderr)
```
vkarak, line 242: Also this is wrong. `_resubmit_on_qos_errors` is a list of QOS errors that we should ignore. So this must revised.

docs/config_reference.rst
```diff
@@ -1006,6 +1006,20 @@ Common scheduler options
...
 
 
+.. js:attribute:: .schedulers[].resubmit_on_qos_errors
+
+   :required: No
+   :default: ``[]``
+
+   This option is relevant to the Slurm backends only.
+
+   When a job is submitted certain errors can be ignored and the framework will try to submit again the job after some seconds.
```
vkarak, line 1016: Not `certain` errors -> QOS errors. And we should give an example in parenthesis.

docs/config_reference.rst
```diff
@@ -1006,6 +1006,20 @@ Common scheduler options
...
+   ReFrame is checking with a regular expression if any of the given expressions is contained in the ``stderr`` of the submission command.
```
vkarak, line 1017: That's an implementation detail; it's not needed.

docs/config_reference.rst
```diff
@@ -1006,6 +1006,20 @@ Common scheduler options
...
+   Keep in mind that the submission is blocking and ReFrame will not continue until the submission is successful or has a different error.
```
vkarak, line 1018:
```suggestion
   Job submission is a synchronous operation in ReFrame.
   If this option is set, it will block the whole execution until the error conditions specified in this list are raised.
```

vkarak (changes requested):

docs/config_reference.rst
vkarak: Or we change the option name to `resubmit_on_errors`, which I think corresponds better to what we actually do.

ekouts: I think it's better to rename the option to `resubmit_on_errors` and keep it more general.

docs/config_reference.rst
ekouts: Initially I wrote it so that users are careful with the string they choose to ignore. I will remove.

reframe/core/schedulers/slurm.py
ekouts: Can I have access to `e.stderr` without it?

vkarak: Ok, I missed that. But then don't `raise e`, do just `raise`.

reframe/core/schedulers/slurm.py
```diff
@@ -236,11 +236,9 @@ def submit(self, job):
...
             except SpawnedProcessError as e:
-                sbatch_error = '|'.join(self._resubmit_on_qos_errors)
-                if (
-                    not self._resubmit_on_qos_errors or
-                    not re.search(sbatch_error, e.stderr)
-                ):
+                sbatch_error = '|'.join(self._resubmit_on_errors)
+                if (not self._resubmit_on_errors or
+                    not re.search(sbatch_error, e.stderr)):
                     raise e
```
vkarak, line 243: I think we should log something here indicating the error condition that we encountered and that we are going to retry the submission in X seconds. Also, I don't remember whether we log the standard output and error of failing commands.

vkarak (changes requested):
I will also do some minor adaptations to the documentation.

vkarak (approved):
Works fine.

PR #1705, opened 2021-01-22
[feat] Add a new family of sanity functions that looks for patterns inside text instead of files

### LINKED ISSUE #1029: Provide sanity function variants to match regexes directly against string inputs
Currently, all regex matching sanity functions are taking a filename as input, which they read it and do the matching.

teojgo:
Fixes #1029


vkarak (changes requested):
To avoid the duplication, I would suggest the following pattern:

```python
@deferrable
def finditer_s(patt, string):
    yield from re.finditer(patt, string, re.MULTILINE)

@deferrable
def finditer(patt, filename, encoding='utf-8'):
    with open(filename, 'rt', encoding=encoding) as fp:
        yield from finditer_s(patt, fp.read())
```

All the logic will be in the `_s` functions (so you can simply adapt the existing ones) and their counterparts getting a filename will simply call them as above.

teojgo:
> To avoid the duplication, I would suggest the following pattern:
> 
> ```python
> @deferrable
> def finditer_s(patt, string):
>     yield from re.finditer(patt, string, re.MULTILINE)
> 
> @deferrable
> def finditer(patt, filename, encoding='utf-8'):
>     with open(filename, 'rt', encoding=encoding) as fp:
>         yield from finditer_s(patt, fp.read())
> ```
> 
> All the logic will be in the `_s` functions (so you can simply adapt the existing ones) and their counterparts getting a filename will simply call them as above.

So does that mean that the handling of the exception in case of the filename opening should happen at each function? Because now it is handled by the `finditer` function.

vkarak:
> So does that mean that the handling of the exception in case of the filename opening should happen at each function? Because now it is handled by the finditer function.

Since this is so common we could have a function or context manager that does the handling once.

reframe/utility/sanity.py
```diff
@@ -503,7 +548,20 @@ def finditer(patt, filename, encoding='utf-8'):
...
+
+@deferrable
+def finditer_s(patt, string):
+    '''Get an iterator over the matches of the regex ``patt`` in ``string``.
+
+    This function is equivalent to :func:`findall_s()` except that it returns
+    a generator object instead of a list, which you can use to iterate over
+    the raw matches.
+
+    .. versionadded:: 3.4
```
ekouts, line 562:
```suggestion
    .. versionadded:: 3.5
```

reframe/utility/sanity.py
```diff
@@ -527,6 +585,26 @@ def findall(patt, filename, encoding='utf-8'):
...
+        <https://docs.python.org/3/library/re.html#regular-expression-syntax>`_
+        is accepted.
+        The `re.MULTILINE
+        <https://docs.python.org/3/library/re.html#re.MULTILINE>`_ flag
+        is set for the pattern search.
+    :arg string: The string to examine.
+    :returns: A list of raw `regex match objects
+        <https://docs.python.org/3/library/re.html#match-objects>`_.
+
+    .. versionadded:: 3.4
```
ekouts, line 603:
```suggestion
    .. versionadded:: 3.5
```

reframe/utility/sanity.py
```diff
@@ -606,6 +684,72 @@ def extractiter(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
+@deferrable
+def extractiter_s(patt, string, tag=0, conv=None):
+    '''Get an iterator over the values extracted from the capturing group
+    ``tag`` of a matching regex ``patt`` in ``string``.
+
+    This function is equivalent to :func:`extractall_s` except that it returns
+    a generator object, instead of a list, which you can use to iterate over
+    the extracted values.
+
+    .. versionadded:: 3.4
```
ekouts, line 745:
```suggestion
    .. versionadded:: 3.5
```

reframe/utility/sanity.py
```diff
@@ -648,6 +792,42 @@ def extractall(patt, filename, tag=0, conv=None, encoding='utf-8'):
...
+        extracted from the corresponding capturing group in ``tag``.
+        If more conversion functions are supplied than the corresponding
+        capturing groups in ``tag``, the last conversion function will be used
+        for the additional capturing groups.
+    :returns: A list of tuples of converted values extracted from the
+         capturing groups specified in ``tag``, if ``tag`` is an iterable.
+         Otherwise, a list of the converted values extracted from the single
+         capturing group specified in ``tag``.
+
+    .. versionadded:: 3.4
```
ekouts, line 826:
```suggestion
    .. versionadded:: 3.5
```

reframe/utility/sanity.py
```diff
@@ -665,6 +845,8 @@ def extractsingle(patt, filename, tag=0, conv=None, item=0, encoding='utf-8'):
     :arg item: the specific element to extract.
     :returns: The extracted value.
     :raises reframe.core.exceptions.SanityError: In case of errors.
+
+    .. versionadded:: 3.4
```
ekouts, line 849: This method was already here, why the `versionadded`?

reframe/utility/sanity.py
```diff
@@ -673,8 +855,39 @@ def extractsingle(patt, filename, tag=0, conv=None, item=0, encoding='utf-8'):
...
+
+    :arg patt: as in :func:`extractall_s`.
+    :arg string: as in :func:`extractall_s`.
+    :arg tag: as in :func:`extractall_s`.
+    :arg conv: as in :func:`extractall_s`.
+    :arg item: the specific element to extract.
+    :returns: The extracted value.
+    :raises reframe.core.exceptions.SanityError: In case of errors.
+
+    .. versionadded:: 3.4
```
ekouts, line 880:
```suggestion
    .. versionadded:: 3.5
```

ekouts (changes requested):

vkarak (changes requested):
What about avoiding the duplication in the unit tests with something like that (not complete):

```python
def fixture(funcs):
    @pytest.fixture(params=funcs)
    def _fixture(request, tempfile, temptxt):
        import functools

        if request.param.__name__.endswith('_s'):
            return functools.partial(request.param, string=temptxt)
        else:
            return functools.partial(request.param, filename=tempfile)

    return _fixture


assert_found = fixture([sn.assert_found, sn.assert_found_s])
assert_not_found = fixture([sn.assert_not_found, sn.assert_not_found_s])


def test_assert_found1(assert_found):
    assert assert_found(r'Step: \d+')
    assert assert_found(r'Step: \d+')
    with pytest.raises(SanityError):
        sn.evaluate(assert_found(r'foo: \d+'))


def test_assert_not_found1(assert_not_found):
    assert assert_not_found(r'foo: \d+')
    assert assert_not_found(r'foo: \d+')
    with pytest.raises(SanityError):
        sn.evaluate(assert_not_found(r'Step: \d+'))
```

The only problem still with that is that it doesn't test the case when the `filename` or `string` arguments are deferrables, but I think we can figure out something.

unittests/test_sanity_functions.py
```diff
@@ -511,41 +511,49 @@ def tempfile(tmp_path):
...
     with open(tempfile, 'r') as fp:
         return fp.read()
 
 
-def test_assert_found(tempfile):
-    assert sn.assert_found(r'Step: \d+', tempfile)
-    assert sn.assert_found(r'Step: \d+', sn.defer(tempfile))
-    with pytest.raises(SanityError):
-        sn.evaluate(sn.assert_found(r'foo: \d+', tempfile))
+def fixture(funcs):
```
vkarak, line 519: I think it's better if we name it `make_fixture`.

unittests/test_sanity_functions.py
```diff
@@ -511,41 +511,49 @@ def tempfile(tmp_path):
...
+    @pytest.fixture(params=funcs)
+    def _fixture(request, tempfile, contents):
+        import functools
```
vkarak: Do you use this?

unittests/test_sanity_functions.py
```diff
@@ -511,41 +511,49 @@ def tempfile(tmp_path):
...
+    return _fixture
 
 
-def test_assert_found_encoding(utf16_file):
-    assert sn.assert_found('Odyssey', utf16_file, encoding='utf-16')
+assert_found = fixture([sn.assert_found, sn.assert_found_s])
+assert_not_found = fixture([sn.assert_not_found, sn.assert_not_found_s])
+findall = fixture([sn.findall, sn.findall_s])
+extractall = fixture([sn.extractall, sn.extractall_s])
+extractsingle = fixture([sn.extractsingle, sn.extractsingle_s])
```
vkarak, line 534:
```suggestion
_assert_found = fixture([sn.assert_found, sn.assert_found_s])
_assert_not_found = fixture([sn.assert_not_found, sn.assert_not_found_s])
_findall = fixture([sn.findall, sn.findall_s])
_extractall = fixture([sn.extractall, sn.extractall_s])
_extractsingle = fixture([sn.extractsingle, sn.extractsingle_s])
```

So that you can later avoid the generic names such as `assert_func`.

unittests/test_sanity_functions.py
```diff
@@ -511,41 +511,49 @@ def tempfile(tmp_path):
...
 
 
-def test_assert_not_found(tempfile):
-    assert sn.assert_not_found(r'foo: \d+', tempfile)
-    assert sn.assert_not_found(r'foo: \d+', sn.defer(tempfile))
+def test_assert_found(assert_found):
+    assert_func, arg = assert_found
```
vkarak, line 538:
```suggestion
    assert_found, where = _assert_found
```

vkarak: I'm open to alternatives for `where`, but I prefer something not so generic as `arg`.

vkarak (changes requested):
Lgtm. I have only styling comments.

vkarak (approved):
lgtm

reframe/utility/sanity.py
```diff
@@ -381,12 +392,30 @@ def assert_found(patt, filename, msg=None, encoding='utf-8'):
...
+
+
+@deferrable
+def assert_found_s(patt, string, msg=None):
+    '''Assert that regex pattern ``patt`` is found in the string ``string``.
+
+    :arg patt: as in :func:`assert_found`.
+    :arg string: The string to examine.
+    :returns: ``True`` on success.
+    :raises reframe.core.exceptions.SanityError: if assertion fails.
```
ekouts, line 409: Should you include the `msg` argument?

reframe/utility/sanity.py
```diff
@@ -401,13 +430,30 @@ def assert_not_found(patt, filename, msg=None, encoding='utf-8'):
...
+
+
+@deferrable
+def assert_not_found_s(patt, string, msg=None):
+    '''Assert that regex pattern ``patt`` is not found in ``string``.
+
+    This is the inverse of :func:`assert_found_s()`.
+
+    :returns: ``True`` on success.
+    :raises reframe.core.exceptions.SanityError: if assertion fails.
```
ekouts, line 446: Here as well, I am not sure how strict we are with documenting all the arguments

ekouts (approved):
Lgtm in general, but I think we should add all the arguments of the new functions in their docstrings.

vkarak:
I agree with @ekouts. I didn't notice that the `msg` arguments were missing.

reframe/utility/sanity.py
```diff
@@ -238,6 +238,7 @@ def contains(seq, key):
 def assert_true(x, msg=None):
     '''Assert that ``x`` is evaluated to ``True``.
 
+    :arg msg: The error message to use if the assertion fails.
```
vkarak, line 241: We should mention that somebody can use placeholders in the message. I can fix that and push.

vkarak (changes requested):

reframe/utility/sanity.py
vkarak: Ok, this is documented at the beginning of the section, globally for all the functions. See here: http://localhost:8000/sanity_functions_reference.html#categories-of-sanity-functions

That's why I haven't put anything about the message in each individual function. It was added later and I was lazy :-)

vkarak (approved):
Lgtm, going in.

PR #1710, opened 2021-01-22
[doc] Enhance documentation about performance log directories

### LINKED ISSUE #1643: Enhance documentation about performance log directories
According to feedback from the Slack channel it can be a bit confusing. We must explain it better and give some concrete examples.

rsarm:
Fixes #1643.


docs/configure.rst
```diff
@@ -192,6 +192,9 @@ For example, if we have set it to ``warning``, no debug or informational message
 Finally, there is a special set of handlers for handling performance log messages.
 These are stored in the ``handlers_perflog`` property.
 The performance handler in this example will create a file per test and per system/partition combination and will append the performance data to it every time the test is run.
+The path to the file is created as follows: ``<basedir>/<prefix>/<testname>log``.
+``<basedir>`` is controlled by the ``--perflogdir`` option, the environment variable :envvar:`RFM_PERFLOG_DIR` or the ``basedir`` configuration parameter in the ``handlers_perflog``, which is ``.`` by default; ``<prefix>`` is what is defined in the ``handlers_perflog``.
+Note that that the performance log file is created only if ``perf_patterns`` is defined on the test.
```
teojgo, line 197:
```suggestion
Note that the performance log file is created only if ``perf_patterns`` is defined on the test.
```

docs/configure.rst
```diff
@@ -192,6 +192,9 @@ For example, if we have set it to ``warning``, no debug or informational message
 Finally, there is a special set of handlers for handling performance log messages.
 These are stored in the ``handlers_perflog`` property.
 The performance handler in this example will create a file per test and per system/partition combination and will append the performance data to it every time the test is run.
+The path to the file is created as follows: ``<basedir>/<prefix>/<testname>log``.
+``<basedir>`` is controlled by the ``--perflogdir`` option, the environment variable :envvar:`RFM_PERFLOG_DIR` or the ``basedir`` configuration parameter in the ``handlers_perflog``, which is ``.`` by default; ``<prefix>`` is what is defined in the ``handlers_perflog``.
```
teojgo, line 196:
```suggestion
``<basedir>`` is controlled by the ``--perflogdir`` option, the environment variable :envvar:`RFM_PERFLOG_DIR` and the ``basedir`` configuration parameter in the ``handlers_perflog``, which is ``.`` by default; ``<prefix>`` is defined in the ``handlers_perflog``.
```

teojgo (changes requested):

docs/configure.rst
```diff
@@ -192,6 +192,9 @@ For example, if we have set it to ``warning``, no debug or informational message
 Finally, there is a special set of handlers for handling performance log messages.
 These are stored in the ``handlers_perflog`` property.
 The performance handler in this example will create a file per test and per system/partition combination and will append the performance data to it every time the test is run.
```
vkarak, line 194:
```suggestion
The performance handler in this example will create a file per test and per system/partition combination and will append the performance data to it every time the *performance test* is run.
```

Then either we define what a performance test is or have link to somewhere that we define it.

docs/configure.rst
vkarak: Imho, all this description is not needed. It described in the configuration reference of the `filelog` handler. So this is a duplication. We do have a link to the reference at the end of this section. The key problem in the documentation, in my opinion, is that we don't say explicitly that performance logging happens *only* for performance tests.

vkarak (changes requested):

vkarak (approved):

Issue #1679, opened 2021-01-11
Generate executable `_job.sh` files


homerdin:
It would be useful to have a configuration setting to generate `_job.sh` files that are executable.  Would this be a potential feature that could be added?

Thanks

vkarak:
Hi @homerdin, yes it should be an easy addition. Thanks for the suggestion.

jgphpc:
@vkarak where is
https://github.com/eth-cscs/reframe/blob/master/reframe/frontend/executors/__init__.py#L270
defined ?

teojgo:
> @vkarak where is
> `compile` in https://github.com/eth-cscs/reframe/blob/master/reframe/frontend/executors/__init__.py#L270
> defined ?

https://github.com/eth-cscs/reframe/blob/13c372ef7aa5d63f6d74c73f61c4c033c0312d53/reframe/frontend/executors/__init__.py#L226

teojgo:
@jg The `compile` of the check, it is defined here: https://github.com/eth-cscs/reframe/blob/13c372ef7aa5d63f6d74c73f61c4c033c0312d53/reframe/core/pipeline.py#L1165

vkarak:
@jgphpc Actually the building phase where @teojgo pointed you has nothing to do with the job script creation. The job script is created by the `ShellScriptGenerator` here:

https://github.com/eth-cscs/reframe/blob/master/reframe/core/shell.py#L114

Currently, setting the executable bit is handled by the scheduler backends (only the `local` scheduler does that):

https://github.com/eth-cscs/reframe/blob/master/reframe/core/schedulers/local.py#L63-L64

I think the best solution is to move that responsibility to the shell script generator by adding an option that controls whether to set `+x` or not. And then we remove this logic from the local backend, which is just a leftover from the past.

PR #1716, opened 2021-01-26
[feat] Set executable bit in generated job script files

jgphpc:
Will fix https://github.com/eth-cscs/reframe/issues/1679


vkarak (changes requested):
I think this should be handled by the `ShellScriptGenerator` in `reframe/core/shell.py`. Then the callers will set it or not accordingly. Also I don't know if it makes more sense to create them always as executable (without being configurable). I don't see a reason why a job script would harm if it's executable.

config/cscs.py
```diff
@@ -619,6 +619,7 @@
                     'name': 'cn',
                     'scheduler': 'slurm',
                     'access': [
+                        # '--partition=debug'
```
vkarak, line 622: Can you please remove this?

config/cscs.py
```diff
@@ -990,16 +991,16 @@
...
+##                 {
+##                     'type': 'graylog',
+##                     'address': 'graylog-server:12345',
+##                     'level': 'info',
+##                     'format': '%(message)s',
+##                     'extras': {
+##                         'facility': 'reframe',
+##                         'data-version': '1.0',
+##                     }
+##                 }
```
vkarak, line 1003: You don't need to comment this for testing. ReFrame will simply not initialize the Graylog handler if it cannot connect to it.

config/cscs.py
```diff
@@ -1041,7 +1042,8 @@
             'check_search_path': [
                 'checks/'
             ],
-            'check_search_recursive': True
+            'check_search_recursive': True,
+            # 'executable_jobscript': True,
```
vkarak, line 1046: Remove this.

reframe/core/shell.py
```diff
@@ -115,6 +117,7 @@ class generate_script:
     def __init__(self, filename, *args, **kwargs):
         self._shgen = ShellScriptGenerator(*args, **kwargs)
         self._file = open(filename, 'wt')
+        os.chmod(filename, os.stat(filename).st_mode | stat.S_IEXEC)
```
vkarak, line 120: I think better if you set the mode after you write the file in `__exit__()` and I would suggest that the mode is taken as an argument after `filename` with default value set to `None`. If `None` then we do the `+x`  as of here.

vkarak (changes requested):

vkarak:
I can fix this PR.

vkarak (approved):

vkarak:
@jenkins-cscs retry dom

PR #1718, opened 2021-01-28
[bugfix] Fix handling of single character scheduler backend options passed through `-J`

### LINKED ISSUE #1693: Single-char scheduler options passed through `-J` are not handled properly
[Here](https://github.com/eth-cscs/reframe/blob/d56e435acb4373447021b29ac4eb707165c09a81/reframe/frontend/cli.py#L905-L911) `opt` is the complete option (something like `A=usup`), not just the flag. The length is not 1, then it never goes to the `elif len(opt) == 1` resulting in adding two minuses (`--A`) when it should be a single one (`-A`):
```
        for opt in options.job_options:
            if opt.startswith('-') or opt.startswith('#'):
                parsed_job_options.append(opt)
            elif len(opt) == 1:
                parsed_job_options.append(f'-{opt}')
            else:
                parsed_job_options.append(f'--{opt}')
```

rsarm:
closes #1693


rsarm:
@jenkins-cscs retry dom

reframe/frontend/cli.py
```diff
@@ -903,9 +903,10 @@ def module_unuse(*paths):
         exec_policy.sched_flex_alloc_nodes = sched_flex_alloc_nodes
         parsed_job_options = []
         for opt in options.job_options:
+            optstr, *_ = re.split(r'=|\s+', opt)
```
vkarak, line 906: This is equivalent, I think:

```suggestion
            optstr = re.split(r'=|\s+', opt)[0]
```

vkarak (changes requested):

vkarak (changes requested):
@rsarm I tested the PR and I verified what I pointed out in the issue passing `-A=foo` is wrong. It's not correct under POSIX and -correctly- Slurm rejects it: 

```
[reframer060@reframe ~]$ srun -ptotal -A reframers hostname
reframe-1
[reframer060@reframe ~]$ srun -ptotal -A=reframers hostname
srun: error: Unable to allocate resources: Invalid account or account/partition combination specified
```

Should we then treat the `-J A=foo` option specially and pass `-Afoo` or `-A foo`?

rsarm:
@vkarak oh sorry! I was convinced that `-A=foo` worked!

I would prefer to do not support translating non-valid options like `-A=foo` to valid ones. I would expect the user to pass options exactly as they would be set on the scheduler script. On the other hand it's very simple to support it, I guess we can do this
```python
        for opt in options.job_options:
            optstr, valstr = re.split(r'=|\s+', opt, maxsplit=1)
            if opt.startswith('-') or opt.startswith('#'):
                parsed_job_options.append(opt)
            elif len(optstr) == 1:
                parsed_job_options.append(f'-{optstr} {valstr}')
            else:
                parsed_job_options.append(f'--{optstr} {valstr}')
```

vkarak:
What I'm proposing I think it's on the same line with the conversion from `-J account=foo` to `--account=foo`. To me it seems valid to have a conversion such as `-J A=foo` to `-A foo`. The lhs is our syntax. Otherwise, the user would have to quote the argument `-J 'A foo'` which is not consistent with the `account=foo` and also not so nice to write. I would vote for supporting just this syntax then.

reframe/frontend/cli.py
```diff
@@ -903,13 +903,15 @@ def module_unuse(*paths):
         exec_policy.sched_flex_alloc_nodes = sched_flex_alloc_nodes
         parsed_job_options = []
         for opt in options.job_options:
-            optstr, *_ = re.split(r'=|\s+', opt)
+            opt_split = re.split(r'=|\s+', opt, maxsplit=1)
```
vkarak, line 906: Since we are sticking with `opt=val` syntax and we do the conversion, does it make sense to split also on spaces? In that cases, the user can always pass the full option `--opt val`. What do you think?

reframe/frontend/cli.py
```diff
@@ -903,13 +903,15 @@ def module_unuse(*paths):
...
+            optstr = opt_split[0]
+            valstr = opt_split[1] if len(opt_split) > 1 else ''
             if opt.startswith('-') or opt.startswith('#'):
                 parsed_job_options.append(opt)
             elif len(optstr) == 1:
-                parsed_job_options.append(f'-{opt}')
+                parsed_job_options.append(f'-{optstr} {valstr}')
             else:
-                parsed_job_options.append(f'--{opt}')
+                parsed_job_options.append(f'--{optstr} {valstr}')
```
vkarak, line 914:
```suggestion
                parsed_job_options.append(f'--{optstr}={valstr}')
```

vkarak (changes requested):

reframe/frontend/cli.py
rsarm: Here I left out the `=` on the long options too to cover cases without like `-J foo` without having to do anything extra. Are `--opt=foo` and `--opt foo` equivalent?

vkarak: Good point about the `-J foo`, so you can can leave it as you have it.

> Are --opt=foo and --opt foo equivalent?

Yes, they are.

vkarak (approved):
lgtm

PR #1719, opened 2021-01-29
[feat] Add support for building tests using EasyBuild

### LINKED ISSUE #300: Create an 'Easybuild' build system
A build system backend that would allow ReFrame to use Easybuild recipes for building a software.

rsarm:
Closes #300


rsarm:
At this point, the test has to define the following:
```python
        self.modules = ['EasyBuild-custom']
        self.build_system = 'EasyBuild'
        self.build_system.easyconfigs = [<easyconfigs>]
        self.build_system.options = [<easybuild-cl-options>]
```

reframe/core/buildsystems.py
```diff
@@ -655,6 +655,35 @@ def emit_build_commands(self, environ):
...
+        self.easyconfigs = []
+        self.options = []
+        self._installpath = None
+
+    def emit_build_commands(self, environ):
+        _easyconfigs = ' '.join(self.easyconfigs)
+        _options = ' '.join(self.options)
+        return [f"eb {_easyconfigs} --installpath {self._installpath} "
+                f"--tmpdir {self._installpath} "
+                f"--sourcepath {self._installpath} "
```
boegel, line 673: Shouldn't these be different paths?

reframe/core/buildsystems.py
```diff
@@ -655,6 +655,35 @@ def emit_build_commands(self, environ):
...
+                f"--buildpath {self._installpath}/build {_options}"]
```
boegel, line 674: Very often the installation directory and the build directory are different filesystem (like NFS for install dir, `/dev/shm` for build dir), so maybe you also need `self._buildpath`?

reframe/core/buildsystems.py
```diff
@@ -655,6 +655,35 @@ def emit_build_commands(self, environ):
...
+
+    def _get_built_modules(self):
+        modules = []
+        modules_dir = os.path.join(self._installpath, 'modules', 'all')
+        if os.path.isdir(modules_dir):
+            for mod in os.listdir(modules_dir):
+                for ver in os.listdir(os.path.join(modules_dir, mod)):
+                    modules.append(f'{mod}/{ver}')
```
boegel, line 682: You're making the assumption here that the default module naming scheme is used.

Good enough as a first step maybe, but worth clarifying in a comment?

reframe/core/pipeline.py
```diff
@@ -1239,6 +1239,9 @@ def compile(self):
             self.build_system.srcfile = self.sourcepath
             self.build_system.executable = self.executable
 
+        if type(self.build_system).__name__ == 'EasyBuild':
+            self.build_system._installpath = os.path.join(self._stagedir, 'easybuild')
```
boegel, line 1243: If you just want to put build/install/sources in the same directory, just use `eb --prefix`?

reframe/core/pipeline.py
```diff
@@ -1345,6 +1348,12 @@ def run(self):
         exec_cmd = [self.job.launcher.run_command(self.job),
                     self.executable, *self.executable_opts]
         commands = [*self.prerun_cmds, ' '.join(exec_cmd), *self.postrun_cmds]
+        if type(self.build_system).__name__ == 'EasyBuild':
```
boegel, line 1351: Usually this is done with:

```python
if isinstance(self.build_system, EasyBuild):
```

(but then you need to be able to import `EasyBuild` in here, not sure if that's possible...)

boegel (changes requested):
Just some feedback, let me know if I can be of any further help with this!

vkarak:
Thanks @boegel for reviewing!

reframe/core/buildsystems.py
rsarm: @boegel thanks for having a look! The idea was to put everything inside reframe's `stage` directory in the same way the other build systems do. When the test passes, everything build-related would go away. Here I want to override eb's system configuration (for us, what the EasyBuild module sets) and move everything inside the `stage` dir.

I had dedicated variables for `tmpdir` and `sourcepath` but then I saw that easybuild creates a directory for each of them given the  path that that's passed: `<install-path>/software`, a random name for the `<install-path>/<tmp-name>`, etc. Then I just give the install dir only and let easybuild create the rest within it.

reframe/core/pipeline.py
rsarm: I'm trying `eb --prefix` on Daint but as it is configured, it still gets the build/install/sources from the environment variables. For instance, 
```
eb zlib-1.2.11.eb --prefix=$HOME/easybuild
```
creates `$HOME/easybuild` with only `ebfiles_repo/zlib/zlib-1.2.11.eb` inside.

akesandgren:
Not sure how you've set things up here (haven't looked closely enough) but I assume it's based on having a pre-existing EB installation with the toolchains and basic stuff? Or was the plan to have it build totally from scratch?

And it needs to handle the existing EB using HMNS too.

vkarak:
> Not sure how you've set things up here (haven't looked closely enough) but I assume it's based on having a pre-existing EB installation with the toolchains and basic stuff? Or was the plan to have it build totally from scratch?

It assumes a pre-existing installation of EB.

> And it needs to handle the existing EB using HMNS too.

Can you elaborate on this?

akesandgren:
I.e., if the existing EB uses HMNS any EB config params you use must make sure to also use it, and it needs to be done in such a way that the site installed EB's modulefiles that extend MODULEPATH can find the modules you install with ReFrame.

vkarak:
Thanks @akesandgren for the explanation. The initial idea behind this feature was not to install the package, but rather stay in the stage directory just for the duration of the test. See it as testing a recipe or using EB to install and test a more complex software. If the build phase also installs it in the final location, what would happen if the sanity or the performance break? Then your test would have an undesirable side effect. Is it possible in EB to "move" an installation? Because then, users could have a post-performance hook, that would install it to its final location. @boegel @victorusu ?

akesandgren:
That's what I mean by setting $HOME.
If the site is using HMNS for EB and has prepared for letting users use EB to install packages themselves in $HOME by setting subdir-user-modules, then if you make sure to set $HOME before running "eb some-package", it will install the modules under $HOME/<subdir-user-modules> and at RunTest you can then actually load the just installed module from $HOME(set-to-some-stage-dir-of-reframe) after loading the required toolchain.

So the package ReFrame installs using EB can easily be thrown away after by just removing the stage-dir where the actual install happened.

You can't (normally) move a package installed by EB somewhere else though, due to path's usually getting upset...

Any real-life site/user installation of the package would still have to be done outside of ReFrame, and that is, as far as I'm concerned, the way it should be.

Does that explain things better?

vkarak:
The initial idea was not to have ReFrame install the package and I'm also concerned about this, especially about the side effects I described above. We had a discussion with @rsarm and @victorusu today and we think that the best way is to build in the stage dir (unless the user really passes specific options to EB to change that, which I don't think is a good idea) and produce the necessary rpms (or any other supported packages from EB). Then, if the test is successful, you can keep this as an artifact with `keep_files` and deploy it by yourself either with the O/S package manager or with EB itself. This would work if you don't use RPATH installations. You could even go a step further and have a test depending on this one that would do the deployment. And, of course, you can parameterize your test on different easyconfig files ;-) There can more improvements (such as having EB spawning build jobs), but we would like to start with a minimum valuable feature and see later on what is the potential of this.

reframe/core/buildsystems.py
```diff
@@ -655,6 +655,49 @@ def emit_build_commands(self, environ):
...
+        easyconfigs = ' '.join(self.easyconfigs)
+
+        if self.emit_package:
+            self.options.append('--package')
+            for key, val in self.package_opts.items():
+                self.options.append(f"--package-{key}={val}")
+
+        options = ' '.join(self.options)
+
+        cmd = [f"EASYBUILD_BUILDPATH={self._eb_sandbox}/build \\",
```
teojgo, line 680: You can use `rf` strings here so you don't have to escape the `\`

reframe/core/buildsystems.py
```diff
@@ -658,43 +659,45 @@ def emit_build_commands(self, environ):
...
+        self.emit_package = False
+        self.eb_modules = []
 
     def emit_build_commands(self, environ):
         easyconfigs = ' '.join(self.easyconfigs)
+
+        if self.emit_package:
+            self.options.append('--package')
+            for key, val in self.package_opts.items():
+                self.options.append(f"--package-{key}={val}")
```
vkarak, line 678:
```suggestion
                self.options.append(f'--package-{key}={val}')
```

reframe/core/buildsystems.py
vkarak: Also use single quotes for strings.

reframe/core/buildsystems.py
```diff
@@ -658,43 +659,45 @@ def emit_build_commands(self, environ):
...
-                for ver in os.listdir(os.path.join(modules_dir, mod)):
-                    modules.append({'name': f'{mod}/{ver}',
+
+        cmd = [f"EASYBUILD_BUILDPATH={self._eb_sandbox}/build \\",
+               f"EASYBUILD_INSTALLPATH={self._eb_sandbox} \\",
+               f"EASYBUILD_PREFIX={self._eb_sandbox} \\",
+               f"EASYBUILD_SOURCEPATH={self._eb_sandbox} \\",
+               f"eb {easyconfigs} {options}"]
+
+        return cmd
```
vkarak, line 688: Just do `return [...]`

reframe/core/pipeline.py
```diff
@@ -1240,17 +1240,12 @@ def compile(self):
             self.build_system.executable = self.executable
 
         if isinstance(self.build_system, EasyBuild):
```
vkarak, line 1242: I don't like that you have to do build system specific stuff here.

reframe/core/pipeline.py
```diff
@@ -1240,17 +1240,12 @@ def compile(self):
...
-            if not self.build_system.buildpath:
-                self.build_system.buildpath = os.path.join(self._stagedir,
-                                                           'rfm_easybuild',
-                                                           'build')
-
-            if not self.build_system.sourcepath:
-                self.build_system.sourcepath = os.path.join(self._stagedir,
-                                                            'rfm_easybuild')
+            self.build_system._eb_sandbox = os.path.join(self._stagedir,
+                                                         'rfm_easybuild')
```
vkarak, line 1244: The `RegressionTest` should not care about this directory. It should the responsibility of the build system. And I don't think that the `stagedir` is needed also here. The compilation is always invoked from the stage directory.

reframe/core/pipeline.py
```diff
@@ -1240,17 +1240,12 @@ def compile(self):
...
+            if self.build_system.emit_package:
+                self.keep_files.append(
+                    f'{self.build_system._eb_sandbox}/packages'
+                )
```
vkarak, line 1248: This should be responsibility of the tests. We should not touch the `keep_files`.

reframe/core/pipeline.py
```diff
@@ -1302,6 +1297,10 @@ def compile_wait(self):
         if self._build_job.exitcode:
             raise BuildError(self._build_job.stdout, self._build_job.stderr)
 
+        if isinstance(self.build_system, EasyBuild):
+            self.build_system.collect_built_modules(
+                os.path.join(self._stagedir, self._build_job.stdout))
```
vkarak, line 1302: This should not be done here. You can cache it inside the build system.

reframe/core/buildsystems.py
```diff
@@ -658,43 +659,45 @@ def emit_build_commands(self, environ):
...
+
+    def collect_built_modules(self, build_stdout):
```
vkarak, line 690: The users should call this functions, but better rename it `generated_modules()`. You don't need the `eb_modules` attributes. In order to avoid opening and reading files every time, you can simply cache the result internally in the build system.

unittests/resources/settings.py
```diff
@@ -90,7 +90,7 @@
     'environments': [
         {
             'name': 'PrgEnv-gnu',
-            'modules': [{'name': 'PrgEnv-gnu', 'collection': False}],
+            'modules': [{'name': 'PrgEnv-gnu', 'collection': False, 'path': None}],
```
vkarak, line 93: If this is to be accepted, for which I'm in favor, we should update the config schema (making sure that the `additionalProperties` is `False` of this, which apparently I have forgotten to put in the current version of the module spec. We should also update the documentation.

vkarak (changes requested):
We also need: 

- [ ] Documentation for the EB build system
- [ ] Unit tests to test the `path` argument of the updated module interface
- [x] Unit tests for the EB build system that will be executed inside a container that will have EB installed.

rsarm:
Now the test should have `self.modules = self.build_system.eb_modules(self.build_job.stdout)`. It requires to pass `self.build_job.stdout` which can be confusing for the users since there is `self.build_stdout` which is a deferrable and wouldn't work there.

reframe/core/buildsystems.py
```diff
@@ -661,42 +661,52 @@ class EasyBuild(BuildSystem):
     options = fields.TypedField(typ.List[str])
     emit_package = fields.TypedField(bool)
     package_opts = fields.TypedField(typ.Dict[str, str])
+    srcdir = fields.TypedField(str, type(None))
```
vkarak, line 664: I would vote against this, since it has a complete different meaning in other build systems. Perhaps the best solution is to call the `emit_build_commands()` after we have changed to the stage directory. Then any build system can be sure that the current working directory is the stage directory.

reframe/core/buildsystems.py
```diff
@@ -661,42 +661,52 @@ class EasyBuild(BuildSystem):
...
 
     def __init__(self):
         super().__init__()
         self.easyconfigs = []
         self.options = []
         self.emit_package = False
-        self.eb_modules = []
+        self.package_opts = {}
+        self._eb_modules = []
```
vkarak, line 672: Better set this to `None`.

reframe/core/buildsystems.py
```diff
@@ -661,42 +661,52 @@ class EasyBuild(BuildSystem):
...
+                self.options.append(f'--package-{key}={val}')
 
         options = ' '.join(self.options)
 
-        cmd = [f"EASYBUILD_BUILDPATH={self._eb_sandbox}/build \\",
-               f"EASYBUILD_INSTALLPATH={self._eb_sandbox} \\",
-               f"EASYBUILD_PREFIX={self._eb_sandbox} \\",
-               f"EASYBUILD_SOURCEPATH={self._eb_sandbox} \\",
-               f"eb {easyconfigs} {options}"]
+        self._eb_sandbox = os.path.join(self.srcdir, 'rfm_easybuild')
```
vkarak, line 684: Just `easybuild` should be fine I think here.

```suggestion
        self._eb_sandbox = os.path.join(self.srcdir, 'easybuild')
```

reframe/core/buildsystems.py
```diff
@@ -661,42 +661,52 @@ class EasyBuild(BuildSystem):
...
 
-        return cmd
+        return [f'EASYBUILD_BUILDPATH={self._eb_sandbox}/build \\',
+                f'EASYBUILD_INSTALLPATH={self._eb_sandbox} \\',
+                f'EASYBUILD_PREFIX={self._eb_sandbox} \\',
+                f'EASYBUILD_SOURCEPATH={self._eb_sandbox} \\',
+                f'eb {easyconfigs} {options}']
```
vkarak, line 690: Why don't you use `export EASYBUILD_XXX` here?

reframe/core/pipeline.py
```diff
@@ -1196,22 +1196,18 @@ def compile(self):
                 else:
                     self.build_system = 'Make'
 
-            self.build_system.srcdir = self.sourcepath
+            if self.sourcepath:
+                self.build_system.srcdir = self.sourcepath
+            else:
+                self.build_system.srcdir = self._stagedir
```
vkarak, line 1202: As I said revert this and just move the part where we are setting the `build_commands` inside the `osext.change_dir()` a bit later on.

vkarak (changes requested):

reframe/core/buildsystems.py
```diff
@@ -655,6 +656,59 @@ def emit_build_commands(self, environ):
         return prepare_cmd + [' '.join(configure_cmd), ' '.join(make_cmd)]
 
 
+class EasyBuild(BuildSystem):
+    easyconfigs = fields.TypedField(typ.List[str])
+    options = fields.TypedField(typ.List[str])
+    emit_package = fields.TypedField(bool)
+    package_opts = fields.TypedField(typ.Dict[str, str])
+    srcdir = fields.TypedField(str, type(None))
```
vkarak, line 664: All these must be documented.

reframe/core/buildsystems.py
vkarak: Ignore this comment. It has to be in a single line.

vkarak: Actually, you can generate multiple lines, so you should be able to use `export`s here.

reframe/core/buildsystems.py
vkarak: Users can pass any argument to `eb` through the build system's `options`.

reframe/core/buildsystems.py
vkarak: This is now changed. We are now parsing EB's output for that.

unittests/resources/settings.py
vkarak: I don't think that the `path` is needed here.

vkarak (approved):

PR #1720, opened 2021-01-29
[test] Constrain flexible node allocation on the HPCG check

jjotero:
The HPCG test requires the number of nodes to be factored into `N=nx*ny*nx`, where `max(nx, ny, nz)/min(nx, ny, nz) >= 8`.
Thus, since this case is a flexible test, some guidance is needed to have the 'raw' slurm node guess educated into something the HPCG tests can work with (for example, if the number of available nodes in the system is a prime number, the above factorisation is impossible).
 
This PR solves this issue by adding a hook into the `pre_run` stage of these tests. The hook simply does the node guess in advance and makes sure that the number of nodes is a sensible one. With this 'constrained' number of nodes we set `self.num_tasks` in the test, which avoids reframe to query for the available number of nodes again at a later stage.

### DIFF
--- a/cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
+++ b/cscs-checks/microbenchmarks/hpcg/hpcg_benchmark.py
@@ -17,12 +17,13 @@ class HPCGHookMixin(rfm.RegressionTest):
         ntasks_per_node = self.num_tasks_per_node or 1
         self.job.num_tasks_per_node = ntasks_per_node
         n = int(self.job.guess_num_tasks()/ntasks_per_node)
+
         def first_factor(x):
             if x <= 1:
                 return 1
 
             for i in range(2, x+1):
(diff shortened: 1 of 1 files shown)


vkarak (changes requested):
Another thing is that we would need to set `self.num_tasks = -8` in the tests. As far as I understood this is the minimum number of tasks, right? This would cause `guess_num_tasks()` to fail if it can't find enough nodes to satisfy this. In all our examples, however, since we always use one task per core and all our systems have more than 8 cores, but we can set this for correctness.

jjotero:
I just checked and the test seems to run fine with only one node (and 2 tasks per node). The only restriction I could find is the aspect ratio must be above 0.125 when doing the domain decomposition. But that's automatically satisfied for any node count below 10.

vkarak (approved):
lgtm

PR #1721, opened 2021-01-29
[dom] Temporary fix for GridTools build on dom

ChristopherBignamini:
This is a temporary fix for the GridTools build issue on Dom reported here: https://jira.cscs.ch/browse/UES-1382

### DIFF
--- a/cscs-checks/libraries/gridtools/gridtools_tests.py
+++ b/cscs-checks/libraries/gridtools/gridtools_tests.py
@@ -28,7 +28,7 @@ class GridToolsBuildCheck(rfm.CompileOnlyRegressionTest):
         ]
         self.build_system.flags_from_environ = False
         self.build_system.make_opts = ['perftests']
-        self.build_system.max_concurrency = 2
+        self.build_system.max_concurrency = 8
         self.postbuild_cmds = ['ls tests/regression/']
         self.sanity_patterns = sn.assert_found(r'perftest',
                                                self.stdout)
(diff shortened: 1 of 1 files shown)


cscs-checks/libraries/gridtools/gridtools_tests.py
```diff
@@ -52,7 +52,13 @@ class GridToolsGPUBuildCheck(GridToolsBuildCheck):
     def __init__(self):
         super().__init__()
         self.descr = 'GridTools GPU build test'
-        self.modules.append('cudatoolkit')
+        if self.current_system.name == 'dom':
+            self.modules.append(
+                'cudatoolkit/10.2.89_3.29-7.0.2.1_3.5__g67354b4')
+            self.modules.append('cdt-cuda/20.10')
+            self.modules.append('gcc/8.3.0')
```
vkarak, line 59:
```suggestion
        if self.current_system.name == 'dom':
            self.modules += [...]
```

So it fails to build with CUDA 11? Could it be that loading simply `cdt-cuda` would fix that?

vkarak (changes requested):

cscs-checks/libraries/gridtools/gridtools_tests.py
ChristopherBignamini: I've tested different configurations, after loading daint-gpu, switching to PrgEnv-gnu and loading Boost: 

1) Default cdt: 
Make Error at CMakeLists.txt:133 (message):
GPU targets where required but we couldn't configure GPU support.  
2) Default cdt+cudatoolkit: link error as reported at the beginning of this issue
3) switch gcc to 8.3.0 (from 9.3, that is loaded by Boost): same link error
4) switch cdt-cuda/20.11 to cdt-cuda/20.10: same link error
5) switch again gcc to 8.3.0: link error
6) switch cudatoolkit to 10: this one works and is the same configuration of the reframe test

vkarak (approved):

Issue #1736, opened 2021-02-02
Container execution w/ SARUS


ChristianKniep:
A couple of ideas around the container execution. Starting with SARUS as that is what I am after.
### Adjustments
- [ ] Honor the configuration of an images when it comes to `ENTRYPOINT` and `CMD`.
- [ ] Use the appropriate runtime options to change the `workdir` and not to `cd self.workdir`

### Quick fixes
- [ ] allow to `skip_pull` an image pull (that might not be something to consider for production, as this kinda violates the reproducibility)

ChristianKniep:
Finally fixed the unittests... :)

PR #1737, opened 2021-02-02
[feat] Improve Sarus backend

ChristianKniep:
Some changes to improve the use of SARUS #1736 . @Madeeks

### DIFF
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -136,7 +136,7 @@ class Sarus(ContainerPlatform):
             return []
         # Using 
         if self.with_metahub:
-            tmpdir = tempfile.TemporaryDirectory()
+            tmpdir = tempfile.mkdtemp()
             return [
                 'docker pull -q mh.qnib.org/'+self.image,
                 'docker save -o '+tmpdir+'/image.tar mh.qnib.org/'+self.image,
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -164,7 +164,7 @@ class Sarus(ContainerPlatform):
         run_cmd = self._command + ' run %s %s' % (' '.join(run_opts),
                                                            self.image)
         if self.commands:
-            return run_cmd + "'" + ' ; '.join(self.commands) + "'"
+            return run_cmd + " '" + ' ; '.join(self.commands) + "'"
         else:
             return run_cmd
 
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -8,7 +8,7 @@ import abc
 import reframe.core.fields as fields
 import reframe.utility.typecheck as typ
 from reframe.core.exceptions import ContainerError
-
+from tempfile import TemporaryDirectory
 
 class ContainerPlatform(abc.ABC):
     '''The abstract base class of any container platform.'''
@@ -116,17 +116,17 @@ class Sarus(ContainerPlatform):
     #: :default: :class:`False`
     with_mpi = fields.TypedField(bool)
 
-    #: Skip pull of images
+    #: Using metahub to fetch instance specific images
     #:
     #: :type: boolean
     #: :default: :class:`False`
-    skip_pull = fields.TypedField(bool)
+    with_metahub = fields.TypedField(bool)
 
     def __init__(self):
         super().__init__()
         self.with_mpi = False
         self._command = 'sarus'
-        self.skip_pull = False
+        self.with_metahub = False
 
     def emit_prepare_commands(self):
         # The format that Sarus uses to call the images is
@@ -134,11 +134,17 @@ class Sarus(ContainerPlatform):
         # locally from a tar file, the <reposerver> is 'load'.
         if self.image.startswith('load/'):
             return []
-        # For testing purposes a way to skip the pull of an image.
-        if self.skip_pull:
-            return []
-
-        return [self._command + ' pull %s' % self.image]
+        # Using 
+        if self.with_metahub:
+            tmpdir = TemporaryDirectory()
+            return [
+                'docker pull -q mh.qnib.org/'+self.image,
+                'docker save -o '+tmpdir+'/image.tar mh.qnib.org/'+self.image,
+                self._command + ' load ${TMP_DIR}/gromacs.tar mh.qnib.org/'+self.image,
+                'rm -rf '+tmpdir
+            ]
+        else:
+            return [self._command + ' pull %s' % self.image]
 
     def launch_command(self):
         super().launch_command()
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -8,7 +8,7 @@ import abc
 import reframe.core.fields as fields
 import reframe.utility.typecheck as typ
 from reframe.core.exceptions import ContainerError
-from tempfile import TemporaryDirectory
+import tempfile
 
 class ContainerPlatform(abc.ABC):
     '''The abstract base class of any container platform.'''
@@ -136,7 +136,7 @@ class Sarus(ContainerPlatform):
             return []
         # Using 
         if self.with_metahub:
-            tmpdir = TemporaryDirectory()
+            tmpdir = tempfile.TemporaryDirectory()
             return [
                 'docker pull -q mh.qnib.org/'+self.image,
                 'docker save -o '+tmpdir+'/image.tar mh.qnib.org/'+self.image,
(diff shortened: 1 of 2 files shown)


reframe/core/containers.py
```diff
@@ -88,9 +88,6 @@ def validate(self):
         if self.image is None:
             raise ContainerError('no image specified')
 
-        if not self.commands:
```
ChristianKniep, line 91: The creator of an images should know best how to execute the command. E.g. for Spack the `ENTRYPOINT` needs to be honored.
I'd like to argue that a container should also include the execution script.

reframe/core/containers.py
```diff
@@ -119,32 +116,45 @@ class Sarus(ContainerPlatform):
     #: :default: :class:`False`
     with_mpi = fields.TypedField(bool)
 
+    #: Skip pull of images
```
ChristianKniep, line 119: This violates the reproducibility; thus, not a keeper.
But (being devil's advocate), an image name like `qnib/spack-gromacs:latest` violates reproducibility as well. :)
To make an image reproducibile you would want to specify an image using the content hash.
```
docker image inspect qnib/spack-gromacs-shp2:c5n_18xl_off |jq '.[]|.RepoDigests'
[
  "qnib/spack-gromacs-shp2@sha256:4f5ae272e343395de2259e2745ab24f20a34688343f9f233cec895915f1704f5"
]
```
and not `qnib/spack-gromacs-shp2:c5n_18xl_off`

ChristianKniep: That might be a thing btw. inspecting the actual hash and putting it into the job output as an strict description.
For some images like `ubuntu:18.04` you actually specify a manifest list (image index) and not a 'real' image.

reframe/core/containers.py
```diff
@@ -119,32 +116,45 @@ class Sarus(ContainerPlatform):
...
+        if self.skip_pull:
+            return []
 
         return [self._command + ' pull %s' % self.image]
 
     def launch_command(self):
         super().launch_command()
         run_opts = ['--mount=type=bind,source="%s",destination="%s"' %
                     mp for mp in self.mount_points]
+        run_opts.append('--workdir='+self.workdir)
```
ChristianKniep, line 147: Cleaner then just changing the dir with `cd`

reframe/core/containers.py
```diff
@@ -119,32 +116,45 @@ class Sarus(ContainerPlatform):
...
         if self.with_mpi:
             run_opts.append('--mpi')
 
         run_opts += self.options
-        run_cmd = self._command + ' run %s %s bash -c ' % (' '.join(run_opts),
+        run_cmd = self._command + ' run %s %s' % (' '.join(run_opts),
```
ChristianKniep, line 152: Again, the entrypoint you specify here (`bash -c`) might mess with the container image and renders it unusable.

ChristianKniep (commented):
I put some inline comments into the PR to explain what I am doing.

teojgo:
@ChristianKniep make sure you adjust also the unittests

reframe/core/containers.py
```diff
@@ -119,32 +116,57 @@ class Sarus(ContainerPlatform):
...
+        if self.image.startswith('load/') or self.skip_pull:
             return []
-
-        return [self._command + ' pull %s' % self.image]
+        # Using 
+        if self.with_metahub:
+            tmpdir = tempfile.mkdtemp()
+            return [
+                'docker pull -q mh.qnib.org/'+self.image,
+                'docker save -o '+tmpdir+'/image.tar mh.qnib.org/'+self.image,
```
vkarak, line 148: Couldn't `sarus pull` work here? I don't understand very much the purpose of the `save` and `load`.

vkarak (changes requested):
I guess we could also extend the Docker backend in a similar way, couldn't we?

reframe/core/containers.py
```diff
@@ -119,32 +116,57 @@ class Sarus(ContainerPlatform):
...
+    #:
+    #: :type: boolean
+    #: :default: :class:`False`
+    skip_pull = fields.TypedField(bool)
+
     def __init__(self):
         super().__init__()
         self.with_mpi = False
         self._command = 'sarus'
+        self.with_metahub = False
```
teojgo, line 134: This is not needed since you can specify some hooks to be run before each step of the test pipeline.

reframe/core/containers.py
```diff
@@ -119,32 +116,57 @@ class Sarus(ContainerPlatform):
...
+                'docker save -o '+tmpdir+'/image.tar mh.qnib.org/'+self.image,
+                self._command + ' load ${TMP_DIR}/gromacs.tar mh.qnib.org/'+self.image,
+                'rm -rf '+tmpdir
+            ]
+        else:
+            return [self._command + ' pull %s' % self.image]
 
     def launch_command(self):
         super().launch_command()
         run_opts = ['--mount=type=bind,source="%s",destination="%s"' %
```
teojgo, line 157:
```suggestion
        run_opts = [f'--mount=type=bind,source="{mp}",destination="{mp}"' for mp in self.mount_points]
```

reframe/core/containers.py
```diff
@@ -119,32 +116,57 @@ class Sarus(ContainerPlatform):
...
                     mp for mp in self.mount_points]
+        run_opts.append('--workdir='+self.workdir)
```
teojgo, line 159:
```suggestion
        run_opts.append(f'--workdir={self.workdir}')
```

reframe/core/containers.py
```diff
@@ -119,32 +116,57 @@ class Sarus(ContainerPlatform):
...
             run_opts.append('--mpi')
 
         run_opts += self.options
-        run_cmd = self._command + ' run %s %s bash -c ' % (' '.join(run_opts),
+        run_cmd = self._command + ' run %s %s' % (' '.join(run_opts),
                                                            self.image)
-        return run_cmd + "'" + '; '.join(
-            ['cd ' + self.workdir] + self.commands) + "'"
+        if self.commands:
+            return run_cmd + " '" + '; '.join(self.commands) + "'"
```
teojgo, line 167: If we are not going to use `bash -c` then we don't need the single quotes nor we need the `;`. I think it's better if the user passes the whole command verbatim.

reframe/core/containers.py
```diff
@@ -119,32 +116,57 @@ class Sarus(ContainerPlatform):
...
         super().launch_command()
         run_opts = ['--mount=type=bind,source="%s",destination="%s"' %
                     mp for mp in self.mount_points]
+        run_opts.append('--workdir='+self.workdir)
         if self.with_mpi:
             run_opts.append('--mpi')
 
         run_opts += self.options
-        run_cmd = self._command + ' run %s %s bash -c ' % (' '.join(run_opts),
+        run_cmd = self._command + ' run %s %s' % (' '.join(run_opts),
```
teojgo, line 164:
```suggestion
        run_cmd = f"{self._command}  run {' '.join(run_opts)} {self.image}"
```

teojgo (changes requested):

reframe/core/containers.py
```diff
@@ -119,32 +116,57 @@ class Sarus(ContainerPlatform):
     #: :default: :class:`False`
     with_mpi = fields.TypedField(bool)
 
+    #: Using metahub to fetch instance specific images
+    #:
+    #: :type: boolean
+    #: :default: :class:`False`
+    with_metahub = fields.TypedField(bool)
```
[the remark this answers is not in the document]
ChristianKniep: Usually the registry is part of the string that is the image name. 
`ubuntu` is actually `docker.io/library/ubuntu:latest`
The first part is the registry, the second the org and the last piece the image name.
I might check the prefix and react accordingly.

But in a greater scheme of things the question becomes how do you deal with container images compared to binaries installed on a system?

reframe/core/containers.py
ChristianKniep: yes, `sarus pull` seems to need a API call that the proxy Metahub is not implementing (yet). working with the SARUS folks to fix that soon(-ish).

teojgo: In this case, you should put this logic on the test as a prerun step. When sarus can pull directly, you just remove the workaround from the particular test.

teojgo: Btw, you can use [skopeo](https://github.com/containers/skopeo) to copy the image directly as a docker-archive and load it with sarus

teojgo:
@ChristianKniep I can take this issue and work on it. Is that fine?

kniec:
@teojgo Sure, let me know if I can help

vkarak:
I am closing this, since it's replaced by #1775.

PR #1750, opened 2021-02-05
[test] Port QuantumESPRESSO checks to Eiger

lucamar:
Setting `num_cpus_per_task=16` and `OMP_NUM_THREADS=8` should yield a better performance with respect to `num_cpus_per_task=8` and `'OMP_NUM_THREADS': str(self.num_cpus_per_task)`.
In order to improve a bit more the performance of the test, I have also added the Slurm directive `#SBATCH --distribution=block:block` and the `srun` command line option `--cpu-bind=cores` using pipeline hooks.


cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -39,17 +39,43 @@ class QuantumESPRESSOCpuCheck(QuantumESPRESSOCheck):
...
+                    'OMP_NUM_THREADS': '8',
+                    'OMP_PLACES': 'cores',
+                    'OMP_PROC_BIND': 'close'
+                }
         else:
-            self.num_tasks = 576
             energy_reference = -11427.09017152
-
-        self.num_tasks_per_node = 36
+            if self.current_system.name in ['daint', 'dom']:
```
ekouts, line 64:
```suggestion
            if self.current_system.name in ['daint']:
```
It doesn't hurt to keep it, but we don't run `large` on `dom`

ekouts (approved):
Lgtm in general, I have only a small comment.

cscs-checks/apps/quantumespresso/quantumespresso_check.py
lucamar: Thanks @ekouts

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -65,26 +91,37 @@ def __init__(self, scale, variant):
...
         self.reference = references[variant][scale]
         self.tags |= {'maintenance' if variant == 'maint' else 'production'}
 
+    @rfm.run_before('run')
+    def set_task_distribution(self):
+        self.job.options = ['--distribution=block:block']
+
+    @rfm.run_before('run')
+    def set_cpu_binding(self):
+        self.job.launcher.options = ['--cpu-bind=cores']
```
vkarak, line 124: Are these meant for eiger only or are they ok for the rest of the systems, too?

lucamar: I tested them on Eiger only for the moment, since they were boosting a bit the performance.

vkarak: I am asking because as you have coded it now, it will run with these options on every system. The CI does not check for performance in this case, because the test is marked as non-strict (which we could change tbh).

lucamar: Thanks for checking @vkarak: as far as I could test on Dom and Piz Daint, the additional Slurm options do not harm the performance of this check. We have been using the same options already with the CP2K check on Dom and Piz Daint.

vkarak:
@jenkins-cscs retry eiger

PR #1752, opened 2021-02-08
[feat] Replace getdep with getdeps method

### LINKED ISSUE #1381: Extend the @require_deps decorator to work better with parameterized tests
It would be maybe useful to be able to pass the name of the class as a parameter to the decorated function and have access to all the parameterized tests (or maybe just the ones that the test depends on). If we implement this we should have a way to get the variant of the parameterized tests.

Not sure what is the best way to support dependencies on parameterized tests, this is just a suggestion.

ekouts:
Closes #1381.

### DIFF
--- a/reframe/core/parameters.py
+++ b/reframe/core/parameters.py
@@ -245,5 +245,8 @@ class ParamSpace:
     def __getitem__(self, key):
         return self._params.get(key, ())
 
+    def __contains__(self, key):
+        return key in self._params
+
     def is_empty(self):
         return self._params == {}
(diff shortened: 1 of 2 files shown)


ekouts:
@vkarak I have some questions about this PR, because I am not sure what is part of it and what we should implement later.
1. Should this wait for the extension of `user_deprecation_warning` and should we deprecate `require_deps` as well?
2. Should we write completely new unittests for `getdeps` or just replace the calls of `getdep`?
3. Should we remove/change the documentation that uses `require_deps` and show how to do it with `getdeps`?

jjotero:
@ekouts @vkarak with #2166 in mind, is this still needed? I think the fixtures also address #1381 (not through the `@require_deps` decorator of course, but the functionality to do a reduction is there).

vkarak:
@jjotero I think so, too. I would suggest closing this PR.

PR #1754, opened 2021-02-08
[bugfix] Fix "New in version" annotation for some sanity functions

vkarak:
I adapted the `.. versionadded::` annotation to the correct version.


victorusu (approved):
`lgtm`

PR #1755, opened 2021-02-08
[test] Adjust regex patterns for CP2K 8.1

lucamar:
The pull request addresses minor changes in the output of CP2K 8.1 with respect to previous releases:
- the step keyword `STEP NUMBER` now becomes `MD| Step number`
- the energy units `(a.u.)` are now enclosed in square branckets `[a.u.]`

### DIFF
--- a/cscs-checks/apps/cp2k/cp2k_check.py
+++ b/cscs-checks/apps/cp2k/cp2k_check.py
@@ -15,14 +15,14 @@ class Cp2kCheck(rfm.RunOnlyRegressionTest):
         self.executable_opts = ['H2O-256.inp']
 
         energy = sn.extractsingle(r'\s+ENERGY\| Total FORCE_EVAL \( QS \) '
-                                  r'energy \(a\.u\.\):\s+(?P<energy>\S+)',
+                                  r'energy [\[\(]a\.u\.[\]\)]:\s+(?P<energy>\S+)',
                                   self.stdout, 'energy', float, item=-1)
         energy_reference = -4404.2323
         energy_diff = sn.abs(energy-energy_reference)
         self.sanity_patterns = sn.all([
             sn.assert_found(r'PROGRAM STOPPED IN', self.stdout),
             sn.assert_eq(sn.count(sn.extractall(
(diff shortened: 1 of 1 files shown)


lucamar:
I don't know how to shorten line 82, as reported by pep8speaks: could you please add a suggestion directly?

lucamar:
I have indented all the `energy` block to reduce the length of line 18, in order to comply with the ReFrame Coding Style Guide as reported by pep8speaks.

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -14,15 +14,17 @@ def __init__(self):
         self.executable = 'cp2k.psmp'
         self.executable_opts = ['H2O-256.inp']
 
-        energy = sn.extractsingle(r'\s+ENERGY\| Total FORCE_EVAL \( QS \) '
-                                  r'energy \(a\.u\.\):\s+(?P<energy>\S+)',
-                                  self.stdout, 'energy', float, item=-1)
+        energy = sn.extractsingle(
+            r'\s+ENERGY\| Total FORCE_EVAL \( QS \) '
+            r'energy [\[\(]a\.u\.[\]\)]:\s+(?P<energy>\S+)',
```
vkarak, line 19: I think this can be written in a simpler way:

```suggestion
            r'energy \[a\.u\.\]:\s+(?P<energy>\S+)',
```

vkarak (changes requested):

cscs-checks/apps/cp2k/cp2k_check.py
lucamar: Would this regex match the old CP2K 7.1 output `(a.u)` as well? I tried to match both outputs since CP2K 7.1 is the current default on Piz Daint (even if for only two more days, the check would still fail).

vkarak: Ok then, it can't be written more simply then :-)

vkarak (approved):

Issue #1742, opened 2021-02-03
Automatically configure ReFrame for the underlying system


vkarak:
It would be nice if ReFrame could automatically detect the system it runs on (if no configuration is given) and generate a configuration file. The following things need to be detected at least:

1. System partitions
    - Modules system
    - Local host architecture (micro-arch + topology)
    - Job schedulers
    - Parallel launchers 
2.  Environments
    - This is a bit challenging. Spack is doing a similar thing in auto-detecting the compilers, so we could use Spack for this, but what about EasyBuild? We could potentially look for module name patterns. But what if the s/w stack is hidden behind a top-level module?
    - Any environments defined should follow a certain naming convention, so that tests can define `valid_prog_environs` accordingly. Similar thing should be done for `valid_systems`.

@ChristianKniep You can help formulate this. What information would you guys need from a system auto-detection?

kniec:
As a MVP, I'd say focusing on single-node jobs with local execution (of containerized IMHO) while detecting the architecture and topology (core count, NUMA layout) is a great first building block.
This allows to define expected performance values for different system setups (on a 36core Skylake with hyperthreading off I expect XYZ). With that you can run this local test as a performance test on deployment. 

Second step (IMHO) is to allow for multi-node jobs by adding the interconnect and add expected performance value per interconnect setup. Same goal; running a performance test after setting up the infrastructure and installing the software.
With containers you skip the module problem (even though I am biased towards containers). :)
Not sure for my use-case we do need to detect the programming environment as compilation won't be done.

vkarak:
> As a MVP, I'd say focusing on single-node jobs with local execution (of containerized IMHO) while detecting the architecture and topology (core count, NUMA layout) is a great first building block.
This allows to define expected performance values for different system setups (on a 36core Skylake with hyperthreading off I expect XYZ). With that you can run this local test as a performance test on deployment.

As a first target, that's doable. The devil might always might be in the details, as always, but I don't see a showstopper right away.

> Second step (IMHO) is to allow for multi-node jobs by adding the interconnect and add expected performance value per interconnect setup. Same goal; running a performance test after setting up the infrastructure and installing the software.

[...]

PR #1756, opened 2021-02-08
[test] New affinity check

jjotero:
This new affinity check reads the node configuration from a reference file which contains the output of `lscpu -e`. This builds the cpu sets for each NUMA region, socket, etc; so all the reframe tests are now independent from the system. However, since slurm settings could potentially affect a given system, each test has the flexibility to prescribe system specific options too.

Now, to add a new system, one only needs to add one reference file per system and extend the `self.valid_systems`.


vkarak:
I just read the description of the PR. Wouldn't `lscpu -J` help, which spits everything in JSON?

vkarak:
You can even combine `lscpu -e -J`

jjotero:
> I just read the description of the PR. Wouldn't `lscpu -J` help, which spits everything in JSON?

Sure, I just didn't have a preference for that. I just updated it to use JSON files.

vkarak:
> Sure, I just didn't have a preference for that. I just updated it to use JSON files.

I proposed JSON because it's much easier to parse within Python. I will have a look in the test later on.

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,520 @@ def __init__(self, variant):
         self.sourcesdir = os.path.join('src/affinity_ref')
         self.prebuild_cmds = ['git clone https://github.com/vkarak/affinity']
```
jgphpc, line 36: why not using /apps/common/UES/reframe/resources/ here ? you can always the link to the repo as a comment for non CSCS users

vkarak: @jgphpc I would like that the tests are self contained, so for me it's better if it doesn't use the resources dir.

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,520 @@ def __init__(self, variant):
...
         self.executable = './affinity/affinity'
-        self.variant = variant
+        self.sanity_patterns = self.assert_consumed_cpu_set()
         self.maintainers = ['RS', 'SK']
```
jgphpc, line 39: you may want to add them as reviewers ?

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,520 @@ def __init__(self, variant):
...
-    @rfm.run_before('sanity')
-    def set_sanity(self):
+        # Dict with the partition's topology - output of "lscpu -e"
+        self.topology = {
+            'dom:gpu':   'topo_dom_gpu.json',
+            'dom:mc':    'topo_dom_mc.json',
+            'daint:gpu': 'topo_dom_gpu.json',
+            'daint:mc':  'topo_dom_mc.json',
+            'eiger:mc':  'topo_eiger_mc.json',
+        }
```
jgphpc, line 49: no file for ault ?

cscs-checks/prgenv/affinity_check.py
```diff
@@ -3,15 +3,29 @@
...
+    file is simply the output of the `lscpu -e` command. For more info on this,
+    see the `read_proc_topo` hook.
+    '''
+
+    def __init__(self):
         self.valid_systems = ['daint:gpu', 'daint:mc',
-                              'dom:gpu', 'dom:mc']
+                              'dom:gpu', 'dom:mc', 'eiger:mc',
+                              'ault:a64fx', 'ault:amdv100']
         self.valid_prog_environs = ['PrgEnv-gnu']
```
jgphpc, line 29: do you expect the same result with other compilers ?

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,520 @@ def __init__(self, variant):
...
+    def assert_aff_set_length(self):
+        '''Only 1 CPU pinned per thread.'''
+        if not all(len(aff_set) == 1 for aff_set in self.aff_cpus):
+            raise SanityError('incorrect affinity set')
+
+
+@rfm.simple_test
+class OneThreadPerSocketOpenMP(AffinityOpenMPBase):
+    '''Pin each OMP thread to a different socket.'''
+    parameter('omp_bind', ['socket'])
```
jgphpc, line 331: I get `libgomp: Invalid value for environment variable OMP_PLACES`
According to https://www.openmp.org/spec-html/5.0/openmpse53.html, it should be:
```suggestion
    parameter('omp_bind', ['sockets'])
```

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,520 @@ def __init__(self, variant):
...
+            'OMP_PROC_BIND': self.omp_proc_bind,
         }
 
+    @rfm.run_before('sanity')
+    def consume_cpu_set(self):
+        raise ValueError('this function must be overridden')
+
+
+@rfm.simple_test
+class OneThreadPerCPUOpenMP(AffinityOpenMPBase):
```
jgphpc, line 241: minor comment: `OneThreadPerCPU` is a bit misleading name as we have here more than 1 thread per "cpu". I would suggest to rename (but feel free to ignore):
```
OneThreadPerCore -> OneThreadPerPhysicalCore
OneThreadPerCPU -> OneThreadPerLogicalCore
```
?

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,520 @@ def __init__(self, variant):
...
+
+    @rfm.run_before('sanity')
+    def assert_aff_set_length(self):
+        '''Only 1 CPU pinned per thread.'''
+        if not all(len(aff_set) == 1 for aff_set in self.aff_cpus):
+            raise SanityError('incorrect affinity set')
+
+
+@rfm.simple_test
+class OneThreadPerSocketOpenMP(AffinityOpenMPBase):
```
jgphpc, line 329: `OneTaskPerSocketOpenMP` looks similar to this class.
What the difference ?

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,520 @@ def __init__(self, variant):
...
         }
 
+    @rfm.run_before('sanity')
+    def consume_cpu_set(self):
+        raise ValueError('this function must be overridden')
+
+
+@rfm.simple_test
+class OneThreadPerCPUOpenMP(AffinityOpenMPBase):
+    '''Pin each OMP thread to a different CPU.'''
```
jgphpc, line 242: Would it make sense to also have a class for `OneThreadPerCPUOpenMPnomultithread` ?

jjotero (commented):
Updated the config file for ault to get the tests passing here for the `amdv100` partition. I've taken out the `a64fx` partition from the `valid_systems`, since I couldn't get it to work. It looks like the settings for ault have to be revised anyways, so those should be fixed later on in a separate PR.
Also, the naming of the checks has been changed to remove ambiguities.

sebkelle1 (approved):

vkarak:
How urgent is that we merge this? Because it can be really simplified after we merge #1773. The idea is that we put all this information in the configuration and make it available under `current_partition`. So I suggest commenting in #1773 what is the information that we need there.

jjotero:
I don't see how adding all this information from the config file would simplify this test. If so, I believe it would make the process of porting this test to a new partition a lot more complex, given that you manually have to add all the data into the config file; whereas now you just dump the JSON file and off you go. 
My idea with this test was that once we start supporting fixtures, the whole process of fetching the partition topology can be made a fixture that any test could use if needed. That way you remove the need to hardcode any settings anywhere.

vkarak:
Check also this #1742. The idea is that we want to generate the configuration automatically. So all this topology discovery machinery could be part of the "auto-configuration". Other tests might want to know at least some high level information in order to configure themeselves. Why should they have to re-write that. Of course, this could be offered separately in a fixture library, but that's something we need to see.

jgphpc (approved):

jjotero:
@vkarak I see. I personally tend to vote against those "auto" features, because then it won't work for some users, and then they complain 😂.
Regarding the urgency of this PR, it is part of the Eiger tests, so I guess we should merge it at some point. We could also push it with a  `FIXME` in there and then we use this test as the "guinea pig" for all that stuff and see which way looks better.

@jgphpc The the fixture idea would transform that into something like
```python
@rfm.fixture
class NodeTopology:
   ...

# and then in the test
@rfm.simple_test
class MyAffinityTest(rfm.RegressionTest):
    topology = fixture(NodeTopology, scope='partition')
    
    def __init__(self):
...
```
or something like that.

vkarak:
>  I personally tend to vote against those "auto" features, because then it won't work for some users, and then they complain 😂.

I understand this and that's why ReFrame didn't have such automations until now. The requested feature is more on auto-detecting the local node, which you can do in a way more controlled way. On the other hand, since the detection happens locally, fixtures could do the job nicely and I would vote for a `cpuinfo` fixture that would give you also the `archspec` information apart from the topology. Fixtures can also be made smart enough to run also on remote partitions. Regarding the syntax, that could be a potential one, but I am also thinking something more pytest-like. But we will discuss this.

> Regarding the urgency of this PR, it is part of the Eiger tests, so I guess we should merge it at some point. We could also push it with a FIXME in there and then we use this test as the "guinea pig" for all that stuff and see which way looks better.

I'm fine with a `FIXME` note for the moment.

cscs-checks/prgenv/affinity_check.py
vkarak: > unless some compilers decide to not comply to the OMP standard

This is a regression, isn't it? In fact, we have seen differences in the past. Keep in mind that the compiler wrappers may also do some magic that's not visible immediately.

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,523 @@ def __init__(self, variant):
...
-            sn.assert_eq(sn.sorted(self.aff_cores), sn.sorted(self.ref_cores))
-        ])
+
+        def parse_cpus(x):
+            return sorted([int(xi) for xi in x.split()])
+
+        with osext.change_dir(self.stagedir):
+            self.aff_cpus = sn.extractall(
+                re_aff_cpus, self.stdout, 'cpus', parse_cpus
+            ).evaluate()
```
vkarak, line 172: Why do you need `evaluate()` here and not in the the rest?

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,523 @@ def __init__(self, variant):
...
+
```
vkarak, line 173: You can remove the blank lines in between.

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,523 @@ def __init__(self, variant):
...
-                'ref_dom:mc': 'mc_omp_bind_cores.txt',
-                'num_cpus_per_task:gpu': 12,
-                'num_cpus_per_task:mc': 36,
-                'ntasks_per_core': 1,
-                'multithreading': None,
-                'OMP_PLACES': 'cores',
-            },
-        }
-        self.variant = variant
+        '''Hook to control multithreading settings for each system.'''
```
vkarak, line 184: Leave a blank line after docstrings.

vkarak (changes requested):

cscs-checks/prgenv/affinity_check.py
jjotero: I just copied those over from the previous test, but they're not used for these new tests. That's why I didn't notice that they had that `evaluate()` missing.

I'm taking `aff_thrds` and `aff_ranks` out.

jjotero (commented):
I've also updated the syntax of the tests to comply with the new `variable` syntax. Now the `AffinityBase` includes three new variables which can be used to control the slurm settings.

cscs-checks/prgenv/affinity_check.py
vkarak: I would probably put explicitly the different programming environments that we have (cray, gnu, intel, pgi, aocc)

vkarak: Because, for example, it's not so relevant for the `builtin`.

jjotero: Makes sense. I'll change that. Would it make sense to extend the `valid_prog_environs` syntax to allow something like 
`self.valid_prog_environs=['^builtin']`, which would use all the programming environments available except for `builtin`?

jjotero: `PrgEnv-aocc` has been left out for now (JG has already opened an internal issue on it).

cscs-checks/prgenv/affinity_check.py
```diff
@@ -21,166 +35,523 @@ def __init__(self, variant):
...
+                lambda y: int(y['cpu']),
+                filter(
+                    lambda x: x[by] == sibiling_id,
+                    self._lscpu
+                )
+            )
+        )
+
+    @sn.sanity_function
+    def assert_consumed_cpu_set(self):
```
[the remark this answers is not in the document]
vkarak: Oh, yes, I meant to comment on the `consume_cpu_set`. Regarding the `rfm.abstracthook`, we could possibly consider it.

cscs-checks/prgenv/affinity_check.py
```diff
@@ -3,184 +3,547 @@
...
-            # Both OMP_PROC_BIND values CLOSE and SPREAD give the same
-            # result as OMP_PROC_BIND=TRUE when all cores are requested.
+            'OMP_NUM_THREADS': str(self.num_omp_threads),
+            'OMP_PLACES': self.omp_bind,
+            'OMP_PROC_BIND': self.omp_proc_bind,
         }
 
+    @rfm.run_before('sanity')
+    def consume_cpu_set(self):
+        raise ValueError('this function must be overridden')
```
vkarak, line 252: If this is meant to act as an abstract method, better raise a `NotImplementedError`.

vkarak (changes requested):

cscs-checks/prgenv/affinity_check.py
```diff
@@ -18,17 +18,37 @@ class AffinityTestBase(rfm.RegressionTest):
...
+    # Variable to specify system specific launcher options. This will override
+    # any of the above global options.
+    #
+    # Example:
+    #
+    # system = {
+    #     'daint:mc': {
+    #         multithreading = False,
+    #         cpu_bind = 'none',
+    #         hint = 'nomultithread',
```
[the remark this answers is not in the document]
jjotero: Oh my ... it actually took me a good minute to figure out what was wrong here 😂. It's funny though how I typed right the outer level, and then decided to throw the whole Python standard out of the window.

vkarak (approved):

cscs-checks/prgenv/affinity_check.py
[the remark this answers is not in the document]
vkarak: You're too much into the ReFrame variable syntax 😂

jgphpc:
1 last minute comment about AlternateSocketFilling/ConsecutiveSocketFilling
On AMD, you have 2 sockets, with 4 numa nodes per socket. 
The name could be AlternateNuma/ConsecutiveNuma

jjotero:
@jgphpc Good catch! In fact, the number of `AlternateSocketFilling` is just wrong. I've been playing with the slurm settings and it seems that there is no easy way to force an actual alternate socket filling in here.

I'll fix that sanity function and rename the tests.

jjotero:
I've now re-run all these tests and they all pass. As I mentioned above, it's not straightforward to me how to actually do an alternate socket filling. This could be quite important for systems that might have (say) 2 devices, and each device with affinity to different sockets. I'll keep experimenting in ault about this, but as far as this PR is concerned, I think it can be merged.

jjotero:
@jenkins-cscs retry eiger

jjotero:
@jenkins-cscs retry all

PR #1757, opened 2021-02-08
[test] Fix minor bug in sanity pattern for GPU burn test

akesandgren:
The perf value may have two space in front of it if <1000.
There is never a space between the GPU number and the (OK) value.


vkarak:
ok to test

jjotero (approved):
lgtm

vkarak (approved):
lgtm too.

vkarak:
Thanks @akesandgren!

PR #1762, opened 2021-02-10
[bugfix] Add Python version check to bootstrap script

### LINKED ISSUE #1743: Misleading Python Bootstrapping Version
One can bootstrap ReFrame but not be able to run it due to wrong Python version.

Example
```
$ git clone https://github.com/eth-cscs/reframe.git
$ cd reframe
$ ./bootstrap.sh
$ ./bin/reframe -C config/ci_config.py -ccscs-checks/prgenv/helloworld.py -r
Unsupported Python version: Python >= 3.6.0 is required
```

Can we have a check for the python version that will be used to bootstrap ReFrame?
What do you think?

teojgo:
Fixes #1743


bootstrap.sh
```diff
@@ -57,7 +64,13 @@ while [ -n "$1" ]; do
...
 done
 
-pyver=$($python -V | sed -n 's/Python \([0-9]\+\)\.\([0-9]\+\)\..*/\1.\2/p')
+pyver=$($python -V 2>&1 | sed -n 's/Python \([0-9]\+\)\.\([0-9]\+\)\..*/\1.\2/p')
+pymajver=$(echo $pyver | cut -f 1 -d '.')
+pyminver=$(echo $pyver | cut -f 2 -d '.')
+
+if [ $pymajver -lt 3 ] || [ $pyminver -lt 6 ]; then
+    PYTHON_REQ $pyver
+fi
```
vkarak, line 73: I don't like that we have to redo some of the logic of the `reframe` module in Bash. We could simply run `./bin/reframe -V` and check if it fails. In any case, `./bootstrap.sh` is meant to be executed from the top-level resource directory.

vkarak: We could have something like: `echo "Bootstrapping ReFrame $(./bin/reframe -V) ..."`

vkarak (changes requested):

bootstrap.sh
teojgo: The problem with this is that it won't respect a particular python passed as a cli option to the bootstrap script.

vkarak: Ok, but the problem is that this is not too portable. This `sed` command, for example, is not valid under OSX. You need GNU sed for this. I would suggest doing the check in a smart way directly through Python:

```bash
# NOTE: We need to exit with a zero code if the Python version is the correct one, so we invert the comparison
if python -c 'import sys; sys.exit(sys.version_info[:2] > (3, 6))'; then
    echo -e "ReFrame requires Python >= 3.6"
fi
```

vkarak (changes requested):

vkarak (approved):

vkarak:
@jenkins-cscs retry eiger

PR #1770, opened 2021-02-11
[ci] Add spack recipe generation test

victorusu:
I added to the Spack recipe generation to the wrong workflow file just to test. This allows me to check the generated file without having to make a release.
I will move the actions to the right place as soon as we have a definite solution.

Currently, we still miss
- [ ] Make a PR to Spack Repo (we need a fork for that - we can use mine)
- [ ] Move the action to the correct workflow path (release stage)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -86,7 +86,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: 3.8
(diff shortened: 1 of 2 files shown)


vkarak (changes requested):
I would vote for Jinja2 templates in a separate requirements file just for the `ci-scripts`.

vkarak:
Will not pursue this further.

PR #1772, opened 2021-02-15
[test] Adapt NetCDF test to Eiger

ajocksch:
according to Cray only dynamic linking is supported on eiger


cscs-checks/libraries/io/netcdf_compile_run.py
```diff
@@ -22,15 +22,23 @@ def __init__(self, lang, linkage):
...
+
         if self.current_system.name in ['daint', 'dom']:
             self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu',
                                         'PrgEnv-intel', 'PrgEnv-pgi']
             self.modules = ['cray-netcdf']
         elif self.current_system.name in ['arolla', 'tsa']:
             self.exclusive_access = True
             self.valid_prog_environs = ['PrgEnv-gnu-nompi', 'PrgEnv-pgi-nompi']
+        elif self.current_system.name in ['eiger']:
+            self.valid_prog_environs = ['builtin']
```
vkarak, line 40: @ajocksch Why don't we use the different programming environments here?

ajocksch: this was a mistake; I deleted 'builtin' and added PrgEnv-cray and PrgEnv-gnu

cscs-checks/libraries/io/netcdf_compile_run.py
```diff
@@ -22,15 +22,23 @@ def __init__(self, lang, linkage):
...
         self.descr = f'{lang_names[lang]} NetCDF {linkage.capitalize()}'
-        self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc',
-                              'arolla:cn', 'tsa:cn']
+        self.valid_systems = ['daint:gpu', 'daint:mc',
+                              'dom:gpu', 'dom:mc',
+                              'arolla:cn', 'tsa:cn',
+                              'eiger:mc']
+        if linkage == 'static':
+            self.valid_systems.remove('eiger:mc')
+
```
ekouts, line 31:
```suggestion
        self.valid_systems = ['daint:gpu', 'daint:mc',
                              'dom:gpu', 'dom:mc',
                              'arolla:cn', 'tsa:cn']
        if linkage == 'dynamic':
            self.valid_systems.append('eiger:mc')

```
Doesn't it make more sense to add it only in case of dynamic instead of removing it?

ekouts (approved):
lgtm. I have one style comment but it could be merged as it is

cscs-checks/libraries/io/netcdf_compile_run.py
vkarak: I thought about it, but it's more explicit if you remove it: you express your intent more clearly: static linkage is not supported in `eiger:mc`.

vkarak (approved):

PR #1773, opened 2021-02-15
[feat] Add processor and device information to the system partitions and make them accessible from the tests

### LINKED ISSUE #1709: Add more properties to the system partitions in the configuration accessible from the tests
For the beginning, such properties can be the following: `num_cores`, `num_sockets`, `memory` etc.

This was originally requested by the [EESSI](https://www.eessi-hpc.org/) guys.

ekouts:
Fixes #1709 .
So far I have only added `arch`, `sockets_per_node`, `cores_per_socket` and `threads_per_core`. User can use these in a test, as well as `self.current_partition.cores_per_node` and `self.current_partition.threads_per_node` that are derived from these. What do you think about the `/scratch-shared/` and `/scratch-local/` that are mentioned on the issue?

Would it make more sense to allow users define custom properties on the configuration?


reframe/schemas/config.json
```diff
@@ -471,7 +475,10 @@
         "systems/partitions/resources/options": [],
         "systems/partitions/modules": [],
         "systems/partitions/variables": [],
-        "systems/partitions/environs": [],
```
ekouts, line 474: I deleted this because it's already defined some lines above.

vkarak:
> What do you think about the /scratch-shared/ and /scratch-local/ that is mentioned on the issue?

These are too specific to the EESSI setup afaiu. I would suggest though to have an attribute that would keep an arbitrary dictionary, so that sites define site-specific stuff in it, which it would be accessible through the `current_partition`.

docs/config_reference.rst
```diff
@@ -316,6 +316,47 @@ System Partition Configuration
...
+   :required: No
+   :default: ``""``
+
+   The microarchitecture of this partition.
+
+
+.. js:attribute:: .systems[].partitions[].sockets_per_node
+
+   :required: No
+   :default: ``-1``
```
rsarm, line 330: What about  `None` instead?

docs/config_reference.rst
```diff
@@ -316,6 +316,47 @@ System Partition Configuration
...
+
+   The number of socket per node on this partition.
```
rsarm, line 332:
```suggestion
   The number of sockets per node on this partition.
```

reframe/core/systems.py
```diff
@@ -187,6 +193,86 @@ def environment(self, name):
...
+
+        .. versionadded:: 3.5
+
+        :type: :class:`str`
+        '''
+        return self._arch
+
+    @property
+    def sockets_per_node(self):
+        '''The number of socket per node on this partition.
```
rsarm, line 208: Same here `sockets`

vkarak (changes requested):
I would ideally go for this information:

```python
                    'processor': {
                        'arch': 'skylake',
                        'num_cpus': 8,
                        'num_cpus_per_core': 2,
                        'num_cpus_per_socket': 8,
                        'num_sockets': 1,
                        'topology': {
                            'numa_nodes': [0x000000ff],
                            'sockets': [0x000000ff],
                            'cores': [0x00000003, 0x0000000c,
                                      0x00000030, 0x000000c0]
                            'caches': [
                                {
                                    'type': 'L3',
                                    'size': 6291456,
                                    'linesize': 64,
                                    'associativity': 0,
                                    'num_cpus': 8,
                                    'cpusets': [0x000000ff]
                                },
                                {
                                    'type': 'L2',
                                    'size': 262144,
                                    'linesize': 64,
                                    'associativity': 4,
                                    'num_cpus': 2,
                                    'cpusets': [0x00000003, 0x0000000c,
                                                0x00000030, 0x000000c0]
                                },
                                {
                                    'type': 'L1',
                                    'size': 32768,
                                    'linesize': 64,
                                    'associativity': 0,
                                    'num_cpus': 2,
                                    'cpusets': [0x00000003, 0x0000000c,
                                                0x00000030, 0x000000c0]
                                }
                            ]
                        }
                    },
                    'devices': [
                        {
                            'type': 'gpu',
                            'arch': 'pascal',
                            'vendor': 'NVIDIA Corporation'
                        }
                    ],
                    'extras': {}
```

By `cpu` I mean here logical CPU. From these numbers one can derive everything else, for which we can have convenience properties in the `SystemPartition`, for example:

- `num_cores = num_cpus // num_cpus_per_core`
- `num_cores_per_socket = num_cores // num_sockets`
- `num_gpus = len(d for d in devices if d['type'] == 'gpu')`

All of this information should be optional, apparently, and the tests must treat properly the cases when this is not provided.

The topology information might seem too much for a user to type, but the idea is that this will be automatically generated. I derived this information directly from `hwloc`, which I think is a better solution than `lscpu` since it's more widespread (for example, I could not find `lscpu` for Mac). I also think that auto-detection is very natural to happen when we create the `SystemPartition`, since this also knows how to access remote nodes (we can think how this can be done efficiently).

Also when updating the schema, do not do that inside the `partition`. Write the definitions of each of these objects under the `definitions` section in the schema and reference them from the other places.

vkarak:
We can rethink the `devices` section, because the way I propose it here will have unnecessary duplication for systems with multiple GPUs, but for the rest I am more confident.

ekouts:
1. How will the user have access to the processor information? He can get `self.current_partition.processor` as a dictionary, and then with something like `self.current_partition.processor['num_cpus']` but then it's not very nice to have the extra attributes `num_cores`, `num_cores_per_socket` and `num_gpus`. What do you think?
2. `  'sockets': [0x000000ff], `
Do we want a list of strings of numbers? Same for `numa_nodes`, `cores` and `cpu_sets`.

vkarak:
> How will the user have access to the processor information? He can get self.current_partition.processor as a dictionary, and then with something like self.current_partition.processor['num_cpus'] but then it's not very nice to have the extra attributes num_cores, num_cores_per_socket and num_gpus. What do you think?

No `self.current_partition.processor`, directly `.num_cpus` etc.

> Do we want a list of strings of numbers? Same for numa_nodes, cores and cpu_sets.

I prefer the cpusets here. They are much more compact and they can be retrieved directly from `hwloc`. I don't expect users to fill the topology manually. Also we retrieve the human-readable lists from special properties in `current_partition`.

ekouts:
> No self.current_partition.processor, directly .num_cpus etc.

The reason I don't like exposing everything is that some could have meaning for both processor and device, like `arch`. And by keeping it a dictionary it is easy to expand in the future without breaking old tests.

> Do we want a list of strings of numbers? Same for numa_nodes, cores and cpu_sets.
I prefer the cpusets here. They are much more compact and they can be retrieved directly from hwloc. I don't expect users to fill the topology manually. Also we retrieve the human-readable lists from special properties in current_partition.

I am sorry, I made a typo. I was only asking if it's going to be a list of string or a list of numbers.

jgphpc:
I would suggest this format: https://github.com/RRZE-HPC/kerncraft/tree/master/examples/machine-files
Let me read the pr to check the differences.

ekouts:
@jenkins-cscs retry all

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+class Processor(jsonext.JSONSerializable):
+    '''A representation of a processor inside ReFrame.
+
+    .. warning::
+       Users may not create :class:`Processor` objects directly.
+    '''
+
+    def __init__(self, processor_info):
+        self._arch = None
+        self._num_cpus = None
```
jgphpc, line 25: - `num_cpus_per_socket`: lscpu, papi_avail and likwid-topology are using `Cores per socket` -> can we use num_cores_per_socket ?

ekouts: The `num_cores_per_socket` value is already available as a derived property `num_cores // num_sockets`.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+        self._num_cpus_per_core = None
+        self._num_cpus_per_socket = None
```
jgphpc, line 27: - `num_cpus_per_core`: lscpu, papi_avail and likwid-topology are using `threads per core` -> can we use `num_threads_per_core` ?

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+        self._num_sockets = None
+        self._topology = None
+        self._info = processor_info
+
```
jgphpc, line 31: can we add `num_numa_nodes` and `num_cores_per_numa_node` ?

reframe/core/systems.py
vkarak: @jgphpc The internal values of this will be the minimum required, from which we can derive other properties. I am strongly in favor of using `cpu` as the the name of any logical processing unit. This is consistent with Linux and I find it more precise than `thread`.

reframe/core/systems.py
vkarak: If we go with the `threads_per_core` terminology we should name our basic unit as "thread," which I don't like it. "Processing Unit" is the best term or "logical cpu". If this is our basic term, then "num_cpus_per_core" makes sense. I have no problem supporting the "threads_per_core" as an alias, but if we do, we should keep in mind that this also breaks another convention: it doesn't start with "num_". I would see the "threads per core" more like a description term rather than a programmatic one.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+        if processor_info == {}:
```
vkarak, line 32:
```suggestion
        if not processor_info:
```

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+            return
+
+        for var in ['arch', 'num_cpus', 'num_cpus_per_core',
+                    'num_cpus_per_socket', 'num_sockets', 'topology']:
```
vkarak, line 36: You don't need to reiterate on that. I'd rather restrict the JSON schema not to allow additional properties, so that you can set the attributes by simply iterating over the `processor_info` dictionary.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+    def num_sockets(self):
+        '''Number of sockets.
+
+        :type: integral
+        '''
+        return self._num_sockets
+
+    @property
+    def topology(self):
+        '''Topology of the architecture.
```
vkarak, line 90:
```suggestion
        '''Processor topology.
```

And we should expand this with an example.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+
+        :type: :class:`Dict[str, obj]`
+        '''
+        return self._topology
+
+    @property
+    def num_cores(self):
+        '''Number of cores. Derived by ``num_cpus // num_cpus_per_core``.
```
vkarak, line 98:
```suggestion
        '''Total number of cores.
```

No need to explain how it is derived.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+
+        :type: integral
```
vkarak, line 100: Or `None` if no information is available. Same to every other property.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+        '''
+        if self._num_cpus and self._num_cpus_per_core:
+            return self._num_cpus // self._num_cpus_per_core
+        else:
+            return None
+
+    @property
+    def num_cores_per_socket(self):
+        '''Number of cores per socket. Derived by
+        ``num_cores // num_sockets``.
```
vkarak, line 110: Remove how the property is derived from everywhere.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+
+        :type: integral
+        '''
+        if self.num_cores and self._num_sockets:
+            return self.num_cores // self._num_sockets
+        else:
+            return None
+
+    def __str__(self):
+        return json.dumps(self._info, indent=2)
```
vkarak, line 120: Why this? You probably need the `__rfm_json_encode__()`.

reframe/core/systems.py
vkarak: Yes, I think these make sense as calculated properties and we should perhaps add more like the cpu lists from the different cpu sets.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+    def __init__(self, device_info):
+        self._type = None
+        self._arch = None
+        self._vendor = None
+        self._info = device_info
+
+        if device_info == {}:
+            return
+
+        for var in ['arch', 'type', 'vendor']:
```
vkarak, line 139: We should revise the way `device_info` is defined. The problem with my original definition is that on systems with multiple GPUs, we will end up having 8 same entries.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+    @property
+    def arch(self):
+        '''The architecture of the device.
+
+        :type: :class:`str`
+        '''
+        return self._arch
+
+    @property
+    def type(self):
```
vkarak, line 160: I would call it `device_type`  everywhere, in order not to confuse with Python's `type`.

reframe/core/systems.py
```diff
@@ -13,6 +13,169 @@
...
+    @property
+    def vendor(self):
+        '''The vendor of the device.
+
+        :type: :class:`str`
+        '''
+        return self._vendor
+
+    def __str__(self):
+        return json.dumps(self._info, indent=2)
```
vkarak, line 176: Why? Perhaps you need `__rfm_json_encode__`.

reframe/core/systems.py
```diff
@@ -194,85 +354,35 @@ def environment(self, name):
         return None
 
     @property
-    def arch(self):
-        '''The microarchitecture of this partition.
+    def processor(self):
+        '''The processor object of the current partition.
 
         .. versionadded:: 3.5
```
vkarak, line 360: Use semantic versioning conformat strings everywhere:

```suggestion
        .. versionadded:: 3.5.0
```

reframe/schemas/config.json
```diff
@@ -139,6 +139,58 @@
...
+        },
+        "processor_info": {
+            "type": "object",
+            "properties": {
+                "arch": {"type": "string"},
+                "num_cpus": {"type": "number"},
+                "num_cpus_per_core": {"type": "number"},
+                "num_cpus_per_socket": {"type": "number"},
+                "num_sockets": {"type": "number"},
+                "topology": {
```
vkarak, line 151: Since this goes too deep, you can define it separately.

vkarak (changes requested):

reframe/core/systems.py
ekouts: It was just for testing, I removed.

reframe/core/systems.py
```diff
@@ -13,6 +13,187 @@
...
+
+    @property
+    def num_cores_per_numa_node(self):
+        if self.num_numa_nodes and self.num_cores:
+            return self.num_cores // self.num_numa_nodes
+        else:
+            return None
+
+
+class Device(jsonext.JSONSerializable):
```
jjotero, line 135: How about changing this class name to `DeviceType` instead? If I understand this correctly, this class represents a type of device that is present in a partition, and this partition might have several devices of this type. So by calling this class `Device` I got the impression that this was referring instead to a single gpu and I found a bit confusing that the class `Device` had the attribute `_num_devices`. If we change this name to `DeviceType`, I'd suggest to also change the class above referring to the processor as `ProcessorType`, just for consistency purposes.

jjotero (commented):
Just a very minor suggestion to rename the `Device` class. Other than that, lgtm.

docs/config_reference.rst
```diff
@@ -316,39 +316,140 @@ System Partition Configuration
...
+
+   Number of logical CPUs per socket.
+
+
+.. js:attribute:: .systems[].partitions[].processor.num_sockets
+
+   :required: No
+   :default: ``None``
+
+   Number sockets on this partition.
```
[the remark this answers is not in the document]
ekouts: I will completely remove "on this partition" since I already removed it from other attributes and I have also forgotten an "of" 😅

reframe/core/systems.py
ekouts: Good point, I will change both `DeviceType` and `ProcessorType`.

jjotero (approved):
lgtm

vkarak (approved):

PR #1775, opened 2021-02-15
[feat] Improve the container platform API exposed to tests

### LINKED ISSUE #1736: Container execution w/ SARUS

teojgo:
* Respect the entrypoint and command defined inside the container image.
* Deprecate `workdir` and `commands` fields.
* Introduce the `command` field.
* Allow to skip pulling of the container image from a container registry.
* Expand container platform tutorial example.

Fixes #1736 

- [x] Need to change also the documentation
- [x] Change the tutorial example
- [x] Add a tutorial example for making files available outside of the container
- [x] Deprecate the `workdir` and `command` fields


teojgo:
@ChristianKniep please give your feedback

reframe/core/containers.py
```diff
@@ -19,11 +19,31 @@ class ContainerPlatform(abc.ABC):
     #: :default: :class:`None`
     image = fields.TypedField(str, type(None))
 
-    #: The commands to be executed within the container.
+    #: The command to be executed within the container.
```
kniec, line 22: , overwriting the CMD setting of the container image.

reframe/core/containers.py
```diff
@@ -19,11 +19,31 @@ class ContainerPlatform(abc.ABC):
...
+    #: If no command is given, then the default command of the corresponding
+    #: container image is going to be executed.
+    #:
+    #: ..versionchanged:: 3.4.2
+    #:   Changed the attribute name from `commands` to `command` and its type
+    #:   to a string.
+    #:
+    #: :type: :class:`str` or :class:`None`
+    #: :default: :class:`None`
+    command = fields.TypedField(str, type(None))
```
kniec, line 33: How about an `entrypoint` type to overwrite the `ENTRYPOINT` of the image?
Maybe by popular demand.

reframe/core/containers.py
```diff
@@ -19,11 +19,31 @@ class ContainerPlatform(abc.ABC):
...
+    #: If an empty string is given as the pull command, then the default
+    #: pull command of the corresponding container platform is going to be
+    #: used to pull the image. If set to :class:`None`, then no pull action
+    #: is going to be performed by the container platform.
+    #:
+    #: ..versionadded:: 3.4.2
+    #:
+    #: :type: :class:`str` or :class:`None`
+    #: :default: ``''``
+    pull_command = fields.TypedField(str, type(None))
```
kniec, line 46: Hmm, I'd say this is not relevant in the future - just as a workaround for my edge-case. 
At the end, you might generalise the pull by using skopeo from RH. They are able to transform in almost all formats.

reframe/core/containers.py
```diff
@@ -88,25 +110,28 @@ def validate(self):
...
+        run_opts = [f'-v "{mp[0]}":"{mp[1]}"' for mp in self.mount_points]
+        run_opts += ['-v "${PWD}":"/rfm_workdir"']
         run_opts += self.options
-        run_cmd = 'docker run --rm %s %s bash -c ' % (' '.join(run_opts),
-                                                      self.image)
-        return run_cmd + "'" + '; '.join(
-            ['cd ' + self.workdir] + self.commands) + "'"
+        workdir_opt = f'--workdir="{self.workdir}" ' if self.workdir else ''
+
+        return (f'docker run --rm {workdir_opt}{" ".join(run_opts)} '
```
kniec, line 133: Thinking about it; in the future you might want to change that ti use the docker-py SDK to create containers within python. Even though that would break with all the other execution environments.

kniec (approved):
Sweat, just some minor comments. I reckon that the `pull_command` is a bandaid until a more sustainable way forward is found. Might be that skopeo is the ultimate swiss army knife for downloading and transforming image. @vrothberg
Valentin might also have an opinion about adding podman here... :)

reframe/core/containers.py
teojgo: In general you can put all the options you might need in the `options` field of the container platform. I am still thinking if we should have a separate `workdir` field.

reframe/core/containers.py
teojgo: Maybe then have a boolean field indicating whether the image should be pulled or not. In any case you can perform custom pulling/loading using the `prerun_cmds`

vrothberg:
> Sweat, just some minor comments. I reckon that the `pull_command` is a bandaid until a more sustainable way forward is found. Might be that skopeo is the ultimate swiss army knife for downloading and transforming image. @vrothberg
> Valentin might also have an opinion about adding podman here... :)

Thanks for pulling me in.  [Skopeo](https://github.com/containers/skopeo) is super powerful and light weight. Podman, on the other hand, has Python bindings: https://github.com/containers/podman-py

unittests/test_containers.py
```diff
@@ -6,15 +6,15 @@
...
 @pytest.fixture(params=[
-    'Docker',
-    'Sarus', 'Sarus+mpi', 'Sarus+localimage',
-    'Shifter', 'Shifter+mpi', 'Shifter+localimage',
-    'Singularity', 'Singularity+cuda'
-])
+    'Docker', 'Docker+nocommand', 'Docker+nopull', 'Sarus', 'Sarus+nocommand',
+    'Sarus+nopull', 'Sarus+mpi', 'Sarus+load', 'Shifter', 'Shifter+nocommand',
+    'Shifter+mpi', 'Shifter+nopull', 'Shifter+load', 'Singularity',
+    'Singularity+nocommand', 'Singularity+cuda'])
```
vkarak, line 17: Put the closing `])` in the next line.

unittests/test_containers.py
```diff
@@ -157,8 +186,67 @@ def test_prepare_command(container_platform, expected_cmd_prepare):
...
+
+
+@pytest.fixture(params=[
+    'Docker', 'Singularity', 'Sarus', 'Shifter'])
+def container_variant_deprecated(request):
+    return request.param
+
+
+@pytest.fixture
+def platform_deprecated(container_variant_deprecated):
```
vkarak, line 201: I don't understand what do you want to achieve with this here?

docs/tutorial_advanced.rst
```diff
@@ -693,26 +693,40 @@ This attribute accepts a string that corresponds to the name of the container pl
 In this case, the test will be using `Singularity <https://sylabs.io>`__ as a container platform.
 If such a platform is not `configured <config_reference.html#container-platform-configuration>`__ for the current system, the test will fail.
 
-As soon as the container platform to be used is defined, you need to specify the container image to use and the commands to run inside the container by setting the :attr:`image <reframe.core.containers.ContainerPlatform.image>` and the :attr:`commands <reframe.core.containers.ContainerPlatform.commands>` container platform attributes.
-These two attributes are mandatory for container-based checks.
+As soon as the container platform to be used is defined, you need to specify the container image to use by setting the :attr:`image <reframe.core.containers.ContainerPlatform.image>`.
+The default command that the container runs can be overwritten by setting the :attr:`command <reframe.core.containers.ContainerPlatform.command>` of the container platform.
```
vkarak, line 697:
```suggestion
The default command that the container runs can be overwritten by setting the :attr:`command <reframe.core.containers.ContainerPlatform.command>` attribute of the container platform.
```

docs/tutorial_advanced.rst
```diff
@@ -693,26 +693,40 @@ This attribute accepts a string that corresponds to the name of the container pl
...
-Users may also change the default mount point of the stage directory by using :attr:`workdir <reframe.core.pipeline.RegressionTest.container_platform.workdir>` attribute:
 Besides the stage directory, additional mount points can be specified through the :attr:`mount_points <reframe.core.pipeline.RegressionTest.container_platform.mount_points>` attribute:
 
 .. code-block:: python
 
     self.container_platform.mount_points = [('/path/to/host/dir1', '/path/to/container/mount_point1'),
                                             ('/path/to/host/dir2', '/path/to/container/mount_point2')]
 
+
+The container filesystem is ephemeral, therefore, ReFrame mounts the stage directory under ``/rfm_workdir`` inside the container where the user can copy directories/files as needed.
```
vkarak, line 718:
```suggestion
The container filesystem is ephemeral, therefore, ReFrame mounts the stage directory under ``/rfm_workdir`` inside the container where the user can copy artifacts as needed.
```

Same for the following occurences.

docs/tutorial_advanced.rst
```diff
@@ -693,26 +693,40 @@ This attribute accepts a string that corresponds to the name of the container pl
...
+The aforementioned directories/files will then be available inside the stage directory after the container execution finishes.
+This is very useful if the above directories/files are going to be used for the sanity/performance checks.
+If this is the case, the user should overwrite the default command executed by the container, using the :attr:`command <reframe.core.containers.ContainerPlatform.command>` to include the appropriate copy commands.
```
vkarak, line 721:
```suggestion
If the copy does not happen inside the container, the user should overwrite the default command executed by the container, using the :attr:`command <reframe.core.containers.ContainerPlatform.command>` to include the appropriate copy commands.
```

docs/tutorial_advanced.rst
```diff
@@ -693,26 +693,40 @@ This attribute accepts a string that corresponds to the name of the container pl
...
+In this test, we have changed the directory that the container is going to run the commands using the `--pwd=/rfm_workdir` through the :attr:`options <reframe.core.containers.ContainerPlatform.options>`.
```
vkarak, line 722: I would prefer if you add the `cp` as you just proposing above. It can be confusing.

vkarak (changes requested):

reframe/core/containers.py
```diff
@@ -13,22 +14,58 @@
 class ContainerPlatform(abc.ABC):
     '''The abstract base class of any container platform.'''
 
+    #: The default mount location of the test case stage directory inside the
+    #: container
+    RFM_STAGEDIR = '/rfm_workdir'
```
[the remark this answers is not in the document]
teojgo: No, I just wanted to make sure that I don't hardcode a value when we add this mount via the pipeline (since you need to know the stage dir)

unittests/test_containers.py
teojgo: I wanted to separate the tests related to deprecated fields so that it is easier to remove afterwards.

docs/tutorial_advanced.rst
teojgo: Do you mean to add also the command I use to copy the artifacts?

reframe/core/containers.py
```diff
@@ -46,16 +83,27 @@ class ContainerPlatform(abc.ABC):
...
+        fields.TypedField(str, type(None)),
+        'The `workdir` field is deprecated, please use the `options` field to '
+        'set the container working directory', 1
+    )
 
     def __init__(self):
         self.image = None
-        self.commands = []
+        self.command = None
+        with warnings.catch_warnings(record=True):
```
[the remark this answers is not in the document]
teojgo: This I did in oder not to throw a warning everytime, since this is called automatically.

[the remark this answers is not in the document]
vkarak: The only thing that `record=True` does is that it returns a list of warnings that were caught, which here you don't use.

unittests/test_containers.py
vkarak: Couldn't you simply add it to the normal test and wrap the deprecated operations with the following?

```python
with pytest.warns(ReframeDeprecationWarning):
    deprecated_op()
```

docs/tutorial_advanced.rst
vkarak: I am suggesting to not use the `--pwd`. Let the container execute on its ephemeral filesystem and then copy the  `releases.txt` to the `/rfm_workdir`, as you are exactly suggesting right above.

reframe/core/containers.py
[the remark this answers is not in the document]
teojgo: If I don't use `record=True`, the warnings are thrown otherwise they are supressed.

docs/tutorial_advanced.rst
```diff
@@ -715,13 +715,19 @@ Besides the stage directory, additional mount points can be specified through th
...
-This is very useful if the above directories/files are going to be used for the sanity/performance checks.
-If this is the case, the user should overwrite the default command executed by the container, using the :attr:`command <reframe.core.containers.ContainerPlatform.command>` to include the appropriate copy commands.
-In this test, we have changed the directory that the container is going to run the commands using the `--pwd=/rfm_workdir` through the :attr:`options <reframe.core.containers.ContainerPlatform.options>`.
-Since this is the directory that ReFrame mounts the stage directory inside the container, the files created there are available after the container execution finishes.
-Therefore, the ``release.txt`` where the output of the ``cat /etc/os-release`` is redirected, can be used in the subsequent sanity function:
+The container filesystem is ephemeral, therefore, ReFrame mounts the stage directory under ``/rfm_workdir`` inside the container where the user can copy artifacts as needed.
+The aforementioned artifacts will then be available inside the stage directory after the container execution finishes.
```
vkarak, line 719:
```suggestion
These artifacts will therefore be available inside the stage directory after the container execution finishes.
```

docs/tutorial_advanced.rst
```diff
@@ -715,13 +715,19 @@ Besides the stage directory, additional mount points can be specified through th
...
+This is very useful if the above artifacts are going to be used for the sanity/performance checks.
```
vkarak, line 720:
```suggestion
This is very useful if the artifacts are needed for the sanity or the performance checks.
```

docs/tutorial_advanced.rst
```diff
@@ -715,13 +715,19 @@ Besides the stage directory, additional mount points can be specified through th
...
+If the copy is not performed by the default container commands, the user should overwrite the default command executed by the container, using the :attr:`command <reframe.core.containers.ContainerPlatform.command>` to include the appropriate copy commands.
```
vkarak, line 721:
```suggestion
If the copy is not performed by the default container command, the user can override this command by setting the :attr:`command <reframe.core.containers.ContainerPlatform.command>` attribute such as to include the appropriate copy commands.
```

docs/tutorial_advanced.rst
```diff
@@ -715,13 +715,19 @@ Besides the stage directory, additional mount points can be specified through th
...
+In the current test, the output of the ``cat /etc/os-release`` is made available outside the container since we have used the command:
+
+.. code-block:: bash
+
+    bash -c 'cat /etc/os-release | tee /rfm_workdir/release.txt'
```
vkarak, line 726: You should mention perhaps that you also print in stdout to demonstrate also how the stdout of the container can be used in sanity and performance checking as usual.

reframe/core/containers.py
```diff
@@ -29,25 +28,27 @@ class ContainerPlatform(abc.ABC):
     #: If no command is given, then the default command of the corresponding
     #: container image is going to be executed.
     #:
-    #: ..versionchanged:: 3.5
+    #: ..versionadded:: 3.5.0
```
vkarak, line 31: You should leave a space after `..`. Otherwise it is not shown correctly. Make sure to fix this everywhere.

reframe/core/containers.py
```diff
@@ -77,29 +78,32 @@ class ContainerPlatform(abc.ABC):
...
     #: The working directory of ReFrame inside the container.
     #:
     #: This is the directory where the test's stage directory is mounted inside
     #: the container. This directory is always mounted regardless if
     #: :attr:`mount_points` is set or not.
     #:
-    #: ..versionchanged:: 3.5
-    #:   The `commands` field is now deprecated.
+    #: ..deprecated:: 3.5
+    #: Please use the `options` field to set the working directory.
```
vkarak, line 89:
```suggestion
    #: .. deprecated:: 3.5
    #:    Please use the `options` field to set the working directory.
```

reframe/core/containers.py
```diff
@@ -77,29 +78,32 @@ class ContainerPlatform(abc.ABC):
...
     )
 
     def __init__(self):
         self.image = None
         self.command = None
-        with warnings.catch_warnings(record=True):
-            self.commands = []
-            self.workdir = self.RFM_STAGEDIR
+
+        # NOTE: Here we set the target fields directly to avoid the warnings
```
vkarak, line 104:
```suggestion
        # NOTE: Here we set the target fields directly to avoid the deprecation warnings
```

reframe/core/containers.py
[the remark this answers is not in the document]
vkarak: Make it private then or even move it at the module level. Also better call it `_STAGEDIR_MOUNT`.

tutorials/advanced/containers/container_test.py
```diff
@@ -16,10 +16,8 @@ def __init__(self):
         self.container_platform = 'Singularity'
         self.container_platform.image = 'docker://ubuntu:18.04'
         self.container_platform.command = (
-            "bash -c 'pwd; cat /etc/os-release > release.txt'"
+            "bash -c 'cat /etc/os-release | tee /rfm_workdir/release.txt'"
+        )
+        self.sanity_patterns = sn.assert_found(
+            r'18.04.\d+ LTS \(Bionic Beaver\)', 'release.txt'
         )
```
vkarak, line 23: Check that the same thing is also in the stdout.

vkarak (changes requested):

teojgo:
@kniec the PR is more or less ready. I have made various changed taking into account your comments. I have also expanded the tutorial for containers to include Sarus.

kniec:
Thanks for the effort - I was busy this week, but I'll carve out some time in the next couple of days to go through it.

docs/tutorial_advanced.rst
```diff
@@ -671,13 +671,14 @@ ReFrame can be used also to test applications that run inside a container.
...
 .. literalinclude:: ../tutorials/config/settings.py
-   :lines: 38-58
-   :emphasize-lines: 15-20
+   :lines: 38-62
+   :emphasize-lines: 15-24
 
 For each partition, users can define a list of container platforms supported using the :js:attr:`container_platforms` `configuration parameter <config_reference.html#.systems[].partitions[].container_platforms>`__.
-In this case, we define the `Singularity <https://sylabs.io>`__ platform, for which we set the :js:attr:`modules` parameter in order to instruct ReFrame to load the ``singularity`` module, whenever it needs to run with this container platform.
+In this case, we define the `Sarus <https://github.com/eth-cscs/sarus>`__ platform for which we set the :js:attr:`modules` parameter in order to instruct ReFrame to load the ``sarus`` module, whenever it needs to run with this container platform.
+Furthermore, we define the `Singularity <https://sylabs.io>`__ platform, for which the ``singularity`` module needs to be loaded.
```
vkarak, line 679:
```suggestion
Similarly, we add an entry for the `Singularity <https://sylabs.io>`__ platform.
```

docs/tutorial_advanced.rst
```diff
@@ -671,13 +671,14 @@ ReFrame can be used also to test applications that run inside a container.
...
 
-The following test will use a Singularity container to run:
+The following parameterized test, will create two test cases, one for each of the supported contaiter platforms:
```
vkarak, line 681:
```suggestion
The following parameterized test, will create two tests, one for each of the supported container platforms:
```

docs/tutorial_advanced.rst
```diff
@@ -686,25 +687,38 @@ The following test will use a Singularity container to run:
...
-   :emphasize-lines: 11-15
+   :emphasize-lines: 11-16
 
 A container-based test can be written as :class:`RunOnlyRegressionTest <reframe.core.pipeline.RunOnlyRegressionTest>` that sets the :attr:`container_platform <reframe.core.pipeline.RegressionTest.container_platform>` attribute.
 This attribute accepts a string that corresponds to the name of the container platform that will be used to run the container for this test.
-In this case, the test will be using `Singularity <https://sylabs.io>`__ as a container platform.
 If such a platform is not `configured <config_reference.html#container-platform-configuration>`__ for the current system, the test will fail.
 
 As soon as the container platform to be used is defined, you need to specify the container image to use by setting the :attr:`image <reframe.core.containers.ContainerPlatform.image>`.
+In the ``Singularity`` test case, we add the ``docker://`` prefix to the image name, in order to instruct ``Singularity`` to pull the image from `DockerHub <https://hub.docker.com/>`__.
```
vkarak, line 697:
```suggestion
In the ``Singularity`` test variant, we add the ``docker://`` prefix to the image name, in order to instruct ``Singularity`` to pull the image from `DockerHub <https://hub.docker.com/>`__.
```

Test cases have a specific definition in ReFrame, so avoid the term here. A parameterized test generates *tests*, not *test cases*.

reframe/core/pipeline.py
```diff
@@ -1290,7 +1290,7 @@ def run(self):
 
             self.container_platform.validate()
             self.container_platform.mount_points += [
-                (self._stagedir, self.container_platform.RFM_STAGEDIR)
+                (self._stagedir, _STAGEDIR_MOUNT)
```
vkarak, line 1293: Can't this happen inside the container platform? For example, just before emitting the commands? Or adding it without touching the user mount points at all?

vkarak (changes requested):

reframe/core/pipeline.py
teojgo: We need to know the stage directory of the test to do so.

vkarak: I think the best way is to change the interface of the `emit_prepare_commands()` and the `launch_command()` to accept a path, where you will pass the stagedir.

vkarak: I fixed that.

vkarak (approved):

PR #1780, opened 2021-02-16
[test] Enable DGEMM regression test on Eiger

ajocksch:
PrgEnv-cray is not considered in this check. It looks like a performance bug in the Cray compiler which has not been fixed.


jjotero (approved):
lgtm

cscs-checks/microbenchmarks/cpu/dgemm/dgemm.py
```diff
@@ -34,6 +37,7 @@ def __init__(self):
             'daint:mc': (860.0, -0.15, None, 'Gflop/s'),
             'dom:gpu': (300.0, -0.15, None, 'Gflop/s'),
             'dom:mc': (860.0, -0.15, None, 'Gflop/s'),
+            'eiger:mc': (650.0, -0.15, None, 'Gflop/s'),
```
vkarak, line 40: How sure are we that this is a decent number? Slower than Broadwell with more threads? Are we not affected by affinity issues?

vkarak (changes requested):

cscs-checks/microbenchmarks/cpu/dgemm/dgemm.py
ajocksch: the thread affinity looks good

ekouts (approved):

cscs-checks/microbenchmarks/cpu/dgemm/dgemm.py
vkarak: I think that this is seriously underperforming. AMD Rome can deliver up to 16 DP flops per cycle (2x FMA units, AVX2). This leads to a 2.304 Tflop/s nominal peak per socket (no turbo boost). Conversely, on Broadwell, the reference is much closer to the nominal, which is 604.8 Gflop/s per socket. I've not a problem accepting it as is, but it doesn't seem "all good." It could be that the dgemm library underperforms.

vkarak: @ajocksch Put a `FIXME` referring to the Jira ticket number, fix the conflicts and then we can merge it.

ajocksch: done

vkarak (approved):
lgtm

PR #1788, opened 2021-02-16
[feat] Allow logging arbitrary job attributes

### LINKED ISSUE #1729: Allow to log easily the job's nodelist in performance log
Currently, this can be done using a custom test field:

```python
    ...
    @rfm.run_after('run')
    def set_nodelist(self):
        self.mynodelist = self.job.nodelist
```

And then use `%(check_mynodelist)s` as a format string.

It would be much easier if we supported directly `%(check_nodelist)s` or `%(check_job_nodelist)s` or even better support any job attribute to be logged with `%(check_job_ATTR)s`.

vkarak:
This PR allows users to log arbitrary attributes of the run job associated with the currently executing test by using the `%(check_job_ATTR)s` log format specifier. For example, users can log the node list that the job has run on by specifying `%(check_job_nodelist)s`.

Fixes #1729.


ekouts (approved):
lgtm

teojgo (approved):

PR #1800, opened 2021-02-19
[test] Adapt VASP check for Eiger

lucamar:
I provide the check on Eiger: the pipeline hooks that set the task distribution and cpu binding should improve the performance bit, also on Dom and Piz Daint.


ekouts (approved):

vkarak (approved):
lgtm

PR #1803, opened 2021-02-23
[feat] Add a new `prepare_cmds` partition configuration parameter which allows to emit commands before loading any environment

### LINKED ISSUE #1734: Option to include commands in the job script before the modules are loaded
For jobs scripts to be accepted on our system we have to include:
```
module purge
module load bluebear
```
and this has to appear before any `module load foo` commands.

Can we have an option to add a custom command (in my case the `module purge` part) early in the job script? (I can handle the `module load bluebear` by adding `bluebear` to the start of the list of modules to load.)

(At the moment I work around the issue by adding the `module purge` into the `prerun_cmds` and then generate my own `module load` list.)

rsarm:
Closes #1734


teojgo (changes requested):
We wil also need a unittest for this.

reframe/core/pipeline.py
```diff
@@ -348,6 +348,19 @@ def pipeline_hooks(cls):
     container_platform = variable(type(None),
                                   field=ContainerPlatformField, value=None)
 
+    #: .. versionadded:: 3.5
```
ekouts, line 351:
```suggestion
    #: .. versionadded:: 3.5.0
```
I think the version should be `3.5.0`, but I am not sure.

vkarak (changes requested):
I don't think this PR is correct in its principle. These commands are partition specific and not test-specific. I don't think we should anything to the test API. Instead, these commands must be part of the configuration of the partition. This is what I understand from the issue: every script submitted to that partition needs to have a `module purge`.

reframe/core/pipeline.py
```diff
@@ -1220,6 +1220,7 @@ def compile(self):
             try:
                 self._build_job.prepare(
                     build_commands, environs,
+                    self._current_partition.preload_cmds,
```
vkarak, line 1223: I would rename the commands everywhere to

```suggestion
                    self._current_partition.prepare_cmds,
```

reframe/core/schedulers/__init__.py
```diff
@@ -330,7 +330,7 @@ def nodelist(self):
     def submit_time(self):
         return self._submit_time
 
-    def prepare(self, commands, environs=None, **gen_opts):
+    def prepare(self, commands, environs=None, preload_cmds=[], **gen_opts):
```
vkarak, line 333: We need to add a unit test for that or adapt the `test_prepare` unit test.

vkarak (changes requested):
The implementation looks good. I have a suggestion about the name of the option and we still need the following:

1. A unit test or adapt `test_prepare`
2. Documentation (both in the configuration reference and [here](https://reframe-hpc.readthedocs.io/en/stable/tutorial_advanced.html#customizing-the-test-job-script) (we need to update the script template).

teojgo:
@rsarm, @vkarak do you think that this is the case for both build/run scripts? I have the feeling that this is needed only for the compute nodes, at least on #1734

vkarak:
@teojgo I think you are correct here. We should use it also for the build scripts, since for the moment we say that we submit to the same partition, so users would expect to see that in the build scripts, too, since they submit in the same partition.

teojgo:
> @teojgo I think you are correct here. We should use it also for the build scripts, since for the moment we say that we submit to the same partition, so users would expect to see that in the build scripts, too, since they submit in the same partition.

Yes for the moment it is used for both.

vkarak (approved):
lgtm

teojgo (approved):

ekouts (approved):

vkarak:
@branfosj With this feature you will now be able to define your partition as follows:

```python
{
    'name': 'mypart',
    'scheduler': 'slurm',
    'launcher': 'srun',
    'prepare_cmds': ['module purge'],
    'modules': ['bluebear']
}
```

The `prepare_cmds` will always be emitted before any commands that load the test environment. Let us know if this is sufficient.

branfosj:
Thanks

PR #1761, opened 2021-02-10
[feat] Comply with the semantic versioning specification

vkarak:
This PR also adds the following:

- Expand deprecation message function to issue deprecation warnings in future releases.
- Old style of version specification is set to deprecate in 3.5.0

TODO:

- [x] Update documentation of the `@required_version` decorator.

~~NOTE: The CI will fail due to an irrelevant reason. A separate PR will fix that (it's Gitlab CI lint failure)~~

PR #1804, opened 2021-02-23
[feat] Allow deprecation of fields in a future version and issue deprecation warnings also in prereleases

vkarak:
Also:

- Make unit tests for deprecation warnings more robust
-  Ignore prereleases when filtering deprecation warnings. For example, if we are on 3.5.0-dev.0 and a feature is set to deprecate in 3.5.0, then we issue the deprecation warning also for the prerelease.
- Fix deprecation warnings introduced by #1761.

This is a follow up of #1761.

### DIFF
--- a/unittests/test_warnings.py
+++ b/unittests/test_warnings.py
@@ -23,7 +23,9 @@ def test_deprecation_warning():
 
 def test_deprecation_warning_from_version():
     next_version = semver.VersionInfo.parse(reframe.VERSION).bump_minor()
-    warn.user_deprecation_warning('deprecated', str(next_version))
+    with warnings.catch_warnings(record=True) as w:
+        warn.user_deprecation_warning('deprecated', str(next_version))
+        assert len(w) == 0
 
 
 def test_deprecation_warning_formatting(with_colors):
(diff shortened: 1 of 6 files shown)


ekouts (approved):

unittests/test_warnings.py
```diff
@@ -23,7 +23,9 @@ def test_deprecation_warning():
 
 def test_deprecation_warning_from_version():
```
teojgo, line 24: Is it possible to test a case where the a deprecation warning is thrown. Like test for a version previous to this one.

teojgo (changes requested):

unittests/test_warnings.py
vkarak: The default `from_version` is `0.0.0`, so what you ask here is implicitly tested, isn't it?

teojgo: I mean a case where the deprecation warning is actually thrown.

vkarak: I think the `test_deprecation_warning` does already this.

teojgo (approved):

PR #1806, opened 2021-02-23
[feat] Support parameter and variable access directly from the class body

jjotero:
This PR implements the attribute access to variables and parameters directly from the class body. Prior to this PR, an attempt to access either variables or parameters from the class body resulted in a `NameError` since none of these built-ins are stored in the public namespace from the class. This PR not only implements the access to these items, but also provides a more robust error handling in case of erroneous user input:

- Accessing a required (undefined) variable in the class body will raise a `ValueError`:
```python
class Foo(rfm.RegressionMixin):
    v = variable(str)
    x = f'{v}' # This will raise an error because `v` has no value
```
- Accessing a defined variable will return its default value:
```python
class Foo(rfm.RegressionMixin):
    v = variable(int, value=4)
    x = v + 1 # this sets x to 5
```
- Accessing a parameter directly from the class body is disallowed.
```python
class Foo(rfm.RegressionMixin):
    p = parameter([1, 2, 3])
    x = f'{p}' # This raises an error. Here `P` is a TestParam object, not a test parameter.
```

### DIFF
--- a/reframe/core/meta.py
+++ b/reframe/core/meta.py
@@ -28,6 +28,38 @@ class RegressionTestMeta(type):
             else:
                 super().__setitem__(key, value)
 
+        def __getitem__(self, key):
+            '''Expose and control access to the local namespaces.
+
+            Variables may only be retrieved if their value has been previously
+            set. Accessing a parameter in the class body is disallowed (the
+            actual test parameter is set during the class instantiation).
+            '''
+            try:
+                return super().__getitem__(key)
+            except Exception as err:
+                try:
+                    # Handle variable access
+                    var = self['_rfm_local_var_space'][key]
+                    if var.is_defined():
+                        return var.default_value
+                    else:
+                        raise ValueError(
+                            f'variable {key!r} does not have an assigned value'
+                        )
+
+                except KeyError:
+                    # Handle parameter access
+                    if key in self['_rfm_local_param_space']:
+                        raise ValueError(
+                            'accessing a test parameter from the class '
+                            'body is disallowed'
+                        )
+                    else:
+                        # If 'key' is neither a variable nor a parameter,
+                        # raise the exception from the base __getitem__.
+                        raise err
+
     @classmethod
     def __prepare__(metacls, name, bases, **kwargs):
         namespace = super().__prepare__(name, bases, **kwargs)
--- a/reframe/core/meta.py
+++ b/reframe/core/meta.py
@@ -37,7 +37,7 @@ class RegressionTestMeta(type):
             '''
             try:
                 return super().__getitem__(key)
-            except Exception as err:
+            except KeyError as err:
                 try:
                     # Handle variable access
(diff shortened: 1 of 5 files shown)


reframe/core/pipeline.py
```diff
@@ -138,9 +138,22 @@ class RegressionMixin(metaclass=RegressionTestMeta):
...
+                not self._rfm_var_space.vars[name].is_defined()):
+                raise AttributeError(
+                    f'required variable {name!r} has not been set'
+                ) from None
+            else:
+                super().__getattr__(name)
 
 
-class RegressionTest(jsonext.JSONSerializable, metaclass=RegressionTestMeta):
+class RegressionTest(jsonext.JSONSerializable, RegressionMixin):
```
vkarak, line 156: I would put `RegressionMixin` first in the row here.

reframe/core/meta.py
```diff
@@ -28,6 +28,38 @@ def __setitem__(self, key, value):
...
+                return super().__getitem__(key)
+            except Exception as err:
+                try:
+                    # Handle variable access
+                    var = self['_rfm_local_var_space'][key]
+                    if var.is_defined():
+                        return var.default_value
+                    else:
+                        raise ValueError(
+                            f'variable {key!r} does not have an assigned value'
```
vkarak, line 48:
```suggestion
                            f'variable {key!r} is not assigned a value'
```

reframe/core/meta.py
```diff
@@ -28,6 +28,38 @@ def __setitem__(self, key, value):
...
+        def __getitem__(self, key):
+            '''Expose and control access to the local namespaces.
+
+            Variables may only be retrieved if their value has been previously
+            set. Accessing a parameter in the class body is disallowed (the
+            actual test parameter is set during the class instantiation).
+            '''
+            try:
+                return super().__getitem__(key)
+            except Exception as err:
```
vkarak, line 40: Why not just a `KeyError`?

reframe/core/meta.py
```diff
@@ -28,6 +28,38 @@ def __setitem__(self, key, value):
...
+                    # Handle parameter access
+                    if key in self['_rfm_local_param_space']:
+                        raise ValueError(
+                            'accessing a test parameter from the class '
+                            'body is disallowed'
+                        )
+                    else:
+                        # If 'key' is neither a variable nor a parameter,
+                        # raise the exception from the base __getitem__.
+                        raise err
```
vkarak, line 61:
```suggestion
                        raise err from None
```

vkarak (changes requested):

reframe/core/meta.py
jjotero: Yeah, makes sense. Changed.

vkarak (approved):
lgtm

ekouts (approved):

teojgo (approved):

PR #1814, opened 2021-02-24
[bugfix] Fix status printout in case tests are retried

### LINKED ISSUE #1812: ReFrame erroneously reports `FAILURE` when tests pass in retries
This is a bug introduced by #1801.

vkarak:
Fixes #1812.


ekouts (commented):
The fix looks good to me but some lines below there is another bug from my maxfail PR:
```python
            num_completed = len(self._stats.completed())
...
            self._printer.status(
                status,
                status,
                f'Ran {num_completed}/{len(testcases)}'
                f' test case(s) from {num_checks} check(s) '
                f'({num_failures} failure(s))',
                just='center'
            )
```
In this line `f'Ran {num_completed}/{len(testcases)}'`, `num_completed` will be only the ones that were completed in the last retry. Not sure how we to fix this simply, since completed are the ones that failed or succeeded and we cannot just sum the completed from all runs.

ekouts (approved):

vkarak:
@jenkins-cscs retry daint dom

PR #1815, opened 2021-02-24
[test] Update AllocSpeedTest and enable it for Eiger

toxa81:
Looks like gcc/10 has more compile-time optimizations. This is fixed now.


vkarak:
ok to test

toxa81:
@ekouts Let me know how it runs on Eiger. I tested it myself and it looks OK.

ekouts (approved):
Thanks @toxa81 , it looks good on eiger and the timings on dom went back to normal. :)

vkarak (approved):

teojgo:
@ekouts, maybe put larger margins for lower values so that we can accept even better values but still not accept negative values.

ekouts:
> @ekouts, maybe put larger margins for lower values so that we can accept even better values but still not accept negative values

@vkarak what do you think? Should we put larger margins? I was looking at older performance values and in general it looks stable.

teojgo:
> > @ekouts, maybe put larger margins for lower values so that we can accept even better values but still not accept negative values
> 
> @vkarak what do you think? Should we put larger margins? I was looking at older performance values and in general it looks stable.

I suggested since I saw a failure on dom at some point where the allocation speed was even faster.

vkarak:
@ekouts Perhaps it makes sense to make the threshold 20% on the fast side.

teojgo (approved):

PR #1816, opened 2021-02-24
[feat] Log reference values of all performance variables in a single line

vkarak:
This also introduces a new function in the `ScopedDict` that returns all the keys in a specific scope.  This is used by the logging mechanism to retrieve and log the correct references for the current system partition.

Now somebody could use `%(check_perf_patterns)s` and `%(check_reference)s` to log all performance variables and their references in a single line.

Addresses #1543.


reframe/utility/__init__.py
```diff
@@ -727,6 +727,22 @@ def update(self, other):
             for k, v in scope_dict.items():
                 self.data[scope][k] = v
 
+    def scope(self, scope):
```
teojgo, line 730: You can mention this on line 673 where you comment on trying to retrieve whole scopes.

teojgo (changes requested):

teojgo (approved):

PR #1819, opened 2021-02-25
[feat] Improve access to the namespaces of both parameter and variable built-ins

### LINKED ISSUE #1841: Make built-in variables behave like normal class attributes
The idea is to expand the functionality of the current built-in variables and make their behaviour more like regular python class attributes. So far, the variables can only be modified once per class body and no attributes can be appended to these variables directly in the class body.
```python
class Foo:
    pass

    v0 = variable(Foo, value=Foo())

    # Add an attribute to the variable is currently not supported
    v0.my_attribute = 'a random attribute'

    v1 = variable(str, value='a')
    ... 
    # Modifying the variable that was assigned before is currently not supported.
    v1 += 'b' 
```

jjotero:
Rather than customising the test attribute access using the `__getattribute__` method, this PR moves this customisation into the `__getattr__` method instead, which is called when the default `__getattribute__` raises an `AttributeError`.

This PR also makes the variable and parameter spaces behave as proper namespaces, which makes operations such as testing if a variable has been declared and so on more intuitive:
```python
my_var_name in self._rfm_var_space
```
instead of
```python
my_var_name in self._rfm_var_space.vars
```
However, adding new items into the variable or parameter spaces is disallowed:
```python
    pass

Foo.param_space['P0'] = (1,2,) # This raises an error
Foo._rfm_var_space['V0'] = 'a' # and this too
```

The corresponding unit tests have been added.

Closes #1841.


ekouts (approved):

jjotero:
Variables now behave like regular python class attributes and multiple modifications within the same class body are now allowed. For example:
```python
class Foo:
    pass

    v = variable(Foo, value=Foo())
    v.var_attribute = 'my attribute'

    # Override a regular attribute with a variable
    a = 4
    a = variable(int, value=40)
    print(a+1) # Prints 41

    # Modify an existing var
    a *= 10 # variable a is now 400

    # Initialize a variable as a function of another variable
    b = variable(float, value=a/3.141592)
    

    def __init__(self):
        print(self.v.var_attribute) # This prints 'my attribute'
```

- Parameters and variables may override regular class attributes
- A parameter may only be defined once per class body and it's value may not be altered after it was declared.
- A variable's value may be modified multiple times within the same class body.
- Overriding a variable with a parameter or the other way around results in undefined behaviour.

reframe/core/variables.py
```diff
@@ -211,6 +534,11 @@ def inject(self, obj, cls):
             if var.is_defined():
                 setattr(obj, name, var.default_value)
 
+                # If the variable value itself has attributes, inject them.
+                value = getattr(obj, name)
+                for attr, attr_value in var.attrs.items():
+                    setattr(value, attr, attr_value)
```
vkarak, line 540: I don't see a unit tests that exercises this or am I missing something?

unittests/test_parameters.py
```diff
@@ -235,3 +246,35 @@ def test_param_access():
...
+        class Bar(Foo):
+            p = 0
+
+
+def test_override_regular_attribute():
+    with pytest.raises(ValueError):
+        class Foo(rfm.RegressionTest):
+            p = 4
+            p = parameter([1, 2])
+            p = p+1
```
vkarak, line 280: Which of the two is the offending line here?

unittests/test_variables.py
```diff
@@ -100,10 +100,10 @@ class MyTest(rfm.RegressionTest):
...
 
 def test_double_action_on_variable():
-    with pytest.raises(ValueError):
-        class MyTest(rfm.RegressionTest):
-            v0 = variable(int, value=2)
-            v0 = 2
+    '''Modifying a variable in the class body is permitted.'''
+    class MyTest(rfm.RegressionTest):
+        v0 = variable(int, value=2)
+        v0 += 2
```
vkarak, line 106: I would put an assert later in the resulting object to check that `v0 == 4`.

unittests/test_variables.py
```diff
@@ -149,9 +149,12 @@ def __init__(self):
 
 
 def test_require_undeclared_variable(NoVarsTest):
-    with pytest.raises(ValueError):
-        class MyTest(NoVarsTest):
-            foo = required
+    '''Setting a regular variable as required is permitted.
+
+    The behaviour of such attribute is undefined.
```
vkarak, line 154: If it's undefined, what do we test? It could throw an exception...

vkarak (changes requested):

vkarak:
Also coverage is a bit unhappy with the fact that none of the special functions are tested. I would add a small test test exercising all operators.

vkarak (commented):
One problem that I see, which would help a lot in eliminating the need for a defining a constructor is that currently you can't use in an expression a built-in that was defined in a base class. Look at this?

```python
class Base(rfm.RegressionTest):
    foo = variable(str, value='hello')


@rfm.simple_test
class HelloTest2(Base):
    descr = 'C Hello World test ' + foo
    valid_systems = ['*']
    valid_prog_environs = ['*']
    sourcepath = 'hello.c'
    tags = {'foo', 'bar'}
    sanity_patterns = sn.assert_found(r'Hello, World\!', 'foo.txt')
    maintainers = ['VK']
```

This will complain about `foo` in the `HelloTest2` definition. I am wondering whether this ca be fixed or not?

jjotero (commented):
I've added a few more unit tests covering the variable operators and the other issues mentioned in the PR review.

unittests/test_parameters.py
```diff
@@ -273,8 +273,26 @@ class Bar(Foo):
 
 
 def test_override_regular_attribute():
+    class Foo(rfm.RegressionTest):
+        p = 4
+        p = parameter([1, 2])
+
+    assert Foo.p == (1, 2,)
```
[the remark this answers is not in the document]
jjotero: This would not work. The parameters are only stored as instances of `TestParam` in the local parameter space, not in the class' final parameter space. This is because the `TestParam` class just stores the values of the parameters and their inheritance behaviour. Since the parameter inheritance is resolved when building the final class' parameter space, all this extra info in `TestParam` is no longer needed and the parameters are just stored in the parameter space dict as tuples with all their values.

jjotero (commented):
> One problem that I see, which would help a lot in eliminating the need for a defining a constructor is that currently you can't use in an expression a built-in that was defined in a base class.

This is standard Python behaviour, isn't it? I have a few ideas to extend this behaviour. Let me do some "exploring" :)

vkarak:
> This is standard Python behaviour, isn't it? I have a few ideas to extend this behaviour. Let me do some "exploring" :)

Yes, looks like that. It seems that a `NameError` is thrown before even looking it up in the `MetaNamespace`. The only reason I want this is to be able to define `sanity_patterns` in the class body. If you give a filename and not `stdout`, it just works with the current machinery. Technically, also, if you need something from the test instance, like the `self.stdout`, you should use a pipeline hook, but sanity patterns with `stdout` are so common and it would be nice to be able to define them in the class body. I am experimenting with a solution just for this. I wouldn't go too far to avoid this `NameError` (there is enough magic in the metaclasses already), I'm fine with that, as soon as I can put the damn `sanity_patterns` inside the class body.

unittests/test_parameters.py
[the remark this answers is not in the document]
vkarak: Right, now I remember that when I was playing with my other PR about the naming scheme.

reframe/core/meta.py
```diff
@@ -76,6 +76,20 @@ def __getitem__(self, key):
...
+                        for b in self['_rfm_bases']:
+                            if (hasattr(b, '_rfm_var_space') and
+                                key in b._rfm_var_space):
+                                    v = b._rfm_var_space[key]
+                                    v.__set_name__(self, key)
+
+                                    # Store a deep-copy of the variable's
+                                    # value and return.
+                                    self._namespace[key] = v.default_value
+                                    return self._namespace[key]
```
vkarak, line 91: I think you should decrease the identation level here.

unittests/test_variables.py
```diff
@@ -227,3 +227,191 @@ class MyTest(rfm.RegressionTest):
...
+def test_var_add_operator():
+    class A(rfm.RegressionTest):
+        v = variable(int, value=3)
+        assert (v - 1) == 2
+        assert (1 - v) == -2
+        v -= 1
+        assert v == 2
+
+
+def test_var_mult_operator():
```
vkarak, line 306:
```suggestion
def test_var_mul_operator():
```

vkarak (changes requested):
Lgtm, I have only a couple of minor style comments.

reframe/core/meta.py
jjotero: Since this is the only place that needs to prematurely access the base classes, I've changed this a bit and filtered out the base classes that don't have a var space before they're added into this namespace. So now the `if` above is a bit simpler.

vkarak (approved):
lgtm

PR #1820, opened 2021-02-25
[doc] Add links to the latest presentations and add Twitter badge as well

vkarak:


teojgo:
Change also the line where the setup.py starts: https://github.com/eth-cscs/reframe/blob/96ce404efe7afabb486fd9e4b5f2f1b79886e241/setup.py#L11

teojgo (changes requested):
Only the setup.py needs to be updated otherwise it's fine.

setup.py
```diff
@@ -8,7 +8,7 @@
 from reframe import VERSION
 
 with open('README.md') as read_me:
-    long_description = ''.join(read_me.readlines()[16:])
+    long_description = ''.join(read_me.readlines()[18:])
```
teojgo, line 11:
```suggestion
    long_description = ''.join(read_me.readlines()[17:])
```

teojgo (changes requested):

setup.py
vkarak: Why 17? The section starts in line 18 and 17 is a blank line. Is there a reason to have the first line blank?

teojgo (approved):

PR #1824, opened 2021-02-26
[test] Remove `@required_version` decorator from tests

teojgo:


teojgo:
@jenkins-cscs retry all

vkarak:
@teojgo For this one, we need to make sure that the versions in the decorator are correct. I'm not sure that we maintained correctly these version numbers...

teojgo:
> @teojgo For this one, we need to make sure that the versions in the decorator are correct. I'm not sure that we maintained correctly these version numbers...

Ok, I can look at the history to see when each test was introduced.

vkarak (changes requested):
As discussed, I think it's better if we remove the annotations completely. We do not actually maintain them. They would make sense if we move the tests in a separate repo.

ekouts (approved):

cscs-checks/prgenv/mpi.py
```diff
@@ -39,7 +38,7 @@ def __init__(self, required_thread):
         self.valid_systems = ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc',
                               'eiger:mc']
         self.valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu', 'PrgEnv-pgi',
-                                    'PrgEnv-intel', 'PrgEnv-aocc']
```
vkarak, line 42: Why did you remove `aocc`?

vkarak (approved):

cscs-checks/prgenv/mpi.py
teojgo: There is an incompatibility with libsci

PR #1827, opened 2021-03-02
[test] Add configuration for Pilatus and VASP check

lucamar:
I propose a configuration of ReFrame on Pilatus with the Hierarchical module naming scheme, introducing the environments `cpeAMD`, `cpeCray`, `cpeGNU` and `cpeIntel` that correspond to the toolchain modules to be loaded before the apps. 
I also provide a modified check for VASP that works on Pilatus: we might use this configuration temporarily.

### DIFF
--- a/config/cscs.py
+++ b/config/cscs.py
@@ -660,10 +660,10 @@ site_configuration = {
             ]
         },
         {
-            'name': 'pilatus',
-            'descr': 'Alps Cray EX Supercomputer (TDS)',
+            'name': 'eiger',
+            'descr': 'Alps Cray EX Supercomputer',
             'hostnames': [
-                'pilatus'
+                'eiger'
             ],
             'modules_system': 'lmod',
             'resourcesdir': '/apps/common/UES/reframe/resources',
@@ -711,10 +711,10 @@ site_configuration = {
             ]
         },
         {
-            'name': 'eiger',
+            'name': 'pilatus',
             'descr': 'Alps Cray EX Supercomputer',
             'hostnames': [
-                'eiger'
+                'pilatus'
             ],
             'modules_system': 'lmod',
             'resourcesdir': '/apps/common/UES/reframe/resources',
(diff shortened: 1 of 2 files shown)


config/cscs.py
```diff
@@ -924,12 +943,48 @@
...
+            ]
+        },
+        {
+            'name': 'cpeAMD',
             'target_systems': [
                 'eiger', 'pilatus'
             ],
             'modules': [
-                {'name': 'PrgEnv-aocc', 'collection': True}
+                {'name': 'cpeAMD', 'collection': False}
```
vkarak, line 960: When a module is not a collection, just use the classical way of defining it: `'modules': ['cpeAMD']`.

config/cscs.py
```diff
@@ -660,10 +660,10 @@
             ]
         },
         {
-            'name': 'pilatus',
-            'descr': 'Alps Cray EX Supercomputer (TDS)',
+            'name': 'eiger',
```
vkarak, line 663: I don't understand this. Pilatus becomes Eiger and Eiger Pilatus?

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -9,7 +9,10 @@
 
 class VASPCheck(rfm.RunOnlyRegressionTest):
     def __init__(self):
-        self.valid_prog_environs = ['builtin']
+        if self.current_system.name == 'pilatus':
+            self.valid_prog_environs = ['cpeIntel']
+        else:
+            self.valid_prog_environs = ['builtin']
```
vkarak, line 15: Add a new line after here.

vkarak (changes requested):

config/cscs.py
lucamar: I did that, but I got an error so I changed it... maybe the error was due to something else, I can revert the change.

config/cscs.py
lucamar: I just moved Pilatus after Eiger, since we have also Dom after Piz Daint.

config/cscs.py
lucamar: I realised that initially I had specified `{'name': 'cpeAMD'}` without the `collection`, therefore I got the error.

vkarak (approved):

PR #1838, opened 2021-03-06
[feat] Improve naming scheme of tests

### LINKED ISSUE #1794: Improve naming scheme of parameterized tests
With the new `parameter()` directive, the test names can become really large and not easy to handle, so we need to find an alternative naming scheme.

vkarak:
This PR enhances the naming scheme of tests by introducing the following concepts:

1. Each test is now associated with a unique id, which is a SHA256 hash. This has is calculated based on the test name and its parameters, if it is a parameterized test.
2. The unique hash is available through the `hash_long` and `hash_short` properties.
3. A test gets a `unique_name`, which is essentially the test class names with the test hash appended, if the test is parameterized, i.e., `ParamTest_<short_hash>`. We could always append the test hash, but we choose not to do so, in order to maintain compatibility with the current naming scheme of non-parameterized tests.
4. The test's `name` becomes a human-readable version of the test's name. If the test is not parameterized, this is simple the qualified class name of the test. If it is parameterized, the parameter names and their values are appended as follows: `MyTest%param1=value1%param2=value2`
5. Modifying the test's `name` is now deprecated.
6. Test listings have been updated to include the test hashes.
7. Tests can also be selected by hash using the `-n` option as follows: `-n /<hash>`.
8. The test's `name` is no more used in paths. The `unique_name`, which is read-only and has no special characters, is used instead.
9. The test's default `executable` uses the `unique_name` instead of `name`.
10. The name of the generated job and build scripts do not contain the test name any more. They are simply `_rfm_build.{sh,out,err}`and `_rfm_run.{sh,out,err}`. The test name was superfluous here, since it is alread part of the path.

### Implementation details

The parameterized tests defined using the `@parameterized_test` decorator are now implemented using the `parameter` built-in machinery. We essentially extend the test's parameter space with the parameters passed in the decorators. By inspecting the test's `__init__()` function we can also assign the right names to the parameters. However, one of the key differences of the `@parameterized_test` decorator compared to the `parameter` built-in is the way multiple parameters are treated.  The `parameter()` directive uses the external product of the parameters to create new tests, whereas the `@parameterized_test` decorator simply zips the parameters and passes them to the test's init function. Supporting this requires to add the possibility to change the iteration scheme of the parameterization space. Another difference between the built-in and the decorator is that the parameterization space is inherited in the case of the built-in, such that a `@simple_test` that derives from a parameterized test is also parameterized. This is not the case, however, when using the decorator. To be able to support this using the built-in machinery, we had to mark a test that is parameterized in the old way, and do not allow inheritance of the parameterization space in this case. In the future, we should consider deprecating the `@parameterized_test` decorator, since it does not offer anything compared to the more powerful `parameter` built-in.

- The `_rfm_init()` function is renamed to `__rfm_init__()` so as to be in sync with other special framework functions.
- The hacks in the series of sleep checks used in `test_policies` are now removed and a parameterized test is used.
- The framework's module import logic is extended so as to allow force reloading of a module. This is quite useful in unit tests, where load multiple times the same tests and we would like every time to start fresh with the creation of its variable and parameter spaces. The addition of this logic to the basic module load utility functions allowed us to simplify the logic of some test loading unit test fixtures.

### Todos:

- [ ] Update documentation
- [ ] Actually deprecate setting the test's `name`.

Fixes #1794.

### DIFF
--- a/reframe/core/decorators.py
+++ b/reframe/core/decorators.py
@@ -22,6 +22,7 @@ import traceback
 import reframe.utility.osext as osext
 from reframe.core.exceptions import ReframeSyntaxError, user_frame
 from reframe.core.logging import getlogger
+from reframe.core.parameters import TestParam
 from reframe.core.pipeline import RegressionTest
 from reframe.utility.versioning import VersionValidator
 
@@ -56,9 +57,11 @@ def _register_test(cls, args=None):
                 ret.append(_instantiate(cls, args))
             except Exception:
                 frame = user_frame(*sys.exc_info())
-                msg = "skipping test due to errors: %s: " % cls.__name__
-                msg += "use `-v' for more information\n"
-                msg += "  FILE: %s:%s" % (frame.filename, frame.lineno)
+                msg  = f"skipping test due to errors: {cls.__qualname__}: "
+                msg += f"use `-v' for more information\n"
+                if frame:
+                    msg += f'  FILE: {frame.filename}:{frame.lineno}'
+
                 getlogger().warning(msg)
                 getlogger().verbose(traceback.format_exc())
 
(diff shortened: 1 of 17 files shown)


vkarak:
I will redo this PR against the `4.x` branch.

PR #1842, opened 2021-03-08
[bugfix] Fix passing of stagedir to container platforms

teojgo:


vkarak:
@jenkins-cscs retry dom

vkarak:
I am merging but we need a unit test to exercise this bug.

vkarak (approved):

PR #1844, opened 2021-03-09
[WIP] Add processor info to configuration file

jgphpc:


config/cscs.py
```diff
@@ -768,19 +768,17 @@
...
-                            'num_cpus_per_socket' 128,
-                            # +num_cores
-                            # +num_cores_per_socket
-                            # +num_numa_nodes
-                            # +num_cores_per_numa_node
-                            # 'num_sockets'
-                        },
-                    ],
+                    'processor': {
+                        'arch': 'AMD EPYC 7742 64-Core Processor',
```
ekouts, line 772: This should be compatible with [archspec](https://archspec.readthedocs.io/en/latest/index.html). After installing you can see the name with
```
python -c "import archspec.cpu; print(archspec.cpu.host())"
```

ekouts: I think it should be `zen2` in pilatus.
```suggestion
                        'arch': 'zen2',
```

jgphpc:
ok, I will use this config file as a temporary solution.

jgphpc:
see https://github.com/eth-cscs/reframe/pull/1991

PR #1847, opened 2021-03-09
[feat] Minimize the list of doc requirements

### LINKED ISSUE #1836: Minimize the list of required dependencies in `docs/requirements.txt`
I think that the current list is populated by a `pip freeze`. We should create a list by hand with only the first-level dependencies

teojgo:
Fixes #1836


vkarak:
Is this indeed the minimal? I.e., if I remove one those, the build will fail? In the Spack package we have only those and it seems to work:

https://github.com/spack/spack/blob/3b1b51e90de380fa9524121f95df989b00eeb6a8/var/spack/repos/builtin/packages/reframe/package.py#L66-L68

teojgo:
> Is this indeed the minimal? I.e., if I remove one those, the build will fail? In the Spack package we have only those and it seems to work:
> 
> https://github.com/spack/spack/blob/3b1b51e90de380fa9524121f95df989b00eeb6a8/var/spack/repos/builtin/packages/reframe/package.py#L66-L68

Well performing `pip install sphinx` brings almost all the above with it.

vkarak (approved):

PR #1848, opened 2021-03-09
[feat] Improve naming scheme of tests

### LINKED ISSUE #1794: Improve naming scheme of parameterized tests

vkarak:
This PR enhances the naming scheme of tests by introducing the following concepts:

1. Each test is now associated with a unique id, which is a SHA256 hash. This hash is calculated based on the test name and its parameters, if it is a parameterized test.
2. The unique hash is available through the `hash_long` and `hash_short` properties.
3. A test gets a `unique_name`, which is essentially the test class name with the test hash appended, if the test is parameterized, i.e., `ParamTest_<hash_short>`. We could always append the test hash, but we choose not to do so, in order to maintain compatibility with the current naming scheme of non-parameterized tests.
4. The test's `name` becomes a human-readable version of the test's name. If the test is not parameterized, this is simply the qualified class name of the test. If it is parameterized, the parameter names and their values are appended as follows: `MyTest%param1=value1%param2=value2`
5. Modifying the test's `name` is now deprecated.
6. Test listings have been updated to include the test hashes.
7. Tests can also be selected by hash using the `-n` option as follows: `-n /<hash>`.
8. The test's `name` is no more used in paths. The `unique_name`, which is read-only and has no special characters, is used instead.
9. The test's default `executable` uses the `unique_name` instead of `name`.
10. The name of the generated job and build scripts do not contain the test name any more. They are simply `_rfm_build.{sh,out,err}`and `_rfm_run.{sh,out,err}`. The test name was superfluous here, since it is already part of the path.

### Implementation details

The parameterized tests defined using the `@parameterized_test` decorator are now implemented using the `parameter` built-in machinery. We essentially extend the test's parameter space with the parameters passed in the decorators. By inspecting the test's `__init__()` function we can also assign the right names to the parameters. However, one of the key differences of the `@parameterized_test` decorator compared to the `parameter` built-in is the way multiple parameters are treated.  The `parameter()` directive uses the external product of the parameters to create new tests, whereas the `@parameterized_test` decorator simply zips the parameters and passes them to the test's `__init__()` function. Supporting this requires to add the possibility to change the iteration scheme of the parameterization space. Another difference between the built-in and the decorator is that the parameterization space is inherited in the case of the built-in, such that a `@simple_test` that derives from a parameterized test is also parameterized. This is not the case, however, when using the decorator. To be able to support this using the built-in machinery, we had to mark a test that is parameterized in the old way, and do not allow inheritance of the parameterization space in this case. In the future, we should consider deprecating the `@parameterized_test` decorator, since it does not offer anything compared to the more powerful `parameter` built-in.

Other implementation changes:

- The `_rfm_init()` function is renamed to `__rfm_init__()` so as to be in sync with other special framework functions.
- The hacks in the series of sleep checks used in `test_policies` are now removed and a parameterized test is used.
- The framework's module import logic is extended so as to allow force reloading of a module. This is quite useful in unit tests, where we load multiple times the same tests and we would like every time to start fresh with the creation of its variable and parameter spaces. The addition of this logic to the basic module load utility functions allowed us to simplify the logic of some test loading unit test fixtures.

### Todos:

- [ ] Update documentation.
- [ ] Actually deprecate setting the test's `name`.
- [ ] Remove stale comments.
- [ ] Implement and test the `have_not_hash()` filter.

Fixes #1794.

### DIFF
--- a/reframe/core/decorators.py
+++ b/reframe/core/decorators.py
@@ -22,6 +22,7 @@ import traceback
 import reframe.utility.osext as osext
 from reframe.core.exceptions import ReframeSyntaxError, user_frame
 from reframe.core.logging import getlogger
+from reframe.core.parameters import TestParam
 from reframe.core.pipeline import RegressionTest
 from reframe.utility.versioning import VersionValidator
 
@@ -56,9 +57,11 @@ def _register_test(cls, args=None):
                 ret.append(_instantiate(cls, args))
             except Exception:
                 frame = user_frame(*sys.exc_info())
-                msg = "skipping test due to errors: %s: " % cls.__name__
-                msg += "use `-v' for more information\n"
-                msg += "  FILE: %s:%s" % (frame.filename, frame.lineno)
+                msg  = f"skipping test due to errors: {cls.__qualname__}: "
+                msg += f"use `-v' for more information\n"
+                if frame:
+                    msg += f'  FILE: {frame.filename}:{frame.lineno}'
+
                 getlogger().warning(msg)
                 getlogger().verbose(traceback.format_exc())
 
@@ -117,6 +120,28 @@ def parameterized_test(*inst):
       This decorator does not instantiate any test.  It only registers them.
       The actual instantiation happens during the loading phase of the test.
     '''
+    def _extend_param_space(cls):
+        if not inst:
+            return
+
+        argnames = inspect.getfullargspec(cls.__init__).args[1:]
+        params = {}
+        for args in inst:
+            if isinstance(args, collections.abc.Sequence):
+                for i, v in enumerate(args):
+                    params.setdefault(argnames[i], ())
+                    params[argnames[i]] += (v,)
+            elif isinstance(args, collections.abc.Mapping):
+                for k, v in args.items():
+                    params.setdefault(k, ())
+                    params[k] += (v,)
+
+        # argvalues = zip(*inst)
+        # for name, values in zip(argnames, argvalues):
+        #     cls.param_space.params[name] = values
+
+        cls.param_space.params.update(params)
+
     def _do_register(cls):
         _validate_test(cls)
         if not cls.param_space.is_empty():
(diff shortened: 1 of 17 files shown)


vkarak:
> I think there is no need to have the full test name 6 times in here. 

Why not? "Repetitio est mater studiorum" 😭

victorusu:
@vkarak, is the CI failure spurious?

vkarak:
@victorusu This PR is outdated and needs to be synced with the master. I don't think that the CI failure comes from the PR, since it happens from time to time.

jjotero (changes requested):
I think there're a few things that need to be addressed before discussing the naming scheme, and perhaps it's a good idea to tackle these in separate PRs:

- Drop `@parameterized_test` decorator. I saw there're a few workarounds for that and I think we should not have this decorator in the 4.0 version. I think it's knife time :)
- Rather than relying on the `_rfm_use_params` constructor argument, this argument must be changed to an index. This will give us control over which point of the parameter space we create with each class instantiation. This is even more important when thinking of fixtures (that'll add a second index for the fixture space). An added benefit of this is that several sections of the code no-longer need to rely on the order of instantiation, and we'll be able to query things into the parameter space "by index" (e.g. what's the `name` for parameter variant `6`, and so on). This will also potentially simplify the implementation of the `getdeps`.

Now, regarding the naming scheme:
- `name`, `_unique_id` and so on must not be variables. The `varable` built-in should only be used for class attributes that the user can interact (edit) with. These are instead better expressed as properties, I think.
- `name` should refer to the internal & unique name for the test and not the human-readable one. The human-readable name is only used for the console output (`fancy_name`?).
- We need a `@classmethod` to query the name of any test variant by test variant ID. The `name` could be a property that binds this function to the test's variant index.

jjotero:
To be redone after the fixtures #2166

PR #1850, opened 2021-03-09
[test] Scalapack check: remove custom code

ChristopherBignamini:


vkarak:
@ChristopherBignamini I think you mixed up this PR with other tests.

ChristopherBignamini:
Yes, sorry...I don't know what happened, can we delete the PR?

ChristopherBignamini:
No need to close, it should be ok now

vkarak (changes requested):
This is not enough we should remove the access to the resourcesdir altogether in the test by moving the original file in the test's `sourcesdir`.

victorusu (approved):
`lgtm`

vkarak (approved):

PR #1861, opened 2021-03-11
[test] Add nvidia-smi check

victorusu:
This introduces the Nvidia SMI tests that check whether Accounting, Compute and ECC modes are set to the correct values.

### DIFF
--- a/cscs-checks/system/nvidia/nvidia_smi_check.py
+++ b/cscs-checks/system/nvidia/nvidia_smi_check.py
@@ -27,7 +27,7 @@ class nvidia_smi_check(rfm.RunOnlyRegressionTest):
     })
 
     @rfm.run_before('run')
-    def set_variant_opts(self):
+    def set_display_opt(self):
         self.executable_opts.append(self.gpu_mode.upper())
 
     @rfm.run_before('sanity')
--- a/cscs-checks/system/nvidia/nvidia_smi_check.py
+++ b/cscs-checks/system/nvidia/nvidia_smi_check.py
@@ -5,114 +5,46 @@
 
 import reframe as rfm
 import reframe.utility.sanity as sn
-
-
-def sanity_evaluation(regex, regex_group, output, refvalue, field_msg, num_nodes, nodelist):
-    all_tested_nodes = sn.evaluate(sn.extractall(regex, output, regex_group))
-    num_tested_nodes = len(all_tested_nodes)
-    node_failure_msg = ('Requested %s node(s), but found %s node(s)' %
-                   (num_nodes, num_tested_nodes))
-    sn.evaluate(sn.assert_eq(num_tested_nodes, num_nodes,
-                                msg=node_failure_msg))
-
-    failures = []
-    for i in range(len(all_tested_nodes)):
-        stats = all_tested_nodes[i]
-        node = nodelist[i]
-        if stats != refvalue:
-            failures.append(node)
-
-    the_failure_msg = '%s is not set to "%s" for node(s): %s' % (field_msg, refvalue, ','.join(failures))
-    sn.evaluate(sn.assert_eq(len(failures), 0, msg=the_failure_msg))
-
-
-class NVIDIASMIBaseCheck(rfm.RunOnlyRegressionTest):
-    '''Base class for NVIDIA SMI tests'''
-
-    def __init__(self):
-        self.valid_systems = ['daint:gpu', 'dom:gpu']
-        self.valid_prog_environs = ['builtin']
-        self.executable = 'nvidia-smi'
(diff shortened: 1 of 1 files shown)


vkarak:
I am not sure that the location of the test is correct. Perhaps remove the `nvidia/` subfolder.

cscs-checks/system/nvidia/nvidia_smi_check.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(self):
+        self.valid_systems = ['daint:gpu', 'dom:gpu']
+        self.valid_prog_environs = ['builtin']
+        self.executable = 'nvidia-smi'
+        self.num_tasks = 0
+        self.num_tasks_per_node = 1
+        self.exclusive = True
+        self.tags = {'nvidia', 'maintenance', 'production', 'single-node'}
+        self.maintainers = ['VH']
```
vkarak, line 40: Set all those in the class body and remove the `__init__()` altogether.

cscs-checks/system/nvidia/nvidia_smi_check.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@rfm.simple_test
+class NVIDIASMIAccountCheck(NVIDIASMIBaseCheck):
```
vkarak, line 44: I would not use subclasses, but rather parameterize on the type of the test, e.g., `ACCOUNTING`, `ECC` etc.

cscs-checks/system/nvidia/nvidia_smi_check.py
```diff
@@ -0,0 +1,118 @@
+# Copyright 2016-2021 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+def sanity_evaluation(regex, regex_group, output, refvalue, field_msg, num_nodes, nodelist):
```
vkarak, line 10: Most or all the logic here can go in a pre-sanity hook that would also set the `sanity_patterns`.

vkarak (changes requested):

vkarak:
In order to put everything to the class body, you need #1862.

vkarak:
@victorusu I simplified and modernized the test. That's the new way! @jjotero Please confirm :-) We are also almost to the point of getting rid completely of `__init__()` in all cases. By the way, I don't why you were checking also the `Pending ECC Mode`. Did you have to? I am not checking it here, because I wanted to keep the test simpler.

cscs-checks/system/nvidia/nvidia_smi_check.py
```diff
@@ -0,0 +1,50 @@
...
+class nvidia_smi_check(rfm.RunOnlyRegressionTest):
+    gpu_mode = parameter(['accounting', 'compute', 'ecc'])
+    valid_systems = ['daint:gpu', 'dom:gpu']
+    valid_prog_environs = ['builtin']
+    executable = 'nvidia-smi'
+    executable_opts = ['-a', '-d']
+    num_tasks = 0
+    num_tasks_per_node = 1
+    exclusive = True
+    tags = {'maintenance', 'production'}
```
vkarak, line 21: @victorusu Flexible tests in maintenance and production? Nah...

vkarak: I would make it `diagnostic`.

vkarak (changes requested):

cscs-checks/system/nvidia/nvidia_smi_check.py
```diff
@@ -0,0 +1,50 @@
...
+    maintainers = ['VH']
+    mode_values = variable(typ.Dict[str, str], value={
+        'accounting': 'Enabled',
+        'compute': 'Exclusive_Process',
+        'ecc': 'Enabled'
+    })
+
+    @rfm.run_before('run')
+    def set_variant_opts(self):
```
vkarak, line 30:
```suggestion
    def set_display_opt(self):
```

cscs-checks/system/nvidia/nvidia_smi_check.py
victorusu: I talked internally with Matteo and he told me that 1 node is enough to test on maintenance, due to the use of the same image.

vkarak (approved):

jjotero (approved):
lgtm

teojgo (approved):

PR #1862, opened 2021-03-12
[feat] Allow setting the test description and executable name directly in the class body

jjotero:
This makes the default behaviour of these two variables with all the others.


vkarak (approved):
I've done a couple of changes and now I can merge it. Let me know if you agree also with them.

Issue #1864, opened 2021-03-12
Add a `depends_on` directive


vkarak:
I suggest adding a `depends_on()` directive (not built-in) that will act in the class body. This could essentially simply store the arguments in a special internal variable, that would then call `obj.depends_on()` during the variable injection. This way, we eliminate the need to define `depends_on()` in the `__init__()` method or in a post-init hook, as suggested by #1863. 

@jjotero What do you think?

vkarak:
Closing, see discussion in #1868.

PR #1865, opened 2021-03-13
[feat] Re-implement pipeline hook mechanism and add a post-init hook

### LINKED ISSUE #1863: Add a post_init pipeline hook
This is what is missing to completely get rid of defining the `__init__()` method. We currently still need the `__init__()` if we need to set variables, related to test filtering, based on some parameter values or for dependencies.

vkarak:
With the new syntax of variable definition we can almost get rid of `__init__()`. Currently, functions like `depends_on()` must still be called during initialisation, but this will be addressed by #1864. However, there will still be cases that something must be executed during the initialisation and this is when you want to have dynamic tags, for example, change the tags based on a test parameter. In order for the frontend to be able to select tests by tags, the tags must be finalised after the initialisation, so you can't practically set them in any other pipeline stage. 

The post-init hook introduced here is fairly simple it can substitute the constructor. The following:

```python
class my_test(...):
    @rfm.run_after('init')
    def init_a(self):
        self.a = 1

    @rfm.run_after('init')
    def init_a(self):
        self.b = 2
```

is equivalent to

```python
class my_test(...):
    def __init__(self):
        self.a = 1
        self.b = 2
```

Of course, the example with `init_a()` and `init_b()` is a bit silly, but it gets more sense if you imagine test hierarchies.
~~If a test defines `__init__()`, any init hooks will be ignored.~~

Finally, there are no pre-init hooks, although it's trivial to implement. I'm not introducing them here, unless there is a need.

Fixes #1863.

## Implementation details

The pipeline mechanism is re-implemented from scratch. Pipeline hooks are no more applied to the pipeline functions with the decorator. This has a number of advantages:

a. Hooks will be automatically applied to any pipeline function that is overriden by a subclass.
b. The hooks will wrap the leaf function and not the call to `super().pipeline_fn()`
c. The hooks will be applied only once in all cases even if a test and its child classes are instantiated multiple times.

Specific implementation tricks and intricacies are explained with comments in the code.


docs/regression_test_api.rst
```diff
@@ -29,6 +29,18 @@ Regression Test Class Decorators
...
+
+Pipeline hooks is an easy way to perform operations while the test traverses the execution pipeline.
+You can attach arbitrary functions to run before or after any pipeline stage, which are called *pipeline hooks*.
+Multiple hooks can be attached before or after the same pipeline stage, in which case the order of execution will match the order in which the functions are defined in the class body of the test.
+A single hook can also be applied to multiple stages and it will be executed multiple times.
+All pipeline hooks of a test class are inherited by its subclasses.
+Subclasses may override a pipeline hook of their parents by redefining the hook function and re-attaching it at the same pipeline stage.
+There are seven pipeline stages where you can attach test methods: ``init``, ``setup``, ``compile``, ``run``, ``sanity``, ``performance`` and ``cleanup``.
+
+
```
victorusu, line 43: Does it make sense to add a sentence that states why/when/what for one can use the `init` hooks, compared to the other hooks?

victorusu (approved):
`lgtm`.  I just wonder if it makes sense to tell the difference between the `run_before('setup')`, compared to `run_after('init')`. Or in other words, what is the `init` hook meant to be used for.

jjotero:
> If a test defines __init__(), any init hooks will be ignored.

I find very very confusing that a standard Python method can magically override a reframe hook. I'm struggling to see what the advantage is of having this `post-init` hook. Is this simply to get rid of `__init__`?

vkarak:
> I'm struggling to see what the advantage is of having this post-init hook. Is this simply to get rid of __init__?

Basically, yes. It is just syntactic sugar. Generally, everything can go into hooks, but there is one case that you can't use hooks: it's when you have to deal with test filtering. There are tests that we change the tags or the `valid_prog_environs` based on some parameter. This cannot happen anywhere else except in init, so there will be tests to define it, if you don't have some sort of hooks. What about defining the hook differently, something like `pre_filter`. I don't know.

vkarak:
I will postpone this PR to the next sprint, since I would like to improve also the current hook mechanism to follow the same pattern as with what we have discussed with `__init__()`, so that the hooks run even if a method has been overriden and make sure that they are executed once at the class lowest in the hierarchy that defines the hooked method.

reframe/core/meta.py
```diff
@@ -143,27 +145,34 @@ def __init__(cls, name, bases, namespace, **kwargs):
         # Set up the hooks for the pipeline stages based on the _rfm_attach
```
jjotero, line 145: How about moving all this hook logic into the `HookRegistry`'s `__init__`?

reframe/core/hooks.py
```diff
@@ -0,0 +1,109 @@
...
+                hook_kind = name
+            else:
+                # Just any name that does not exist
+                hook_kind = 'xxx'
+
+            if hook_kind not in hooks:
+                return []
+
+            return [h for h in hooks[hook_kind]
+                    if h.name not in obj._disabled_hooks]
```
jjotero, line 30:
```suggestion
def _attach_hooks(hooks, phase=None):
    def _deco(func):
        def select_hooks(obj, kind):
            if phase is None:
                fn_name = func.__name__
                if fn_name == '__init__':
                    fn_name = 'init'

                phase = kind + fn_name
            elif phase is not None and not phase.startswith(kind):
                # Just any name that does not exist
                phase = 'xxx'

            if phase not in hooks:
                return []

            return [h for h in hooks[phase]
                    if h.name not in obj._disabled_hooks]
```

jjotero: Renaming `name` and `hook_kind` as `phase` makes it consistent with the rest of the code and IMO makes this a lot easier to understand.

reframe/core/pipeline.py
```diff
@@ -799,7 +749,13 @@ def __new__(cls, *args, _rfm_use_params=False, **kwargs):
                         os.path.dirname(inspect.getfile(cls))
                     )
 
-        obj._rfm_init(name, prefix)
+        # Apply the hooks only once (not when deep-copying)
+        if not hasattr(cls, '_rfm_first_init'):
+            cls._rfm_pipeline_hooks.apply(obj)
+            cls._rfm_first_init = True
+
```
jjotero, line 756: What I'm missing here is some kind of health check that makes sure that all the hooks are properly attached to a legit stage of the pipeline. I think this could be done by passing here a list with the supported phases that a hook can live on into a sanity function from the `HookRegistry`. Then, in order to not have redundant executions of this sanity function, we could use the `_rfm_use_params` variable, which is only set to `True` by the `simple_test` decorator (it would not run when deep-copying). Of course, if we go for that option, it'd be sensible to rename `_rfm_use_params` to something like `_rfm_registered`.

jjotero (commented):
I like the idea of the `HookRegistry`, but I'm not convinced by the new way the pipeline stages are dynamically decorated to pick up the hooks. I see no reason to do that and I feel that it makes more sense to have the pipeline stages decorated in the pipeline itself. See comments below.

vkarak:
> I see no reason to do that and I feel that it makes more sense to have the pipeline stages decorated in the pipeline itself. See comments below.

The "problem" is that we need to handle disabled hooks. Users can disable hooks on-demand and this happens *after* the test has been created. Regardless if the `_rfm_disabled_hooks` is a class or instance attribute, it is set only after the object is created, i.e., after `__rfm_init__()`. For this reason, running the hooks cannot be static and I also converted the `_rfm_disabled_hooks` to an object attribute, because it didn't have any meaning to be a class attribute. It was only complicating the implementation of inheritance.

reframe/core/hooks.py
```diff
@@ -0,0 +1,109 @@
...
+    def __getitem__(self, key):
+        return self.__hooks[key]
+
+    def __setitem__(self, key, name):
+        self.__hooks[key] = name
+
+    def __getattr__(self, name):
+        return getattr(self.__hooks, name)
+
+    def update(self, hooks):
```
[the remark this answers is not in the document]
vkarak: I'm not sure about these type checks here. This is an internal method. If the wrong type is passed, it's certainly an implementation bug.

reframe/core/pipeline.py
```diff
@@ -799,7 +749,13 @@ def __new__(cls, *args, _rfm_use_params=False, **kwargs):
                         os.path.dirname(inspect.getfile(cls))
                     )
 
-        obj._rfm_init(name, prefix)
+        # Apply the hooks only once (not when deep-copying)
+        if not hasattr(cls, '_rfm_first_init'):
```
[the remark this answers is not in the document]
vkarak: > I think the decoration of the pipeline stages should happen in the pipeline

What do you mean by this? In the frontend that executes the tests?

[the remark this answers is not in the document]
vkarak: > (Reading this _rfm_first_init immediately triggered horror memories of Fortran 77 code 😱😂)

Practically, this is not too different from the `_rfm_use_params` argument 😬

reframe/core/hooks.py
```diff
@@ -0,0 +1,109 @@
...
+        for phase, hks in hooks.items():
+            self.__hooks.setdefault(phase, util.OrderedSet())
+            for h in hks:
+                self.__hooks[phase].add(h)
+
+    def apply(self, obj):
```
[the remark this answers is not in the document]
vkarak: You have a point here and I am kind of starting understanding your major concern.

reframe/core/meta.py
vkarak: Perhaps it makes sense. I can give it a try.

reframe/core/hooks.py
vkarak: I did it, except that I did not rename the `name` argument of `_attach_hooks`. You have to copy that in the `select_hooks`, because otherwise you would need to mark `phase` as `nonlocal`, which is not correct, because you will be sharing it across invocations of `select_hooks()` (and unit tests failed as a result).

reframe/core/pipeline.py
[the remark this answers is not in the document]
vkarak: Indeed, it guards also against re-decorating an inherited function. I will record the attached hooks and do not re-attach them, so that we can avoid this, and also be able to do the sanity check that you propose.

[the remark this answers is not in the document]
vkarak: This guard, apart from "old-fashioned", is also wrong, because although it will not reattach existing hooks, it will fail to attach new hooks of the derived class. I'm working on alternative designs.

vkarak:
Moving to next sprint. It's much more complicated than it seems.

jjotero (approved):
lgtm

reframe/core/decorators.py
```diff
@@ -226,19 +226,51 @@ def run_before(stage):
...
     '''
+
+    # Map user stage names to the actual pipeline functions if needed
+    if stage == 'init':
+        stage = '__init__'
+    elif stage == 'compile':
+        stage = 'compile_wait'
+    elif stage == 'run':
+        stage = 'run_wait'
+
```
vkarak, line 273: I will also add a sanity check here and to the `run_before` so that we issue an error if the user tries to attach a hook to an arbitrary function: this will also match the documentation.

unittests/test_pipeline.py
```diff
@@ -525,6 +526,42 @@ def set_resources(self):
...
+            self.z = 3
+
+    t0 = _T0()
+    assert t0.x == 1
+    assert t0.y == 2
+
+    t1 = _T1()
+    assert t1.x == 1
+    assert t1.y == 2
+    assert t1.z == 3
```
jjotero, line 562: What about concatenating strings here and then asserting for `t1.x == 'abc'`? That would test the order of execution too.

jjotero (approved):

vkarak:
I fixed the unit tests in `test_policies.py`, but this problem is in different files, too, but it doesn't get triggered. I have created #1913 to address this separately.

unittests/test_pipeline.py
vkarak: Correct, I'll do that.

vkarak:
@jenkins-cscs retry all

PR #1871, opened 2021-03-17
[feat] Add a new build system backend for building test code with Spack

### LINKED ISSUE #1466: Use Spack as a build system
It would be nice to be able to use Spack as a build system for building the source code of a test.

This relates to #300.

rsarm:
Closes #1466


reframe/core/buildsystems.py
```diff
@@ -779,6 +782,89 @@ def generated_modules(self):
...
+
+    def __init__(self):
+        super().__init__()
+        self.packages = []
+        self.environment = None
+        self.upstreams = []
+        self._prefix_save = None
+
+    def _prepare_scope(self):
+        os.mkdir('rfm_scope')
```
victorusu, line 798: I would write `rfm_spack_scope`. If you accept the idea, do not forget to change the `rfm_scope` from everywhere you have it... :-)

reframe/core/buildsystems.py
```diff
@@ -779,6 +782,89 @@ def generated_modules(self):
...
+
+        matches = re.findall(r'\[\+\] \S+/spack/opt/spack/(\S+)/\S+/(\S+)',
+                             out)
+
+        self._spack_modules = [{'name': os.path.basename(mod),
+                                'collection': False,
+                                'path': None}
+                               for _, mod in matches]
+
+        spack_arch = matches[0][0]
```
victorusu, line 850: This `spack_arch` is not really the architecture used in your build because it you are using Lmod, then it will be always `x86_64`, so I think it would be better to name it `spack_module_arch` or just `module_arch`.
Furthermore there is an additional `Core` folder for Lmod, if one is using the hierarchical naming scheme...

victorusu (changes requested):
@rsarm, it works great for me, except when I am using with Lmod with hierarchical naming scheme... I think we should have a dedicated solution for that.

reframe/core/buildsystems.py
```diff
@@ -779,6 +782,89 @@ def generated_modules(self):
...
+        self.packages = []
+        self.environment = None
+        self.upstreams = []
+        self._prefix_save = None
+
+    def _prepare_scope(self):
+        os.mkdir('rfm_scope')
+        scope_config = {
+            'config': {
+                'install_tree': 'spack/opt/spack',
```
victorusu, line 801: Having only the `install_tree` and the `module_roots` is not enough to allow for a proper parallel execution of Spack. The builds can be blocked due to filesystem locks.
One still needs to add the `source_cache` and `misc_cache` entries. And they should point to unique paths for that particular Spack instance. There way we will run completely independent Spack instances. For example entries for these variables, please take a look at https://github.com/eth-cscs/production/blob/6d1dca77df8b7926a63cd5a0a39e063bb9f8b815/spack/daint/config.yaml#L19 and https://github.com/eth-cscs/production/blob/6d1dca77df8b7926a63cd5a0a39e063bb9f8b815/spack/daint/config.yaml#L22

victorusu (changes requested):
@rsarm, I forgot to add the issue that I have experienced with this build system. Please take a look at the comment about `source_cache` and `misc_cache`.

rsarm:
Example checks:
```python
import reframe as rfm

class SpackMixin(rfm.RegressionTest):
    @rfm.run_before('run')
    def prepare_run(self):
        self.prerun_cmds = [
            f'spack env activate -d {self.build_system.environment}'
        ]
        self.prerun_cmds.append([
            f'spack load {mod}'
            for mod in self.build_system.generated_modules
        ])


@rfm.simple_test
class SpackEnvCheck(SpackMixin):
    def __init__(self):
        self.descr = 'Spack buildsystem example.'
        self.valid_systems = ['*']  # ['daint:gpu', 'dom:gpu']
        self.valid_prog_environs = ['*']  # ['builtin']
        self.sourcesdir += '/spack'
        self.build_system = 'Spack'
        self.executable = 'iconv'
        self.executable_opts = ['--version']
        self.build_system.environment = 'spack_env'
        self.sanity_patterns = sn.assert_found(
            r'iconv \(GNU libiconv 1\.16\)', self.stdout
        )
```
with the environment
```yaml
spack:
  specs:
    - zlib@1.2.3
    - libiconv@1.16
  concretization: together
  config:
    install_tree: spack/opt/spack
    module_roots:
      tcl: spack/share/spack/modules
      lmod: spack/share/spack/lmod
```
that goes in `src/spack/spack.yaml`.

This test runs fine.


The following test using 'scopes' gives the error as if spack was not properly setup (at least with the local scheduler):
```python
import reframe as rfm

class SpackMixin(rfm.RegressionTest):
    @rfm.run_before('run')
    def prepare_run(self):
        self.prerun_cmds = [
            f'spack -C {self.build_system.scope_dir} load {mod}'
            for mod in self.build_system.generated_modules
        ]
        # self.executable = 'spack'
        # self.executable_opts = [f'-C {self.build_system.scope_dir} find']


@rfm.simple_test
class UlimitCheck(SpackMixin):
    def __init__(self):
        self.descr = 'Checking the output of ulimit -s in node.'
        self.valid_systems = ['*']  # ['daint:gpu', 'dom:gpu']
        self.valid_prog_environs = ['*']  # ['builtin']
        self.sourcesdir += '/ulimit'
        self.build_system = 'Spack'
        self.build_system.packages = ['libiconv@1.16',
                                      'zlib@1.2.3',
                                      'openmpi@4.0.5']
        self.build_system.upstreams = ['/home/sarafael/spack/opt/spack']
        self.executable = 'iconv'
        self.executable_opts = ['--version']
        self.sanity_patterns = sn.assert_found(
            r'iconv \(GNU libiconv 1\.16\)', self.stdout
        )
```

It creates the script
```
#!/bin/bash
spack -C rfm_spack_scope load libiconv@1.16/af5tdk6ilv6mah2ntgb5odryvlosijnz
spack -C rfm_spack_scope load zlib@1.2.3/5e4xv5gq7liaii4iwktakwk3gqjchlll
spack -C rfm_spack_scope load openmpi@4.0.5/622d22gh5n7vhilmy6seea53cyo3z523
 iconv --version
```
which gives the error:
```
Error: `spack load` requires spack's shell support.
```

reframe/core/buildsystems.py
```diff
@@ -782,6 +785,110 @@ def generated_modules(self):
...
+    def post_build(self, buildjob):
+        # Store the modules generated by Spack
+
+        with open(os.path.join(self._prefix_save, buildjob.stdout)) as fp:
+            out = fp.read()
+
+        matches = re.findall(r'\[\+\] \S+/spack/opt/spack/\S+/(\S+)',
+                             out)
+
+        self._modules = [] #[os.path.basename(mod.split('-')[-1])
```
vkarak, line 881: I guess the comment is a leftover, right?

vkarak (changes requested):
Generally, looks good to me. We need to expand the documentation, as well as write a minimal unit test the check the emitted commands as we for the EB. Finally, did we resolve the issue about `spack load`?

reframe/core/buildsystems.py
```diff
@@ -782,6 +782,72 @@ def generated_modules(self):
...
 
 
+class Spack(BuildSystem):
+    '''A build system for building test code using `Spack
+    <https://spack.io/>`__.
+
+    ReFrame will use Spack to build and install the code given an environment
+    and a set of specs.
+
+    .. versionadded:: 3.6.0
```
vkarak, line 792:
```suggestion
    .. versionadded:: 3.6.1
```

reframe/core/buildsystems.py
```diff
@@ -782,6 +782,72 @@ def generated_modules(self):
...
+    def __init__(self):
+        super().__init__()
+        self.specs = []
+        self.environment = None
+        self._prefix_save = None
+
+    def emit_build_commands(self, environ):
+        self._prefix_save = os.getcwd()
+        if not self.environment:
+            raise BuildSystemError(f"'environment' must not be empty")
```
vkarak, line 821: You can avoid this check by setting the type of the `environment` field to `typ.Str[r'.+']`.

vkarak (changes requested):
What we are still missing is a way to automatically activate the environment and load the modules when the test runs. One way I see this happening is by expanding the `BuildSystem` basic interface with a function that would return commands to be emitted before running the test. And we could also add a boolean field to the Spack backend to enable or disable the automatic generation of these commands.

And, finally, we also miss the documentation.

vkarak (changes requested):
I tried the PR and I have a comment:

I am using the fish shell, so in my case I have to source Spack's `setup-env.sh` in each of the scripts (build and run). Although, sourcing it in `prebuild_cmds` it will work for the build phase, it won't do so if I add it to `prerun_cmds` for the run phase, because the `spack env activate` is issued before these commands. Could we automatically generate the sourcing of the `setup-env.sh` script? Could we get the prefix of the spack installation for example?

reframe/core/pipeline.py
```diff
@@ -1328,9 +1328,9 @@ def run(self):
         if not self.current_system or not self._current_partition:
             raise PipelineError('no system or system partition is set')
 
+        prepare_run = []
```
vkarak, line 1331: I want to combine this with with the `prepare_container` commands. @teojgo Is there a reason that the `prepare_container` gets appended to `prerun_cmds`. Could it be prepended instead?

reframe/core/buildsystems.py
```diff
@@ -782,6 +792,94 @@ def generated_modules(self):
...
+        if self.specs:
+            specs_str = ' '.join(self.specs)
+            ret.append(f'spack add {specs_str}')
+
+        ret.append('spack install')
+        return ret
+
+    def _env_activate_cmds(self):
+        return [f'. $SPACK_ROOT/share/spack/setup-env.sh',
+                f'spack env activate -d {self.environment}']
```
victorusu, line 854: @rsarm, in order to avoid issues related to view creation, I would add the flag `--without-view` this command line.
More info can be found here https://spack.readthedocs.io/en/latest/environments.html

reframe/core/buildsystems.py
```diff
@@ -782,6 +792,94 @@ def generated_modules(self):
...
+
+        self._modules = []
+        for mod in matches:
+            mod = os.path.basename(mod)
+            name, version, hashid = mod.split('-')
+            self._modules.append(f'{name}@{version}/{hashid}')
+
+    @property
+    def generated_modules(self):
+        return self._modules
```
vkarak, line 880: @victorusu @rsarm Do we really need this now? I have modified the code to automatically generate `spack load` for running. I'm asking because I'm getting a strange error when we are trying to retrieve the name, version and hash id above:

```
  File "/Users/karakasv/Repositories/reframe/reframe/frontend/executors/__init__.py", line 269, in _safe_call
    return fn(*args, **kwargs)
  File "/Users/karakasv/Repositories/reframe/reframe/core/pipeline.py", line 119, in _fn
    func(obj, *args, **kwargs)
  File "/Users/karakasv/Repositories/reframe/reframe/core/pipeline.py", line 133, in _wrapped
    return fn(*args, **kwargs)
  File "/Users/karakasv/Repositories/reframe/reframe/core/pipeline.py", line 1304, in compile_wait
    self.build_system.post_build(self._build_job)
  File "/Users/karakasv/Repositories/reframe/reframe/core/buildsystems.py", line 875, in post_build
    name, version, hashid = mod.split('-')
ValueError: too many values to unpack (expected 3)
```

vkarak (approved):
lgtm now. I have done some improvements and added a couple of improvements.

victorusu (approved):
`lgtm`

vkarak:
@jenkins-cscs retry daint

PR #1873, opened 2021-03-17
[feat] Show a preview of the standard error in cases of build failures

### LINKED ISSUE #1735: Print first N lines of a failed build's stderr
Now we request users to always go and look into the stored stderr. It'd be more convenient if we printed the first N lines (N could be configurable).

teojgo:
Fixes #1735


vkarak (approved):
I reimplemented according to what we have discussed offline.

ekouts (approved):

PR #1874, opened 2021-03-17
[bugfix] Do not reset `sourcesdir` to `None` if it is set to a URL

jjotero:
Fixed a bug in the pipeline's `_rfm_init`, where `sourcesdir` was reset to `None` if the directory didn't exist. Consequently, when `sourcesdir` is a git repo and is set in the class body, `_rfm_init` would just reset it to `None`.


vkarak:
> Hmm, this behaviour is a feature actually. Here is the feature request, if I remember correctly: #1094.

Well, at least we need to make sure that we don't revert the feature in #1094.

jjotero:
> Well, at least we need to make sure that we don't revert the feature in #1094.

I think that should be fixed now. But I guess that we should move all the stuff that affects the vars to a post-init hook in the future.

teojgo (approved):

vkarak (approved):

vkarak:
@jenkins-cscs retry daint

PR #1879, opened 2021-03-23
[bugfix] Set stdin to null when launching processes from ReFrame, so that processes do no think they are on a tty

### LINKED ISSUE #1877: RunOnlyRegressionTest with executables that treat stdin like a tty fail on local scheduler
I wrote a small RunOnlyRegressionTest test that runs using the local scheduler and launcher, running a Python program. The program unexpectedly fails with this error output:
```
OSError: [Errno 25] Inappropriate ioctl for device
```
Running the command, as well as the job script created by ReFrame, does not cause the error. Neither does running the job script in a fresh python shell using subprocess.Popen.

I eventually tracked it back and was able to reproduce the error, with this as the executable
```test.py
#!/usr/bin/env python3

import sys
import os
if sys.stdin.isatty():
    print("stdin is a tty")
    # do some tty-only thing
    print(os.tcgetpgrp(sys.stdin.fileno()))
else:
    print("stdin is not a tty")
```
and the Reframe test:
```bug.py
@rfm.simple_test
class TestBuggyStdin(rfm.RunOnlyRegressionTest):
    def __init__(self):
        self.descr = 'test buggy stdin'
        self.valid_systems = ['*']
        self.executable = 'test.py'
        self.executable_opts = []
        self.sanity_patterns = sn.assert_found('stdin is not a tty', self.stdout)
```

rfm_TestBuggyStdin_job.out contains only ```stdin is a tty```.  The error file contains, as expected
```
  File "/verafs/home/joeuser/reframe_venv_3.8/bin/test.py", line 8, in <module>
    print(os.tcgetpgrp(sys.stdin.fileno()))
OSError: [Errno 25] Inappropriate ioctl for device
```
So, as far as I can tell, commands forked off by the local scheduler think that their stdin is a tty, when in reality it is not.
Just to rule out this being a python-only issue, I repeated the test with an equivalent Ruby script and got the same result.  

Adding `< /dev/null` as an entry to `executable_opts` (overriding whatever stdin reframe is feeding to the script for the command in question) lets the tests run as expected. 

Additionally, changing the call to `run_command_async` in [`reframe.core.schedulers.local.submit`](https://github.com/eth-cscs/reframe/blob/e5dc0d07acb2d570fbac34577da0c74a04f8a7b3/reframe/core/schedulers/local.py#L68) to pass either `stdin=subprocess.PIPE` or `stdin=subprocess.DEVNULL` to the underlying subprocess.Popen call makes things work.

rsarm:
Close #1877

### DIFF
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -29,13 +29,6 @@ def test_command_success():
     assert completed.stdout == 'foobar\n'
 
 
-def test_command_stdin_isatty():
-    completed = osext.run_command(
-        "python -c 'import sys; print(sys.stdin.isatty())'")
-    assert completed.returncode == 0
-    assert completed.stdout == 'False\n'
-
-
 def test_command_success_cmd_seq():
     completed = osext.run_command(['echo', 'foobar'])
     assert completed.returncode == 0
--- a/unittests/test_utility.py
+++ b/unittests/test_utility.py
@@ -29,6 +29,13 @@ def test_command_success():
     assert completed.stdout == 'foobar\n'
 
 
+def test_command_stdin_isatty():
+    completed = osext.run_command(
+        "python -c 'import sys; print(sys.stdin.isatty())'")
+    assert completed.returncode == 0
+    assert completed.stdout == 'False\n'
+
+
 def test_command_success_cmd_seq():
     completed = osext.run_command(['echo', 'foobar'])
     assert completed.returncode == 0
--- a/reframe/utility/osext.py
+++ b/reframe/utility/osext.py
@@ -77,6 +77,7 @@ def run_command(cmd, check=False, timeout=None, shell=False, log=True):
 def run_command_async(cmd,
                       stdout=subprocess.PIPE,
                       stderr=subprocess.PIPE,
+                      stdin=subprocess.PIPE,
                       shell=False,
                       log=True,
                       **popen_args):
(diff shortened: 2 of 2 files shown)


ekouts (approved):

victorusu (approved):
`lgtm`

vkarak (changes requested):
I would suggest to add a unit test based on the logic of the bug reported in #1877.

unittests/test_utility.py
```diff
@@ -29,6 +29,13 @@ def test_command_success():
     assert completed.stdout == 'foobar\n'
 
 
+def test_command_stdin_isatty():
```
vkarak, line 32: This unit test does not fail if you run it with the current master.

vkarak (changes requested):

unittests/test_utility.py
vkarak: In principle, the test is correct, but it somehow interferes with the stdout capturing of pytest. If you run `-s`, it fails as expected, but not if you let pytest capture the output.

vkarak: I don't think that we can reliably reproduce this from within the unit tests, because pytest always sets stdin to null, unless you pass `-s`. I will remove the unit test then.

vkarak (approved):

Issue #1577, opened 2020-11-04
Support for ReFrame test fixtures


vkarak:
We need to support the notion of fixtures for a ReFrame test. A fixture is different from a test dependency and different from a test parameter. The will execute where ReFrame runs and perform operations such as cloning a repo, connecting to a resource that a test needs etc. 

Test dependencies in ReFrame differ from fixtures in that they have to deal with the individual test cases (where a test can run and with which programming environments). A fixture will run regardless and will run on the same node that ReFrame itself executes. 

Test parameters as discussed in #1567 differ again to fixtures in that the parameterization space is attached to a test we want it to be inherited by its children. Having parameters as fixtures, would require the children to redefine the whole parameter space everytime. We also want parameter attributes to be set. 

A typical fixture example, is when you want a test to clone a repo and you don't want to do that for each test case. The current solution is by using test dependencies: have a test to run on a single partition and programming environment and the test doing the actual work depend fully on that one. The problem with this is that you can end up with unresolved dependencies if you run ReFrame with `--system=x:part` where `x:part` is not the partition that the test that clones the source is meant to run. This is in fact a restriction that should not exist. You want to clone the repo regardless of where the test executes. That's why we need to introduce the concept of fixtures. Here is an example that shows potential syntax (inspired by pytest fixtures):

```python
import reframe as rfm


@rfm.fixture
def tmp_path():
    yield a_dir_under_stage_prefix()


@rfm.fixture
def clone_repo(tmp_path):
    def _clone_repo(repo):
        return git_clone_repo()

    return clone_repo


class MyTest(rfm.RegressionTest):
    use_fixture('repo', clone_repo('url'))

    def __init__(self):
        self.valid_systems = ['dom:gpu', 'dom:mc']
        self.valid_prog_environs = ['gnu', 'cray']
        self.sourcesdir = self.repo
        ...

```

Fixtures can be quite intelligent so as not to clone the same repo again and do some reference counting before deleting paths.

In this example, `MyTest` will clone the test once for all the test cases of `MyTest` without having to rely on dependencies for this sort of task.

akesandgren:
What about a completely different test (in another dir) wants to use the same downloaded repo? would this handle that too?
I.e., tests/testA/testA.py uses repo x, tests/testB/testB.py also uses repo x, will repo x stay available once downloaded or will it eventually be cleaned up after A or B has been finished (with all of their dependencies)?

jjotero:
[...]

PR #1880, opened 2021-03-23
[test] Add PrgEnv checks for HIP

jjotero:
This PR adds a test that builds the HIP ecosystem around the existing `nvcc` compiler, and another hello world test that checks that the build was successful.

The second test makes obvious the need for the changes to be introduced with #1868. However, the dependency of the hello world test on the build test would be better explained through fixtures (#1577).


cscs-checks/prgenv/hip/build_hip.py
```diff
@@ -0,0 +1,80 @@
...
+    maintainers = ['JO']
+    tags = {'production'}
+
+    def __init__(self):
+        # FIXME: PR #1868 should move this into the class body.
+        self.depends_on('BuildHip')
+
+    @rfm.require_deps
+    def get_hip_path(self, BuildHip):
+        self.hip_path = BuildHip().hip_full_path
```
vkarak, line 71: I guess that with fixtures, you would need to dump that in a file, right?

jjotero: Do you mean the `BuildHip` test? I guess that's up to the implementation. In any case, the `HelloHip` test should have a handle to `BuildHip`, so that the `BuildHip` data can be access naturally such as in purest fixtures (e.g. `BuildHip.hip_full_path`)

cscs-checks/prgenv/hip/build_hip.py
```diff
@@ -0,0 +1,80 @@
...
+    tags = {'production'}
+
+    @rfm.run_before('compile')
+    def set_compile_options(self):
+        self.hip_full_path = os.path.abspath(
+            os.path.join(self.stagedir, self.hip_path)
+        )
+        self.build_system.builddir = 'build'
+        self.build_system.config_opts = [
+            f'-DCMAKE_INSTALL_PREFIX={self.hip_full_path}',
```
victorusu, line 37: How do you plan to run this test? The `hip_full_path` doesn't point anywhere...

jjotero: What do you mean? `hip_full_path` is defined in line 32, no?

victorusu: Sorry, I missed it!

victorusu (approved):
`lgtm`

vkarak:
It seems that the test now fails on Dom.

jjotero:
It seems that the options for some env vars were updated during a minor release ... 😭

vkarak (approved):

vkarak:
@jenkins-cscs retry dom

vkarak (approved):

PR #1881, opened 2021-03-23
[bugfix] Fix the call order on the `__set_name__` hook for built-in variables

jjotero:
When declaring a variable with the `variable` built-in, the corresponding `__set_name__` hook was only called if needed by the `__getitem__` method of the metaclass. This is because the variables are temporarily hidden away in their own namespace, and the `__set_name__` hooks from these variables don't get called by the base metaclass `type`.

Hence, it is a bit counterintuitive (if not just plainly wrong) that things get set when calling the `__getitem__` method; so now the `__set_name__` hook is called unconditionally for all variables in the `__setitem__` method of the metaclass.

### DIFF
--- a/reframe/core/meta.py
+++ b/reframe/core/meta.py
@@ -82,9 +82,8 @@ class RegressionTestMeta(type):
                             if key in b._rfm_var_space:
                                 # Store a deep-copy of the variable's
                                 # value and return.
-                                self._namespace[key] = (
-                                    b._rfm_var_space[key].default_value
-                                )
+                                v = b._rfm_var_space[key].default_value
+                                self._namespace[key] = v
                                 return self._namespace[key]
 
                         # If 'key' is neither a variable nor a parameter,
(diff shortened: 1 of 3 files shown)


reframe/core/meta.py
```diff
@@ -81,12 +80,11 @@ def __getitem__(self, key):
...
                             if key in b._rfm_var_space:
-                                v = b._rfm_var_space[key]
-                                v.__set_name__(self, key)
-
                                 # Store a deep-copy of the variable's
                                 # value and return.
-                                self._namespace[key] = v.default_value
+                                self._namespace[key] = (
+                                    b._rfm_var_space[key].default_value
+                                )
```
vkarak, line 87: Just a style thing (unless it's again too long):

```suggestion
                                v = b._rfm_var_space[key].default_value
                                self._namespace[key] = v
```

vkarak (approved):

Issue #1887, opened 2021-03-24
error running tutorial: decorated test ('MyTest') is an abstract test


mabraham:
While exploring the tutorials, I tried to run the example

```
import reframe as rfm
import reframe.utility as util


@rfm.simple_test
class MyTest(rfm.RegressionTest):
    module_info = parameter(util.find_modules('GROMACS'))

    def __init__(self):
        s, e, m = self.module_info
        self.valid_systems = [s]
        self.valid_prog_environs = [e]
        self.modules = [m]
```
found at  https://reframe-hpc.readthedocs.io/en/v3.5.1/tutorial_advanced.html but I got an error:

```
$ reframe -c gromacs.py -r
[ReFrame Setup]
  version:           3.5.1
  command:           '/nethome/mabraham/anaconda3/envs/gmxbm/bin/reframe -c gromacs.py -r'
  launched by:       mabraham@dev-gpu06
  working directory: '/nethome/mabraham/amd-gpu-benchmarks/runs'
  settings file:     '<builtin>'
  check search path: '/nethome/mabraham/amd-gpu-benchmarks/runs/gromacs.py'
  stage directory:   '/nethome/mabraham/amd-gpu-benchmarks/runs/stage'
  output directory:  '/nethome/mabraham/amd-gpu-benchmarks/runs/output'

/nethome/mabraham/anaconda3/envs/gmxbm/bin/reframe: run session stopped: value error: ../../anaconda3/envs/gmxbm/lib/python3.9/site-packages/reframe/core/decorators.py:88: decorated test ('MyTest') is an abstract test
        raise ValueError(f'decorated test ({cls.__qualname__!r}) is an'
```

Did I do something wrong? How should I make the example work?

jjotero:
Hi @mabraham. My first guess is that `util.find_modules('GROMACS')` can't find any gromax modules. A test will not run if any of the parameters do not have any values.

mabraham:
Thanks for the rapid response! My environment does have `gromacs` modules, and when I change the tutorial to search for that with the correct case, I get the same problem:

```
$ module avail gromacs
----- /nethome/mabraham/amd-gpu-benchmarks/installed_binaries/modulefiles ------
gromacs/2020.3-amd-hip          gromacs/2020.3-clang-11-opencl  
gromacs/2020.3-clang-10-opencl  gromacs/2020.3-gcc-10-opencl    

--------------------------- /opt/tcbsys/modulefiles ----------------------------
[...]

PR #1888, opened 2021-03-24
[feat] Improve error message when parameters are undefined

jjotero:
When a test with one or more undefined parameters was decorated, it the decorator would raise an error saying that the test is an `abstract test`. This error message might not be very clear to the user, so it is now changed to something more intuitive.

The "abstract test" nomenclature has been kept in the unit tests to keep things short, and a definition has also been added there for clarity.


mabraham:
This would have helped me help myself better in #1887

vkarak (approved):

PR #1900, opened 2021-03-31
[bugfix] Fix bootstrap script for `pip`-less Python installations

vkarak:
Jumping from Python 3.6.10 to 3.6.12 on Dom has caused the bootstrap script to fail.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -70,15 +70,16 @@ fi
 # Check if ensurepip is installed
 $python -m ensurepip --version &> /dev/null
 
+export PATH=$(pwd)/external/usr/bin:$PATH
+
 # Install pip for Python 3
 if [ $? -eq 0 ]; then
-    CMD $python -m ensurepip --root external/ --default-pip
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -70,15 +70,16 @@ fi
...
 
+export PATH=$(pwd)/external/usr/bin:$PATH
+
 # Install pip for Python 3
 if [ $? -eq 0 ]; then
-    CMD $python -m ensurepip --root external/ --default-pip
+    CMD $python -m ensurepip --root $(pwd)/external/ --default-pip
 fi
 
 # ensurepip installs pip in `external/usr/` whereas the --target option installs
```
teojgo, line 80:
```suggestion
# ensurepip installs pip in `external/usr/` whereas the --root option installs
```

teojgo (changes requested):

teojgo (approved):

vkarak:
@jenkins-cscs retry dom

PR #1905, opened 2021-04-02
[packaging] Bump versions of requirements

teojgo:


requirements.txt
```diff
@@ -1,13 +1,13 @@
 argcomplete==1.12.2
-coverage==5.3
-importlib_metadata==2.0.0
+coverage==5.5
+importlib_metadata==3.10.0
```
vkarak, line 3: Add the Python constraint as in #1906.

boegel, line 3: `importlib-metadata` is only needed with older Python versions:

```suggestion
importlib_metadata==3.10.0; python_version < '3.8'
```

vkarak (changes requested):

vkarak (approved):

victorusu (approved):
`lgtm`

PR #1909, opened 2021-04-02
[ci] Pull base images for CI instead of building them from scratch

teojgo:


victorusu (approved):
I tested the Dockerfiles and they are awesome!

ci-scripts/dockerfiles/Lmod77_base.dockerfile
```diff
@@ -0,0 +1,39 @@
...
+  tar xzf lmod.tar.gz && \
+  cd Lmod-${_LMOD_VER} && \
+  ./configure && make install && \
+  cd .. && rm -rf lmod.tar.gz Lmod-${_LMOD_VER} && \
+  apt-get clean && \
+  rm -rf /var/lib/apt/lists/*
+
+ENV BASH_ENV=/usr/local/lmod/lmod/init/profile
+
+USER rfmuser
```
vkarak, line 39: I would rather create the `rfmuser` in the reframe images and not those that install the different modules systems. Also I prefer hyphens over underscores in filenames, so can you rename all the base images to `-base`?

vkarak (changes requested):

ci-scripts/dockerfiles/Lmod77_base.dockerfile
teojgo: This just changes the user from this step and beyond. Do you mean to move the actual user creation to the reframe images?

vkarak: > Do you mean to move the actual user creation to the reframe images?

Yes, I would say so, because otherwise you can't expand the image with things requiring elevated privileges, can you?

teojgo: No after changing the user you cannot install more things with `apt-get` for example.

vkarak: Yes, that's why I suggest moving this out from the base image.

vkarak (approved):
lgtm

PR #1910, opened 2021-04-02
[bugfix] Fix crash when `sourcesdir` is set to `None` in class body

### LINKED ISSUE #1869: ReFrame crashes if `sourcesdir` is set to `None` in class body
The following test fails:

```python
import reframe as rfm


    valid_systems = ['*']
    valid_prog_environs = ['*']
    sourcesdir = None
    executable = 'echo'

    @rfm.run_before('run')
    def set_exec_opts(self):
        self.executable_opts = [self.name]

    @rfm.run_before('sanity')
    def set_sanity_check(self):
        self.sanity_patterns = sn.assert_found(rf'{self.name}', self.stdout)


@rfm.simple_test
class T0(Base):
    pass
```

Stacktrace:

```
  File "/Users/karakasv/Repositories/reframe/reframe/core/decorators.py", line 56, in _instantiate_all
    ret.append(_instantiate(cls, args))
  File "/Users/karakasv/Repositories/reframe/reframe/core/decorators.py", line 43, in _instantiate
    return cls(_rfm_use_params=True)
  File "/Users/karakasv/Repositories/reframe/reframe/core/meta.py", line 199, in __call__
    obj = cls.__new__(cls, *args, **kwargs)
  File "/Users/karakasv/Repositories/reframe/reframe/core/pipeline.py", line 790, in __new__
    obj._rfm_init(name, prefix)
  File "/Users/karakasv/Repositories/reframe/reframe/core/pipeline.py", line 834, in _rfm_init
    if not os.path.isdir(os.path.join(self._prefix, self.sourcesdir)):
  File "/usr/local/Cellar/python@3.9/3.9.1_6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/posixpath.py", line 90, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/local/Cellar/python@3.9/3.9.1_6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/genericpath.py", line 152, in _check_arg_types
    raise TypeError(f'{funcname}() argument must be str, bytes, or '
TypeError: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'
```

The fix seems to be easy.

vkarak:
I've also changed the automatic assignment of `sourcesdir` to `None` so as to avoid setting it so, in cases that the user has explicitly passed a directory that does not exist. For this one, I'm taking advantage of the default value that is being set by the variables.

Fixes #1869.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -845,8 +845,8 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
 
         # Static directories of the regression check
         self._prefix = os.path.abspath(prefix)
-        if (not os.path.isdir(os.path.join(self._prefix, self.sourcesdir)) and
-            not osext.is_url(self.sourcesdir)):
+        if (self.sourcesdir == 'src' and
+            not os.path.isdir(os.path.join(self._prefix, self.sourcesdir))):
             self.sourcesdir = None
 
         # Runtime information of the test
--- a/unittests/test_pipeline.py
+++ b/unittests/test_pipeline.py
@@ -272,6 +272,20 @@ def test_run_only_no_srcdir(local_exec_ctx):
     _run(test, *local_exec_ctx)
 
 
+def test_run_only_srcdir_set_to_none(local_exec_ctx):
+    @fixtures.custom_prefix('foo/bar/')
+    class MyTest(rfm.RunOnlyRegressionTest):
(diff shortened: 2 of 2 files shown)


vkarak:
@jenkins-cscs retry daint

jjotero (approved):
lgtm

Issue #1730, opened 2021-02-01
Abbreviate host lists when reporting node lists


vkarak:
This is a feature request that has come up from Slack discussions. A candidate syntax is that of Slurm and other tools.

- See also: https://www.nsc.liu.se/~kent/python-hostlist/

PR #1912, opened 2021-04-06
[feat] Abbreviate node lists in `FAILURE INFO` reports

jgphpc:
Closes https://github.com/eth-cscs/reframe/issues/1730

EDIT vkarak: This PR adds an algorithm for compressing node lists and returning a a condensed string representation for them. This abbreviated form of node lists is used in `FAILURE INFO` reports, but not in the full JSON run report. The reason for this is that the JSON report is meant as raw report info that other tools can process, thus it makes more sense imho not to abbreviate the node lists there.


jgphpc:
note: we could prefer hostlist format over nodelist and/or switch to hostlist if len(nodelist) > 1000 ?

reframe/core/schedulers/__init__.py
```diff
@@ -326,6 +327,18 @@ def nodelist(self):
...
+    def hostlist(self):
+        '''The hostlist of this job.
+
+        The value of this field is scheduler-specific.
+
+        .. versionadded:: 3.6
+
+        :type: :class`str` or :class:`None`
+        '''
+        return self._hostlist
```
victorusu, line 340: Is it guaranteed that the `self._hostlist` is in the preferred format?

reframe/core/schedulers/__init__.py
```diff
@@ -326,6 +327,18 @@ def nodelist(self):
         '''
         return self._nodelist
 
+    @property
+    def hostlist(self):
+        '''The hostlist of this job.
+
+        The value of this field is scheduler-specific.
+
+        .. versionadded:: 3.6
```
vkarak, line 336: Version strings should be conformant to version semantic syntax, i.e., this one should be 3.6.0.

vkarak (changes requested):
This solution is entirely Slurm specific. I would prefer that we converted any host list to an abbreviated sequence ourselves. Algorithmically, it should not be difficult as soon as we sort the nodes. It's practically a run-length encoding of the node ids. As for the node name format, we can assume a generic pattern that ends in a sequence of consecutive numbers.

Also we don't need command line options etc. for this. We only need a configuration parameter and an associated environment variable `RFM_ABBREV_HOSTLIST=<n>`. If `<n>` is zero then we don't abbreviate, otherwise we abbreviate any list with size `>= n`.

vkarak:
@jgphpc I took care of the algorithm and it works nicely now. Can you address the rest of the comments?

> Also we don't need command line options etc. for this. We only need a configuration parameter and an associated environment variable RFM_ABBREV_NODELIST=<n>. If <n> is zero then we don't abbreviate, otherwise we abbreviate any node list with size >= n.

Also I don't think that this conversion should be done at the backends, so all the change in the scheduler backends should be reverted. This conversion is purely a presentation thing, so it has to go into the frontend, when we generate the final report.

unittests/test_utility.py
```diff
@@ -1712,3 +1712,42 @@ def foo():
...
+    assert nodelist(nid_nodes) == 'nid00[0-4],nid01[0-4],nid02[0-4]'
+    assert nodelist(['nid01', 'nid10', 'nid20']) == 'nid01,nid10,nid20'
+    assert nodelist([]) == ''
+    assert nodelist(['nid001']) == 'nid001'
+
+    with pytest.raises(TypeError, match='nodes argument must be a Sequence'):
+        nodelist(1)
+
+    with pytest.raises(TypeError, match='nodes argument cannot be a string'):
+        nodelist('foo')
```
vkarak, line 1753: We also need to add a test for node duplicates, which actually we don't address. I don't even know the outcome :-D

jjotero: Is such thing even possible? 😱

jjotero: I just tested this and the outcome is that it doesn't compress the duplicate nodes. I think that's fair enough!

jjotero (approved):
Looks good!

victorusu:
@vkarak, if I understand the last changes correctly, they imply that we will always use the abbreviated node list. Is it what we want?

vkarak:
@victorusu Check the modified description of the PR. Yes, in `FAILURE INFO` there is no reason to use the non-abbreviated form. Conversely, the JSON report contains the full node list.

reframe/frontend/statistics.py
```diff
@@ -216,8 +217,9 @@ def print_failure_report(self, printer):
             printer.info(f"  * System partition: {r['system']}")
             printer.info(f"  * Environment: {r['environment']}")
             printer.info(f"  * Stage directory: {r['stagedir']}")
-            nodelist = ','.join(r['nodelist']) if r['nodelist'] else None
-            printer.info(f"  * Node list: {nodelist}")
+            printer.info(
+                f"  * Node list: {util.nodelist_abbrev(r['nodelist'])}"
```
victorusu, line 221: Is it guaranteed that the `r['nodelist']` will always contain an entry? Because there are assertion points down the road.
Previously there a check whether the `r['nodelist']` was not empty.

vkarak (approved):

victorusu (approved):
`lgtm`. I just have one question related to whether the `r['nodelist']` array is *always* populated, because it used to have a check if it was empty or not. There are some assertions requesting that it has at least one entry. So, it got me confused a bit.

reframe/frontend/statistics.py
vkarak: Good point, I'll have a look, but I think that it has to be a list according to the JSON schedma.

vkarak: It is guaranteed to be a list. Check here:

https://github.com/eth-cscs/reframe/blob/e8a81b3191b2443973562c621cd432611656659a/reframe/frontend/statistics.py#L114

And here

https://github.com/eth-cscs/reframe/blob/e8a81b3191b2443973562c621cd432611656659a/reframe/frontend/statistics.py#L142

And nodelist must a be a list in the JSON report:

https://github.com/eth-cscs/reframe/blob/e8a81b3191b2443973562c621cd432611656659a/reframe/schemas/runreport.json#L50-L52

PR #1923, opened 2021-04-08
[refactor] Consolidate unit test fixtures and other refactoring of unit tests

### LINKED ISSUE #1551: Consolidate the different unit test fixtures
Currently, there is some duplication of some common fixtures, such as the `temp_runtime()`. We need to check if it is possible to move them in a common place shared by all unit tests.

### LINKED ISSUE #1913: Fix unit test fixture to create a temporary runtime in unit tests
The way it is implemented in the different unit tests is buggy. The correct implementation is now in `test_policies.py`. This should be propagated to other tests or be part of a fixture library as described in #1551.

vkarak:
- The `temp_runtime` fixture that was defined in different files is now replaced by the more general `make_exec_ctx` fixture and defined inside `unittests.fixtures`. All unit tests that were making use of the old fixtures are now updated.
- The `unittests.fixtures` module now contains only pytest fixtures. The rest utilities are moved to a separate module `unittests.utility`.
- Linted the unit tests.

Fixes #1913.
Fixes #1551.


teojgo:
You can also take a look at `conftest.py` which provides a way of sharing fixtures across multiple files: https://docs.pytest.org/en/stable/fixture.html#conftest-py-sharing-fixtures-across-multiple-files

teojgo (approved):

jjotero (approved):

vkarak:
@jenkins-cscs retry daint

Issue #651, opened 2019-01-18
Generate report in XML JUnit format


teojgo:
This feature would allow ReFrame to generate an xml report in JUnit format used by Jenkins. The schema for the xml can be found [here](https://github.com/junit-team/junit5/blob/master/platform-tests/src/test/resources/jenkins-junit.xsd). A command line option which accepts the name of the output file should be added to make ReFrame generate the report after running the tests.

vkarak:
Regarding the command line options, I think we can have two options: one for specifying the output type, e.g., (Junit, standard ReFrame report) and one for specifying the output file. The latter could be technically a different issue, because it could also replace the `reframe.out` file that we now specify in the logging settings.

vkarak:
Now, that I'm thinking it again. This option cannot replace `reframe.out`. It is specifically for the report not the output.

vkarak:
I've created issue #653 for creating the corresponding command line options.

vkarak:
I'm raising the priority of this issue, because this feature would also be useful for reporting within Gitlab pipelines. I think we should pursue it.

PR #1925, opened 2021-04-09
[feat] Add command-line option to generate a JUnit XML report

jgphpc:
Closes https://github.com/eth-cscs/reframe/issues/651

Example in https://git.cscs.ch/c4k/c4kcust/example-junit/-/pipelines/23226/test_report


reframe/frontend/cli.py
```diff
@@ -26,6 +26,7 @@
 import reframe.frontend.filters as filters
 import reframe.frontend.runreport as runreport
 import reframe.utility.jsonext as jsonext
+# import reframe.utility.xmlext as xmlext
```
teojgo, line 29: Remove this line

reframe/frontend/statistics.py
```diff
@@ -5,6 +5,7 @@
 
 import inspect
 import traceback
+import xml.etree.ElementTree as ET
```
teojgo, line 8: Add an empty line after this.

teojgo (changes requested):
I would say to either have a separate option for the xml report or choose the format based on the extension of the filename passed by the user. Also, it could be nice if we could somehow pretty print the xml with proper indentation.

jgphpc:
> I would say to either have a separate option for the xml report or choose the format based on the extension of the filename passed by the user. 

- I would suggest `--junit-xml` to be consistent with pytest
- what if you want both ?

teojgo:
> > I would say to either have a separate option for the xml report or choose the format based on the extension of the filename passed by the user.
> 
> * I would suggest `--junit-xml` to be consistent with pytest
> * what if you want both ?

Ok, `--junit-xml` sounds good to me.

reframe/frontend/statistics.py
```diff
@@ -197,6 +199,41 @@ def json(self, force=False):
...
+        # https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd
+        xml_testsuites = ET.Element('testsuites')
+        xml_testsuite = ET.SubElement(
+            xml_testsuites, 'testsuite',
+            attrib={
+                'name': 'rfm',
+                'errors': '0',
+                'failures': str(json_report['session_info']['num_failures']),
+                'tests': str(json_report['session_info']['num_cases']),
+                'time': str(json_report['session_info']['time_elapsed']),
```
victorusu, line 212: Does it make sense to add additional fields, like the `hostname` the reframe invocation command? Or other stuff?

victorusu (commented):
@jgphpc, I am sure if this is possible, I am not familiar with JUnit, but I wonder if it makes sense and whether if this is also possible to add additional information to the JUnit... It would be nice to know the hostname. The schema that you shared in the code (https://github.com/windyroad/JUnit-Schema/blob/162a883ac631f61ca1d14a3d472e1fbb53c1c05f/JUnit.xsd#L171) shows a hostname entry. 
But I wonder if we can also add additional fields...

What do you think?

reframe/frontend/cli.py
```diff
@@ -1021,6 +1021,18 @@ def module_unuse(*paths):
                     f'failed to generate report in {report_file!r}: {e}'
                 )
 
+            # Build xml report
+            xml_data = runner.stats.junit(json_report).decode()
+            xml_report_file = os.path.splitext(report_file)[0]+'.xml'
```
jgphpc, line 1026: @teojgo would this be possible (if yes, how) ?
- `--junit-xml` then xml_rpt_file=`'rpt_file'.xml`
- `--junit-xml=0.xml`  then xml_rpt_file=`0.xml`
- no `--junit-xml` then no xml_rpt_file

vkarak: I have some comment about the options. I am not entirely in favour of these options as such. I'll get back to it later.

reframe/frontend/statistics.py
jgphpc: https://github.com/eth-cscs/reframe/pull/1925/commits/75f1bb4856604d0cda89e4eb9db259599e44ed25

reframe/frontend/cli.py
vkarak: Why not something like `--report-junit[=FILE]` very similar to what we do with the `--report-file`? We should also support the exact same syntax in this case, as we do with the report file. Check [here](https://reframe-hpc.readthedocs.io/en/stable/manpage.html?highlight=--report-file#cmdoption-report-file).

jgphpc: https://github.com/eth-cscs/reframe/pull/1925/commits/286cc91447a985e497ced5e53fbde52ef147e975

vkarak:
> Example in https://git.cscs.ch/c4k/c4kcust/example-junit/-/pipelines/23226/test_report

@jgphpc Is this report generated from this PR?

reframe/frontend/statistics.py
```diff
@@ -197,6 +199,42 @@ def json(self, force=False):
...
+                    'time': str(tid['time_total']),
+                }
+            )
+            if not tid['result'] == 'success':
+                testcase_msg = ET.SubElement(
+                    testcase, 'failure', attrib={'message': tid['fail_phase']}
+                )
+                testcase_msg.text = tid['fail_reason']
+
+        return ET.tostring(xml_testsuites, encoding='utf8', method='xml')
```
vkarak, line 236: Can't the output be formatted in multiple lines instead of everything in a single line? PS: I'm just judging from the example XML file you've posted in the description of this PR.

reframe/frontend/statistics.py
```diff
@@ -197,6 +199,42 @@ def json(self, force=False):
...
+                'name': 'rfm',
+                'errors': '0',
+                'failures': str(json_report['session_info']['num_failures']),
+                'tests': str(json_report['session_info']['num_cases']),
+                'time': str(json_report['session_info']['time_elapsed']),
+                'hostname': json_report['session_info']['hostname'],
+            }
+        )
+
+        for testid in range(len(json_report['runs'][0]['testcases'])):
```
vkarak, line 217: Here you are just taking the first run. How are we going to deal with the retries? Perhaps, we can completely ignore them since their future is not sure, but we should somehow mention it or issue a warning if the `--report-junit` option is passed along with `--max-retries`.

reframe/frontend/statistics.py
```diff
@@ -197,6 +199,42 @@ def json(self, force=False):
...
+            tid = json_report['runs'][0]['testcases'][testid]
+            name = (
+                f"{tid['name']} on {tid['system']} using {tid['environment']}"
+            )
+            testcase = ET.SubElement(
+                xml_testsuite, 'testcase',
+                attrib={
+                    'classname': tid['filename'],
```
vkarak, line 225: Better use our test name here.

reframe/frontend/statistics.py
```diff
@@ -197,6 +199,42 @@ def json(self, force=False):
...
+                'time': str(json_report['session_info']['time_elapsed']),
+                'hostname': json_report['session_info']['hostname'],
+            }
+        )
+
+        for testid in range(len(json_report['runs'][0]['testcases'])):
+            tid = json_report['runs'][0]['testcases'][testid]
+            name = (
+                f"{tid['name']} on {tid['system']} using {tid['environment']}"
+            )
```
vkarak, line 221: I would prefer a less verbose text, e.g., `MyTest[dom:gpu, PrgEnv-gnu]`

vkarak: Also name the variable differently, because it's confusing: `casename` perhaps.

reframe/frontend/statistics.py
```diff
@@ -197,6 +199,42 @@ def json(self, force=False):
...
+            testcase = ET.SubElement(
+                xml_testsuite, 'testcase',
+                attrib={
+                    'classname': tid['filename'],
+                    'name': name,
+                    'time': str(tid['time_total']),
+                }
+            )
+            if not tid['result'] == 'success':
```
vkarak, line 230:
```suggestion
            if tid['result'] == 'failure':
```

And I'm not sure if we can report `error`. I think yes, in which case you should create a separate entry afaics in the schema.

reframe/frontend/statistics.py
```diff
@@ -197,6 +199,42 @@ def json(self, force=False):
...
+                testcase_msg = ET.SubElement(
+                    testcase, 'failure', attrib={'message': tid['fail_phase']}
```
vkarak, line 232: What would happen if we used `type` instead of `message` here? Just asking.

vkarak (changes requested):

vkarak:
Also, we will need a unit test. Ideally, the unit test would be validating the generated XML against the schema. I think for validation we might need the [lxml](https://lxml.de/) library, which also support the `etree` that you are using. So we could move everything over to this (it looks like an import change, but I'm not sure).

jgphpc:
> @jgphpc Is this report generated from this PR?

I updated it: https://git.cscs.ch/c4k/c4kcust/example-junit/-/pipelines/23916/test_report

reframe/frontend/statistics.py
```diff
@@ -264,8 +302,8 @@ def print_failure_stats(self, printer):
         stats_header = row_format.format('Phase', '#', 'Failing test cases')
         num_tests = len(self.tasks(current_run))
         num_failures = 0
-        for l in failures.values():
-            num_failures += len(l)
+        for ll in failures.values():
+            num_failures += len(ll)
```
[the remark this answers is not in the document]
jgphpc: pycodestyle gave me a warning.

reframe/frontend/statistics.py
jgphpc: If you click on `View details` in [c4k](https://git.cscs.ch/c4k/c4kcust/example-junit/-/pipelines/23916/test_report), it shows:

> sanity error: pattern 'x' not found in 'rfm_V100_Test_1_job.out'

Do you want more ?

reframe/frontend/statistics.py
jgphpc: I believe the schema expects `message`, see [L82](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L82)

reframe/frontend/statistics.py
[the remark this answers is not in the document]
vkarak: Yes, because you should ignore E741 as it says in the [coding style](https://github.com/eth-cscs/reframe/wiki/coding-style-guide).

docs/manpage.rst
```diff
@@ -276,7 +276,7 @@ Options controlling ReFrame output
    This option can also be set using the :envvar:`RFM_REPORT_JUNIT` environment
    variable or the :js:attr:`report_junit` general configuration parameter.
 
-   .. versionadded:: 3.1
+   .. versionadded:: 3.6
```
vkarak, line 279:
```suggestion
   .. versionadded:: 3.6.0
```

reframe/schemas/config.json
```diff
@@ -487,7 +487,7 @@
         "general/non_default_craype": false,
         "general/purge_environment": false,
         "general/report_file": "${HOME}/.reframe/reports/run-report.json",
-        "general/report_junit": "${HOME}/.reframe/reports/run-report.xml",
+        "general/report_junit": "",
```
vkarak, line 490: Why not `none`?

reframe/frontend/statistics.py
vkarak: Yes, I can see this, but I would like to see other type of failures as well: build failures, performance failures. How do they show up?

vkarak (changes requested):
I think the unit tests are overly complicated. Imho, it would be enough to generate the XML file and simply validate it against the schema here: https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd

Check the "XML Schema" section [here](https://lxml.de/validation.html) on how to do it. Also we don't need a separate test for this. It can go in the `test_runall()` of `test_policies.py`:

https://github.com/eth-cscs/reframe/blob/45fbacb23210c757724882a13c4a53f33af04800/unittests/test_policies.py#L208

There we also do the same thing for the JSON report, which we validate it against our report schema.

vkarak (changes requested):
I updated the PR by fine tuning the implementation. It's a bit simpler now. Nonetheless, the generated XML does not validate against the schema. See the message of the failing unit tests. If we fix that and finalise the actual information of the output, it'll be ready to be merged.

requirements.txt
```diff
@@ -1,7 +1,9 @@
 argcomplete==1.12.2
 coverage==5.5
+dateutil
```
vkarak, line 3: I don't understand why is this needed.

unittests/test_policies.py
```diff
@@ -226,6 +237,11 @@ def test_runall(make_runner, make_cases, common_exec_ctx, tmp_path):
     with open(report_file, 'w') as fp:
         jsonext.dump(report, fp)
 
+    # Validate the junit report
+    xml_report = runreport.junit_xml_report(report)
+    # print(f'xml_report={xml_report}')
```
vkarak, line 242: This should be removed.

unittests/test_policies.py
```diff
@@ -466,7 +482,7 @@ def assert_dependency_run(runner):
     assert_runall(runner)
     stats = runner.stats
     assert 10 == stats.num_cases(0)
-    assert 4  == len(stats.failed())
+    assert 4 == len(stats.failed())
```
vkarak, line 485: Revert the formatting :-) I think you should tell your editor to ignore E221 and W504.

reframe/frontend/runreport.py
```diff
@@ -185,10 +192,41 @@ def junit_xml_report(json_report):
...
+    # testcase_skip_msg = etree.SubElement(
+    #     testcase_skip, 'skipped',
+    #     attrib={'message': 'no test skipped', 'type': 'skipped'}
+    # )
+    # testcase_skip_msg.text = 'S'
+
+    # debug_str = etree.tostring(xml_testsuites, encoding='utf8',
+    #                            pretty_print=True, method='xml',
+    #                            xml_declaration=True).decode()
+    # print(f'xml_testsuites={debug_str}')
```
vkarak, line 229: Remember to remove those before merging.

reframe/frontend/runreport.py
```diff
@@ -158,18 +159,24 @@ def junit_xml_report(json_report):
     '''Generate a JUnit report from a standard ReFrame JSON report.'''
 
     xml_testsuites = etree.Element('testsuites')
+    isotime = parser.parse(json_report['session_info']['time_start'][:-6])
```
vkarak, line 162: I don't think that this is needed. You should be able to do this with the standard library of Python.

reframe/frontend/runreport.py
```diff
@@ -158,18 +159,24 @@ def junit_xml_report(json_report):
...
             'errors': '0',
             'failures': str(json_report['session_info']['num_failures']),
+            'hostname': json_report['session_info']['hostname'],
+            'id': '0',
+            'name': 'reframe',
+            'package': 'reframe',
             'tests': str(json_report['session_info']['num_cases']),
             'time': str(json_report['session_info']['time_elapsed']),
-            'hostname': json_report['session_info']['hostname'],
+            'timestamp': isotime.isoformat(),
```
vkarak, line 174: The timestamps of the JSON report are already in ISO format.

reframe/schemas/junit.xsd
```diff
@@ -5,7 +5,7 @@
 	 attributeFormDefault="unqualified">
 	<xs:annotation>
 		<xs:documentation xml:lang="en">JUnit test result schema for the Apache Ant JUnit and JUnitReport tasks
-Copyright © 2011, Windy Road Technology Pty. Limited
+Copyright 2011, Windy Road Technology Pty. Limited
```
vkarak, line 8: I don't like very much that we need to change this. What type of error where you getting?

reframe/frontend/runreport.py
```diff
@@ -171,7 +170,7 @@ def junit_xml_report(json_report):
             'package': 'reframe',
             'tests': str(json_report['session_info']['num_cases']),
             'time': str(json_report['session_info']['time_elapsed']),
-            'timestamp': isotime.isoformat(),
+            'timestamp': json_report['session_info']['time_start'][:-5],
```
vkarak, line 173: I guess it doesn't like the timezone? We should put a short comment then explaining this.

reframe/frontend/runreport.py
```diff
@@ -187,7 +186,7 @@ def junit_xml_report(json_report):
             attrib={
                 'classname': tid['filename'],
                 'name': casename,
-                'time': str(tid['time_total']),
+                'time': str(decimal.Decimal(tid['time_total'])),
```
vkarak, line 189: What was the problem here?

jgphpc: I got UnicodeDecodeError, let me retry.

reframe/schemas/junit.xsd
jgphpc: I got UnicodeDecodeError, let me retry.

jgphpc: can not reproduce it, reverting to original: 42285eb

reframe/frontend/runreport.py
jgphpc: The error message is:
```
E   AssertionError: Element 'testsuite', attribute 'timestamp': '2021-05-01T12:09:31+0000' is not a valid value of the atomic type 'ISO8601_DATETIME_PATTERN'.
```

reframe/frontend/runreport.py
jgphpc: The error message is:
```
E   AssertionError: Element 'testcase', attribute 'time': '8.130073547363281e-05' is not a valid value of the atomic type 'xs:decimal'.
```

vkarak: Perhaps, you can add a small comment explaining why we have to use `Decimal`.

reframe/schemas/junit.xsd
vkarak: Yes, because your locale could have been the ANSI C. But to avoid such failures, you can force the encoding in `open()` by passing `encoding='utf-8'`.

reframe/frontend/statistics.py
vkarak: For this one we should add a note in the command line option description for the report.

reframe/frontend/statistics.py
vkarak: And this one we need to verify it before merging.

vkarak: I checked this, we have "success," "failure" and "aborted". So the check about "failure" here is correct.

docs/manpage.rst
```diff
@@ -274,6 +274,17 @@ Options controlling ReFrame output
    .. versionadded:: 3.1
 
 
+.. option:: --report-junit=FILE
+
+   Instruct ReFrame to generate a JUnit XML report in ``FILE``.
+   The ``FILE`` argument may contain the special placeholder ``{sessionid}``, in which case ReFrame will generate a new report each time it is run by appending a counter to the report file.
```
vkarak, line 280: This is not valid; I will remove it.

vkarak (approved):

PR #1929, opened 2021-04-09
[test] Factor out `cscs-checks` into a reusable library of tests

jjotero:
A few tests have been ported into a library style format. These are under the `library` directory.

A symlink to `cscs-checks` has been created as `cscslib` to work around python module import errors.


cscs-checks/microbenchmarks/gpu/hooks.py
```diff
@@ -0,0 +1,55 @@
+
+import reframe as rfm
+
+
+class SetCompileOpts(rfm.RegressionMixin):
```
victorusu, line 5: I would call this hook something different, because - based on the name - I was  expecting compiler cli flags. In this case, it only sets the GPU related modules and set some GPU related cli flags. I think it should be called `SetArchandModules` or something related. What do you think?

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -8,123 +8,49 @@
 import reframe as rfm
 import reframe.utility.sanity as sn
 
+from library.microbenchmarks.gpu.gpu_burn import GpuBurnBase
+import cscslib.microbenchmarks.gpu.hooks as hooks
```
victorusu, line 12: One question that the hook name rises is whether how one classify hooks. I am referring to classical distinction of  `hooks.gpu` vs `gpu.hooks`... What do you think?

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
```diff
@@ -6,163 +6,122 @@
...
-@rfm.simple_test
-class GpuBandwidthCheck(rfm.RegressionTest):
-    def __init__(self):
-        self.valid_systems = ['daint:gpu', 'dom:gpu',
-                              'arolla:cn', 'tsa:cn',
-                              'ault:amdv100', 'ault:intelv100',
-                              'ault:amda100', 'ault:amdvega']
-        self.valid_prog_environs = ['PrgEnv-gnu']
+
+class PrgEnvMixin(rfm.RegressionMixin):
```
victorusu, line 13: The name of this mixing is not very generic. I wonder if it makes sense to leave it like this, or set more specific names and have the mixings be reused by other library classes too.
Because the name `PrgEnvMixin` tells me nothing and this one is very, very specific for Tsa and Arolla. But I wonder how this particular one will evolve!

What do you think? Do you think that this mixing will have additional functions in the future that can be generically used by other classes or that this mixing is meant to be used just to fix the Tsa and Arolla programming environment definitions?

victorusu (changes requested):
@jjotero, I have a few questions! But I really, loved this PR! The libraries will rock! 😂 
I am struggling with the evolution and reuse of the mixing and hooks.
Of course, this is because we haven't used them yet and this is the first PR of a very long series of PRs... 😂 

But I wonder if we can solve some of this questions before we start making additional PRs... What do you think?

cscs-checks/microbenchmarks/gpu/hooks.py
jjotero: Another thing that bothers me a bit is having the hooks in separate classes. I'm not sure if it would make sense to stash all of them into the same class.

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
jjotero: I have thought of having site-wide hooks that could be used for all the tests that we run at CSCS. The issue with this is that it'd be rather difficult to have a Mixin that all tests are happy with. We might need to structure the Mixin classes into their own hierarchies as well.

Regarding the naming of this mixin, so far this is the only test that uses it, so I'm not sure it makes much difference to change it now.

jjotero (commented):
@victorusu I think we should really crunch all those questions in this PR and think of a plan to move forward with this. Doing changes further along the line will result into a lot more work I reckon. So better get it out of the way now.

The hooks and the mixin classes are also something I'm not entirely convinced with. I find it a bit hard to keep track of the hooks once you import them from somewhere else.

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
victorusu: I would still change the name to make it more meaningful. This also helps in understanding what the Check is depending on. It is much easier to read/reason about a test that inherits from `TsaArollaNoMPIPrgEnvMixin`, then from just `PrgEnvMixin`. Even though the name that I propose is not very good! 🤔

cscs-checks/microbenchmarks/gpu/pointer_chase/pointer_chase.py
```diff
@@ -3,16 +3,18 @@
...
+
 import reframe.utility.typecheck as typ
 import reframe as rfm
 
-import os
+import library.microbenchmarks.gpu.pointer_chase as pchase
+import cscslib.microbenchmarks.gpu.hooks as hooks
 
 
 class PchaseGlobal(rfm.RegressionMixin):
```
victorusu, line 15: I am really struggling with the name choices... `PchaseGlobal`... Does it make sense to have it more specific?

cscs-checks/microbenchmarks/gpu/hooks.py
victorusu: In this case, I vote for what you have implemented. Stashing everything into a single class hinders the code reuse

jjotero: Perhaps we need to think about some mechanism that makes this hook reuse more explicit on the class that uses those hooks. I'm not sure if it would be good to establish a convention where we don't decorate the hooks on base classes. For example, here `SetCompileOpts` would have the member functions undecorated, and then the tests that use these class would explicitly decorate the hooks there ... but I'm not convinced of that approach either 😅

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
jjotero: I agree. But I think it's not such a big deal here because everything is on this same file, and also this is our final specialisation of the test. In any case, I don't mind changing this.

cscs-checks/microbenchmarks/gpu/pointer_chase/pointer_chase.py
jjotero: Yeah, not the best.  Let me change this to something better. I'll also change the imports to avoid having the `chase. ...` and `hooks. ...`

jjotero: Changed to `BaseCSCS`

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
jjotero: Changed this one to `BaseCSCS`

jjotero:
I have updated the way the hooks are shared. Instead of being declared in a `rfm.RegressionMixin` class, the hooks are declared as free unary functions, and they're then bound into the class that uses them with the `run_after` and `run_before` decorators. 

Perhaps it makes sense now to have a wrapper on top of these decorator such that someone could do
```python
   hook_a = rfm.inject_before('compile', foo) 
   # Instead of
   # hook_a = rfm.run_before('compile')(foo)

   hook_b = rfm.inject_after('run', bar)
   # Instead of
   # hook_b = rfm.run_after('run')(bar)
```
Or perhaps not even have these as decorators and make them available through the metaclass .... #DisruptiveChange 😂

jjotero:
@jenkins-cscs retry dom

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+     - temp: the GPU's temperature in degC after the run.
+
+    '''
+
+    # GPU build variables
+    gpu_build = variable(str)
+    gpu_arch = variable(str, type(None), value=None)
+
+    # Required variables
+    executable_opts = required
```
vkarak, line 43: This must be documented in a library. An example of options would help.

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    num_tasks = required
+
+    descr = 'GPU burn test'
+    build_system = 'Make'
+    executable = './gpu_burn.x'
+    num_tasks_per_node = 1
```
vkarak, line 49: This test should set or require the `num_gpus_per_node`. When I run a simple flexible test, I get a sanity error, because `num_tasks_assigned` is 0. If I set `num_gpus_per_node`, then it passes.

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+    The variables required by the run stage are:
+     - executable_opts: Defines the options to be passed to the executable.
+     - num_tasks: Number of tasks to use for this test.
+
+    -- Sanity --
+    Checks that the output matches the number of num_gpus_per_node*num_tasks.
+
+    -- Performance --
+    The performance patterns are:
+     - perf: Gflops/s achieved on average for the duration of the run.
```
vkarak, line 33: The minimum from all nodes is reported.

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+     - temp: the GPU's temperature in degC after the run.
```
vkarak, line 34: The maximum of all nodes.

vkarak: Performance library tests should also define a dummy reference just to set the units. This will cause ReFrame to automatically show the units in its run report as well as with the `--performance-report` option. You can achieve this by simply setting:

```python
reference = {
    '*': {
        'perf': (0, None, None, 'Gflop/s'),
        'temp': (0, None, None, 'degC')
    }
}
```

vkarak: I would even add this always as a pre-performance hook. This will allow users to define references partially. Otherwise, ReFrame complains about unresolved references.

vkarak (changes requested):
I like the idea and the tests really work as a library. Here is an example I wrote using the `GpuBurnBase`:

```python
import reframe as rfm
from library.microbenchmarks.gpu.gpu_burn import GpuBurnBase


@rfm.simple_test
class gpu_burn_test(GpuBurnBase):
    valid_systems = ['dom:gpu']
    gpu_build = 'cuda'
    modules = ['craype-accel-nvidia60', 'cdt-cuda']
    executable_opts = ['-d', '20']
    num_tasks = 0
    num_gpus_per_node = 1
    reference = {
        'dom:gpu': {'perf': (4100, -0.05, None, 'Gflop/s')}
    }
```

Which is really simple and straightforward. One important thing with the library is proper documentation of test variables and what the test does or assumes.

One thing that we should also probably address is how to deal with tests that are valid for multiple programming environments. Imagine a library of the "Hello, World" tests. Each test can be compiled with different compilers. Do the tests need to define `valid_prog_environs` assuming a naming convention about the environments?

library/microbenchmarks/gpu/gpu_burn/__init__.py
jjotero: I think I like more the option of setting it into the class body. Setting the `references` with a hook would prevent any of the derived classes from setting the references directly in the class body; unless of course you disable the hook. But that just seems like a lot more work than trying to reuse these 2 lines of code from the dummy references 😂

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class GpuBurnBase(rfm.RegressionTest, pin_prefix=True):
+    '''Base class for the GPU Burn test.
+
+    -- Compile --
+    The tests sources can be compiled for both CUDA and HIP. This is set with
+    the `gpu_build` variable, which must be set by a derived class. This
+    source code can also be compiled for a specific device architecture by
+    defining the `gpu_arch` variable.
```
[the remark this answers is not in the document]
jjotero: I've added a sentence here saying that this can be any of the supported architecture codes from NVIDIA or AMD.

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+                self.build_system.cxxflags = [f'-arch=compute_{self.gpu_arch}',
+                                              f'-code=sm_{self.gpu_arch}']
+        elif self.gpu_build == 'hip':
+            self.build_system.makefile = 'makefile.hip'
+            if self.gpu_arch:
+                self.build_system.cxxflags = [
+                    f'--amdgpu-target={self.gpu_arch}'
+                ]
+        else:
+            raise ValueError('unknown gpu_build option')
```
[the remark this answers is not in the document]
jjotero: See comment above.

library/microbenchmarks/gpu/memory_bandwidth/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+import reframe.utility.sanity as sn
+import reframe as rfm
+
+
+__all__ = ['GpuBandwidthSingle', 'GpuBandwidthMulti']
+
+
+class GpuBandwidthBase(rfm.RegressionTest, pin_prefix=True):
```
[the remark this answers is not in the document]
jjotero: The `base` here is not public, so I guess it's fine. I'll remove it from the public ones.

I've changed the name to `GPU_bandwidth_base` in this class.

library/microbenchmarks/gpu/memory_bandwidth/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+    '''Base class to the gpu bandwidth test.
+
+    The source code can be compiled for both CUDA and HIP. This option must be
+    set by derived class defining the `gpu_build` variable. A specific target
```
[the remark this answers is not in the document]
jjotero: Moved these docs to right above the variable itself.

library/microbenchmarks/gpu/memory_bandwidth/__init__.py
```diff
@@ -0,0 +1,170 @@
...
+    architecture can be specified through the variable `gpu_arch`.
+
+    By default, these benchmarks have default values for the size of the copy
+    buffer (`copy_size`) and the number of times this copy is performed
+    (`num_copies`). These values can also be overriden by a derived class.
+    '''
+
+    # Default copy variables
+    copy_size = variable(int, value=1073741824)
+    num_copies = variable(int, value=20)
```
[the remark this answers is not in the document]
jjotero: Done for these and all the rest 👍

library/microbenchmarks/gpu/pointer_chase/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @rfm.run_before('sanity')
+    def set_sanity(self):
+        self.sanity_patterns = sn.assert_found(r'pChase.x', self.stdout)
+
+
+class RunGpuPChaseBase(rfm.RunOnlyRegressionTest, pin_prefix=True):
+    '''Base RunOnly class for the gpu pointer chase test.
+
+    This runs the pointer chase algo on the linked list from the code compiled
```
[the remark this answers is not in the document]
jjotero: Extended the doctoring on the public classes below.

cscs-checks/microbenchmarks/gpu/pointer_chase/pointer_chase.py
```diff
@@ -24,162 +26,47 @@ class PchaseGlobal(rfm.RegressionMixin):
...
+    set_gpu_arch = rfm.run_after('setup')(hooks.set_gpu_arch)
+    set_gpus_per_node = rfm.run_before('run')(hooks.set_gpus_per_node)
 
 
 @rfm.simple_test
-class CompileGpuPointerChase(rfm.CompileOnlyRegressionTest, PchaseGlobal):
+class CompileGpuPChase(pchase.BuildGpuPChaseBase, BaseCSCS):
+    ''' Build the executable.'''
+
     def __init__(self):
```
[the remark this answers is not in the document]
jjotero: I guess this piece of code was from before the post-init hooks time :) Changed them.

jjotero:
> One thing that we should also probably address is how to deal with tests that are valid for multiple programming environments. Imagine a library of the "Hello, World" tests. Each test can be compiled with different compilers. Do the tests need to define valid_prog_environs assuming a naming convention about the environments?

@vkarak Tests that live in libraries should not specify neither `valid_systems` nor `valid_prog_environs`. It's up to the derived classes to do that.

If we're happy with the hooks being free functions, I think we should merge this PR and then address a the issue of a nicer way to inject external hooks in a separate PR.

library/microbenchmarks/gpu/gpu_burn/__init__.py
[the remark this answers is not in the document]
vkarak: The only downside of this is the late failure. But it's an advantage in different contexts. So it's a matter of library design decision.

library/microbenchmarks/gpu/memory_bandwidth/__init__.py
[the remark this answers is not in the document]
vkarak: I would vote for `gpu_bandwidth_base`.

library/microbenchmarks/gpu/pointer_chase/__init__.py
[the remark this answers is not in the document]
vkarak: doctoring? 😂

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -8,11 +8,11 @@
 import reframe as rfm
 import reframe.utility.sanity as sn
 
-from library.microbenchmarks.gpu.gpu_burn import GpuBurnBase
+from library.microbenchmarks.gpu.gpu_burn import GPU_burn
```
vkarak, line 11: I would vote for all lowercase everywhere. It's easier to type in the command line.

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
```diff
@@ -5,23 +5,28 @@
 
 import reframe as rfm
 
-import library.microbenchmarks.gpu.memory_bandwidth as mb
 import cscslib.microbenchmarks.gpu.hooks as hooks
+from library.microbenchmarks.gpu.memory_bandwidth import *
 
 
-class BaseCSCS(rfm.RegressionMixin):
+class Base_CSCS(rfm.RegressionMixin):
```
vkarak, line 12: This name is too generic. I would make that simply private and name it anything. Also before changing the names, let's all settle on a style. I'm either all lowercase with underscores of camel. For sure, the actual tests should be lowercase with underscores, because they are more readable when showing up in listings and are easier to type. For internal classes we can keep the Python convention. For library test bases? I've no strong opinion.

library/microbenchmarks/gpu/gpu_burn/__init__.py
vkarak: Perhaps we need a feature to set the units of the performance variables without defining the `reference`. Let me think a bit about it.

jjotero: Took me a while to find it, but I was convinced it was not the first time I had heard that! 😂 See #1568

library/microbenchmarks/gpu/memory_bandwidth/__init__.py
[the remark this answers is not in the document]
jjotero: I'll resolve this one and move the naming conversation yo your other comment below.

cscs-checks/microbenchmarks/gpu/memory_bandwidth/memory_bandwidth.py
jjotero: I agree with the internal CSCS convention to name the tests all lowercase and with underscores. I would also add to this convention that we should append `_check` at the end. This would be for the classes that are decorated with the `@rfm.simple_test` decorator.

For the libraries and other non-decorated classes, I would also vote for the underscore style, but here allowing upper cases when appropriate and mandatory at the beginning. It feels to me that we're breaking the law by naming a class with all lowercase. Other library users might have a strong opinion there.

jjotero: For example, in the gpu burn test, we would import the `GPU_burn` class from the library, but the decorated test would be 
```python
from library.microbenchmarks.gpu.gpu_burn import GPU_burn

@rfm.simple_test
class gpu_burn_check(GPU_burn):
   ...
```

vkarak (changes requested):
Looks good overall, but we need a couple more things:

- Make the documentation RST compatible
- We should find better names for `library` and `cscslib`.
- We should settle on a naming convention for tests.

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -8,48 +8,76 @@
...
+    #: This variable will be passed to the compiler to generate the
+    #: arch-specific code.
+    #:
+    #: :default: ``None``
     gpu_arch = variable(str, type(None), value=None)
 
     # Required variables
-    executable_opts = required
     num_tasks = required
+    num_gpus_per_node = required
```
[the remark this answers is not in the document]
jjotero: These two required vars are mentioned in the above in the class docstring. As discussed, the library docs will be addressed in a separate PR.

After we settle on this PR, I think it'll be good if we port a few more tests before writing any docs, so we can get a broader perspective on how to best document these tests.

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -8,48 +8,76 @@
...
+    reference = {
+        '*': {
+            'perf': (0, None, None, 'Gflop/s'),
+            'temp': (0, None, None, 'degC')
+        }
+    }
 
     @rfm.run_before('compile')
     def set_gpu_build(self):
+        '''Set the build options.
```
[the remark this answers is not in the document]
jjotero: Changed this to `Set the build options before the compile pipeline stage`. Is this what you meant?

library/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -8,48 +8,76 @@
...
 
-__all__ = ['GpuBurnBase']
+__all__ = ['GPU_burn']
 
 
-class GpuBurnBase(rfm.RegressionTest, pin_prefix=True):
+class GPU_burn(rfm.RegressionTest, pin_prefix=True):
     '''Base class for the GPU Burn test.
 
     -- Compile --
```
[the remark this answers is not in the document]
jjotero: I've now updated the docstrings to a more standard format.

jjotero:
Regarding the dir structure, how about just `tests` and `cscs`?

library/microbenchmarks/gpu/gpu_burn/__init__.py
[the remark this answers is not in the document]
vkarak: Even more explicitly: `Set the build options [pre-compile hook]`

vkarak (approved):
Looks good to me and we can merge it and proceed with other tests to get the feeling. We can address the documentation in another PR, as you proposed. A side comment for that PR to come, single backquotes in RST are producing text in *italic* not in `monospace` as with markdown. Also as food for thought for the future, users are asking also for final tests, even minimal, so if the library was to come with simple test instantiations with `valid_systems = ['*']` and `valid_prog_environs = ['*']` would be rather helpful.

Regarding the name of the library, let's leave it as of now for the moment, since it is not released yet. We can come up with better ones soon.

vkarak:
Regarding the names about the library and the tests, what about this?

- `hpctestlib` for the library,
- `hpctests` for the generic, ready-to-use tests, 
- `cscstests` for the CSCS-specific tests.

victorusu (approved):
I agree with the naming scheme proposed by @vkarak. 
And I also agree that discussing the order of module imports here is not fruitful because we need to have more experience with the libraries.
Having said that, it `lgtm`

library/microbenchmarks/gpu/pointer_chase/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @sn.sanity_function
+    def average_P2P_latency(self):
+        '''
+        Extract the average P2P latency. Note that the pChase code
+        returns a table with the cummulative latency for all P2P
+        list traversals, and the last column of this table has the max
+        values for each device.
+        '''
+        return int(sn.evaluate(
```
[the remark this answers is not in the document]
jjotero: Yup, you're right. I don't need that evaluate in there either.

library/microbenchmarks/gpu/pointer_chase/__init__.py
```diff
@@ -0,0 +1,159 @@
...
+                r'the chase took on average (\d+) '
+                r'cycles per node jump.', self.stdout, 1, int)
+            ),
+        }
+
+
+class RunGpuPChaseP2P(RunGpuPChaseBase):
+    '''Base class for inter-GPU (P2P) latency tests.'''
+
+    executable_opts = ['--multi-gpu']
```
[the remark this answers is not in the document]
jjotero: Not necessarily. The aim of this test is to get the D2D latency and by setting the `--multi-gpu` flag, the output of the app is changed into a table format with all the combinations (hence the need for a new perf function below). However, one can still run this on a single-device node, and they'll get a 1-by-1 matrix :)

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -8,123 +8,49 @@
...
-            self.num_gpus_per_node = 8
-        elif cp in {'ault:amda100', 'ault:intelv100'}:
-            self.num_gpus_per_node = 4
-        elif cp in {'ault:amdv100'}:
-            self.num_gpus_per_node = 2
-        elif cp in {'ault:amdvega'}:
-            self.num_gpus_per_node = 3
-        else:
-            self.num_gpus_per_node = 1
+class GpuBurnTest(GpuBurnBase, hooks.SetArchAndModules, hooks.SetGPUsPerNode):
```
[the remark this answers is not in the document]
jjotero: You've commented some old code there. The latest version no-longer uses this :)

jjotero:
Also agree with the proposed directory names by @vkarak . I have updated the branch with this new dir names and also fixed the backquote issue in the docstrings. I have also addressed the comments from @victorusu

vkarak (approved):

vkarak:
@jenkins-cscs retry daint tsa

PR #1933, opened 2021-04-16
[feat] Issue a warning if a module operation is attempted on the `nomod` modules system

### LINKED ISSUE #1892: Issue warning when a test tries to invoke a module operation on the `nomod` pseudo modules system
This is a common pitfall. Users tend to forget setting the `modules_system` in their system configuration, because it's not required.

This is a follow up of #1887.

vkarak:
In order to avoid multiple warnings, this PR introduces also a caching mechanism for warnings. Warnings that are intended to be cached, must be issued with `cache=True`. Currently, only warnings from `nomod` are issued in such a way.

Fixes #1892.

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -408,10 +408,10 @@ class Logger(logging.Logger):
         return self.log(ERROR, msg, *args, **kwargs)
 
     def warning(self, msg, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


victorusu (approved):
`lgtm`

jjotero (approved):
lgtm

Issue #1833, opened 2021-03-04
Drop the `@parameterized_test` decorator in favor of the `parameter` built-in


vkarak:
Depends on #1849 (mostly to avoid unnecessary conflicts in the `4.x` branch).

PR #1934, opened 2021-04-17
[feat] Deprecate the use of the `@parameterized_test` decorator

### LINKED ISSUE #1849: Deprecate the use of the `@parameterized_test` decorator
This depends on #1691.

We should deprecate this as soon as we have ported the documentation and the tutorial to the new syntax.

vkarak:
Fixes #1849.

## Implementation details

Since I had to convert everything to the built-in `parameter` in the unit tests, I introduced the notion of force loading the test, in which case the loader first unloads the test module if it has already been loaded and then reloads it. This change is implemented in a larger extent in #1848, but here I am only using it in one unit test.


reframe/core/decorators.py
```diff
@@ -60,10 +61,12 @@ def _instantiate_all():
                 getlogger().warning(f'skipping test {cls.__name__!r}: {e}')
             except Exception:
                 frame = user_frame(*sys.exc_info())
+                filename = frame.filename if frame else 'n/a'
+                lineno = frame.lineno if frame else 'n/a'
```
vkarak, line 65: This is an irrelevant fix that came up as I was working on this PR.

unittests/test_loader.py
```diff
@@ -48,13 +48,6 @@ def test_load_all(loader_with_path):
     assert 12 == len(checks)
 
 
-def test_load_new_syntax(loader):
-    checks = loader.load_from_file(
-        'unittests/resources/checks_unlisted/good.py'
```
vkarak, line 53: This file has no meaning any more, so I removed it.

jjotero (approved):
I think we should also add a few fixmes or perhaps even create an issue to do the cleanup after the deprecation is done.
The main thing that I have in mind is the `_register_test` function, which can be massively simplified to just forward whichever `*args` and `**kwargs` from the decorator to the test constructor.
Also, (this one is for 4.0) I think it would also make sense to change the `simple_test` decorator to just `@test`.

vkarak:
@jjotero Yes, all these should be addressed properly by #1833 which is scheduled for 4.x

PR #1944, opened 2021-04-22
[bugfix] Mark jobs properly as completed in PBS scheduler

### LINKED ISSUE #1930: OpenPBS job queue hang
When running jobs in OpenPBS 20.0.1 we have found what looks to be partly related to #1301 & #1473 

Currently our PBS implementation uses job history output, when running jobs in PBS with reframe the job will error out due to this: 
```
[CMD] 'qstat -f 1015.sch1 1016.sch1 1017.sch1 1018.sch1'
[  FAILED  ] Ran 0/2 test case(s) from 3 check(s) (0 failure(s))
[==========] Finished on Wed Apr 14 16:32:55 2021
/usr/bin/reframe: run session stopped: job scheduler error: qstat failed with exit code 35 (standard error follows):
qstat: 1015.sch1 Job has finished, use -x or -H to obtain historical job information
qstat: 1016.sch1 Job has finished, use -x or -H to obtain historical job information
qstat: 1017.sch1 Job has finished, use -x or -H to obtain historical job information
qstat: 1018.sch1 Job has finished, use -x or -H to obtain historical job information
```

if we disable the job history support (not ideal), the job just hangs infinitely regardless of if the PBS jobs fail or run successfully

Reframe version: 
`reframe --version
3.5.0`

SLE 15 SP2
lmod
OpenPBS 20.0.1

vkarak:
In case `qstat` returns 153, although we do set the job status as completed, we did not set the `_completed` variable. This is the variable that is queried to check if the job has actually completed. The reason we don't check the job state only is that, if a test has finished, we must wait until its stdout and stderr are copied back to the test's working dir.

Addresses #1930.

@cblackworth Could you check if this fix works for you?


ekouts (approved):

vkarak:
@cblackworth Hmm, looking again at the issue, this will likely solve the hang that you get if you disable the job history. It will not solve the previous error.

vkarak:
@cblackworth I've pushed another patch that supposedly fixes the original issue. Let me know if this works for you. I don't have an access to a PBS scheduler that supports the `-x` option, but I understood what is the problem.

vkarak:
I will merge this one, since it indeed fixes a bug, but I will keep the linked issue open to follow up on the discussions.

PR #1949, opened 2021-04-26
[bugfix] Warn when a check search path does not exist

### LINKED ISSUE #1928: ReFrame does not warn or give an error if a test file does not exist
It will simply ignore it without any warning:

```
./bin/reframe -c /foo.py -l
```

```
[ReFrame Setup]
  version:           3.6.0-dev.1+ff33e201
  command:           './bin/reframe -c /foo.py -l'
  settings file:     '<builtin>'
  check search path: '/foo.py'

[List of matched checks]

Found 0 check(s)

Log file(s) saved in: '/tmp/rfm-x4c_6wmq.log'```

### PR DESCRIPTION
**Pull request by ekouts**
Fixes #1928.

### DIFF
--- a/reframe/frontend/loader.py
+++ b/reframe/frontend/loader.py
@@ -203,6 +203,7 @@ class RegressionCheckLoader:
         for d in self._load_path:
             getlogger().debug(f'Looking for tests in {d!r}')
             if not os.path.exists(d):
+                getlogger().warning(f"check path {d} doesn't exist")
                 continue
 
             if os.path.isdir(d):
--- a/reframe/frontend/loader.py
+++ b/reframe/frontend/loader.py
@@ -203,7 +203,7 @@ class RegressionCheckLoader:
         for d in self._load_path:
             getlogger().debug(f'Looking for tests in {d!r}')
             if not os.path.exists(d):
-                getlogger().warning(f"check path {d!r} doesn't exist")
+                getlogger().warning(f'check path {d!r} does not exist')
                 continue
 
             if os.path.isdir(d):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by vkarak — CHANGES_REQUESTED** [review]
I think the warning should be issued by the `load_all()` function of the loader:

https://github.com/eth-cscs/reframe/blob/23dc03bd2b85adcae62fb69be51a047e5cef1be5/reframe/frontend/loader.py#L205-L206

**Comment by teojgo** [comment]
I agree with @vkarak, this way you will also take into account the case where the checkpath is given by the env variable.

**Comment by ekouts** [comment]
> I think the warning should be issued by the `load_all()` function of the loader:
> 
> https://github.com/eth-cscs/reframe/blob/23dc03bd2b85adcae62fb69be51a047e5cef1be5/reframe/frontend/loader.py#L205-L206

@vkarak  @teojgo  What about the case that the user doesn't pass any path? Should the warning appear when he passes nothing and there is no `${RFM_INSTALL_PREFIX}/checks/`? That's why I made the check earlier.

**Comment by vkarak** [comment]
@ekouts I think the `load_all()` will cover that case, too. `load_all()` is always called to load the tests.

**Comment by ekouts** [comment]
Ok, then I will change it. I thought we don't want a warning in that case.

**Comment by vkarak** [comment]
We do want a warning there, too.

**Review by teojgo — APPROVED** [review]

File: reframe/frontend/loader.py [hunk 1]
```diff
@@ -203,6 +203,7 @@ def load_all(self, force=False):
         for d in self._load_path:
             getlogger().debug(f'Looking for tests in {d!r}')
             if not os.path.exists(d):
+                getlogger().warning(f"check path {d!r} doesn't exist")
```
vkarak, line 206: Do not use abbreviated forms in error messages. I will fix this and merge it.

```suggestion
                getlogger().warning(f'check path {d!r} does not exist')
```

vkarak (approved):

PR #1951, opened 2021-04-26
[feat] Drop the `@parameterized_test` decorator

### LINKED ISSUE #1833: Drop the `@parameterized_test` decorator in favor of the `parameter` built-in

vkarak:
Tests should use the `parameter()` built-in instead.

Fixes #1833.

I don't know if we should change also the `@simple_test` name to something else. And if we do, do we deprecate it from now? Or should we better deprecate `@simple_test` in 4.0 and remove it in 5.x?

### DIFF
--- a/reframe/core/decorators.py
+++ b/reframe/core/decorators.py
@@ -20,7 +20,6 @@ import sys
 import traceback
 
 import reframe.utility.osext as osext
-import reframe.core.warnings as warn
 from reframe.core.exceptions import (ReframeSyntaxError,
                                      SkipTestError,
                                      user_frame)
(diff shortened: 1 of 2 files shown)


ekouts (approved):
Looks fine to me. If we change the name from `simple_test` to `frm_test` I would deprecate in 4.0 and remove in 5.0. But I am not sure it's worth it, since it will affect literally everyone with old tests.
Should we also update the parameterized tests that use the decorator in cscs-check in a different PR? Or have a missed any open PR about this?

victorusu (approved):
`lgtm`

vkarak:
> Should we also update the parameterized tests that use the decorator in cscs-check in a different PR? Or have a missed any open PR about this?

This should go in a different PR and be updated in the master branch, not in 4.x.

PR #1941, opened 2021-04-21
[feat] Add a configuration option to disable automatic module conflict resolution

vkarak:
Fixes #1846.

@branfosj Can you try this PR by setting `export RFM_RESOLVE_MODULE_CONFLICTS=no` ?

PR #1956, opened 2021-04-29
[bugfix] Fix regression in setting the graylog server address

teojgo:
The bug was introduced in #1941


vkarak (approved):
Damn copy-paste :-D

vkarak:
But I don't think that it affected the behaviour, did it? Because I think that it is not accessed through the argparser's options.

teojgo:
> But I don't think that it affected the behaviour, did it? Because I think that it is not accessed through the argparser's options.

It did not allow to set the option via the environment variable.

PR #1957, opened 2021-04-29
[test] Specify the ipcmagic version currently working with jupyterlab

teojgo:


rsarm (approved):

victorusu (approved):
`lgtm`

victorusu:
@teojgo, can we merge this?

victorusu (approved):
The ipcmagic module definition is consistent with the Horovod one. But I wonder if there is anything we can do about passing this modules like this...

PR #1958, opened 2021-04-30
[test] Modernize Scalapack test

ChristopherBignamini:
Changes:
- parameterized decorator removed, parameter builtin is now used 
- init function removed
- original base/derived class separation removed: I don't remember why I've created the test with that separation but it looks useless


vkarak (approved):
I did a small fix and a couple of style fixes. Lgtm.

PR #1959, opened 2021-05-02
[feat] Add SGE scheduler backend

### LINKED ISSUE #1937: Support SGE scheduler
It'd be great to support SGE as scheduler.  I'm trying to use PBS as they are somewhat similar, but polling doesn't seem to really work: at
https://github.com/eth-cscs/reframe/blob/45fbacb23210c757724882a13c4a53f33af04800/reframe/core/schedulers/pbs.py#L186
the command `qstat -f Your` is executed for me, I guess there is something wrong going on.

I might be able to work on this implementation in the near future, it shouldn't be too different from PBS, but I'll likely need some assistance along the way :slightly_smiling_face:

giordano:
This should be mostly working, but I need to fix some details, so I'm marking the pull request as draft.

I probably also need to update the documentation to mention SGE is a supported scheduler?

Fix #1937


reframe/schemas/config.json
```diff
@@ -232,7 +232,7 @@
                                     "type": "string",
                                     "enum": [
                                         "local", "pbs", "slurm",
-                                        "squeue", "torque"
+                                        "sge", "squeue", "torque"
```
giordano, line 235: Is there a way to update this file automatically?  I did it manually :slightly_smiling_face:

ekouts: Afaik no, we update it manually too.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,225 @@
...
+                elif state in ['s', 'ts', 'S', 'tS', 'T', 'tT', 'Rs',
+                               'Rts', 'RS', 'RtS', 'RT', 'RtT']:
+                    jobinfo[job_number] = 'SUSPENDED'
+                elif state in ['Eqw', 'Ehqw', 'EhRqw']:
+                    jobinfo[job_number] = 'ERROR'
+                elif state in ['dr', 'dt', 'dRr', 'dRt', 'ds',
+                               'dS', 'dT', 'dRs', 'dRS', 'dRT']:
+                    jobinfo[job_number] = 'DELETING'
+                elif state == 'z':
+                    jobinfo[job_number] = 'COMPLETED'
```
ekouts, line 215: You can make a mapping similar to the `JOB_STATES` variable and set directly the `job._state`.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,225 @@
...
+
+        for job in jobs:
+            if job.jobid not in jobinfo:
```
ekouts, line 218: If you are using `jobinfo` only to keep track of what you have encountered, you can keep a set of jobids instead of a dictionary.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,225 @@
...
+                self.log(f'Job {job.jobid} not known to scheduler, '
+                         f'assuming job completed')
+                job._state = 'COMPLETED'
+                job._completed = True
+                continue
+
+            job._state = jobinfo[job.jobid]
```
ekouts, line 225: I think you can set directly the state in the previous loop and omit this here.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,225 @@
...
+
+        if not jobs:
+            return
+
+        user = osext.osuser()
+        completed = osext.run_command(
+            f'qstat -xml -u {user}'
+        )
+
+        if completed.returncode != 0:
```
ekouts, line 171: I am not familiar with SGE so I am just checking here. Does squeue return information for old jobs indefinitely? If not, what is the return code in case (a) all the job ids are old (b) one of the job ids is old? I am looking at PBS and it returns (a) status code 153 and (b) status code 0, but only includes recent IDs in the output.

ekouts (commented):
In general it looks good to me, thank you @giordano for the contribution! I have some comments but I think you should wait also for @vkarak 's feedback before changing things. 

First of all I was wondering if you really need to rewrite everything. As far as I see most of the methods are very similar to the PBS scheduler, so it would make sense to me to inherit from PBS and change only the methods that are different, like we do for the `torque` scheduler.  You would have to change `__init__`, `submit` and `poll`. In the same logic I don't think you need a special `_SgeJob`, `SGE_CANCEL_DELAY` and `SGE_OUTPUT_WRITEBACK_WAIT`, you can just reuse PBS's variables.

reframe/core/schedulers/sge.py
ekouts: Sorry, my bad. I had a look again at the code and you don't have immediate access to the job in this part. You should leave it as it is.

reframe/core/schedulers/sge.py
giordano: > Does squeue return information for old jobs indefinitely?

SGE's `qstat` doesn't show old jobs at all, as far as I can see.  [`qacct`](https://linux.die.net/man/1/qacct) shows old jobs but not active ones (so it can't be used as replacement for `qstat`), and it can be _slow_, depending on whether how often (if ever) sysadmins rotate flat files.

> If not, what is the return code in case (a) all the job ids are old (b) one of the job ids is old? I am looking at PBS and it returns (a) status code 153 and (b) status code 0, but only includes recent IDs in the output.

I'm not passing the list of jobs to `qstat`.

giordano: > I'm not passing the list of jobs to `qstat`.

To elaborate on this, the reason why I'm simply calling [`qstat`](https://arc.liv.ac.uk/SGE/htmlman/htmlman1/qstat.html) with all jobs for the current user is because SGE's `qstat` is very weird, if you call it with `qstat -f -j <comma-separated list of jobs>` you get _less_ information about the job itself, for example you don't get its status.

ekouts: Ok, I understand now, thanks! Then it is fine I think to ask for all the jobs of the user 👍

giordano:
Thanks for the suggestion about inheriting `SgeJobScheduler` from `PbsJobScheduler`, I just did it!  The last remaining bit I believe is emitting the preamble: at the moment I'm skipping all settings like number of cores/number of tasks/number of CPUs, because I'm not really sure how to set them in a portable way.

What are exactly the settings that need to be in the preamble?

vkarak:
ok to test

vkarak:
Thanks @giordano for the PR! I agree with @ekouts on her comments.

> The last remaining bit I believe is emitting the preamble: at the moment I'm skipping all settings like number of cores/number of tasks/number of CPUs, because I'm not really sure how to set them in a portable way.

What about setting the `TASKS_OPT` for the SGE scheduler such that it doesn't need the extra information? Then I guess that PBS' `_emit_lselect_option()` would do the job.

I am now looking into your PR in more detail and I'll get back to you with some more comments.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # Store information for each job separately
+        jobinfo = {}
+        for queue_info in root:
+
+            # Reads the XML and prints jobs with status belonging to user.
+            if queue_info is None:
+                raise JobSchedulerError('Decomposition error!\n')
+
+            for job_list in queue_info:
```
vkarak, line 124: From what I understand from the code, this is not a `job_list` but actually the job info XML Element. What about naming it as `job_elem`?

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,162 @@
...
+                if job_list.find("JB_owner").text != user:
+                    # Not a job of this user.
+                    continue
+
+                job_number = job_list.find("JB_job_number").text
+
+                if job_number not in [job.jobid for job in jobs]:
```
vkarak, line 131: This check is not very efficient. I would suggest keeping the jobs submitted by ReFrame in a set that you populate as you `submit()` them and here you simply check `if job_number in self._submitted_jobs`

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,162 @@
...
+            # Reads the XML and prints jobs with status belonging to user.
+            if queue_info is None:
+                raise JobSchedulerError('Decomposition error!\n')
+
+            for job_list in queue_info:
+                if job_list.find("JB_owner").text != user:
+                    # Not a job of this user.
+                    continue
+
+                job_number = job_list.find("JB_job_number").text
```
vkarak, line 129:
```suggestion
                jobid = job_list.find("JB_job_number").text
```

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,162 @@
...
+
+        for job in jobs:
+            if job.jobid not in jobinfo:
+                self.log(f'Job {job.jobid} not known to scheduler, '
+                         f'assuming job completed')
+                job._state = 'COMPLETED'
+                job._completed = True
+                continue
+
+            job._state = jobinfo[job.jobid]
```
vkarak, line 162: I don't think that you need this loop here. If you keep track of the jobs submitted by reframe, you can just directly update the job state as you iterate over the XML output.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,162 @@
...
+        if completed.returncode != 0:
+            raise JobSchedulerError(
+                f'qstat failed with exit code {completed.returncode} '
+                f'(standard error follows):\n{completed.stderr}'
+            )
+
+        root = ET.fromstring(completed.stdout)
+
+        # Store information for each job separately
+        jobinfo = {}
```
vkarak, line 117: I don't think you need this. See my comments below.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,162 @@
...
+        for queue_info in root:
```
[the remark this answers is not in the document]
vkarak: Also from what I can see, I think this refers to the list of jobs, right? If yes, perhaps a job-related name would be more relevant.

vkarak (changes requested):
@giordano Thanks again for your PR! I think that the way you look into the XML output of the jobs can be simplified. Here is an outline of what I think would be more suitable:

```python
@register_scheduler('sge')
class SgeJobScheduler(PbsJobScheduler):
    def __init__(self):
        ...
        self._submitted_jobs = {}

    def submit(self, job):
        ...
        self._submitted[job.jobid] = job

    def poll(self, *jobs):
        ...
        for queue_info in root:
            ...
            for job_elem in queue_info:
                jobid = job_list.find("JB_job_number").text
                if jobid not in self._submitted_jobs:
                    # Not a reframe job
                    continue

                job = self._submitted_jobs[jobid]
                job._state = _sge_state(job_list.find("state").text)
                if job.state in {'COMPLETED', 'SUSPENDED', 'ERROR'}:
                    job._completed = True
```

Where `_sge_state()` will actually do the mapping of the states, as you exactly do now. I don't think that anything else is needed from what I can imagine by reading through the code.

reframe/core/schedulers/sge.py
ekouts: The only thing I don't like about this is that in theory now `poll` can work with jobs submitted by a different instance of a sge scheduler (I think). Not that we use this like that, but this is how it is in other schedulers as well. To avoid going through the list every time maybe a compromise is to make a set in the beginning of the `poll`.

vkarak: I updated my proposal. You don't need to keep track of the submitted jobs anymore, but simply index the jobs passed to `poll()`, because otherwise the complexity of polling becomes O(N*P), where N is the total jobs in the queue and `P` the number of jobs to poll.

giordano:
The problem of the proposed approach is that a completed job may never enter the `z` status and just disappear from the list.  I implemented locally what you suggested but reframe hangs forever.  Now I'm missing this whole block which made sure to account for jobs which simply disappear from `qstat`:
https://github.com/eth-cscs/reframe/blob/624c44e2b2ae64fbf383d03a1aed0ae925effca1/reframe/core/schedulers/sge.py#L154-L160
Always keep in mind that `qstat` is weird :slightly_smiling_face:

reframe/core/backends.py
```diff
@@ -19,7 +19,8 @@
 _scheduler_backend_modules = [
     'reframe.core.schedulers.local',
     'reframe.core.schedulers.slurm',
-    'reframe.core.schedulers.pbs'
+    'reframe.core.schedulers.pbs',
+    'reframe.core.schedulers.sge',
```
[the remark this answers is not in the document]
vkarak: Never thought about it :-) I was mostly attached to the style.

reframe/core/schedulers/sge.py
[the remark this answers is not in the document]
vkarak: Is the second entry (that with the pending job) in this example indeed a `job_info`? I would expect it to be `queue_info` as well. Otherwise, the output is totally weird :-D

reframe/core/schedulers/sge.py
vkarak: Ok, you can leave it as of now. I didn't expect the SGE's output to be so weird!

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,162 @@
...
+
+            # Reads the XML and prints jobs with status belonging to user.
+            if queue_info is None:
+                raise JobSchedulerError('Decomposition error!\n')
```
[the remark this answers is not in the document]
vkarak: Looking into the XML output sample, I don't understand why this is a hard error. And besides, I think we should still change the error message :-)

vkarak:
@giordano You're right on your comment about the last loop. I've opened a PR to your branch with some small improvements. If it works, you can merge it and then we update the documentation for this one (plus the "decomposition" error message) and we're good to go, I think.

vkarak:
And I will add a unit test for the parts that do not require the scheduler, e.g., the `emit_preamble()`.

reframe/core/schedulers/sge.py
```diff
@@ -0,0 +1,169 @@
...
+        self._submit_timeout = rt.runtime().get_option(
+            f'schedulers/@{self.registered_name}/job_submit_timeout'
+        )
+
+    def _emit_lselect_option(self, job):
+        num_tasks_per_node = job.num_tasks_per_node or 1
+        num_cpus_per_task = job.num_cpus_per_task or 1
+        num_nodes = job.num_tasks // num_tasks_per_node
+        num_cpus_per_node = num_tasks_per_node * num_cpus_per_task
+        select_opt = ''
```
vkarak, line 43: You are ignoring completely the `num_nodes` etc. here. I don't know how practical this is? This backend can then only run single node jobs?

giordano: I know :slightly_smiling_face:  This is the very last bit.  SGE does support multi-node jobs, the problem is more that, as far as I could understand, there isn't a general way to do it, the details depends on the setup of the system.

vkarak: Ok, I'm leaving it to @victorusu to have a further look, since he has some experience with SGE.

victorusu: @giordano, AFAIK SGE comes with the default parallel environments (`smp`, `make` and `mpi`) and they can be accessed using the `-pe` option. An example SGE preamble would be:
```
#$ -S /bin/bash
#$ -N example_job
#$ -o sge.out
#$ -e sge.err
#$ -pe smp 12
```
You are right that one can modify as much as you want and even remove those parallel environments. That exactly what I used to do back in the day... I used to have the `local`, `mpich`, `openmpi`, `mvapich` and `intelmpi` parallel environments set.

Depending on how you configure them, they can allocate multiple nodes, that what I used to do with the MPI-related parallel environments.

IMHO, the issue is that SGE doesn't know the concept of `tasks` and `cpus-per-tasks`. It only knows the concept of SLOTS. The number of SLOTS is passed after the parallel environment definition (in the example it is set to 12). 

Because of this limitation, people started to use custom launchers with SGE. A very famous one is called `ibrun` that was adopted by many US sites (https://github.com/TACC/lariat/blob/master/bin/ibrun.sge).

Fun fact: Back in the day, I wrote my own launcher that I named `srun` and it had the flags `-n` and `-c`. But at that time, I had no clue about the SLURM scheduler. 😀 That was a fun coincidence!

My proposal is to set the SLOTS to be configurable in the virtual partition.
We could have a virtual partition defined as:
```
{
    'name': 'smp',
    'time_limit': '10m',
    'scheduler': 'sge',
    'access': [
        f'-pe smp %(num_cpus_per_task)s'
    ],
    'name': 'mpi',
    'time_limit': '10m',
    'scheduler': 'sge',
    'access': [
        f'-pe mpi %(num_task)s'
    ],
    'name': 'mpismp',
    'time_limit': '10m',
    'scheduler': 'sge',
    'access': [
        f'-pe mpi %(num_task)s*%(num_cpus_per_task)s'
    ],
}
```

Please note that we need to have a syntax that is different than the one that I propose here because the product (`*` for the `mpismp` partition) must be evaluated. We cannot emit the preamble with the product.

Since the SGE scheduler doesn't have its own launcher, I think that It is the responsibility of the site to configure the launcher the way they want.
We could later introduce support for the `ibrun` launcher, for instance. But this is a different discussion because we are not sure which version of `ibrun` to support...

If that's the case, then virtual partition configuration could be
```
'scheduler': 'sge+ibrun',
```

What do you guys think?
How do you configure the parallel environments in your site?

giordano: >  How do you configure the parallel environments in your site?

Yes, we do use parallel environments and the default one is called `mpi`, but as far as I could see, and you confirmed it, sysadmins can change the name of the environment or add new ones, which makes it harder to write a general script which can run on different systems.

vkarak: Thanks @victorusu for the insight and I didn't know that `ibrun` was to workaround SGE 😂 As a matter of fact, we do have an `ibrun` launcher inn ReFrame (I made it for TACC at some point). I am not sure though that the syntax you propose for the `access` is the correct. I was thinking that this configuration should be part of scheduler-specific configuration. `access` has the meaning of how you get access to that system partition, and here you overload that with something completely different I think. I would rather see it as follows:

```python
    'schedulers': [
        {
            'name': 'sge',
            'slots': [
                {
                    'name': 'smp',
                    'value': 'num_cpus_per_task'
                },
                {
                    'name': 'mpi',
                    'value': 'num_tasks'
                },
                {
                    'name': 'mpismp',
                    'value': 'num_tasks*num_cpus_per_task'
                }
            ],
            'target_systems': ['system_with_sge_scheduler']
        }
    ],
```

The SGE backend will then access the slots through the configuration:

```python
from reframe.core.runtime import runtime

slots = runtime.get_option('schedulers/@sge/slots')
```

From this point on, you can use this information to emit `-pe` command. What I am missing is what thing determines the slot to use. Any ideas?

To evaluate the expression for each slot, you can do this `eval(slot_value, job.__dict__)`.

NOTE: @victorusu I think that in your example the `-pe` option for the `mpismp` slot is wrong. It should be `-pe mpismp`, right?

vkarak: @giordano I think the slots concept that @victorusu described can be support from what ReFrame already offers with the concept of scheduler resources. Try define something along the following lines in your SGE scheduler partition (here called `foo`):

```python
                {
                    'name': 'foo',
                    'time_limit': '10m',
                    'scheduler': 'sge',
                    'resources': [
                        {
                            'name': 'smp',
                            'options': ['-pe', 'smp', '{num_slots}']
                        },
                        {
                            'name': 'mpi',
                            'options': ['-pe', 'mpi', '{num_slots}']
                        },
                        {
                            'name': 'mpismp',
                            'options': ['-pe', 'mpismp', '{num_slots}']
                        }
                    ]
                }
```

Each resource here is *slot* which then you can use in your test as follows:

```python
self.extra_resouces = {
    'smp': {'num_slots': self.num_cpus_per_task},
    'mpi': {'num_slots': self.num_tasks},
    'mpismp': {'num_slots': self.num_tasks*self.num_cpus_per_task}
}
```

@giordano Can you give it a try? If it works, then we only need to document this for the SGE scheduler and we can merge this PR.

Of course, you need to specify only the resources that you need. The important thing is that the test remains portable to systems with other schedulers, because if a partition does not define resources, `extra_resources` will be ignored.

victorusu: @vkarak, this is actually perfect, because this is a simple solution.

giordano: Sorry for the delay, we had some issues on our clusters in the previous days, today they seem to be behaving better.

So, I think the problem is that at the moment this PR is not writing these options correctly, I get this in the preamble:
```sh
#$ :smp:1:mpi:4:mpismp:4
#$ -pe
#$ -pe
#$ -pe
```
which is clearly wrong.  As far as I understand, the first line comes from
https://github.com/eth-cscs/reframe/blob/d90431cacf792e7a586246ea0d5810a6fb93586d/reframe/core/schedulers/sge.py#L38-L63  I believe the function `_emit_lselect_option` doesn't make sense for SGE.  I'm missing what is writing the `-pe` lines (and the arguments don't seem to be passed correctly).  Any hints? :slightly_smiling_face:

vkarak: Yes, `_emit_lselect_option()` must be updated for the SGE scheduler. Since I have already started writing some unit tests for this, I can update this as well.

vkarak: Ciao @giordano, can you try again now? I removed the `-lselect` option (it has no meaning) and I'm supposed to emit correctly the options passed from the cmd line or the extra resources.

giordano: Just to see how the job file is written, I'm using this test:
```py
import reframe as rfm


@rfm.simple_test
class HelloTest(rfm.RegressionTest):
    valid_systems = ['*']
    valid_prog_environs = ['*']
    sourcepath = 'hello.c'
    executable_opts = ['> hello.out']
    sanity_patterns = sn.assert_found(r'Hello, World\!', 'hello.out')
    num_cpus_per_task = 1
    num_tasks = 4
    extra_resources = {
        'smp': {'num_slots': num_cpus_per_task},
        'mpi': {'num_slots': num_tasks},
        'mpismp': {'num_slots': num_tasks*num_cpus_per_task}
    }
```
but this is writing
```sh
#$ -pe
#$ smp
#$ 1
#$ -pe
#$ mpi
#$ 4
#$ -pe
#$ mpismp
#$ 4
```
which is wrong because it should be
```sh
#$ -pe smp 1
#$ -pe mi 4
#$ -pe mpismp 4
```

vkarak: My bad in the implementation 🤦 I will fix that now!

vkarak (approved):

giordano:
Before the merge I was trying to run some benchmarks using also Spack as build system to test my full pipeline (had to fix some things), but it seems to be working as expected.  Thanks a lot!

PR #1965, opened 2021-05-07
[feat] Support test retries in the JUnit report

### LINKED ISSUE #1963: Support retries when generating a JUnit report
I think that we should support retries, because otherwise the test report is deceiving. Retries are used in scheduled runs mostly (not in CI), so the report must be accurate also in these cases.

teojgo:
* Each retry(run) is passed as an individual testsuite to the junit report.

Fixes #1963

### DIFF
--- a/reframe/frontend/runreport.py
+++ b/reframe/frontend/runreport.py
@@ -167,7 +167,7 @@ def junit_xml_report(json_report):
                 'failures': str(rfm_run['num_failures']),
                 'hostname': json_report['session_info']['hostname'],
                 'id': str(run_id),
-                'name': f'ReFrame run {run_id + 1}',
+                'name': f'ReFrame run {run_id}',
                 'package': 'reframe',
                 'tests': str(rfm_run['num_cases']),
                 'time': str(json_report['session_info']['time_elapsed']),
--- a/reframe/frontend/runreport.py
+++ b/reframe/frontend/runreport.py
@@ -159,51 +159,52 @@ def junit_xml_report(json_report):
     '''Generate a JUnit report from a standard ReFrame JSON report.'''
 
     xml_testsuites = etree.Element('testsuites')
-    xml_testsuite = etree.SubElement(
-        xml_testsuites, 'testsuite',
-        attrib={
-            'errors': '0',
-            'failures': str(json_report['session_info']['num_failures']),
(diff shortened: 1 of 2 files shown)


vkarak (approved):
Lgtm. Do you have an example of how it shows up on Gitlab and Jenkins?

reframe/frontend/runreport.py
```diff
@@ -159,51 +159,52 @@ def junit_xml_report(json_report):
...
-
-                # XSD schema does not like the exponential format and since we
-                # do not want to impose a fixed width, we pass it to `Decimal`
-                # to format it automatically.
-                'time': str(decimal.Decimal(tid['time_total'])),
+                'errors': '0',
+                'failures': str(rfm_run['num_failures']),
+                'hostname': json_report['session_info']['hostname'],
+                'id': str(run_id),
+                'name': f'ReFrame run {run_id + 1}',
```
vkarak, line 170:
```suggestion
                'name': f'ReFrame run {run_id}',
```

vkarak (changes requested):

vkarak (approved):

vkarak (changes requested):
We need also to remove the sentence in the documentation stating that this option does not work with the retries.

jgphpc:
> Lgtm. Do you have an example of how it shows up on Gitlab and Jenkins?

see https://git.cscs.ch/c4k/c4kcust/example-junit/-/pipelines/26043/test_report

jgphpc:
> Have you ran with the `--retries=<number>` option?

see https://git.cscs.ch/c4k/c4kcust/example-junit/-/pipelines/26045/test_report

vkarak:
Thanks @jgphpc. It seems that Gitlab combines the results of everything. Anyway, it's up to each program how interprets the results, at least we encode everything.

teojgo:
> Thanks @jgphpc. It seems that Gitlab combines the results of everything. Anyway, it's up to each program how interprets the results, at least we encode everything.

I could append a `run{id}` to the name of each test to differentiate.

vkarak:
@jenkins-cscs retry daint dom

vkarak:
> I could append a run{id} to the name of each test to differentiate.

No, that would mess with the actual test naming.

vkarak (approved):

PR #1966, opened 2021-05-08
[test] Remove VC test

jgphpc:
https://github.com/VcDevel/Vc
is switching to 
https://github.com/VcDevel/std-simd


teojgo:
@jgphpc why don't you adapt it to the new repository location?

jgphpc:
I will create a new test (later)

teojgo (approved):

PR #1969, opened 2021-05-12
[feat] Make hook-related machinery available as class directives

jjotero:
This PR builds on #1929 to simplify and improve the syntax when writing test libraries. The `run_before`, `run_after` and `require_deps` decorators are now available as class directives, so there is no need for the preceding `rfm.` anymore:
```python
    ...
    @run_before('compile')
    def a_pre_compile_hook(self):
        # do something

    # The standard way is still supported
    @rfm.run_after('run')
    def a_post_run_hook(self):
        # do something else
``` 

This PR also introduces the `bind` directive, which binds an external function with the regression test class. This allows to write pipeline hooks as standalone unary functions, which can then be safely "imported" by several other classes.
```python
# file handy_hooks.py
def ext_fn(x):
    # do something handy

# file mytest.py
from handy_hooks import ext_fn

    run_before('compile')(bind(ext_fn))

    # It is possible to bind the hook with a different name
    run_after('run')(bind(ext_fn, name='new_name'))
```
 Not using the bind directive here might lead to problems because the function object `ext_fn` would be shared by multiple classes.

### Issues with the current implementation

 By moving these directives into the `reframe/core/hooks.py` I feel that the mapping of the argument to the `run_before` and `run_after` should be done in the metaclass and not by any of the functions in the `hooks.py` file. It's a bit strange to me that these hook functions are aware of the pipeline stages. It would feel a lot more natural if the directives exposed by the metaclass would preprocess this argument and do the mapping before calling any of the functions in `hooks.py`.

Alternatively, we could add some syntactic sugar into this and get rid of the mapping problem altogether by exposing directives to the `pre_` and `post_` steps of each individual pipeline stage. IMO, this would vastly improve the syntax above:
```python
from handy_hooks import ext_fn

    pre_compile(bind(ext_fn))
    post_run(bind(ext_fn, name='new_name'))
```


victorusu:
@jjotero, I am not sure how this feature will evolve or be utilized...
I understand that there is potential to reuse hooks. I do not see any other benefit and I wonder if the gain compensates for the costs...
Do you have a proper use case for this feature?
Because when I check the `cscs-checks` folder, I  see that the hooks are test-dependent and cannot be used in any other check. Ok, this may be because we miss this feature, but at first glance, I cannot find any test that would really benefit from "hook reuse". It is possible that in my search I may have missed the key tests that would benefit, but still,  it looks to me that they are not the majority. 

So, what's the other benefit of this feature? 

Furthermore, I agree that the syntactic-sugar syntax is required. And I wonder if, in the simplified form, there could be an implicit `bind` call. So, instead of having
```python
from handy_hooks import ext_fn

    pre_compile(bind(ext_fn))
    post_run(bind(ext_fn, name='new_name'))
```
one could write
```python
from handy_hooks import ext_fn

    pre_compile(ext_fn)
    post_run(ext_fn, name='new_name')
```

jjotero:
> So, what's the other benefit of this feature?

In the gpu microbenchmarks, the `dgemm`, `shmem`, `kernel_latency`, `pointer_chase`, `memory_bandwidth` and `gpu_burn` checks share the two hooks in `/cscs-checks/microbenchmarks/hooks.py`. In the current version, these checks manually bind the external hook by wrapping the external function; but that's a lot of typing :) So there the `bind` directive becomes very handy.

There is quite a lot of room at the `hpctestlib` level to use the `bind` directive there also on the same microbenchmarks from above. All the GPU microbenchmarks there also share some hooks.

Regarding the syntax, I agree that binding an external hook doing 
```python
    run_before('compile')(bind(ext_fn, name='local_fn'))
```
is perhaps worse than a manual bind such as
```python
    @run_before('compile')
    def local_fn(self):
        ext_fn(self)
```
So that's why I think that the least we should do here is expose the `pre_` & `post_` decorators bound to a specific stage so we can do
```python
from handy_hooks import ext_fn

    pre_compile(bind(ext_fn))
    post_run(bind(ext_fn, name='new_name'))
```
If you want to go further and make the `bind` transparent to the user, I'm fine with that, but I'd probably go for a more explicit name such as this
```python
from handy_hooks import ext_fn

    inject_before_compile(ext_fn)
    inject_after_run(ext_fn, name='new_name'))
```
or instead
```python
from handy_hooks import ext_fn

    inject_before('compile', ext_fn)
    inject_after('run', ext_fn, name='new_name'))
```

jjotero:
Here's another and perhaps a bit of a different use-case. Look at the `affinity_check`. It's got a few "utility" functions in there such as `get_sibling_cpus`.
https://github.com/eth-cscs/reframe/blob/8bd06ea60a55b00c31c1fc999749ad3619823148/cscs-checks/prgenv/affinity_check.py#L154
If this function is factored out into a test utils library, other tests could use it as follows
```python
from hpctestlib.utils.topo import get_sibling_cpus

@rfm.simple_test
    ...
    bind(get_sibling_cpus)

    @run_before('run')
    def get_logical_cpus_per_core(self):
        self.num_cpus_per_core = len(self.get_sibling_cpus(0, by='core'))
```

vkarak:
I personally see the value of this proposal in this PR. Here are my general comments:

1. Do we still allow the `rfm.run_before` syntax?
2. I would vote against yet another syntax with the same meaning, such as `pre_compile`, `inject_pre_compile` etc. It will create more confusion. I like that the current proposal uses the same syntax and functionality: `run_before()` does exactly what it is expected to do as a decorator.
3. I don't see a problem with the `bind()` function: it takes a free function accepting at least one argument and binds it to the object.
4. I wouldn't argue that the one liner syntax is worse than the multiline, but again this is a matter of taste and, in fact, users can choose which one to use! I would definitely be one of those choosing the one liner.
5. I think adding the option to set the name of the bounded function is good, because it would avoid conflicts with other local methods.

jjotero:
1. I'd vote to schedule the drop for 4.0.
2. I think the current `run_before` & `run_after` decorators would make the one-liner syntax very (very) hard to read for non-Python people. When you use a decorator as `@deco`, all you need to know is that this decorator does something to the function below and the user might not know/care how that happens. However, if you use it as `deco(...)(...)`, the user must know how decorators work in order to understand the syntax. That's something far more advanced, I think. In any case, I think that the syntax we use for the decorators is a full discussion on its own (I agree that the `inject` syntax probably causes more problems than solutions). So I'd say we get this PR done and we have that chat at some other time :)

vkarak:
> I think the current run_before & run_after decorators would make the one-liner syntax very (very) hard to read for non-Python people. When you use a decorator as @deco, all you need to know is that this decorator does something to the function below and the user might not know/care how that happens. 

That's why there is the more verbose option to define a function and decorate it. I don't see the point in exposing the hooks as `pre_compile`, `post_compile` etc... That would mean another major syntax change with very little value and without a strong reason.

vkarak:
> I'd vote to schedule the drop for 4.0.

Meanwhile, we need to deprecate it since 3.7.0, I guess.

jjotero:
> Meanwhile, we need to deprecate it since 3.7.0, I guess.

Done

I've also intercepted all these directives to not make their way into the class and the instance, since something like `self.bind(...)` or `self.parameter(...)` does not make sense.

vkarak:
I think there is sth wrong with this PR git-wise. I see 67 files changed 🤔

jjotero:
@vkarak I've updated all the `@rfm.run_before` and so on to the `@run_before` syntax in all the cscs-checks. I can revert that if you think that should go in a separate PR.

vkarak:
Oh, I see! No, keep it, because we don't want tons of deprecation warnings. We already have enough with the `parameterized_test` decorator :-)

reframe/core/meta.py
```diff
@@ -93,6 +95,48 @@ def __getitem__(self, key):
...
+        with the desired name for the bound-method. Since this class is a
+        descriptor, the `__get__` method will return the right bound-method
+        when accessed from a class instance.
+        '''
+
+        __slots__ = ('fn')
+
+        def __init__(self, fn, name=None):
+            @functools.wraps(fn)
+            def fn_(*args, **kwargs):
```
vkarak, line 112:
```suggestion
            def _fn(*args, **kwargs):
```

reframe/core/meta.py
```diff
@@ -93,6 +95,48 @@ def __getitem__(self, key):
...
+            )
+            if obj is None:
+                return self.fn
+
+            return types.MethodType(self.fn, obj)
+
+        def __getattr__(self, name):
+            return getattr(self.fn, name)
+
+        def __setattr__(self, name, value):
```
vkarak, line 134: Why are you using `__setattr__` here and not `__set__`?

vkarak (changes requested):
Lgtm in general, but I don't see unit tests for the new functionality introduced:

- The `bind()` function
- Removal of the blacklisted directives after the class is created.

reframe/core/meta.py
jjotero: I think this should remain as a non-data descriptor. The method `__setattr__` is to attach attributes into `self.fn`; whereas `__set__` would be used to change `self.fn` itself, which is something that Python does not allow. For example:
```python
class Foo:
  def fn(self):
    print('fn')

  def __init__(self):
    self.fn = 4 # This inserts 'fn' directly in self.__dict__['fn'], because the method fn is a non-data descriptor.

f = Foo()
print(f.fn) # Prints 4 and fn is stored under self.__dict__['fn'].
```
I just noticed that this is missing the `__call__` method to be a "proper" Python function object.

unittests/test_meta.py
```diff
@@ -0,0 +1,67 @@
...
+        run_after('run')(ext)
+        require_deps(ext)
+
+        def __init__(self):
+            assert not hasattr(self, 'parameter')
+            assert not hasattr(self, 'variable')
+            assert not hasattr(self, 'bind')
+            assert not hasattr(self, 'run_before')
+            assert not hasattr(self, 'run_after')
+            assert not hasattr(self, 'require_deps')
```
vkarak, line 30: These assertions are not tried; you need to instantiate `MyTest`.

unittests/test_meta.py
```diff
@@ -0,0 +1,67 @@
...
+
+        # Test __setattr__ and __getattr__
+        assert hasattr(ext, '_rfm_foo')
+        ext._rfm_foo = False
+        assert ext._rfm_foo == ext.fn._rfm_foo
+        assert ext_fn._rfm_foo
+
+        def __init__(self):
+            assert isinstance(self.ext_fn(), MyTest)
+            assert isinstance(self.ext(), MyTest)
```
vkarak, line 60: Why not?

```suggestion
            assert self.ext_fn is self
            assert self.ext is self
```

reframe/core/meta.py
```diff
@@ -162,8 +170,17 @@ def __prepare__(metacls, name, bases, **kwargs):
...
-        # Directive to bind a free function into the class
         def bind(fn, name=None):
+            '''Directive to bind a free function to a class.
+
+            By default, the function is bound with the same name as the free
+            function. However, the function can be bound using a different name
+            with the ``name`` argument.
+
+            :param fn: external function to be bound to a class.
+            :param name: bind the function under a diferent name.
```
vkarak, line 181: We need a `.. versionadded:: 3.6.2` here.

vkarak: And actually this is not documented publicly. We need to add this file to the corresponding rst file.

vkarak (changes requested):
Still a couple of minor comments.

reframe/core/meta.py
jjotero: Good shout. The docs from the `run_before`, `run_after` and `require_deps` decorators was also missing. I have now rearranged this a bit and moved that to the `builtins` section of the docs.

jjotero:
Implementation-wise, I'm still not 100% happy that the file `hooks.py` is aware of the pipeline stages. Perhaps it would make sense to have the `run_after` and `run_before` functions specialising the `_runx` function inside the metaclass.

jjotero:
I feel this now looks a lot better. The backend `_runx` function to attach the hooks to the stages has now been renamed to `attach_to`. Decorators `run_before` and `run_after` have now live in the metaclass, and they use the `attach_to` function after mapping the pipeline stages.

vkarak (changes requested):
Lgtm, except the documentation needs a little bit of fine-tuning. Also the fact that the `run_before` etc. docs are in three places. I am working on that.

vkarak:
@jjotero I've fine tuned the documentation as follows:

- Remove duplicate documentation; use pointers in the code to the actual user documentation.
- Move pipeline hooks section under the builtins.
- Add a warning box for the deprecation of former syntax.

If you agree with the changes, I am ready to merge it.

vkarak (approved):

jjotero:
> If you agree with the changes, I am ready to merge it.

Yeah, docs look far cleaner now.

I've also made the parameters and variables to show as `RegressionMixin.parameter` (instead of `RegressionTest.parameter`) for consistency with all the other builtins.

PR #1975, opened 2021-05-17
[test] Add test for JobReport

jjotero:
This PR introduces a check for the GPU output of the job report in Piz Daint.

Depends on #1929.

- [x] Merge #1929


jjotero:
@victorusu I see that the job report almost always returns the info for all nodes. Right now, the lower performance threshold is `-100%`. Do you think it makes sense to be a bit more strict and raise this to, say, `-20%`?

victorusu:
@jjotero, there is no guarantee that there will be a report at the end of the job... 😅 
So, having -20% is good enough. Because the failure will be a sanity one if that doesn't happen...

jjotero:
Updated the refs to `-20%`.

victorusu (approved):
`lgtm`

jjotero:
@jenkins-cscs retry daint

cscs-checks/system/jobreport/gpu_report.py
```diff
@@ -0,0 +1,111 @@
...
+        much later after job has already completed (this could be up to 25s).
+        However, this wait might not be necessary for a small node count, so
+        we use the try/except block below to wait only if a first call to the
+        sanity function does not succeed.
+        '''
+
+        super().set_sanity_patterns()
+        self.sanity_patterns = sn.all([
+            self.sanity_patterns, self.gpu_usage_sanity()
+        ])
```
vkarak, line 60: Which part of the sanity patterns is expected to fail? That from GPU burn or the one from this test?

jjotero: The one from this test. I've updated the try/except block from below to just test the `self.gpu_usage_sanity()` function instead of the full sanity_patterns.

vkarak (changes requested):
Something is mixed up with this PR. I see changes to `meta.py` and `variables.py`.

Also, can sb answer my other comment. Depending on the answer, I have a follow up :-)

vkarak:
Ok, ignore my comment about not responding to my other comment :-D I just saw the answer.

jjotero:
That was weird. Those changes seemed to have leaked somehow from the other bugfix PR. Should be fixed now.

vkarak (approved):

vkarak:
@jenkins-cscs retry daint

Issue #563, opened 2018-11-07
Pass command line options to a regression test


teojgo:
It would be nice to have the ability to pass command line options directly to a test. A use case is the following:

```
.
.
.

@rfm.cmd_parameterized_test:
class VariableGpuTest(rfm.RegressionTest):
    def __init__(self, num_gpus_per_node):
        super().__init__()
        self.num_gpus_per_node = num_gpus_per_node

.
.
.
```

Then we could run the test with `reframe -c variable_gpu_test.py --rfm-option='[1, 2, 3, 4]'`
which will make the test run with 4 different values of `num_gpus_per_node`.

vkarak:
@teojgo Can you be more precise with this feature request. How would you like to program such a test? What you suggest in your description, it won't work.

- How the option should be named for each test?
- How a test specifies its arguments?

A more detailed user story is needed.

vkarak:
Now with the variable syntax we can extend the API to simple select variable to expose at the command line, e.g., `num_nodes = variable(int, expose=True, conv=int)`

jjotero:
I think that the strongest point from this feature is that it will enable us to register tests that have `required` variables, and then running the test without explicitly setting the variable in the test itself.
```python
@rfm.simple_test
class MyTest(rfm.RegressionTest):
    num_tasks = required
    ...
```
This is awesome because it just enables to run tests without coding any Python at all! 🚀

However, if the underlying test does explicitly set the value, I see three different scenarios for this depending on where this value is set. Here is how I think this should behave:

1. __The variable is set at the class level__: If a test just defines a given variable at the class level, the value provided through the cli will override the variable's original (default) value.

```python
@rfm.simple_test
class A(rfm.RunOnlyRegressionTest):
    ...
    num_tasks = 1
```
If a user runs this as
```bash
./bin/reframe ... -c <path_path_to_file> -r --set-defaults={num_tasks: 0}
```
(or whatever other cli option we go for to do this) the test `A` becomes a flexible test.

2. __The variable is set during the object construction__: If a test defines a given variable at the instance level before entering the pipeline (this is either in the `__init__` method or a post-init hook), the value provided through the cli will be overridden during construction.

```python
@rfm.simple_test
class B(rfm.RunOnlyRegressionTest):
    num_tasks = 1
    ...
    def __init__(self):
        print(self.num_tasks) # Prints 0 (the value passed from the cli)
        self.num_tasks = 10
```
[...]

PR #1988, opened 2021-05-27
[bugfix] Fix variable and parameter access as class attributes

jjotero:
This PR fixes a bug that was causing variables and parameters set as class attributes to be silently overridden. Bug reproducer:
```python
class Base(rfm.RegressionTest):
    ...
    p = parameter([1])
    v = variable(int, value=1)

Base.p = [2] # This PR will disallow updating parameters like this.
Base.v = 2

@rfm.simple_test
class Derived(Base):
    def __init__(self):
        print(self.p) # Prints 1
        print(self.v) # Prints 1. With this PR this prints 2.
```
This PR disables setting/updating parameters when accessed as class attributes. This is because such parameter update would break the internal iterators on the parameter space. If this feature is required, explicit `insert_parameter` or `update_parameter` class methods could be added.

This might be needed by #563

### DIFF
--- a/reframe/core/meta.py
+++ b/reframe/core/meta.py
@@ -62,6 +62,7 @@ class RegressionTestMeta(type):
             set. Accessing a parameter in the class body is disallowed (the
             actual test parameter is set during the class instantiation).
             '''
+
             try:
                 return super().__getitem__(key)
             except KeyError as err:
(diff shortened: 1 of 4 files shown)


vkarak (approved):
Just a couple of minor comments.

Issue #1843, opened 2021-03-08
Add a tutorial example showcasing the use of build automation tools through ReFrame


vkarak:
Since now we will have both EasyBuild and Spack supported, we should have a tutorial example for both.

vkarak:
Fixed by #1989

PR #1989, opened 2021-05-28
[doc] Add tutorial to showcase the use of build automation tools through ReFrame

rsarm:
Addresses #1843


rsarm:
@jenkins-cscs retry eiger

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+
+   export RFM_CONFIG_FILE=$(pwd)/tutorials/config/mysettings.py
+
+
+
+Basic test using EasyBuild
+----------------------------
+
+Let's consider a simple ReFrame test that installs ``bzip2-1.0.6`` given the easyconfig `bzip2-1.0.6.eb <https://github.com/eth-cscs/production/blob/master/easybuild/easyconfigs/b/bzip2/bzip2-1.0.6.eb>`__ and checks that the installed version is correct.
+The following code block shows the check highlighting the lines specific to this tutorial:
```
teojgo, line 20:
```suggestion
The following code block shows the check, highlighting the lines specific to this tutorial:
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+
+For this test, ReFrame generates the following command to build and install the easyconfig:
+
+.. code-block:: console
+
+   eb bzip2-1.0.6.eb -f
+
+All the files generated by this command i.e. the sources, temporary files, installed software and the corresponding modules, are located in the folder ``easybuild`` in the stage directory.
+
+Now we know all related to building and installing the code so we can move to the part dealing with running it.
```
teojgo, line 44:
```suggestion
Now that we know everything related to building and installing the code, we can move to the part dealing with running it.
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+To run the code, the generated modules need to be loaded in order to make the software available.
+The modules can be accessed through :attr:`~reframe.core.buildsystems.Easybuild.generated_modules`, however, they are available only after EasyBuild is done wit the installation.
```
teojgo, line 46:
```suggestion
The modules can be accessed through :attr:`~reframe.core.buildsystems.Easybuild.generated_modules`, however, they are available only after EasyBuild completes the installation.
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+This means that :attr:`~reframe.core.pipeline.RegressionTest.modules` can be set only after the build phase finishes.
+For that, we can set :attr:`~reframe.core.pipeline.RegressionTest.modules` in a class method wrapped by the :py:func:`~reframe.core.decorators.run_before` decorator, specifying the ``run`` phase.
+In this test this is done with the function ``prepare_run()`` in the mixin class ``EasybuildMixin`` that serves as a parent class for the regression test.
```
teojgo, line 49:
```suggestion
In this test this is done by the function ``prepare_run()`` of the mixin class ``EasybuildMixin`` which is a parent class of the regression test.
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+Writing this as a mixin class is a convenient way of creating a block of code that can be easily reused for other tests.
+This test will run then the following commands:
```
teojgo, line 51:
```suggestion
This test will then run the following commands:
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. code-block:: console
+
+   module load bzip/1.06
+   bzip2 --help
+
+Since all EasyBuild files are in the stage directory, unless ReFrame was run with ``--keep-stage``, they are going to be wiped out after the test passes.
```
teojgo, line 58:
```suggestion
Since all the EasyBuild files are in the stage directory, unless ReFrame was run with ``--keep-stage-files``, they are going to be wiped out if the test is successful.
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+:attr:`~reframe.core.pipeline.RegressionTest.keep_files` can be used to avoid that:
```
teojgo, line 59:
```suggestion
The :attr:`~reframe.core.pipeline.RegressionTest.keep_files` attribute can be used to avoid that:
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+.. code-block:: python
+
+   self.keep_files = ['easybuild']
+
+
+
+Packaging the installation
+--------------------------
+
+The EasyBuild build system offers a way of packaging the installation by using EasyBuild's packaging support.
```
teojgo, line 70:
```suggestion
The EasyBuild build system offers a way of packaging the installation via EasyBuild's packaging support.
```

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. code-block:: python
+
+   self.keep_files = ['easybuild/packages']
+   self.build_system.package_opts = {
+           'type': 'rpm',
+        }
+
+The packages are generated by EasyBuild in the stage directory.
+To keep them after the test passes, :attr:`~reframe.core.pipeline.RegressionTest.keep_files` needs to be set.
```
teojgo, line 83:
```suggestion
To retain them after the test succeeds, :attr:`~reframe.core.pipeline.RegressionTest.keep_files` needs to be set.
```

teojgo (changes requested):

teojgo (approved):

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
+============================================
+Tutorial 2: Using the EasyBuild build system
```
jjotero, line 2:
```suggestion
Tutorial 5: Using the EasyBuild build system
```

tutorials/build_systems/easybuild/eb_test.py
```diff
@@ -0,0 +1,25 @@
...
+        self.valid_prog_environs = ['gnu']
+        self.modules = ['daint-gpu',
+                        'EasyBuild-custom']
+        self.build_system = 'EasyBuild'
+        self.build_system.easyconfigs = ['bzip2-1.0.6.eb']
+        self.build_system.options = ['-f']
+        self.sanity_patterns = sn.assert_found(r'Version 1.0.6',
+                                               self.stderr)
+        self.executable = 'bzip2'
+        self.executable_opts = ['--help']
```
vkarak, line 25: Use the new syntax without the `__init__` method here.

tutorials/build_systems/easybuild/eb_test.py
```diff
@@ -0,0 +1,25 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+class EasybuildMixin(rfm.RegressionTest):
+    @rfm.run_before('run')
+    def prepare_run(self):
+        self.modules = self.build_system.generated_modules
```
vkarak, line 8: I don't think that the mixin offers something in this example. I would suggest moving everything into the test class.

tutorials/build_systems/easybuild/eb_test.py
vkarak: Also I'd use `--version` instead of `--help`.

tutorials/build_systems/easybuild/eb_test.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@rfm.simple_test
+class BZip2Check(EasybuildMixin):
+    def __init__(self):
+        self.descr = 'This demonstrates the EasyBuild build system.'
+        self.valid_systems = ['daint:gpu']
+        self.valid_prog_environs = ['gnu']
+        self.modules = ['daint-gpu',
+                        'EasyBuild-custom']
```
vkarak, line 18: These make the test Daint-specific which is not very good for a tutorial. I would remove them. If users want to run the test on Daint, they should either load these modules outside ReFrame or edit the test file.

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+
+   module load bzip/1.06
+   bzip2 --help
+
+Since all the EasyBuild files are in the stage directory, unless ReFrame was run with ``--keep-stage-files``, they are going to be wiped out if the test is successful.
+The :attr:`~reframe.core.pipeline.RegressionTest.keep_files` attribute can be used to avoid that:
+
+.. code-block:: python
+
+   self.keep_files = ['easybuild']
```
vkarak, line 63: This can be a bit misleading, because the `keep_files` will copy the files to the output directory and still remove the stage directory. If users don't want a re-installation, I would suggest to use the `--keep-stage-files --dont-restage`. I will propose that also in the corresponding Spack tutorial.

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,83 @@
...
+
+   eb bzip2-1.0.6.eb -f
+
+All the files generated by this command i.e. the sources, temporary files, installed software and the corresponding modules, are located in the folder ``easybuild`` in the stage directory.
+
+Now that we know everything related to building and installing the code, we can move to the part dealing with running it.
+To run the code, the generated modules need to be loaded in order to make the software available.
+The modules can be accessed through :attr:`~reframe.core.buildsystems.Easybuild.generated_modules`, however, they are available only after EasyBuild completes the installation.
+This means that :attr:`~reframe.core.pipeline.RegressionTest.modules` can be set only after the build phase finishes.
+For that, we can set :attr:`~reframe.core.pipeline.RegressionTest.modules` in a class method wrapped by the :py:func:`~reframe.core.decorators.run_before` decorator, specifying the ``run`` phase.
```
vkarak, line 48:
```suggestion
For that, we can set :attr:`~reframe.core.pipeline.RegressionTest.modules` in a class method wrapped by the :func:`~reframe.core.RegressionTest.run_before` built-in, specifying the ``run`` phase.
```

vkarak (changes requested):

tutorials/build_systems/easybuild/eb_test.py
rsarm: Here I put `--help` because `bzip2 --version` crashes for bzip2-1.0.6. With bzip2-1.0.8 it doesn't crash, but that version doesn't install with easybuild because of a problem fetching the sources.
```
bzip2: I won't write compressed data to a terminal.
bzip2: For help, type: `bzip2 --help'.
```

rsarm:
@jenkins-cscs retry eiger

docs/tutorial_easybuild.rst
```diff
@@ -1,5 +1,5 @@
 ============================================
-Tutorial 2: Using the EasyBuild build system
+Tutorial 5: Using the EasyBuild build system
```
vkarak, line 2:
```suggestion
Tutorial 5: Using package managers as a build system
```

vkarak: And generally I would remove the mention of EasyBuild in this introductory section, because we are going to add another section about Spack.

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,81 @@
...
+Then, the software to be installed is passed as a list to :attr:`~reframe.core.buildsystems.EasyBuild.easyconfigs`.
+Here only one easyconfig is given, but more than one can be passed.
+Finally, through :attr:`~reframe.core.buildsystems.EasyBuild.options`, command line options can be passed to the ``eb`` executable.
+In this test we pass ``-f`` to make sure that ``bzip2`` will be built even if the module already exists externally.
+
+For this test, ReFrame generates the following command to build and install the easyconfig:
+
+.. code-block:: console
+
+   eb bzip2-1.0.6.eb -f
```
vkarak, line 40: I think that we are missing the environment variables that are set, aren't we? Because in the next line we say that everything will be installed in the `{stagedir}/easybuild`, so the readers will wonder how this is happening?

vkarak: We need also to mention that ReFrame expects EasyBuild to be installed already on the system.

docs/tutorial_easybuild.rst
```diff
@@ -0,0 +1,81 @@
...
+
+   module load bzip/1.06
+   bzip2 --help
+
+Since all the EasyBuild files are in the stage directory, unless ReFrame was run with ``--keep-stage-files``, they are going to be wiped out if the test is successful.
+The attribute :attr:`~reframe.core.pipeline.RegressionTest.keep_files` can be use to copy the files generated by EasyBuild to the output directory once the test has successfully finished:
+
+.. code-block:: python
+
+   self.keep_files = ['easybuild']
```
vkarak, line 61: I don't think that this offers something to the test, is it? Also have you tried to `--dont-restage` option?

vkarak (changes requested):
I have only a couple of minor comments.

docs/tutorial_easybuild.rst
rsarm: `self.keep_files` is used to keep the installed software once the test passes and reframe finishes. I tried `--dont-restage`, but it doesn't seem to do that. That option would be used to rerun a test that failed and do not create a new stage directory for it it since stage directory exists.

rsarm:
@jenkins-cscs retry dom

docs/tutorial_easybuild.rst
vkarak: Here I mean that the combination `--keep-stage-files` and `--dont-restage` should do it. `self.keep_files` simply copies the `easybuild/` folder to the output directory.

tutorials/build_systems/easybuild/eb_test.py
```diff
@@ -0,0 +1,26 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+@rfm.simple_test
+class BZip2Check(rfm.RegressionTest):
+    descr = 'This demonstrates the EasyBuild build system.'
+    valid_systems = ['daint:gpu']
+    valid_prog_environs = ['gnu']
```
vkarak, line 9: Why gnu and not builtin here?

tutorials/build_systems/easybuild/eb_test.py
```diff
@@ -0,0 +1,26 @@
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+@rfm.simple_test
+class BZip2Check(rfm.RegressionTest):
+    descr = 'This demonstrates the EasyBuild build system.'
+    valid_systems = ['daint:gpu']
```
vkarak, line 8: I would also use `*` here.

tutorials/build_systems/easybuild/eb_test.py
```diff
@@ -0,0 +1,26 @@
...
+    valid_prog_environs = ['gnu']
+    executable = 'bzip2'
+    executable_opts = ['--help']
+
+    @run_before('compile')
+    def set_makefile(self):
```
vkarak, line 14:
```suggestion
    def setup_build_system(self):
```

docs/tutorial_easybuild.rst
rsarm: Ah I see now. I thought that what we wanted was to copy the easybuild folder to the output directory.

rsarm:
@jenkins-cscs retry daint

tutorials/build_systems/easybuild/eb_test.py
vkarak: Ok, we can update the tutorial later on.

docs/tutorial_build_automation.rst
```diff
@@ -0,0 +1,174 @@
...
+   . $SPACK_ROOT/share/spack/setup-env.sh
+   spack env activate -V -d myenv
+   bzip2 --help
+
+From this point on, sanity and performance checking are exactly identical to any other ReFrame test.
+
+.. tip::
+
+   While developing a test using Spack or EasyBuild as a build system, it's good to run ReFrame with :option:`--keep-stage-files` and :option:`--dont-restage` options.
+   These options will not remove the test stage files upon successful completion of the test but, more importantly, ReFrame will not wipe out the stage directory every time you retry the test, which is quite useful for avoiding rebuilding the dependencies of the packages built by the test.
```
jjotero, line 174:
```suggestion
   While developing a test using Spack or EasyBuild as a build system, it can be useful to run ReFrame with :option:`--keep-stage-files` and :option:`--dont-restage` options to prevent ReFrame from removing the test's stage directory upon successful completion of the test.
   For this particular type of test, these options will avoid having to rebuild the required package dependencies every time the test is retried.
```

tutorials/build_systems/spack/spack_test.py
```diff
@@ -0,0 +1,25 @@
...
+class BZip2SpackCheck(rfm.RegressionTest):
+    descr = 'Demo test using Spack to build the test code'
+    valid_systems = ['*']
+    valid_prog_environs = ['builtin']
+    executable = 'bzip2'
+    executable_opts = ['--help']
+
+    @run_before('compile')
+    def setup_build_system(self):
+        self.build_system = 'Spack'
```
jjotero, line 20: How about moving the `build_system` assignment to the class body and renaming this hook to `set_spack_environment`?

tutorials/build_systems/easybuild/eb_test.py
```diff
@@ -0,0 +1,30 @@
...
+class BZip2EBCheck(rfm.RegressionTest):
+    descr = 'Demo test using EasyBuild to build the test code'
+    valid_systems = ['*']
+    valid_prog_environs = ['builtin']
+    executable = 'bzip2'
+    executable_opts = ['--help']
+
+    @run_before('compile')
+    def setup_build_system(self):
+        self.build_system = 'EasyBuild'
```
jjotero, line 20: How about moving the `build_system` assignment to the class body and renaming this hook to `set_easibuild_options`? Either way, this is just a comment and I don't have a strong opinion on this. However, if this is changed, the spack tutorial should also be changed for consistency reasons.

jjotero (changes requested):

tutorials/build_systems/easybuild/eb_test.py
vkarak: I will still keep `setup_build_system` as the hook name.

vkarak:
@victorusu Can you have a look into this PR now?

jjotero (approved):
lgtm

docs/tutorial_build_automation.rst
```diff
@@ -0,0 +1,183 @@
...
+This example is the equivalent to the previous one, except that it uses Spack to build ``bzip2``.
+Here is the test's code:
+
+.. literalinclude:: ../tutorials/build_systems/spack/spack_test.py
+   :lines: 6-
+   :emphasize-lines: 12,14-16
+
+
+When :attr:`~reframe.core.pipeline.RegressionTest.build_system` is set to ``'Spack'``, ReFrame will leverage Spack environments in order to build the test code.
+For this reason, currently user must specify an environment.
```
teojgo, line 112:
```suggestion
For this reason, the current user must specify an environment.
```

docs/tutorial_build_automation.rst
```diff
@@ -0,0 +1,183 @@
...
+ReFrame treats Spack environments as *test resources* so it expects them under the test's :attr:`~reframe.core.pipeline.RegressionTest.sourcesdir`, which defaults to ``'src'``.
```
teojgo, line 113:
```suggestion
ReFrame treats Spack environments as *test resources* so it expects to find them under the test's :attr:`~reframe.core.pipeline.RegressionTest.sourcesdir`, which defaults to ``'src'``.
```

docs/tutorial_build_automation.rst
```diff
@@ -0,0 +1,183 @@
...
+.. code:: bash
+
+   . $SPACK_ROOT/share/spack/setup-env.sh
+   spack env activate -V -d myenv
+   spack install
+
+Any additional specs specified inside the ReFrame test will be added using the ``spack add`` command.
+As you might have noticed ReFrame expects that Spack is already installed on the system.
+The packages specified in the environment and the tests will be installed in the test's stage directory, where the environment is copied before building.
+Here is the the stage directory structure:
```
teojgo, line 145:
```suggestion
Here is the stage directory structure:
```

teojgo (changes requested):

docs/tutorial_build_automation.rst
vkarak: Here, I do mean "currently", in terms of time. Because there is also this PR: #2031.

vkarak (approved):

docs/tutorial_build_automation.rst
victorusu: I would write:
```
For this reason, users must specify an environment.
```

victorusu (approved):
`lgtm`. I would only change one sentence. But this is not important.

teojgo (approved):

docs/tutorial_build_automation.rst
vkarak: You are right it confuses as it is written now. In fact, I wanted to write this:

```suggestion
For this reason, currently, users must specify an environment.
```

PR #1991, opened 2021-05-29
[feat] Auto-detect cpu topology

### LINKED ISSUE #1742: Automatically configure ReFrame for the underlying system

vkarak:
This integrates @ekouts' CPU auto-detection work into the framework. ReFrame will try to auto-detect the system's topology the first time it runs on it and will reuse that in subsequent runs on the same system. More specifically:

- If the processor info is specified in the configuration file, nothing is tried.
- If the processor info is not specified in the configuration file, ReFrame looks processor info metadata files (which it might have produced in a previous run) and loads them. These metadata files are looked up in `{configdir}/_meta/{system}-{part}/processor.json` or in `~/.reframe/topology/{system}-{part}/processor.json` in case of the builtin configuration file. If the file is found, the topology information is loaded from there.
- If the topology file is not found, the topology will be autodetected. If the system partition is local (i.e., local scheduler + local launcher), the topology is auto-detected unconditionally. If the partition is remote, ReFrame will not try to auto-detect it unless the `RFM_REMOTE_DETECT` (or the corresponding configuration option) is set.
- For detecting remote topologies, ReFrame will generate a job script based on the partition information and launch itself on the remote system with `{launcher} reframe --detect-host-topology=topo.json`. The `--detect-host-topology` option causes ReFrame to detect the topology of the current host. More specifically, ReFrame creates a temporary directory under `.` (by default) where it copies itself and re-bootstraps itself *inside* the job script, i.e., on the remote nodes. This is to account for environment differences between the local and remote hosts.
- The temporary directory prefix can be changed by setting the `RFM_REMOTE_WORKDIR` environment variable.
- In case of errors during auto-detection, ReFrame will simply issue a warning and continue.

I tested it on Dom, Eiger and Ault and works beautifully.

### Todos

- [x] Write documentation
- [x] Add unit tests
- [x] Review and fine tune the implementation
- [x] Implement the mechanism for loading device configuration from a file, despite the lack of device auto-detection

### Future work

Auto-detection of GPU devices is left for future work.

Fixes #1742.


reframe/frontend/autodetect.py
```diff
@@ -0,0 +1,153 @@
...
+        with tempfile.TemporaryDirectory(dir='.') as dirname:
+            job = Job.create(part.scheduler,
+                             part.launcher_type(),
+                             name='rfm-detect-job',
+                             sched_access=part.access)
+            with osext.change_dir(dirname):
+                launcher_cmd = job.launcher.run_command(job)
+                job.prepare([f'{launcher_cmd} {rfm_exec} '
+                             f'--detect-host-topology=topo.json'],
+                            trap_errors=True)
```
jjotero, line 63: This assumes that the Python environment from the login node is also valid in the compute node, which is not necessarily true (I think). When I run this on Ault's `a64fx` partition, the cpu topology doesn't get detected because the above job gives me the classic `couldn't find module blah blah ...`. So I think this section here might need to pull the whole of ReFrame into the temp directory and there do the `sh bootstrap.sh`. This would make it more portable, I think.

jjotero (commented):
Just very minor comments and a bit of a corner case where this fails (see the comment below on the `autodetect.py` file).

reframe/frontend/autodetect.py
jjotero: I just hacked the code to manually reinstall ReFrame in the temper from scratch and it worked like a charm. Since this will cache the output the first time around, I think that the extra overhead from the full reinstall is not a problem at all.

vkarak: Good point. When writing this, I assumed that the same environment was available on the compute nodes, but I didn't document it. In any case, I could update the generated script to reinstall reframe.

reframe/frontend/autodetect.py
```diff
@@ -0,0 +1,153 @@
...
+
+def _load_info(filename):
+    try:
+        with open(filename) as fp:
+            return json.load(fp)
+    except OSError as e:
+        getlogger().warning(
+            f'could not load file: {filename!r}: {e}'
+        )
+        return {}
```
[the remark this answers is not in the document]
vkarak: Good catch, I'm not doing any validation for it now :-D

reframe/frontend/autodetect.py
vkarak: How did you install it? I see three different ways, each on with its own advantages and disadvantages:

a. Using `pip` by doing something like `PYTHONPATH=external/ python3 -m pip install . --target=${tmp_path}`
b. Doing a fresh clone into a temp dir and bootstrap again
c. Copy the current installation to tempdir and re-bootstrap it.

vkarak: Also I guess that the temp dir is on the remote host, because we can't make the assumption that a temp dir on the local host exists on the remote one.

reframe/frontend/autodetect.py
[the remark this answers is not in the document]
vkarak: The only problem that I see currently with this is that we most likely need to duplicate the part of the configuration schema that handles the processor info. Currently, this is part of the global configuration schema, but the files we are loading here have only the processor and device parts, so we need a separate schema file for those. I will try to see whether the "big" schema can include other subschemas from different files. I want to avoid the duplication in the schemas in all cases.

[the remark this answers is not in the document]
vkarak: Ok, I see that I can simply selects parts of the big schema to validate against. We're doing that already for validating the default values in the configuration.

reframe/frontend/autodetect.py
```diff
@@ -0,0 +1,153 @@
...
+        'general/0/detect_remote_system_topology'
+    )
+    config_file = rt.site_config.filename
+    if config_file == '<builtin>':
+        config_prefix = os.path.join(
+            os.getenv('HOME'), '.reframe/topology'
+        )
+    else:
+        config_prefix = os.path.dirname(config_file)
+        config_prefix = os.path.join(config_prefix, '_meta')
```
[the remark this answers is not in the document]
vkarak: The proposed change is not correct. The directory structure is the following:

```
config/
├── _meta
│   └── tresa-default
│       └── processor.json
├── cscs.py
└── tresa.py
```

And I want to create the `config/_meta` prefix.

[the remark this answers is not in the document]
vkarak: Oh, ignore my comment, I was changing only the second line 😅  I adopted the proposal now.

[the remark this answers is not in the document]
jjotero: Yeah, but they both do the same, right? My proposed change just inlines 92 into 93.

reframe/frontend/autodetect.py
jjotero: I think I went for the option c to avoid dealing with the checking out the correct branch. In that case option b would just not work because my changes were only locally and I would clone a version of reframe which does not have the feature I was playing around with.

reframe/frontend/autodetect.py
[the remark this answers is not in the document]
jjotero: Yeah, but they both do the same, right? My proposed change just inlines 92 into 93. Or am I missing something else here?

reframe/frontend/autodetect.py
vkarak: Yes, I was thinking about (c) as well, but the downside is that if we copy the whole dir, we may also copy trash, that's why I was thinking about (a), which again has some other problems. But in any case, the TMPDIR you're copying into from the host may not be valid in the remote host.

reframe/frontend/autodetect.py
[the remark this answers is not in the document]
vkarak: Ok, I've done it. We can simply pick a part of a loaded schema and validate something against it.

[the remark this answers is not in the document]
vkarak: I was a bit quick in responding. There is a side effect due to the way we have structured the schema. The solution is not that straightforward, because we are using cross-references in the schema. So, in practice, we need to create a custom schema on the fly with all the references that we need...

reframe/frontend/autodetect.py
jjotero: I mean, once implemented, b should always work. The tricky bit is how to test it while developing it 😅

victorusu: I like better the solution that makes use of `cp (option `c`), rather than `a` and `b`.
`a` has the problem that @jjotero has mentioned and `b` may be a different version than the one that I am running, which can cause problems if the file format or fields change between ReFrame versions.
Copying is more consistent. But I am not sure how one can bootstrap after copying if the compute nodes do not have `pip` or no internet access, which is a problem for all solutions...

vkarak: I am personally a bit reluctant with copy, because you don't know what you are going to copy. We will be copying `external/` which can be hundreds of MB and even some stale stage directories which could be several GB.

jjotero: I agree. I guess we could try/except the current option, then b and then c?

vkarak: I wouldn't go for (c), because that would mean additional configuration. If we cannot run the remote command, we will issue a warning and the user can find a way to manually run `--detect-host-topology` and copy the metaconfig file.

vkarak: So, current solution + a fresh clone with bootstrap as a fallback.

reframe/frontend/autodetect.py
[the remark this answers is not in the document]
vkarak: The validation pays off. I already caught a small bug in how `numa_nodes` were reported for Macs ;-)

reframe/frontend/autodetect.py
vkarak: Also, I'm not passing the `FILE` argument to the `--detect-host-topology` anymore, because that would imply that we need to have a common shared directory between the local and the remote hosts. I'm just using the standard output and if it's not a JSON (case of errors), I'm issuing a warning and continue with the next option to try.

vkarak: This is not resolved yet. I'm still working on it...

vkarak: Finally, this is done 🎉  Actually, I'm taking the option (b) because it has the less side effects if you do it in a controlled way, as I do. Actually, "installing" reframe anywhere is pretty easy: 1. Copy `['bin/', 'reframe/', 'bootstrap.sh', 'requirements.txt']` to the target directory, 2. Run `./bootstrap.sh` from that directory. So this is what I'm doing. The `./bootstrap.sh` is always run from the generated job script, so it covers nicely the Ault case. By default, I'm copying to a temporary directory under `.`, assuming this directory is most likely shared with the compute nodes (I want to avoid `$TMPDIR` since most likely -if not always- is not shared). But if you want to run from anywhere else than `.`, you can set the prefix to `RFM_REMOTE_WORKDIR`. Let me know what do you think now.

jjotero (commented):
Need to update the docs to have the right name for the config options and env vars. Other than this, looks good!

jjotero (approved):
lgtm

vkarak:
@jenkins-cscs retry daint eiger pilatus

PR #1996, opened 2021-06-01
[test] Add cpu microbenchmarks to the test library

jjotero:
Ported the `stream`, `strided_bandwidth`, `dgemm` and `latency` checks to the `hpctestlib` library.

This PR also extends the `cscs.py` config file to support the `PrgEnv-fujitsu` on the above microbenchmarks.

For now, I have ported the tests as they were before keeping the performance metrics are the same. However, these could all be flexible tests, and some of them report the performance on a per-node basis and others don't. We should establish a convention to have them all do the same thing.


vkarak:
This still needs substantial work, so I'm closing the PR due to time constraints. It will be added to the library at a later time.

PR #2005, opened 2021-06-08
[test] Update cudatoolkit version in GridTools check

ChristopherBignamini:


teojgo (changes requested):
Make this change for both dom/daint. Also can you adapt to the new parameter syntax to avoid the deprecation warnings?

teojgo (approved):

PR #2012, opened 2021-06-11
[bugfix] Restrict use of `required` keyword

### LINKED ISSUE #2011: Make illegal to set a regular Python class attribute as `required`
Users should get an error when this is attempted. Otherwise the code below would be legal (and it probably shouldn't!)
```python
@rfm.simple_test
class MyTest(rfm.RunOnlyRegressionTest):
    ...
    my_attr = required

    @run_after('init')
    def my_hook(self):
        print(self.my_attr) # This would print the reframe.core.variables._UndefinedType object.
```

jjotero:
See linked issue for details.

Closes #2011.

### DIFF
--- a/reframe/core/meta.py
+++ b/reframe/core/meta.py
@@ -243,7 +243,7 @@ class RegressionTestMeta(type):
 
         blacklist = [
             'parameter', 'variable', 'bind', 'run_before', 'run_after',
-            'require_deps'
+            'require_deps', 'required'
         ]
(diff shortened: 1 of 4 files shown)


ekouts (approved):

jjotero:
@jenkins-cscs retry dom

vkarak (approved):

jjotero:
@jenkins-cscs retry tsa

PR #2015, opened 2021-06-14
[test] Fix QE check on Eiger and Pilatus

lucamar:
By mistake `OMP_NUM_THREADS` was set equal to `str(self.num_cpus_per_task)` (which is 16), rather than 8: in this way the performance of the job is almost half than expected. Defining `self.num_cpus_per_task = 16` and then `OMP_NUM_THREADS = 8` for some reason gives a slightly better performance with respect to setting directly `self.num_cpus_per_task = 8`.


teojgo (approved):

teojgo:
@jenkins-cscs retry pilatus

Issue #1947, opened 2021-04-26
Display the report file in the failure statistics


victorusu:
Currently, ReFrame doesn't report the report file - if any - in the failure statistics section.

It would be nice ReFrame also emits the command to use the report file - similar to what one has for the individual failed tests - something like
```
Rerun the failed tests using '.....'
```

A proposal for the failures statistics is
```
==============================================================================
FAILURE STATISTICS

Total number of test cases: 5
Total number of failures: 3

Phase         #     Failing test cases
------------- ----- ------------------------------------------------------------
sanity        1     [IPCMagicCheck, PrgEnv-xxx, xxx:xxx]
performance   2     [P2PCPULatencyTest, PrgEnv-xxx, xxx:xxx]
                    [P2PCPULatencyTest, PrgEnv-xxx, xxx:xxx]
------------------------------------------------------------------------------
Rerun the failed tests using '.....'

Log file(s) saved in: '/path/to/reframe.log', '/path/to/reframe.out'
```

PR #2016, opened 2021-06-16
[feat] Print the run report location at the end of a session

jgphpc:
Closes https://github.com/eth-cscs/reframe/issues/1947


reframe/frontend/statistics.py
```diff
@@ -281,6 +281,17 @@ def print_failure_stats(self, printer):
             for f in l[1:]:
                 stats_body.append(row_format.format('', '', str(f)))
 
+        for p, l in failures.items():
+            stats_body.append(stats_hline)
```
ekouts, line 285: I am not sure whether we want a `stats_hline` here or a solid line. The result now is:
```
FAILURE STATISTICS

Total number of test cases: 10
Total number of failures: 4

Phase         #     Failing test cases
------------- ----- ------------------------------------------------------------
sanity        1     [T2, builtin, generic:default]
startup       2     [T7, None, None]
                    [T9, None, None]
setup         1     [T8, builtin, generic:default]
------------- ----- ------------------------------------------------------------
Rerun the failed tests with:
 -n T2 -p builtin --system generic:default -r
------------- ----- ------------------------------------------------------------
Rerun the failed tests with:
 -n T7 -p None --system None -r
 -n T9 -p None --system None -r
------------- ----- ------------------------------------------------------------
Rerun the failed tests with:
 -n T8 -p builtin --system generic:default -r
```

ekouts (changes requested):
I am not sure this was what @victorusu originally asked for in issue #1947. Victor maybe you can shed some light:
1. I think he meant to print something like this
`Rerun the failed tests using '--restore-session <report-file> -n <name1> -p <partition1> --system <system> -n <name2> -p <partition2>  ...'`. Or we can make it simpler with 
`Rerun the failed tests using '--restore-session --failed -r'`.
2. I don't agree to print it only in the failure statistic, when we pass `--failure-stats`. This is not really part of the statistics and I prefer that we print it always. What do you guys think?

vkarak (changes requested):
I agree with @ekouts. I don't think that this is what @victorusu requested in the corresponding issue. I interpret it as a simple line at the vey end (after the failure statistics) saying the something like "Rerun all failed tests with `reframe --restore-session=report-file --failed`. I personally also thing that the `report-file` is not needed. But @victorusu, please explain more what's your feature request is about.

victorusu (approved):
`lgtm`

vkarak:
I updated this PR to print the report file always at the end of the session. I am against adding a `Rerun with` type of message, cos it introduces unnecessary complexity to the implementation. We report the file here, and somebody can use it with the `--restore-session` option.

vkarak (approved):

PR #2018, opened 2021-06-17
[test] Update Amber test to new syntax and add support for Eiger and Pilatus

victorusu:
I am adding support for Eiger and Pilatus, where the number of cores per node is so high that one cannot run the simulations with the same amount of nodes. Therefore, the `small` tests run with 4 nodes instead of 6 and the `large` ones with 8 nodes instead of 16. We could alternatively, on Eiger just test the small one with 6 nodes. Let me know what you think.

I have also extended the number of tests from 1 to 4 and I have appended the name of the input files to the test name.  
The former Amber test was based on the system [Cellulose_production_NVE](https://github.com/victorusu/amber_benchmark_suite/blob/main/amber_16_benchmark_suite/PME/Cellulose_production_NVE.tar.bz2).

Furthermore, I have simplified the logic of the tags. Before we were having one dimension to describe two different sets of tests (`maint` and `prod`), which were exactly the same. Now we just have one dimension (`small` and `large`). And both have the `maintenance` and `production` tags. As before, the `health` tag is only added to the single node GPU tests.

Additionally, I have made the tests more reusable outside CSCS.
I have created the GitHub repo [Amber Benchmark Suite](https://github.com/victorusu/amber_benchmark_suite), which holds the input files for the Amber tests.
Now the tests download directly the required input files.

I could alternatively, have downloaded the original [benchmark suite](https://ambermd.org/Amber16_Benchmark_Suite.tar.bz2), but the file is 84.1 MB and expands to a bit more than 500 MB because it is comprised of the full suite. In this case, we would have all files for a bunch of systems that we do not test in every test, which IMHO is a waste.

I have only added the NVE part of the benchmark suite, the reason being the fact that this is the only one that we can reliably test the sanity. The NPT simulations depend on the pseudo-random numbers and I would need to change the original input files in order to make their energy output more reproducible. I can change the input files but for now, I would like to experiment with these NVE systems first.

Please take a look and let me know what you think.


victorusu:
The failed CI is unrelated to test

victorusu:
@jjotero, I have updated the test based on our off-line conversation...

teojgo (approved):

cscs-checks/apps/amber/amber_check.py
```diff
@@ -3,95 +3,190 @@
...
+
+    @run_after('init')
+    def set_hierarchical_prgenvs(self):
+        if self.current_system.name in ['eiger', 'pilatus']:
+            self.valid_prog_environs = ['cpeIntel']
+
+    @run_after('setup')
+    def set_perf_reference(self):
+        if self.scale == 'small':
+            self.reference = REFERENCE_CPU_PERFORMANCE_SMALL
```
jjotero, line 170:
```suggestion
        if self.scale == 'small':
            self.reference = REFERENCE_CPU_PERFORMANCE_SMALL
        else:
            self.reference = REFERENCE_CPU_PERFORMANCE_LARGE
```

cscs-checks/apps/amber/amber_check.py
```diff
@@ -3,95 +3,190 @@
...
+
+@rfm.simple_test
+class AmberCPUCheck(AmberBaseCheck):
+    scale = parameter(['small', 'large'])
+    valid_systems = ['daint:mc', 'eiger:mc']
+    executable = 'pmemd.MPI'
+    input_file = 'mdin.CPU'
+    output_file = 'amber.out'
+    tags = {'maintenance', 'production'}
+    reference = REFERENCE_CPU_PERFORMANCE_LARGE
```
jjotero, line 151: It feels a bit weird that the `REFERENCE_CPU_PERFORMANCE_LARGE` is the default one. Since you have the hook `set_perf_reference` below already, I'd set the `reference` in this hook for all cases.
```suggestion
```

jjotero (changes requested):

jjotero:
@victorusu What about all those PEP complaints?

victorusu:
@jjotero, I hadn't seen them! I will fix them!

victorusu:
The fix is done... The CI error is spurious... Can we merge this?

jjotero (approved):

victorusu:
The CI error is spurious... I can remove `pilatus` from the available systems and the CI would pass. Should I do that?

jjotero:
@victorusu I'd say you go ahead and click that merge button! 😉

victorusu:
@jjotero, thanks! Done!

PR #2020, opened 2021-06-17
[test] Update `cuda-samples` check

jjotero:


config/cscs.py
```diff
@@ -1074,6 +1075,13 @@
                 'PrgEnv-pgi'
             ]
         },
+        {
+            'name': 'PrgEnv-nvidia',
+            'modules': [
+                'PrgEnv-nvidia',
+                'cdt/21.05'
```
jgphpc, line 1082:
```suggestion
                
```
use `-m cdt/21.05` or `-m cdt-cuda/21.05` on the command line ?

jjotero: @jgphpc If I remove this one, my coda-aware MPI checks fail

jgphpc (approved):

victorusu (approved):
`lgtm`

teojgo (changes requested):
The only change I would put is some `FIXME` comments so that we make the corresponding changes for daint as well when the systems are in sync.

config/cscs.py
victorusu: Can we add a `#FIXME` on top of this `cdt` entry? 
We should handle this differently in the future...

victorusu (changes requested):
can you add please a `#FIXME`?

teojgo (approved):

victorusu (approved):
`lgtm`

teojgo:
@jenkins-cscs retry dom

PR #2027, opened 2021-06-23
[bugfix] Fix deep-copy of `_Undefined` constant

giordano:
The condition `self.environment is _Undefined` is always `False`, so the error
is actually never raised.

Without this change, I get a much more obscure error message:
```
SUMMARY OF FAILURES
FAILURE INFO for Test 
  * Test Description: Test
  * System partition: catalina:default
  * Environment: gnu
  * Stage directory: /home/mose/repo/reframe/stage/catalina/default/gnu/Test
  * Node list: 
  * Job type: local (id=None)
  * Dependencies (actual): []
  * Maintainers: []
  * Failing phase: compile_wait
  * Rerun with '-n Test -p gnu --system catalina:default -r'
  * Reason: build error: stdout: 'rfm_Test_build.out', stderr: 'rfm_Test_build.err'
--- rfm_Test_build.err (first 10 lines) ---
/home/mose/repo/reframe/stage/catalina/default/gnu/Test/rfm_Test_build.sh: line 13: syntax error near unexpected token `newline'
/home/mose/repo/reframe/stage/catalina/default/gnu/Test/rfm_Test_build.sh: line 13: `spack env activate -V -d <reframe.core.buildsystems._UndefinedType object at 0x7effb15a5c70>'
--- rfm_Test_build.err --- 
Log file(s) saved in: '/tmp/rfm-mtmrgbz_.log'
```
With this change I get
```
SUMMARY OF FAILURES
FAILURE INFO for Test 
  * Test Description: Test
  * System partition: catalina:default
  * Environment: gnu
  * Stage directory: /home/mose/repo/reframe/stage/catalina/default/gnu/Test
  * Node list: 
  * Job type: local (id=None)
  * Dependencies (actual): []
  * Maintainers: []
  * Failing phase: compile
  * Rerun with '-n Test -p gnu --system catalina:default -r'
  * Reason: build system error: no Spack environment is defined
Log file(s) saved in: '/tmp/rfm-93cirj79.log'
```
which I guess was the point of this check

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -14,6 +14,10 @@ from reframe.core.exceptions import BuildSystemError
 
 class _UndefinedType:
     '''Used as an initial value for undefined values instead of None.'''
+    __slots__ = ()
+
+    def __deepcopy__(self, memo):
+        return self
 
 
 _Undefined = _UndefinedType()
@@ -866,7 +870,7 @@ class Spack(BuildSystem):
 
     def emit_build_commands(self, environ):
         self._prefix_save = os.getcwd()
-        if isinstance(self.environment, _UndefinedType):
+        if self.environment is _Undefined:
             raise BuildSystemError(f'no Spack environment is defined')
 
         ret = self._env_activate_cmds()
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -866,7 +866,7 @@ class Spack(BuildSystem):
 
     def emit_build_commands(self, environ):
         self._prefix_save = os.getcwd()
-        if self.environment is _Undefined:
+        if isinstance(self.environment, _UndefinedType):
             raise BuildSystemError(f'no Spack environment is defined')
 
         ret = self._env_activate_cmds()


reframe/core/buildsystems.py
```diff
@@ -866,7 +866,7 @@ def __init__(self):
 
     def emit_build_commands(self, environ):
         self._prefix_save = os.getcwd()
-        if self.environment is _Undefined:
+        if isinstance(self.environment, _UndefinedType):
```
jjotero, line 869:
```suggestion
        if self.environment is _Undefined:
```

jjotero (changes requested):
Hi @giordano. Thanks for spotting this. 

I have the feeling that the line you changed was correct before. Instead, the definition of `_UndefinedType` should be as follows:
```python
class _UndefinedType:
    '''Used as an initial value for undefined values instead of None.'''
    __slots__= ()

    def __deepcopy__(self, memo):
        return self
```
This would ensure that the `_Undefined` constant will always be the same instance of `_UndefinedType` regardless of deep copies and so on.

Could you please try this approach instead? Thanks!

giordano:
Yes, I confirm this also does the trick, thanks!

victorusu:
@giordano, thanks. I personally like this change much more. I was about to ask you why not work on the `_Undefined` since it is `=` to `_UndefinedType`?

This fix implies that we had a bug because `_UndefinedType` is not deeply copiable... 🤔

giordano:
Also, I wanted to add a test, but I think
https://github.com/eth-cscs/reframe/blob/9dbb9899d371e22f325feb9adeac379b398511a9/unittests/test_buildsystems.py#L281-L284
is already meant to check this situation and the test was already passing, so... this test is catching some other `BuildSystemError`?

jjotero (approved):
@giordano This test is checking this exact same `BuildSystemError`. The difference is that the `Spack(BuildSystem)` instance is deep-copied when you actually run ReFrame, and such deep-copy is not done by the unit test. Hence, the deep-copy
```python
if self.environment is _Undefined:
    ...
```
compares that `self.environment` is the same instance of `_UndefinedType` as `_Undefined`. But because the build system class has been deep-copied, they are not he same instance and this evaluates to `False`. This is what overriding the `__deepcopy__` method of `_UndefinedType` fixes.

This being said, we already have a mechanism in place to deal with these types of classes (`reframe/core/variables.py`) and it probably makes sense to make all these `BuildSystem` classes use all that machinery too. I'll create an issue to discuss that internally.

Thanks again for spotting the bug!

@victorusu are you happy with the changes?

PR #2029, opened 2021-06-24
[test] Add PrgEnv-nvidia to halo_cell_exchange check

ajocksch:
halo_cell_exchange check PrgEnv-nvidia
https://jira.cscs.ch/browse/UES-1641


jgphpc (approved):

jjotero:
@jgphpc There are heaps of failures for this test. At least, the performance references should be adjusted to a less optimistic value.

jgphpc (changes requested):
I get the same failure (in performance) with almost all PrgEnv.
Can we relax the performance limits or is there an issue ?

jjotero:
@jgphpc I don't think there is any performance issue. I just think that the tolerances are way too strict for this case. I also vote for relaxing them.

cscs-checks/mch/collectives_halo.py
```diff
@@ -84,6 +84,7 @@ def __init__(self, bench_reference):
         }
 
         self.maintainers = ['AJ', 'MKr']
+        self.strict_check = False
```
jjotero, line 87: I think you've added the `strict_check` into the wrong test. This should go into the `halo_cell_exchange.py` instead.

jjotero (changes requested):

jgphpc (approved):

jjotero (approved):

PR #2031, opened 2021-06-24
[feat] Automatically create a Spack environment when not set

### LINKED ISSUE #2028: Automatically create the spack environment needed for the spack build system
As far as I understand, at the moment the [Spack build system](https://github.com/eth-cscs/reframe/blob/9dbb9899d371e22f325feb9adeac379b398511a9/reframe/core/buildsystems.py#L802-L893) _requires_ passing in a non-`None` environment (and if you don't you get an unclear error message: #2027), but you also have to create it beforehand (BTW, as far as I can tell the environment is searched in `stagedir`, not `sourcesdir` as claimed in the documentation), otherwise the build will fail because the environment doesn't exist.

What about automatically creating an environment in the staging directory if the environment is `None`, thus making the field optional?  If they already have an environment directory that they absolutely want to use that can still do so, but most simple cases users don't need to care about it

giordano:
Fix #2028.

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -787,6 +787,28 @@ class Spack(BuildSystem):
     #: :default: ``None``
     environment = variable(typ.Str[r'\S+'], type(None), value=None)
 
+    #: The directory where Spack will install the packages requested by this
+    #: test.
+    #:
+    #: After activating the Spack environment, ReFrame will set the
+    #: `install_tree` Spack configuration in the given environment with the
+    #: following command:
+    #:
+    #: .. code-block:: bash
+    #:
+    #:    spack config add "config:install_tree:root:<install tree>"
+    #:
+    #: Relative paths are resolved against the test's stage directory.  If this
+    #: field and the Spack environment are both `None`, the default, the
+    #: install directory will be automatically set to `opt/spack`.  If this
+    #: field `None` but the Spack environment is not, then `install_tree` will
+    #: not be set automatically and the install tree of the given environment
+    #: will not be overridden.
+    #:
+    #: :type: :class:`str` or :class:`None`
+    #: :default: ``None``
+    install_tree = variable(typ.Str[r'\S+'], type(None), value=None)
+
     #: A list of additional specs to build and install within the given
     #: environment.
     #:
@@ -818,6 +840,12 @@ class Spack(BuildSystem):
 
     def emit_build_commands(self, environ):
         ret = self._env_activate_cmds()
+
+        if not self.environment:
+            install_tree = self.install_tree or 'opt/spack'
+            ret.append(f'spack config add '
+                       f'"config:install_tree:root:{install_tree}"')
+
         if self.specs:
             specs_str = ' '.join(self.specs)
             ret.append(f'spack add {specs_str}')
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -780,11 +780,12 @@ class Spack(BuildSystem):
     #: ReFrame looks for environments in the test's
     #: :attr:`~reframe.core.pipeline.RegressionTest.sourcesdir`.
     #:
-    #: This field is required.
+    #: If this field is `None`, the default, the environment name will
+    #: be automatically set to `rfm_spack_env`.
     #:
     #: :type: :class:`str` or :class:`None`
-    #: :default: :class:`required`
-    environment = variable(typ.Str[r'\S+'])
+    #: :default: ``None``
+    environment = variable(typ.Str[r'\S+'], type(None), value=None)
 
     #: A list of additional specs to build and install within the given
     #: environment.
@@ -816,9 +817,6 @@ class Spack(BuildSystem):
     install_opts = variable(typ.List[str], value=[])
 
     def emit_build_commands(self, environ):
-        if not hasattr(self, 'environment'):
-            raise BuildSystemError(f'no Spack environment is defined')
-
         ret = self._env_activate_cmds()
         if self.specs:
             specs_str = ' '.join(self.specs)
@@ -832,8 +830,15 @@ class Spack(BuildSystem):
         return ret
 
     def _env_activate_cmds(self):
-        return [f'. $SPACK_ROOT/share/spack/setup-env.sh',
-                f'spack env activate -V -d {self.environment}']
+        cmds = ['. $SPACK_ROOT/share/spack/setup-env.sh']
+        if self.environment:
+            environment = self.environment
+        else:
+            environment = 'rfm_spack_env'
+            cmds.append(f'spack env create -d {environment}')
+
+        cmds.append(f'spack env activate -V -d {environment}')
+        return cmds
 
     def prepare_cmds(self):
         cmds = self._env_activate_cmds()
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -785,6 +785,11 @@ class Spack(BuildSystem):
     #:
     #: :type: :class:`str` or :class:`None`
     #: :default: ``None``
+    #:
+    #: .. note::
+    #:     .. versionchanged:: 3.7.3
+    #:        The field is no longer required and the Spack environment will be
+    #:        automatically created if not provided
     environment = variable(typ.Str[r'\S+'], type(None), value=None)
 
     #: The directory where Spack will install the packages requested by this
@@ -807,6 +812,8 @@ class Spack(BuildSystem):
     #:
     #: :type: :class:`str` or :class:`None`
     #: :default: ``None``
+    #:
+    #: .. versionadded:: 3.7.3
     install_tree = variable(typ.Str[r'\S+'], type(None), value=None)
 
     #: A list of additional specs to build and install within the given
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -784,12 +784,12 @@ class Spack(BuildSystem):
     #: be automatically set to `rfm_spack_env`.
     #:
     #: :type: :class:`str` or :class:`None`
-    #: :default: ``None``
+    #: :default: :class:`None`
     #:
     #: .. note::
     #:     .. versionchanged:: 3.7.3
     #:        The field is no longer required and the Spack environment will be
-    #:        automatically created if not provided
+    #:        automatically created if not provided.
     environment = variable(typ.Str[r'\S+'], type(None), value=None)
 
     #: The directory where Spack will install the packages requested by this
@@ -811,7 +811,7 @@ class Spack(BuildSystem):
     #: will not be overridden.
     #:
     #: :type: :class:`str` or :class:`None`
-    #: :default: ``None``
+    #: :default: :class:`None`
     #:
     #: .. versionadded:: 3.7.3
     install_tree = variable(typ.Str[r'\S+'], type(None), value=None)
--- a/docs/tutorial_build_automation.rst
+++ b/docs/tutorial_build_automation.rst
@@ -109,8 +109,8 @@ Here is the test's code:
 
 
 When :attr:`~reframe.core.pipeline.RegressionTest.build_system` is set to ``'Spack'``, ReFrame will leverage Spack environments in order to build the test code.
-For this reason, currently, users must specify an environment.
-ReFrame treats Spack environments as *test resources* so it expects to find them under the test's :attr:`~reframe.core.pipeline.RegressionTest.sourcesdir`, which defaults to ``'src'``.
+If the environment is not specified by the user, ReFrame will automatically create one inside the stage directory.
+ReFrame treats Spack environments specified by the user as *test resources* so it expects to find them under the test's :attr:`~reframe.core.pipeline.RegressionTest.sourcesdir`, which defaults to ``'src'``.
 Here is the directory structure for the test in this particular example that we show here:
 
 .. code:: console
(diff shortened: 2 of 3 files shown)


reframe/core/buildsystems.py
```diff
@@ -826,7 +826,8 @@ class Spack(BuildSystem):
     #: ReFrame looks for environments in the test's
     #: :attr:`~reframe.core.pipeline.RegressionTest.sourcesdir`.
     #:
-    #: This field is required.
+    #: If this field is `None`, the default, the environment name will
+    #: be automatically set to `rfm_spack_env`.
```
giordano, line 830: I left the reference to `None` because it was already there, but I'm not sure that's accurate:
https://github.com/eth-cscs/reframe/blob/24321bb7e5a9f7a7e4e4486882d326c6809cad0f/reframe/core/buildsystems.py#L833
https://github.com/eth-cscs/reframe/blob/24321bb7e5a9f7a7e4e4486882d326c6809cad0f/reframe/core/buildsystems.py#L866

reframe/core/buildsystems.py
```diff
@@ -826,7 +826,8 @@ class Spack(BuildSystem):
     #: ReFrame looks for environments in the test's
     #: :attr:`~reframe.core.pipeline.RegressionTest.sourcesdir`.
```
giordano, line 827: Also, I don't think this is true, as far as I can tell the environment is looked in the stage directory

vkarak: The contents of the `sourcesdir` are copied over to the stage directory before entering the compile stage. The environment is a test "resource", so it has to reside in the `sourcesdir` if given.

giordano:
```console
% spack env create -d tutorials/basics/stream/spack-env
==> Updating view at /home/mose/repo/reframe/tutorials/basics/stream/spack-env/.spack-env/view
==> Created environment in /home/mose/repo/reframe/tutorials/basics/stream/spack-env
==> You can activate this environment with:
==>   spack env activate /home/mose/repo/reframe/tutorials/basics/stream/spack-env
% ./bin/reframe -c tutorials/basics/stream/stream1.py -r --performance-report
[ReFrame Setup]
  version:           3.7.0-dev.2+1bc108c6
  command:           './bin/reframe -c tutorials/basics/stream/stream1.py -r --performance-report'
  launched by:       mose@dream
  working directory: '/home/mose/repo/reframe'
  settings file:     '/home/mose/repo/reframe/tutorials/config/mysettings.py'
  check search path: '/home/mose/repo/reframe/tutorials/basics/stream/stream1.py'
  stage directory:   '/home/mose/repo/reframe/stage'
  output directory:  '/home/mose/repo/reframe/output'

[==========] Started on Thu Jun 24 23:28:26 2021 

[----------] started processing Test (Test)
[ RUN      ] Test on catalina:default using gnu
[     FAIL ] (1/1) Test on catalina:default using gnu [compile: 0.784s run: n/a total: 0.793s]
==> test failed during 'compile_wait': test staged in '/home/mose/repo/reframe/stage/catalina/default/gnu/Test'
[----------] finished processing Test (Test)

[  FAILED  ] Ran 1/1 test case(s) from 1 check(s) (1 failure(s), 0 skipped)
[==========] Finished on Thu Jun 24 23:28:27 2021 

SUMMARY OF FAILURES
FAILURE INFO for Test 
  * Test Description: Test
  * System partition: catalina:default
  * Environment: gnu
  * Stage directory: /home/mose/repo/reframe/stage/catalina/default/gnu/Test
  * Node list: 
  * Job type: local (id=None)
  * Dependencies (actual): []
  * Maintainers: []
  * Failing phase: compile_wait
  * Rerun with '-n Test -p gnu --system catalina:default -r'
  * Reason: build error: stdout: 'rfm_Test_build.out', stderr: 'rfm_Test_build.err'
--- rfm_Test_build.err (first 10 lines) ---
==> Error: No such environment: 'spack-env'
==> Error: `spack add` requires an environment
  activate an environment first:
      spack env activate ENV
  or use:
      spack -e ENV add ...
--- rfm_Test_build.err --- 

Log file(s) saved in: '/tmp/rfm-rg3uuekm.log'
```
Either I'm doing something wrong (totally possible but documentation would help :slightly_smiling_face:) or that doesn't work

giordano: Oooh, sorry, now I see: I need to set `sourcesdir` in the test.  Well, I think this is one more reason for this pull request :slightly_smiling_face:

reframe/core/buildsystems.py
```diff
@@ -886,8 +885,12 @@ def emit_build_commands(self, environ):
         return ret
 
     def _env_activate_cmds(self):
-        return [f'. $SPACK_ROOT/share/spack/setup-env.sh',
-                f'spack env activate -V -d {self.environment}']
+        cmds = ['. $SPACK_ROOT/share/spack/setup-env.sh']
+        if self.environment is _Undefined:
+            self.environment = 'rfm_spack_env'
+            cmds.append(f'spack env create -d {self.environment}')
```
vkarak, line 891: Leave a blank line after this one.

vkarak (commented):
I tried this PR with the following test and it works:

```python
import reframe as rfm


@rfm.simple_test
class reframe_spack_env_check(rfm.RegressionTest):
    valid_systems = ['*']
    valid_prog_environs = ['*']
    build_system = 'Spack'
    executable = 'bzip2'
    executable_opts = ['--help']

    @run_before('sanity')
    def set_sanity(self):
        self.sanity_patterns = sn.assert_found(r'Version 1.0.7', self.stderr)

    @run_before('compile')
    def setup_build_system(self):
        self.build_system.specs = ['bzip2@1.0.7']
```

However, when I run the test and then do `spack find bzip2`, I see 1.0.7 installed in my system. Is this expected? I'm not that familiar with Spack, so I might miss something. @victorusu can you also have a look?

vkarak (changes requested):
So the problem with the automatically generated environment is that it doesn't set correctly the installation paths. Ideally, we want this to be installed in the stage directory of the test and not "pollute" the user's installation, unless explicitly required.

giordano:
How is this different from an environment manually created from the user?  I guess I'm missing something, but then we're back to the fact documentation isn't helpful at the moment and I'm trying to reverse engineer what I'm supposed to do :slightly_smiling_face:

victorusu:
@giordano, I think the issue is that we haven't documented enough the current goals of this feature.

AFAIK, what we wanted in the beginning is to test Spack packages and have them installed in test's `stagedir`. Then, we would see how it works in the wild and evolve it to achieve other things.
Therefore, we left the environments out-of-the-box to have this discussion later and told people to enable the environment before starting reframe. In this case, the user can set the environment as they want and still leverage the feature from reframe.

One of the reasons for this decision was that we did not know, in principle, which options of the `Spack.yaml` file we would need to add to the build system. 
So we were waiting for two things:
1. To have feedback from he Community
2. As you know Spack evolves very fast and so we would like to wait a bit until environments also become more and more stable in Spack.

So, we postponed environment's inclusion/discussion.

Now, as you have pointed out, the time to have this discussion has come!😄 

@vkarak and @giordano, what do you think about adding a build system option to set the `install_dir`?
In this case, it would create on the stage dir a config file from Spack that would set the `install_dir` and we could make it point to the `stagedir`, by default?

giordano:
To be clear, what you're suggesting is to automatically run something like
```
spack config add "config:install_tree:${install_dir}"
```
after activating the environment, where `${install_dir}` is the chosen directory, something like `${stagedir}/prefix` by default?  That'd be fine with me as long as there is a reasonable default which doesn't require the user to do anything extra: as a user and someone who will need to tell other people how to use this, I wouldn't be particularly happy about _having_ to manually create the environment with the right configuration, maybe outside of reframe.

I'd love to have something as simple as this
```python
import reframe as rfm


@rfm.simple_test
class Test(rfm.RegressionTest):
    valid_systems = ['*']
    valid_prog_environs = ['*']
    build_system = 'Spack'
    executable = 'myprogram'

    @run_before('compile')
    def set_compiler_flags(self):
        self.build_system.specs = ['program@1.0']

    @run_before('sanity')
    def set_sanity_patterns(self):
        self.sanity_patterns = sn.assert_found(r'SUCCESS', self.stdout)
```
to work without anything extra needed.  As far as I understand, at the moment this isn't possible, and that's why I'm trying to hack the Spack build system with this pull request :slightly_smiling_face:

victorusu:
> To be clear, what you're suggesting is to automatically run something like
> 
> ```
> spack config add "config:install_tree:${install_dir}"
> ```
> 
> after activating the environment, where `${install_dir}` is the chosen directory, something like `${stagedir}/prefix` by default? That'd be fine with me as long as there is a reasonable default which doesn't require the user to do anything extra: as a user and someone who will need to tell other people how to use this, I wouldn't be particularly happy about _having_ to manually create the environment with the right configuration, maybe outside of reframe.
> 
> I'd love to have something as simple as this
> 
> ```python
> import reframe as rfm
> import reframe.utility.sanity as sn
> 
> 
> @rfm.simple_test
> class Test(rfm.RegressionTest):
>     valid_systems = ['*']
>     valid_prog_environs = ['*']
>     build_system = 'Spack'
>     executable = 'myprogram'
> 
>     @run_before('compile')
>     def set_compiler_flags(self):
>         self.build_system.specs = ['program@1.0']
> 
>     @run_before('sanity')
>     def set_sanity_patterns(self):
>         self.sanity_patterns = sn.assert_found(r'SUCCESS', self.stdout)
> ```
> 
> to work without anything extra needed. As far as I understand, at the moment this isn't possible, and that's why I'm trying to hack the Spack build system with this pull request 🙂

@giordano, that's exactly what I am saying. @vkarak what do you think?

vkarak:
@giordano Yes, the rationale from @victorusu is correct about the initial purpose of the feature. We have been seeing it also as feature that would help in CI or to do parameter exploration on different builds. We never thought about it seriously as a means of deployment, because there can be too many parameters to consider there. Also I totally agree with what you'd like to see from a user's perspective. My initial objection was that, as a user, I would not like to see my global environment to be implicitly modified after running a reframe test. So by default the `install_dir` of the environment you create should point to the stage dir.

@victorusu For this PR, I would say that emitting the `spack config add ...` should be enough. If users want to customize the environment, they can simply create one by themselver for the moment. As far as I can see if `spack config` is done inside the environment then it's scoped in that. So that should be fine, imho.

giordano:
Ok, I added an `install_tree` (used same name as Spack) to the build system.  This can be either a non-empty string or `None`:
* if `environment` and `install_tree` are both `None`, then `install_tree` is automatically set to `install_tree` and this is used to set Spack's install tree
* if `environment` is not `None` but `install_tree` is, nothing happens (I feel like if the user provides their own environment, they're going to configure it themselves)

reframe/core/buildsystems.py
vkarak: Sorry I meant before this one... I can fix that, no worries.

reframe/core/buildsystems.py
vkarak: Actually, this can be stale. I would change it to `None`. @rsarm does it ring a bell to you?

reframe/core/buildsystems.py
```diff
@@ -865,16 +885,19 @@ def __init__(self):
...
         self._prefix_save = None
 
     def emit_build_commands(self, environ):
         self._prefix_save = os.getcwd()
-        if self.environment is _Undefined:
-            raise BuildSystemError(f'no Spack environment is defined')
 
         ret = self._env_activate_cmds()
+        if self.install_tree:
+            ret.append('spack config add '
```
vkarak, line 898: Just for symmetry:

```suggestion
            ret.append(f'spack config add '
```

reframe/core/buildsystems.py
```diff
@@ -887,8 +910,15 @@ def emit_build_commands(self, environ):
         return ret
 
     def _env_activate_cmds(self):
-        return [f'. $SPACK_ROOT/share/spack/setup-env.sh',
-                f'spack env activate -V -d {self.environment}']
+        cmds = ['. $SPACK_ROOT/share/spack/setup-env.sh']
+        if self.environment is _Undefined:
+            self.environment = 'rfm_spack_env'
+            if self.install_tree is None:
+                self.install_tree = 'install_tree'
```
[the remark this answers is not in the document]
vkarak: Also I would not set the `install_tree` like that but rather do

```python
install_tree = self.install_tree or 'opt/spack'
```

And then use `install_tree`. The reason for that is that we don't want to change the stage of the object: user had it as `None` and then it magically becomes `opt/spack`.

vkarak (changes requested):

reframe/core/buildsystems.py
giordano: This was changed in #2095

giordano:
Ok, I finally got back to this pull request, rebased on latest `master`, and addressed all comments.  Main relevant change now is that the default value of `environment` and `install_tree` is an empty string (which is a bit like going the opposite direction of #2095) :slightly_smiling_face:

vkarak (changes requested):
Lgtm. I just have some minor suggestions. I could also have a look at the tutorial to see if it needs to be adjusted.

giordano:
I mentioned in the tutorial that the environment isn't required and it'll be automatically created by ReFrame if not provided.  I didn't mention `install_tree` there, do you think it's necessary?

vkarak (approved):
> I mentioned in the tutorial that the environment isn't required and it'll be automatically created by ReFrame if not provided. I didn't mention install_tree there, do you think it's necessary?

No, it's not necessary. It's enough the we mention it in the reference.

Lgtm now! Going in.

vkarak:
ok to test

giordano:
Thanks!

PR #2040, opened 2021-06-28
[test] Update OSU benchmark to include prgenv nvidia

ekouts:
Add `PrgEnv-nvidia` in OSU benchmark and update syntax of the tests

### DIFF
--- a/cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
+++ b/cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
@@ -38,6 +38,7 @@ class AlltoallTest(rfm.RegressionTest):
             'num_switches': 1
         }
     }
+
     @run_after('init')
     def set_tags(self):
         self.tags = {self.variant, 'benchmark', 'craype'}
@@ -45,7 +46,6 @@ class AlltoallTest(rfm.RegressionTest):
     @run_before('compile')
     def set_makefile(self):
         self.build_system.makefile = 'Makefile_alltoall'
-
 
     @run_before('run')
     def set_num_tasks(self):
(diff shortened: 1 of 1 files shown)


cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -3,100 +3,139 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
+from reframe.core.decorators import run_after
```
jjotero, line 6: We don't need this, right?

jjotero (changes requested):

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
ekouts: Good point, I removed it

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -7,96 +7,134 @@
...
+    valid_systems = ['daint:gpu', 'dom:gpu']
+    descr = 'Alltoall OSU microbenchmark'
+    build_system = 'Make'
+    executable = './osu_alltoall'
+    # The -m option sets the maximum message size
+    # The -x option sets the number of warm-up iterations
+    # The -i option sets the number of iterations
+    executable_opts = ['-m', '8', '-x', '1000', '-i', '20000']
+    valid_prog_environs = ['PrgEnv-cray', 'PrgEnv-gnu',
+                           'PrgEnv-intel', 'PrgEnv-nvidia']
```
jgphpc, line 23: For some reason, reframe does not unload PrgEnv-cray in my scripts, only:
```
module load PrgEnv-nvidia
module load cdt/21.05
```
Do you get the same ?

ekouts: Strange, I am looking now at the stage directory of one of the tests (in  dom) and it looks like it does
```
...
module load daint-gpu
module unload PrgEnv-cray
module load PrgEnv-nvidia
module load cdt/21.05
...
```
Where did you run it?

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -194,36 +235,37 @@ def __init__(self):
...
-                'latency': (2.33, None, 0.15, 'us')
-            },
-            'pilatus:mc': {
-                'latency': (2.33, None, 0.15, 'us')
-            },
-            # keeping as reference:
-            # 'monch:compute': {
-            #     'latency': (1.27, None, 0.1, 'us')
-            # },
+    executable = './p2p_osu_latency'
```
jgphpc, line 238: valid_systems missing ?

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -233,70 +275,102 @@ def __init__(self):
...
+        self.build_system.cppflags = ['-D_ENABLE_CUDA_']
+
+    @run_before('compile')
+    def set_modules(self):
         if self.current_system.name in ['daint', 'dom']:
             self.num_gpus_per_node  = 1
-            self.modules = ['craype-accel-nvidia60']
             self.variables = {'MPICH_RDMA_ENABLED_CUDA': '1'}
+            if self.current_environ.name == 'PrgEnv-nvidia':
+                self.modules = ['cdt-cuda/21.05']
```
jgphpc, line 313: I did not know before (sorry) but can you try with adding `cudatoolkit/21.3_11.2` to the list of modules and remove the ldflags to see if it works ?

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -233,70 +275,102 @@ def __init__(self):
...
```
jgphpc, line 363: same comment

jgphpc (approved):
lgtm

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
ekouts: you are right, I deleted by mistake, thanks!

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
ekouts: It is working with `cudatoolkit/21.3_11.2`, I will update the test 👍 The test is also passing without `cdt-cuda/21.05`, should I remove it as well?

jgphpc: I would say yes, remove it.

jjotero: But having the `ldflags` seems a more maintainable solution for the future than hardcoding a specific `cudatoolkit` version, right?

jgphpc: I agree (but then who will test the cudatoolkit/cray wrapper ?).
Both work, what do you think the user will prefer ?

jjotero: Not sure, but we need to stay consistent here. Now some tests do the `ldflags` by hand and others use the `cudatoolkit` wrapper. If you want to check that all those work, we should probably write a check under `prgenv`, and not test that under with a `microbenchmark`.

jjotero (approved):

jjotero:
This should get the PEP issues fixed before merging.

PR #2045, opened 2021-06-30
[feat] Show % of reference value in performance report

InternetUnexplorer:
This adds a little section in the performance report that compares each value to its corresponding reference value. It's a tiny change, but something I thought might be useful, and it's also simple enough that I thought it might be a good first contribution.

As an example, the performance report for `tutorials/basics/stream/stream2.py` now looks like this:
```text
PERFORMANCE REPORT
StreamWithRefTest
- prism:default
   - gnu
      * num_tasks: 1
      * Copy: 34477.1 MB/s (137% of 25200 MB/s)
      * Scale: 22701.7 MB/s (135% of 16800 MB/s)
      * Add: 24816.0 MB/s (134% of 18500 MB/s)
      * Triad: 24675.3 MB/s (131% of 18800 MB/s)
```
Also, as part of this change, `(no unit specified)` is no longer shown when there's no unit.

Feedback is welcome, especially since this is a visual change. Here are some possible changes I thought of (reference values changed for demonstration purposes):
<details>
 <summary>Increasing spacing before details</summary>

```text
PERFORMANCE REPORT
StreamWithRefTest
- prism:default
   - gnu
      * num_tasks: 1
      * Copy: 34496.6 MB/s  (3% of 1000000 MB/s)
      * Scale: 22695.0 MB/s  (45% of 50000 MB/s)
      * Add: 24926.0 MB/s  (125% of 20000 MB/s)
      * Triad: 24424.8 MB/s  (2442% of 1000 MB/s)
```
</details>
<details>
 <summary>Two columns</summary>

```text
PERFORMANCE REPORT
StreamWithRefTest
- prism:default
   - gnu
      * num_tasks: 1
      * Copy: 34496.6 MB/s          (3% of 1000000 MB/s)
      * Scale: 22695.0 MB/s         (45% of 50000 MB/s)
      * Add: 24926.0 MB/s           (125% of 20000 MB/s)
      * Triad: 24424.8 MB/s         (2442% of 1000 MB/s)
```
</details>
<details>
 <summary>Aligned details</summary>

```text
PERFORMANCE REPORT
StreamWithRefTest
- prism:default
   - gnu
      * num_tasks: 1
      * Copy: 34496.6 MB/s  (3% of 1000000 MB/s)
      * Scale: 22695.0 MB/s (45% of 50000 MB/s)
      * Add: 24926.0 MB/s   (125% of 20000 MB/s)
      * Triad: 24424.8 MB/s (2442% of 1000 MB/s)
```
</details>
<details>
 <summary>Aligned everything</summary>

```text
PERFORMANCE REPORT
StreamWithRefTest
- prism:default
   - gnu
      * num_tasks: 1
      * Copy:  34496.6 MB/s (   3% of 1000000 MB/s)
      * Scale: 22695.0 MB/s (  45% of 50000   MB/s)
      * Add:   24926.0 MB/s ( 125% of 20000   MB/s)
      * Triad: 24424.8 MB/s (2442% of 1000    MB/s)
```
</details>

…or some combination of the above. The goal is to avoid making things too visually noisy/hard to read at a glance.

Again, feedback is welcome! I did my best to adhere to the style/contribution guidelines, but I might've missed things. And feedback on whether the output is easy to read as-is or if could be improved (or even whether this feature is even worth adding at all) would be appreciated too.

### DIFF
--- a/reframe/frontend/statistics.py
+++ b/reframe/frontend/statistics.py
@@ -296,32 +296,43 @@ class TestStats:
         previous_name = ''
         previous_part = ''
         for t in self.tasks():
-            if t.check.perfvalues.keys():
-                if t.check.name != previous_name:
-                    report_body.append(line_width * '-')
-                    report_body.append(t.check.name)
-                    previous_name = t.check.name
+            if not t.check.perfvalues.keys():
+                continue
 
-                if t.check.current_partition.fullname != previous_part:
-                    report_body.append(
-                        f'- {t.check.current_partition.fullname}')
-                    previous_part = t.check.current_partition.fullname
+            rows = []
+            if t.check.name != previous_name:
+                report_body.append(line_width * '-')
+                report_body.append(t.check.name)
+                previous_name = t.check.name
 
-                report_body.append(f'   - {t.check.current_environ}')
-                report_body.append(f'      * num_tasks: {t.check.num_tasks}')
+            if t.check.current_partition.fullname != previous_part:
+                report_body.append(f'- {t.check.current_partition.fullname}')
+                previous_part = t.check.current_partition.fullname
+
+            report_body.append(f'   - {t.check.current_environ}')
+            report_body.append(f'      * num_tasks: {t.check.num_tasks}')
 
             for key, ref in t.check.perfvalues.items():
                 var = key.split(':')[-1]
                 val = ref[0]
                 unit = ' ' + ref[4] if ref[4] else ''
-                ref_val = ref[1]
-                if ref_val:
-                    percent = round(val / ref_val * 100)
-                    details = f' ({percent}% of {ref_val}{unit})'
-                else:
+                if ref[1:4] == (0, None, None):
+                    # No reference value present to compare with
                     details = ''
+                else:
+                    ref_val = ref[1]
+                    percent = (round(val / ref_val * 100) if ref_val != 0
+                               else 'N/A')
+                    details = f' ({percent:3}% of {ref_val}{unit})'
 
-                report_body.append(f'      * {var}: {val}{unit}{details}')
+                rows.append((f'{var}:', f'{val}{unit}', details))
+
+            var_width = max(map(lambda row: len(row[0]), rows))
+            val_width = max(map(lambda row: len(row[1]), rows))
+            for var, val, details in rows:
+                var_just = var.ljust(var_width)
+                val_just = val.ljust(val_width)
+                report_body.append(f'      * {var_just} {val_just}{details}')
 
         if report_body:
             return '\n'.join([report_start, report_title, *report_body,
--- a/reframe/frontend/statistics.py
+++ b/reframe/frontend/statistics.py
@@ -313,12 +313,15 @@ class TestStats:
             for key, ref in t.check.perfvalues.items():
                 var = key.split(':')[-1]
                 val = ref[0]
-                try:
-                    unit = ref[4]
-                except IndexError:
-                    unit = '(no unit specified)'
+                unit = ' ' + ref[4] if ref[4] else ''
+                ref_val = ref[1]
+                if ref_val:
+                    percent = round(val / ref_val * 100)
+                    details = f' ({percent}% of {ref_val}{unit})'
+                else:
+                    details = ''
 
-                report_body.append(f'      * {var}: {val} {unit}')
+                report_body.append(f'      * {var}: {val}{unit}{details}')
 
         if report_body:
             return '\n'.join([report_start, report_title, *report_body,
--- a/unittests/test_cli.py
+++ b/unittests/test_cli.py
@@ -377,7 +377,7 @@ def test_performance_report(run_reframe):
         more_options=['-t', 'PerformanceFailureCheck', '--performance-report']
     )
     assert r'PERFORMANCE REPORT' in stdout
-    assert r'perf: 10 Gflop/s (50% of 20 Gflop/s)' in stdout
+    assert r'perf: 10 Gflop/s ( 50% of 20 Gflop/s)' in stdout
 
 
 def test_skip_system_check_option(run_reframe):
--- a/unittests/test_cli.py
+++ b/unittests/test_cli.py
@@ -377,7 +377,7 @@ def test_performance_report(run_reframe):
         more_options=['-t', 'PerformanceFailureCheck', '--performance-report']
     )
     assert r'PERFORMANCE REPORT' in stdout
-    assert r'perf: 10 Gflop/s' in stdout
+    assert r'perf: 10 Gflop/s (50% of 20 Gflop/s)' in stdout
 
 
 def test_skip_system_check_option(run_reframe):


reframe/frontend/statistics.py
```diff
@@ -313,12 +313,15 @@ def performance_report(self):
...
                 var = key.split(':')[-1]
                 val = ref[0]
-                try:
-                    unit = ref[4]
-                except IndexError:
-                    unit = '(no unit specified)'
+                unit = ' ' + ref[4] if ref[4] else ''
+                ref_val = ref[1]
+                if ref_val:
+                    percent = round(val / ref_val * 100)
```
victorusu, line 319: I think that we should handle the case, where `ref_val` is `0`.
What do you think?

unittests/test_cli.py
```diff
@@ -377,7 +377,7 @@ def test_performance_report(run_reframe):
         more_options=['-t', 'PerformanceFailureCheck', '--performance-report']
     )
     assert r'PERFORMANCE REPORT' in stdout
-    assert r'perf: 10 Gflop/s' in stdout
+    assert r'perf: 10 Gflop/s (50% of 20 Gflop/s)' in stdout
```
victorusu, line 380: Is the `20 Gflop/s` generic enough? 
Would be the same if we run on different systems?

victorusu:
@InternetUnexplorer, thanks for proposing this change! I like it! 
I just have a couple of questions. 
1. I am not sure how you want to deal with the `ref_val = 0`.
2. Not sure if the unit test will pass on every system.

unittests/test_cli.py
ekouts: @victorusu I think this is the reference we set in `PerformanceFailureCheck` so it should be fine, right?

jjotero:
Hmm... I'm not sure there is a need for this. What's the benefit of reporting all this extra stuff? Usually, reference performance values are set on a ballpark approach, so knowing that a test performed at 90% or 120% relative to that value doesn't mean anything. And in those situations, all these extra data just buries the actual performance results with extra noise.

In fact, if you want that percentage reported, you can simply create a new performance variable that reports just that.

brandongc:
@jjotero I don't think all reference values are so arbitrary. For example, something like stream has fairly definite value based on the hardware. Also, at NERSC we are working on filling our references out with more rigorous statistics now that we have a good collection of runs over time to draw from. 

As for the benefit, I think having the reference value available at the time of the report is helpful for the person running reframe to decide if the "ballpark reference" is reasonable, the percentage aspect is just a bonus for convenience.  Otherwise, currently I have to go look into the test and find what the reference was.

InternetUnexplorer:
As @brandongc said, the goal is to make it easier to compare the values to the reference values, since currently one has to read the checks to see what the reference values are. At the same time, I definitely don't want to make the performance report significantly harder to read…

reframe/frontend/statistics.py
InternetUnexplorer: I believe this is already handled since `bool(0) == False`; in this case the comparison will be skipped altogether.
However, maybe I should edit the condition/add a comment to make it more obvious that this case is handled?

teojgo:
Ok to test

vkarak:
Thanks @InternetUnexplorer for your PR and I think it's useful. I can see also @jjotero's point. His concern -if I interpret it correctly- is that often you might want to run a performance test on a new system as a benchmark, for which you don't have a reference and therefore showing anything related to reference values will seem like noise. But in this case, the reference value is set to zero (by convention), so the "reference" part of the output will not be printed (I will need to try it out). As of the formatting style, I'd vote for the "aligned everything."

reframe/frontend/statistics.py
vkarak: Yes and no. What about the case when the reference value is actually zero? On the other hand, as I mentioned in my other comment, we need a way to distinguish the "unknown reference" situation. By convention the unknown reference is the tuple `(0, None, None)`, so in that case now reference data should be printed.

InternetUnexplorer: I'm not sure how common the case where the reference value is zero would be. I guess it _could_ be useful in theory to know that the reference value is zero, I'm just not sure about in practice. But if you think it's something that should be handled then I'm happy to add it!

In this case, maybe the details section could be:
```text
(ref. = 0 units)
```
…or maybe this? (more consistent with the ref ≠ 0 case)
```text
(N/A of 0 units)
```

InternetUnexplorer: Also, @vkarak am I correct in understanding that the way to distinguish the unknown reference from the zero reference is by checking whether the upper/lower bounds are `None`?

InternetUnexplorer:
Thanks for the feedback @vkarak and everyone else!

I made a few tweaks to the patch (still changing things a bit so haven't pushed yet) to make the formatting a sort of middle ground where things are partially aligned into three columns (the variables, the values, and the details). Here's what it looks like:
```text
PERFORMANCE REPORT
StreamWithRefTest
- prism:default
   - gnu
      * num_tasks: 1
      * Copy:  37339.6 MB/s (4% of 1000000 MB/s)
      * Scale: 24644.0 MB/s (49% of 50000 MB/s)
      * Add:   27235.6 MB/s (136% of 20000 MB/s)
      * Triad: 26993.7 MB/s (2699% of 1000 MB/s)
```
Note that all of the values here are the same length, but even if they weren't the details would still be aligned.

It probably wouldn't be too hard to align the different parts of the details as well, I just wasn't sure how useful it would actually be in practice since it already seems pretty easy to read in my opinion. However, I'm willing to give it a go if others want it.

Also, when making changes to this branch, should I amend and force-push, or just make new commits (and maybe squash them later?). I wasn't quite sure after reading the contribution guidelines.

InternetUnexplorer:
Ok, I made the following changes:
- The name, value, and details sections are now vertically aligned
- The percentage value in the details section is left-padded to 3 digits
  - It's very simple this way, and I assume the probability of a value being ≥10x the reference value is quite low 
- If the reference value is present but zero, `N/A%` is displayed

The performance report now looks like this:
```text
PERFORMANCE REPORT
ExampleTest
- prism:default
   - gnu
      * num_tasks: 1
      * Unitless: 120.3         (120% of 100)
      * Foo:      27092.9 MB/s  (135% of 20000 MB/s)
      * ZeroRef:  1034.4 iter/s (N/A% of 0 iter/s)
```
The code currently checks whether the upper and lower bounds are present to tell the difference between the reference value being zero and the reference value being absent. I'm not sure if this is the correct way to do it, since it _is_ possible to pass `None` as both the upper and lower bounds, but I haven't been able to find a better way to do it.

InternetUnexplorer:
Whoops, forgot to update `test_performance_report` to match, fixed.

jjotero:
Hi @InternetUnexplorer ! Apologies for dragging this PR for so long. I wanted to say that we've been working on a new syntax to express the performance variables in regression tests, and this PR conflicts with some stuff we had in mind. Perhaps expanding a bit on my previous comment, some of the ReFrame users are just interested in logging performance data and not in comparing it with some reference. Currently, ReFrame enforces users to set a reference in order to get the performance variables from `perf_patterns` reported, and this is a limitation for those users just interested in the performance reporting part, which is what this new syntax that we have in the works is trying to address. This new syntax will make the `reference` field entirely optional, so I think it would be rather weird if we have this new field that you're proposing in the performance report which always expects the `reference` to be present. So if someone were to write a minimal ReFrame check without their reference, they would get that `(N/A% of 0<the_perf_var_units>)` by default.

I have a few ideas on how to integrate your proposed changes with the new syntax though, so I'd appreciate if you can have a look at #2083 and let us know what you think. This new perf syntax introduces the concept of a `performance_function` which is basically a decorated method which ReFrame automatically runs during the performance stage. So what I had in mind is that perhaps we could add an extra argument to this `performance_function` decorator where we pass a callable that produces whichever metadata the user wants into the performance report. This would make that additional field fully customisable on a per-variable basis. What do you think?

brandongc:
Would that mean we would need to use this decorator syntax instead of the "simple"?

```python
        self.perf_patterns = {"flops":
                              sn.extractsingle(r'^(\S+\s){10}.*PERF_SUMMARY.*$', self.stdout, 1, float) }
```

Since most of our performance tests have references (or will) then we would need to make a callable, replicate/import that across all the tests and then pass it to each one?

InternetUnexplorer:
Hi @jjotero! Thanks for the feedback, and no worries. I'll close this now for then and take a look at how #2083 works to see if there's a way to adapt this to that.

jjotero:
That would still remain simple with the new syntax. The only thing that changes is that the utils from the sanity module have to be converted into performance functions, but that can also be done inline. The syntax (given the current state of the other PR) would be:
```python
@performance_variables
def set_perf_vars(self):
    ...
    return {
        "flops": self.make_performance_function(
            "FLOPS", sn.extractsingle(r'^(\S+\s){10}.*PERF_SUMMARY.*$', self.stdout, 1, float)
        )
    }
```
but it could also be just 
```python
@performance_function('FLOPS')
def flops(self):
    return sn.extractsingle(r'^(\S+\s){10}.*PERF_SUMMARY.*$', self.stdout, 1, float)
```
Whether you want a reference in here for your `flops` or not is up to you. Also, you can keep adding as many `performance_function`s as you'd like, and reframe will run them all during the performance stage without needing to use the `@performance_variables` decorated function. That `@performance_variables` decorator is only there just in case you want to customise what you report as performance variables, or excluding some of the defined `performance_function`s. This is particularly useful if you think of composable tests where you inherit a base class from someone else and you want to give them some other use.

We are aware that this is a bit of an overhead for simple examples like this one, but the advantage on more complex systems outweighs by far this extra little bit of work.

PR #2046, opened 2021-07-01
[feat] Make the `@required_version` decorator a class definition parameter

### LINKED ISSUE #1797: Make the `@required_version` decorator a built-in
The `@required_version` decorator is an interesting tool to have ReFrame skip incompatible tests. Although we currently don't use it so much, it will become more relevant when we start creating test libraries and tests may be run by different ReFrame versions that may or may not support some features that the test require. The problem with the decorator now is that it is executed too late, i.e., after the class creation, so we couldn't capture features like the pipeline hooks when they were added. Therefore, I suggest making a  built-in, such as the `variable()` and `parameter()`, which execute even before the class body is executed.

ekouts:
Closes #1797.
I liked a bit more the argument option, than the built-in syntax, but we can change it. The new syntax is:
```python
class BaseFoo(rfm.RegressionTest, require_version=['x.x.x', '>y.y.y']):
   ...

@rfm.simple_test
class Foo(BaseFoo):
   ...
```
`Foo` and all descendants will inherit the required versions, unless they specify `require_version` again.


reframe/core/decorators.py
```diff
@@ -91,6 +91,16 @@ def _validate_test(cls):
         raise ValueError(f'decorated test ({cls.__qualname__!r}) has one or '
                          f'more undefined parameters')
 
+    conditions = [VersionValidator(v) for v in cls._rfm_required_version]
+    if (cls._rfm_required_version and
+        not any(c.validate(osext.reframe_version()) for c in conditions)):
+
+        getlogger().info('skipping incompatible test defined'
+                         ' in class: %s' % cls.__name__)
```
jjotero, line 99: I think we need a more explicit message here such as `"test 'MyTest' is incompatible with the current ReFrame version (x.x.x)"`

jjotero (changes requested):
Just a small comment on the warning message. The rest looks good.

docs/tutorial_tips_tricks.rst
```diff
@@ -583,3 +583,34 @@ The following figure shows one part of the automatically generated pipeline for
...
+A version specification string can have one of the following formats:
+
+1. ``VERSION``: Specifies a single version.
+2. ``{OP}VERSION``, where ``{OP}`` can be any of ``>``, ``>=``, ``<``, ``<=``, ``==`` and ``!=``. For example, the version specification string ``'>=3.5.0'`` will allow the following test to be loaded only by ReFrame 3.5.0 and higher. The ``==VERSION`` specification is the equivalent of ``VERSION``.
+3. ``V1..V2``: Specifies a range of versions.
+
+The test will be selected if *any* of the versions is satisfied, even if the versions specifications are conflicting.
+
+.. note::
+   Even though this feature was available already through a class decorator, now you can set it as a class argument.
```
jjotero, line 615: This reads to me like this decorator is still supported. Since we're deprecating it, I would just not mention it at all.

jjotero (commented):
We need to document these arguments in the manuals. I think that the best place is by adding the `required_version` (and `pin_prefix` and `special`) in the docstring of the `RegressionTest` base class.

jjotero:
Another thing that just crossed my mind is that we should probably be consistent with how we name the class arguments. For example, `pin_prefix` indicates an action and `required_version` does not. Changing this arg to `require_version` would make this naming consistent, I think.

reframe/core/decorators.py
```diff
@@ -91,6 +91,17 @@ def _validate_test(cls):
         raise ValueError(f'decorated test ({cls.__qualname__!r}) has one or '
                          f'more undefined parameters')
 
+    conditions = [VersionValidator(v) for v in cls._rfm_required_version]
+    if (cls._rfm_required_version and
+        not any(c.validate(osext.reframe_version()) for c in conditions)):
+
+        getlogger().info(f"skipping test '{cls.__name__}' because it is "
+                         f"incompatible with the current ReFrame version "
+                         f"({osext.reframe_version().split('-')[0]})")
```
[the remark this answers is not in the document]
vkarak: I would also go for a less verbose message. Something like `skipping incompatible test 'X': not valid for ReFrame version Y.Z`.

reframe/core/pipeline.py
```diff
@@ -798,10 +798,16 @@ def __getattribute__(self, name):
         return super().__getattribute__(name)
 
     @classmethod
-    def __init_subclass__(cls, *, special=False, pin_prefix=False, **kwargs):
+    def __init_subclass__(cls, *, special=False, pin_prefix=False,
+                          required_version=[], **kwargs):
```
vkarak, line 802: Don't use mutable objects as default arguments...

```suggestion
                          required_version=None, **kwargs):
```

vkarak (changes requested):
I agree with @jjotero that we should find a proper location to document the class arguments. The natural place would be when documenting the `RegressionTest` class, but I don't know how we can inject in the documentation the arguments of the `__init_subclass__`.

ekouts:
The documentation of the parameters is not ideal. Besides the signature of the `RegressionTest`, the documentation of the parameters doesn't appear exactly right.
Screenshot 2021-07-13 at 08 32 05
The second format of `require_version` has some grey background and the description ("is a list of ReFrame version...") starts from the next line

docs/tutorial_tips_tricks.rst
```diff
@@ -583,34 +583,3 @@ The following figure shows one part of the automatically generated pipeline for
...
-.. code-block:: python
-
-   class BaseFoo(rfm.RegressionTest, required_version=['x.x.x', '>y.y.y']):
-     ...
-
-   @rfm.simple_test
-   class Foo(BaseFoo):
-     ...
-
-
```
vkarak: I would still leave this part (not the detailed description below) but make it more concrete, for example use `>3.7.0' as a version string. And then point to the documentation.

reframe/core/decorators.py
```diff
@@ -199,7 +199,7 @@ def required_version(*versions):
     '''
     warn.user_deprecation_warning(
         'the @required_version decorator is deprecated; '
-        'please set the required_version in the class arguments instead',
+        'please set the require_version in the class arguments instead',
```
vkarak, line 202:
```suggestion
        "please set the 'require_version' parameter in the class definition instead",
```

reframe/core/pipeline.py
```diff
@@ -132,6 +132,25 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
     This class provides the implementation of the pipeline phases that the
     regression test goes through during its lifetime.
 
+    :param require_version: is a list of ReFrame version specifications that
```
vkarak, line 135: We should make explicit that these parameters are set at the *class definition* (not instantiation). The Python documentation also uses that term to classify such parameters.

vkarak (changes requested):
I have a couple of minor comments. I will have a look at the formatting problem you mentioned and I'll try to fix them too.

docs/tutorial_tips_tricks.rst
vkarak: Oh, ok, ignore this comment. I thought that we had this in master.

vkarak (approved):

reframe/core/pipeline.py
```diff
@@ -138,33 +138,49 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
...
+    class can be defined as follows:
+
+    .. code-block:: python
+
+       class MyTest(RegressionTest, param='foo', ...):
+
+    where ``param`` is one of the following:
+
+    :param pin_prefix: the prefix to pin the test to if the test lives in a
+        test library with resources in it.
```
jjotero, line 151: To me this reads as if this parameter requires an actual prefix instead of being a bool flag. How about something like "Lock the test prefix to the directory where the current class lives."?

jjotero (approved):

PR #2047, opened 2021-07-01
[bugfix] Hide data-descriptors from `variable` attributes

### LINKED ISSUE #2000: `variable` descriptor exposed as class attribute after the first class instantiation.
Reproducer:
```python
    v = variable(int, value=1)

print(Foo.v) # Prints 1
Foo()
print(Foo.v) # Prints the variable descriptor object
Foo.v = 2
print(Foo.v) # Still prints the variable descriptor object
```
Even though this was tested for in the unit tests, the behaviour should be consistent regardless of the number of class instantiations.
In short, the variable descriptor should never be exposed like this and the variable's default value should be returned instead.

jjotero:
Closes #2000


vkarak (approved):
lgtm

ekouts (approved):

PR #2048, opened 2021-07-01
[bugfix] Fix pipeline hook override

jjotero:
This PR fixes a hook attachment leak occurring when a given hook from a base class is overridden by a derived class.

When overriding a pipeline hook, the method in the derived class would still get attached to the pipeline despite that it is not decorated to do so. See the example below:
```python
@rfm.simple_test
    valid_systems = ['*']
    valid_prog_environs = ['*']
    executable = 'echo bananas'

    @sanity_function
    def assert_bananas(self):

    @run_after('setup')
    def hh(self):
        print('bananas')

@rfm.simple_test
class Derived(Base):
    def hh(self):
        print('not bananas')
```
Running this gives:
```bash
[==========] Started on Thu Jul  1 13:06:40 2021 

[----------] started processing Base (Base)
[ RUN      ] Base on dom:login using PrgEnv-gnu
bananas
[----------] finished processing Base (Base)

[----------] started processing Derived (Derived)
[ RUN      ] Derived on dom:login using PrgEnv-gnu
not bananas
[----------] finished processing Derived (Derived)

[       OK ] (1/2) Base on dom:login using PrgEnv-gnu [compile: 0.006s run: 1.239s total: 1.266s]
[       OK ] (2/2) Derived on dom:login using PrgEnv-gnu [compile: 0.014s run: 1.316s total: 1.349s]

[  PASSED  ] Ran 2/2 test case(s) from 2 check(s) (0 failure(s), 0 skipped)
[==========] Finished on Thu Jul  1 13:06:42 2021 
```
The method `Derived.hh` still gets attached to the `post_setup` stage, despite that this method has not even been decorated with any `run_before` or `run _after` decorators.


victorusu (changes requested):
@jjotero, I think we need a unit test for this PR.
This behavior should be guaranteed to not break.
What do you think about adding your example as a unit test? 
And also another unit test that re-attaches the hook?

ekouts (approved):

reframe/core/hooks.py
```diff
@@ -170,10 +170,15 @@ def __contains__(self, key):
     def __getattr__(self, name):
         return getattr(self.__hooks, name)
 
-    def update(self, hooks):
+    def update(self, hooks, *, denied_hooks={}):
```
vkarak, line 173:
```suggestion
    def update(self, hooks, *, denied_hooks=None):
```

And set it in the body with `denied_hooks = denied_hooks or set()`. Otherwise (a) the default arg. is not a set and (b) more importantly, the default argument is mutable here.

reframe/core/hooks.py
```diff
@@ -170,10 +170,15 @@ def __contains__(self, key):
...
+        '''Update the hook registry with the hooks from another hook registry.
+
+        The optional ``denied_hooks`` argument takes a set of disallowed
+        hook names, preventing their inclusion into the current hook registry.
+        '''
         for phase, hks in hooks.items():
             self.__hooks.setdefault(phase, util.OrderedSet())
-            for h in hks:
+            for h in (hk for hk in hks if hk.__name__ not in denied_hooks):
                 self.__hooks[phase].add(h)
```
vkarak, line 182:
```suggestion
            for h in hks:
                if h.__name__ not in denied_hooks:
                    self.__hooks[phase].add(h)
```

We only need to iterate over `hks` once.

docs/regression_test_api.rst
```diff
@@ -204,7 +204,7 @@ You can attach arbitrary functions to run before or after any pipeline stage, wh
 Multiple hooks can be attached before or after the same pipeline stage, in which case the order of execution will match the order in which the functions are defined in the class body of the test.
 A single hook can also be applied to multiple stages and it will be executed multiple times.
 All pipeline hooks of a test class are inherited by its subclasses.
-Subclasses may override a pipeline hook of their parents by redefining the hook function and re-attaching it at the same pipeline stage.
+Subclasses may override a pipeline hook of their parents by redefining the hook function, where the overriding function may be reattached to any pipeline stage.
```
vkarak, line 207: What about something like this?

```suggestion
Subclasses may override a pipeline hook of their parents by redefining the hook function. The overriding function will not be a pipeline hook anymore, unless explicitly decorated, in which case it can be attached to any pipeline stage.
```

vkarak (changes requested):
I have some minor comments + I agree with @victorusu to provide a unit test.

docs/regression_test_api.rst
jjotero: What about this? _The overriding function will not be a pipeline hook unless explicitly decorated, and it may be reattached to any pipeline stage._

reframe/core/hooks.py
jjotero: Oh my ... what a rookie mistake 🙈!

docs/regression_test_api.rst
vkarak: Sorry, I just found the time to reply to this. Yes, your version reads better.

unittests/test_meta.py
```diff
@@ -160,3 +160,44 @@ def my_deferrable(self):
...
+                return False
+
+    assert Foo.hook_in_stage('hook_a', 'post_setup')
+    assert Foo.hook_in_stage('hook_b', 'pre_compile')
+    assert Foo.hook_in_stage('hook_c', 'post_run_wait')
+
+    class Bar(Foo):
+        @run_before('sanity')
+        def hook_a(self):
+            "Convert to a pre-sanity hook"
```
vkarak, line 195:
```suggestion
            '''Convert to a pre-sanity hook'''
```

unittests/test_meta.py
```diff
@@ -160,3 +160,44 @@ def my_deferrable(self):
...
+
+        def hook_b(self):
+            "No longer a hook"
```
vkarak, line 198:
```suggestion
            '''No longer a hook'''
```

vkarak (changes requested):
Lgtm, just a very minor comment.

vkarak (approved):

victorusu (approved):
`lgtm`

PR #2051, opened 2021-07-02
[packaging] Add `--no-deps` option to ignore dependency conflicts when bootstrapping

jjotero:


teojgo (approved):

victorusu (approved):
`lgtm`

bootstrap.sh
```diff
@@ -35,6 +35,7 @@ usage()
     echo "  -h       Print this help message and exit"
     echo "  +docs    Build also the documentation"
     echo "  +pygelf  Install also the pygelf Python package"
+    echo "  +nodeps  Ignore incompatible Python package dependencies"
```
vkarak, line 38: This should be starting with `-`. The `+` is for features to enable.

vkarak (changes requested):

bootstrap.sh
victorusu: Great catch!

jjotero: Uops! Done!

bootstrap.sh
```diff
@@ -35,6 +35,7 @@ usage()
     echo "  -h       Print this help message and exit"
     echo "  +docs    Build also the documentation"
     echo "  +pygelf  Install also the pygelf Python package"
+    echo "  -nodeps  Ignore incompatible Python package dependencies"
```
vkarak, line 38: I would use `--no-deps` for POSIX compatibility and move it up in the help message.

vkarak (changes requested):

vkarak:
What about instead of `--no-deps` adding a `--pip-opts` option that would pass through any option to `pip`?

vkarak (approved):

PR #2058, opened 2021-07-07
[feat] Expose `@final` decorator in the class body

### LINKED ISSUE #2001: Expose the `final` decorator through the metaclass
This should follow the same logic as all the other buitins. We need to move the `final` decorator into the metaclass and also deprecate its use as `rfm.final`.

jjotero:
One now can do
```python
    @final
    def foo(self):
        ...
```
instead of having to use the `@rfm.final` syntax.

### Other changes:

- Renamed `_rfm_special_test` to `_rfm_override_final`. This flag is only used in the metaclass to allow the override of final methods, so this name is more descriptive. Also, the "special" naming of it should just be down to the `RegressionTest` class.
- The unit test of this functionality is now in `unit tests/tests_meta.py` instead of the file with the loader tests.
- The `run` stage of the `RunOnlyRegressionTest` was calling `super().run.__wrapped__(self)`. I've replaced this line for `super.run()` since now the `final` decorator does not create a `functors.wrap` wrapper.

Closes #2001

### DIFF
--- a/unittests/test_meta.py
+++ b/unittests/test_meta.py
@@ -52,6 +52,7 @@ def test_directives(MyMeta):
         deferrable(ext)
         sanity_function(ext)
         v = required
+        final(ext)
 
         def __init__(self):
             assert not hasattr(self, 'parameter')
@@ -63,6 +64,7 @@ def test_directives(MyMeta):
             assert not hasattr(self, 'deferrable')
             assert not hasattr(self, 'sanity_function')
             assert not hasattr(self, 'required')
+            assert not hasattr(self, 'final')
 
     MyTest()
 
@@ -201,3 +203,15 @@ def test_hook_attachments(MyMeta):
     assert not Bar.hook_in_stage('hook_b', 'pre_compile')
     assert Bar.hook_in_stage('hook_c', 'post_run_wait')
     assert Bar.hook_in_stage('hook_a', 'pre_sanity')
+
+
+def test_final(MyMeta):
+    class MyBase(MyMeta):
+        @final
+        def foo(self):
+            pass
+
+    with pytest.raises(ReframeSyntaxError):
+        class MyDerived(MyBase):
+            def foo(self):
+                '''Override attempt'''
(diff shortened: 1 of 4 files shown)


jjotero:
@jenkins-cscs retry daint

unittests/test_meta.py
```diff
@@ -201,3 +203,22 @@ def hook_b(self):
...
+            pass
+
+    with pytest.raises(ReframeSyntaxError):
+        class Derived(Base):
+            def foo(self):
+                '''Override attempt.'''
+
+    class AllowFinalOverride(Base):
+        '''Use flag to bypass the final override check.'''
+        _rfm_override_final = True
```
vkarak, line 221: Why not like this, which is the documented way?

```suggestion
    class AllowFinalOverride(Base, special=True):
        '''Use flag to bypass the final override check.'''
```

vkarak (changes requested):
Lgtm. I have just a minor comment.

unittests/test_meta.py
jjotero: This test just checks that the metaclass machinery to override `final` methods works. The `special=True` belongs to the API defined in the `RegressionTest` class and should not be tested for here. That check is already in `unittests/test_loader.py`.

ekouts (approved):

victorusu:
@jjotero and @vkarak, I cannot understand why we have a way to override the `@final`.
If it is final, then it should be final, no?
I would vote for not having `@final` at all, if one can bypass it in a way that is supported by the framework...
What do you think?
Am I being too radical?

jjotero:
@victorusu This is needed by the `RunOnlyRegressionTest` and `CompileOnlyRegressionTest` classes. The pipeline methods in the base `RegressionTest` are marked as `final` to ensure that users do not accidentally override the methods and break the pipeline, but a "master override" method is needed to specialise these two derived classes.

victorusu (approved):
`lgtm`

vkarak (changes requested):
The only thing we don't check in the unit tests is that a deprecation warning is thrown in case of the old syntax. I think that's why "coverage" is not very happy.

vkarak (approved):

vkarak:
@jenkins-cscs retry daint

PR #2059, opened 2021-07-08
[bugfix] Fix `TypeError` in logger when using tuples as dict keys

### LINKED ISSUE #2022: Set `skipkeys=True` as default arg in `json.dump`
At the moment it defaults to `False` and it causes problems when creating a dict with tuples as keys.

jjotero:
When defining a custom variable of type `dict`, the logger raised a `TypeError` from the `json.dumps` call if the dict had tuples as keys. This error is now caught and the dict keys are converted to strings using the `toalphanum` utility.

Note that fixing this in the `default` hook would not work, because dictionaries are recognised types by json. The alternative fix through the encoder would be a total overkill just for this minor issue (see https://stackoverflow.com/questions/16405969/how-to-change-json-encoding-behaviour-for-serializable-python-object for a similar issue).

Closes #2022

### DIFF
--- a/reframe/utility/jsonext.py
+++ b/reframe/utility/jsonext.py
@@ -24,7 +24,10 @@ class JSONSerializable:
 
 
 def encode_dict(obj, recursive=False):
-    '''Transform non-compatible dict keys into strings.'''
+    '''Transform non-compatible dict keys into strings.
+
+    Use the recursive option to also check the keys in nested dicts.
+    '''
     if isinstance(obj, MutableMapping):
         _valid_keys = (str, int, float, bool, type(None))
         if recursive or not all(isinstance(k, _valid_keys) for k in obj):
@@ -56,7 +59,7 @@ def encode(obj, **kwargs):
     if inspect.istraceback(obj):
         return traceback.format_tb(obj)
 
-    newobj = encode_dict(obj)
+    newobj = encode_dict(obj, recursive=True)
     if newobj:
         return newobj
 
--- a/reframe/utility/jsonext.py
+++ b/reframe/utility/jsonext.py
@@ -23,18 +23,18 @@ class JSONSerializable:
         return _ret
 
 
-def encode_dict(obj, recursive=False):
-    '''Transform non-compatible dict keys into strings.
+def encode_dict(obj, *, recursive=False):
+    '''Transform tuple dict keys into strings.
 
     Use the recursive option to also check the keys in nested dicts.
     '''
+
     if isinstance(obj, MutableMapping):
-        _valid_keys = (str, int, float, bool, type(None))
-        if recursive or not all(isinstance(k, _valid_keys) for k in obj):
+        if recursive or any(isinstance(k, tuple) for k in obj):
             newobj = type(obj)()
             for k, v in obj.items():
-                _key = str(k) if not isinstance(k, _valid_keys) else k
-                _v = encode_dict(v)
+                _key = str(k) if isinstance(k, tuple) else k
+                _v = encode_dict(v, recursive=recursive)
                 newobj[_key] = _v if _v else v
 
             return newobj
@@ -59,7 +59,7 @@ def encode(obj, **kwargs):
     if inspect.istraceback(obj):
         return traceback.format_tb(obj)
 
-    newobj = encode_dict(obj, recursive=True)
+    newobj = encode_dict(obj)
     if newobj:
         return newobj
 
--- a/reframe/utility/jsonext.py
+++ b/reframe/utility/jsonext.py
@@ -6,6 +6,7 @@
 import inspect
 import json
 import traceback
+from collections.abc import MutableMapping
 
 import reframe.utility as util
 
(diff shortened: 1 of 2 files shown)


reframe/utility/jsonext.py
```diff
@@ -40,14 +41,35 @@ def encode(obj):
     return None
 
 
+def _make_json_friendly(obj):
```
vkarak, line 44: I would rather call this something that shows that it's for the keys, because the `encode()` above is meant to make the JSON values at least friendly. The problem as I see it is that `json.dump` does not delegate to the `default` callable when it doesn't know how to convert keys and issues an error.

reframe/utility/jsonext.py
```diff
@@ -40,14 +41,35 @@ def encode(obj):
...
+    '''Fallback type converter if json dump(s) raises a TypeError.'''
+
+    _valid_keys = {str, int, float, bool, type(None)}
+    if isinstance(obj, MutableMapping):
+        if not all((any(isinstance(k, t) for t in _valid_keys) for k in obj)):
+            newobj = type(obj)()
+            for k, v in obj.items():
+                newobj[util.toalphanum(str(k))] = v
```
vkarak, line 52: I'd prefer to use here the `encode()` function above, because it encodes correctly ReFrame types + some other builtin types and it promotes consistency between keys and values. The shortcoming is that `json.dump` does not treat consistently keys and values, so that the `encode()` will not be called for tuples if they are encountered as values, but we would call it here if tuples are encountered as keys. We could expand `encode()` to treat tuples as well, but I wouldn't go as far as to allow anything to be converted. That's why `encode()` simply returns `None` instead of `str(obj)` (n.b. it doesn't raise a `TypeError` also on purpose, in order to allow unknown objects to dumped as `"null"` instead of crashing). In any case, I think we should use `encode()` here for consistency with the values. Also, I wouldn't convert all the keys, but only those that are not in `_valid_keys`.

vkarak (changes requested):
We would also need a unit test for this.

reframe/utility/jsonext.py
jjotero: Makes sense. I moved this into the `encode` function and all of a sudden, the test I had started failing again ... and the reason is be cause this patch was buggy! Say that we have the following tests:
```python
@rfm.simple_test
class MyTest(rfm.RunOnlyRegressionTest):
    ...
    custom_dict = variable(dict, value={
            ('a', 'b'): 'bananas',
            'a': 'bananas',
            ('a', 'c', 'b'): 'bananas',
    })
```
Here ReFrame will log the `custom_dict` variable independently and also will log the `MyTest` class (including its `self.__dict__`) which also contains the `custom_dict` variable. Before, this fix was only catching the `TypeError` when logging the `custom_dict`, and it would just return `None` when a `TypeError` was raised from attempting to log the `MyTest` class. Now that we have moved this functionality into the `encode` function, this function would "catch" the `MyTest` class after the `TypeError` and reattempt the logging without fixing the issues with the `self.__init__`, raising a `TypeError` (again).
The fix to this is to have a separate `encode_dict` function (which `encode` also uses) and this function will massage the `self.__dict__` from the `__rfm_json_encode__` function. This `encode_dict` can be forced to recur through potentially nested dictionaries, so it will also catch tuples as dict keys in sub-dictionaries, even when all the keys from the top-level dict are json-friendly (which was exactly the case of the `self.__dict__` from `MyTest`).

I'll push the changes and the unit tests in a sec.

unittests/test_utility.py
```diff
@@ -1567,6 +1567,9 @@ def __init__(self, x, y):
         self.z = None
         self.w = {1, 2}
 
+        # Dump dict with tuples as keys
+        self.v = {(1, 2): 1}
```
teojgo, line 1571: This has to be also included in the equality test.

reframe/utility/jsonext.py
```diff
@@ -17,10 +18,31 @@ def __rfm_json_encode__(self):
...
         }
         ret.update(self.__dict__)
-        return ret
+        encoded_ret = encode_dict(ret, recursive=True)
+        _ret = encoded_ret if encoded_ret else ret
+        return _ret
 
 
-def encode(obj):
+def encode_dict(obj, *, recursive=False):
```
teojgo, line 26: There could be a need for an equivalent `decode_dict` to convert tuple like strings to tuples when loading the json.

teojgo (changes requested):

reframe/utility/jsonext.py
```diff
@@ -38,19 +59,8 @@ def encode(obj):
...
-
-    _valid_keys = {str, int, float, bool, type(None)}
-    if isinstance(obj, MutableMapping):
-        if not all((any(isinstance(k, t) for t in _valid_keys) for k in obj)):
-            newobj = type(obj)()
-            for k, v in obj.items():
-                newobj[util.toalphanum(str(k))] = v
-
+    newobj = encode_dict(obj)
+    if newobj:
```
vkarak, line 63: Better check explicitly for None here:

```suggestion
    if newobj is not None:
```

reframe/utility/jsonext.py
```diff
@@ -18,10 +18,31 @@ def __rfm_json_encode__(self):
...
+    Use the recursive option to also check the keys in nested dicts.
+    '''
+
+    if isinstance(obj, MutableMapping):
+        if recursive or any(isinstance(k, tuple) for k in obj):
+            newobj = type(obj)()
+            for k, v in obj.items():
+                _key = str(k) if isinstance(k, tuple) else k
+                _v = encode_dict(v, recursive=recursive)
+                newobj[_key] = _v if _v else v
```
vkarak, line 38: Better be explicit here about None:

```suggestion
                newobj[_key] = _v if _v is not None else v
```

unittests/test_utility.py
vkarak: In which case it would call for your other point above about `decode()`.

vkarak (changes requested):

jjotero:
@teojgo @vkarak I see the point with the `decode_dict`, but that is a brand new feature that perhaps should go in a separate issue/PR. This one is just meant to fix a crash when users were writing valid Python inside a test class.

vkarak:
@jjotero Yes, I agree with this. We could just add a FIXME note in the code to remember it.

jjotero:
@jenkins-cscs retry daint

teojgo (approved):

vkarak (approved):
lgtm

vkarak:
@jenkins-cscs retry daint

vkarak (approved):

PR #2061, opened 2021-07-08
[feat] Report whether measured performance was {greater,less} than reference

giordano:
When a performance sanity check fails because the measured performance was
beyond allowed range, explicitly say whether the value was too small or too
large.  When the reference number has several digits, it's hard to quickly
compare the values by eyes.

This is orthogonal to #2045, but it's meant to improve the performance report in
a similar way, specifically when the performance benchmark fails because the
reference value isn't met.

Example with this pull request:
```
  * Reason: performance error: 21560000.0 is less than reference value 100000000.0 (l=90000000.0, u=110000000.00000001)
```
instead of the current
```
  * Reason: performance error: failed to meet reference: l_0=21560000.0, expected 100000000.0 (l=90000000.0, u=110000000.00000001)
```

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -1628,8 +1628,7 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
                     sn.evaluate(
                         sn.assert_reference(
                             val, ref, low_thres, high_thres,
-                            msg=('failed to meet reference: %s={0}, '
-                                 'expected {1} (l={2}, u={3})' % tag))
+                        )
                     )
                 except SanityError as e:
                     raise PerformanceError(e)
--- a/reframe/utility/sanity.py
+++ b/reframe/utility/sanity.py
@@ -566,7 +566,8 @@ def assert_reference(val, ref, lower_thres=None, upper_thres=None, msg=None):
     try:
         evaluate(assert_bounded(val, lower, upper))
     except SanityError:
-        error_msg = msg or '{0} is beyond reference value {1} (l={2}, u={3})'
+        error_msg = msg or ('{0} is %s than reference value {1} (l={2}, u={3})'
+                            % ('lower' if val < lower else 'greater'))
         raise SanityError(_format(error_msg, val, ref, lower, upper)) from None
     else:
         return True
--- a/unittests/test_sanity_functions.py
+++ b/unittests/test_sanity_functions.py
@@ -451,25 +451,25 @@ def test_assert_reference():
     # Check upper threshold values greater than 1
     assert sn.assert_reference(20.0, 10.0, None, 3.0)
     assert sn.assert_reference(-50.0, -20.0, -2.0, 0.5)
-    with pytest.raises(SanityError, match=r'0\.5 is beyond reference value 1 '
-                                          r'\(l=0\.8, u=1\.1\)'):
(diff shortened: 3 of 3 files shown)


reframe/core/pipeline.py
```diff
@@ -1623,8 +1623,7 @@ def check_performance(self):
                     sn.evaluate(
                         sn.assert_reference(
                             val, ref, low_thres, high_thres,
-                            msg=('failed to meet reference: %s={0}, '
-                                 'expected {1} (l={2}, u={3})' % tag))
+                        )
```
giordano, line 1626: Uhm, I just realised the `tag` isn't passed anymore, so there is no way to read the name of the failing variable.  This looks tricky since the logic for greater/less than is only in  `reframe.utility.sanity.assert_reference` :confused:

reframe/utility/sanity.py
```diff
@@ -566,7 +566,15 @@ def calc_bound(thres):
...
-        error_msg = msg or '{0} is beyond reference value {1} (l={2}, u={3})'
+        if msg:
+            error_msg = msg
+        else:
+            if val < lower:
+                error_msg = '{0} is less than reference value {1} ' \
+                    + '(l={2}, u={3})'
+            elif val > upper:
+                error_msg = '{0} is greater than reference value {1} ' \
+                    + '(l={2}, u={3})'
```
jjotero, line 577:
```suggestion
        error_msg = msg or '{0} is %s than reference value {1} (l={2}, u={3})' % (
            'lower' if val < lower else 'greater'
        )
```

jjotero (changes requested):

reframe/utility/sanity.py
```diff
@@ -566,7 +566,9 @@ def calc_bound(thres):
     try:
         evaluate(assert_bounded(val, lower, upper))
     except SanityError:
-        error_msg = msg or '{0} is beyond reference value {1} (l={2}, u={3})'
+        error_msg = msg or '{0} is %s than reference value {1} (l={2}, u={3})' % (
+            'lower' if val < lower else 'greater'
+        )
```
jjotero, line 571: Need to change it as below to avoid the PEP complaints. Also, note that I've suggested to change "less" by "lower" (just to be consistent with the "greater" counterpart), so you also need to update the unit tests.
```suggestion
        error_msg = msg or ('{0} is %s than reference value {1} (l={2}, u={3})'
                            % ('lower' if val < lower else 'greater'))
```

jjotero (changes requested):

vkarak (changes requested):
It turns out the problem is more difficult than it seems. We have faced that in #791. The easy solution would have been to add a `tag` argument to the end of the argument list of `assert_reference()` but that would break the symmetry of all sanity functions, namely accepting `msg` as their last argument. One solution would be to add the `tag` as kwarg-only argument, but again I'm not sure if that's the best solution. Perhaps a new function sharing the internals of `assert_reference()` with a bit of refactoring might be a better idea. I'll try to come up with something.

vkarak:
@giordano I will close this PR, if you don't mind. I think this is more of a minor cosmetic change and I don't think that it is enough to justify refactoring the `assert_reference()` in order to accommodate this.

giordano:
Agreed, this isn't ideal.  I realised the problem with the tag not being only after opening this pull request.

PR #2066, opened 2021-07-13
[ci] Add option to test only modified user tests in ci

teojgo:


vkarak:
Not relevant for the moment.

PR #2068, opened 2021-07-13
[ci] Serialize unit tests everywhere

vkarak:
For a couple of reasons:

1. `pytest-parallel` is not yet stable (0.x versions)
2. Running multiple instances of ReFrame at the same time is not supported and can have unpredictable side effects.


jjotero (approved):

teojgo (approved):

teojgo:
@jenkins-cscs retry daint

PR #2074, opened 2021-07-15
[bugfix] Always use `$HOME` for storing processor auto-detection results

### LINKED ISSUE #2072: ReFrame crashes when trying to autodetect topology and does not have write permissions in `config/_meta`
This is triggered when ReFrame is installed with e.g EasyBuild system-wide, since at the first execution ReFrame tries to create `config/_meta` if this is not already there.

vkarak:
Fixes #2072.


teojgo (approved):

victorusu (approved):
`lgtm`

teojgo:
@vkarak actually now looking also the documentation, you may have to adapt also the auto-detecting part.

PR #2080, opened 2021-07-15
[ci] Fix sanity function deprecation unit test

vkarak:
When testing for deprecation warnings in unit tests, we should force them to be raised, because we might end up with CI failures on master when doing patch-level version bumps.


jgphpc (approved):

victorusu (approved):
`lgtm`

jjotero (approved):
lgtm

Issue #2030, opened 2021-06-24
Make the `BuildSystem` classes use the machinery from the `variables`.


jjotero:
The `BuildSystem` classes have the same structure as the `RegressionTest` classes used to before the variables were introduced. We could spin out a lighter version of the `RegressionTestMeta` and make use of the variables also in these `BuildSystem` classes. That would standardise the use of the `required` keyword throughout ReFrame, instead of each module defining their own `_UndefinedType`.

 #2027 shows why this would be beneficial

jjotero:
From a preliminary look into this, this would require to move the the parameter and variable injection from the `__new__` method in the `RegressionTest` class to the `__call__` method of the `RegressionTestMeta`. Otherwise, each class using this metaclass would be forced to explicitly call the injection every time, which doesn't make much sense. For the same effort, we can also move the hook decoration into `RegressionTestMeta`'s `__call__` method, which would look as follows:
```python
class Meta(type):
    ...
    def __call__(cls, *args, **kwargs):
        obj = cls.__new__(cls, *args, **kwargs)
        
        # Inject the parameters and variables, and also decorate the hooks
        cls._rfm_var_space.inject(obj, cls)                                                                                
        cls._rfm_param_space.inject(obj, cls, _rfm_use_params)
        for stage in _PIPELINE_STAGES:
            cls._add_hooks(stage)    

        obj.__init__(*args, **kwargs)
        return obj
```
The `__rfm_init__` method from `RegressionTest` could be then attached as a "pre-init" hook. The `prefix` and `name` logic in `__new__` could go also into `__rfm_init__`, except for the old naming of the `@parameterized_test` decorator. In short, this should be done after the `@parameterized_test` deco is out.

Issue #1936, opened 2021-04-19
Dynamic pipeline stages


jjotero:
With the changes introduced with #1865, it is now possible to extend the hook functionality and attach hooks to other hooks. The use case for this becomes trivial with test libraries, where one may want to extend the functionality of a hook from a base class. For example

```python
# File 1
class Base(rfm.RegressionTest):
    ...
    @run_before('compile')
    def set_compiler_flags(self):
        self.build_system.cxxflags = ['-O3']

# File 2
from file1 import Base
class Derived(Base):
    ...
    @run_after('set_compiler_flags')
    def extend_compiler_flags(self):
        self.build_system.cxxflags += ['--amdgpu-target=gfx900']
```

jjotero:
Won't do

Issue #1899, opened 2021-03-30
Issue more user-friendly errors on variable and parameter errors


vkarak:
My example is particular for the required variables, but the title is more generic, since I believe we need check the error messages also in these cases.

If a variable is required or a parameter is abstract and the framework tries to instantiate the test, the user should not get a full stack trace unconditionally. As a user, when I see a stack trace, I am expecting a programming failure and not an error from the framework. If these errors were some type of `ReframeError`s, the user can get the stack trace by increasing the verbosity of the output.

jjotero:
This is the current order on which these checks are made:

1. Check that parameters are defined: This is done in the `simple_test` decorator because the decorator itself needs the full length of the parameter space. At the moment, this throws an error and perhaps it would make sense to just skip the test.
2. Check the test defines `valid_systems` and `valid_prog_environs`: This is done by the loader, but it could also be done by the `simple_test` decorator. These two variables and the parameters are needed for the test's instantiation.
3. Check that any other variable is set: This is done when a variable is accessed during the test's execution. A required variable can be set at a later stage by a hook, before its value is needed.

The first two checks are the framework's responsibility and they should never be an `AttributeError`. I think that just skipping the tests in these scenarios and raising a warning would make sense. That being said, it might seem that points 1 & 2 could be moved into the `simple_test` decorator, but these two checks are conceptually different. A parameter determines how many instances are to be created from a class, and the variables `valid_systems` and `valid_prog_environs` determine whether a given instance is well-defined to be executed by reframe. The following valid scenario shows why the check done by point 2 should remain in the loader. 
```python
@rfm.simple_test
class MyTest(rfm.RegressionTest):
    executable = 'echo bananas'
    sanity_patterns = sn.assert_true(1)

MyTest.valid_systems._default_value = ['daint:login']
MyTest.valid_prog_environs._default_value = ['PrgEnv-gnu']
```
If the check on `valid_prog_environs` and `valid_systems` was done in the decorator, Reframe would skip this test despite that the test is good to execute.

Now onto point 3. If this check gets triggered, it's because of a programming error on the user side. Say the user does something like the following:
```python
@rfm.simple_test
class A(rfm.RunOnlyRegressionTest):
    executable = f'echo {v}'
    sanity_patterns = sn.assert_true(1)
```
This this is just Python and the user gets a `NameError` here. Now, say that the user has the following slight variation of the above example:
```python
@rfm.simple_test
class B(rfm.RunOnlyRegressionTest):
[...]

PR #2081, opened 2021-07-16
[bugfix] Fix hook override with multiple inheritance

### LINKED ISSUE #2055: Incomplete hook override with multiple inheritance
Reproducer:
```python
    valid_systems = ['dom:login']
    executable = 'echo bananas'
         
    @sanity_function
    def assert_bananas(self): 

    @run_after('setup')
    @run_after('compile')
    def hh(self):
        print('bananas')

class Base2(rfm.RegressionMixin): 
    @run_after('setup')
    def hh(self):
        print('not bananas') 
    
@rfm.simple_test
class Derived(Base2, Base):
    pass
```
This gives:
```bash
[==========] Started on Mon Jul  5 13:30:09 2021 

[----------] started processing Derived (Derived)
[ RUN      ] Derived on dom:login using PrgEnv-gnu
not bananas
not bananas
[----------] finished processing Derived (Derived)

[       OK ] (1/1) Derived on dom:login using PrgEnv-gnu [compile: 0.006s run: 1.306s total: 1.341s]

[==========] Finished on Mon Jul  5 13:30:10 2021 
```
The hook `hh` from class `Base2` overrides the post-setup hook from class `Base`. However, class `Base` also decorates `hh` as a post-compile hook, and this leaks into `Derived`, which calls `hh` (inherited from `Base2`) both as a post-setup and post-compile hooks. `hh` here should only be called as a post-setup hook.

jjotero:
This PR rearranges the instantiation/construction process of classes using the `RegressionTestMeta` metaclass. The `RegressionTestMeta` is now agnostic of pipeline-specific concepts, which facilitates its use in other classes which could also benefit from using all the machinery implemented in this metaclass (such as the `BuildSystem` - see #2030):

- variables and parameters are now injected into a class from the `__call__` method in the metaclass, instead of the `__new__` method from the class. This means that now any class that uses this `RegressionTestMeta` can use the variables and parameters "for free" without implementing anything else.
- The injection and stage-specific mapping of hooks now happens in the `__new__` method of the `RegressionTest` class. This moves all pipeline-specific concepts into the pipeline itself. This rearrangement paves the way for #1936, and for the same money, it closes #2055.


reframe/core/pipeline.py
```diff
@@ -917,6 +912,55 @@ def __rfm_init__(self, name=None, prefix=None):
...
+        else:
+            return ''
+
+    @classmethod
+    def _process_hook_registry(cls):
+        '''Process and validate the pipeline hooks.'''
+
+        _pipeline_hooks = {}
+        for stage, hooks in cls.pipeline_hooks().items():
+            stage_name = '_'.join(stage.split('_')[1:])
```
vkarak, line 928: I don't understand the purpose of `'_'.join(...)` here...

reframe/core/pipeline.py
```diff
@@ -808,37 +801,29 @@ def __new__(cls, *args, _rfm_use_params=False, **kwargs):
                         os.path.dirname(inspect.getfile(cls))
                     )
 
+        # Prepare initialization of test defaults
+        obj.__deferred_rfm_init = obj.__rfm_init__(*args,
+                                                   name=cls.__qualname__,
+                                                   prefix=prefix, **kwargs)
```
vkarak, line 807: Perhaps a little explanation why we need to defer `__rfm_init__` here could help. If I understand correctly, it's because `__rfm_init__` takes arguments whereas hooks do not and we want to execute it as a hook.

unittests/test_meta.py
```diff
@@ -201,9 +204,28 @@ def hook_b(self):
 
     assert not Bar.hook_in_stage('hook_a', 'post_setup')
     assert not Bar.hook_in_stage('hook_b', 'pre_compile')
-    assert Bar.hook_in_stage('hook_c', 'post_run_wait')
+    assert Bar.hook_in_stage('hook_c', 'post_run')
     assert Bar.hook_in_stage('hook_a', 'pre_sanity')
 
+    class Baz(MyMeta):
```
vkarak, line 210: Can you add a comment here that you are testing attaching hooks to a random class?

vkarak (changes requested):
Lgtm except a couple of minor comments + the following one. Since now the hooks are checked upon class instantiation, if somebody tries to attach a hook to an inexistent state, they get:

```
./bin/reframe: skipping test 'HelloTest' due to errors: use `-v' for more information
    FILE: n/a:n/a
```

The first part of the message, I think should be addressed by #1899, but the second one might be related to this PR. If not, we address that in #1899 as well.

vkarak:
You can ignore the comments about the error message here. They will be fixed by #2087.

reframe/core/pipeline.py
jjotero: With this PR we move the injection of the variables and the parameters to the `__call__` of the metaclass, which means that these are only available after this `__new__` has returned, and `__rfm_init__` needs some of the variables and parameters to be present.

reframe/core/pipeline.py
vkarak: I mean, I understand the `split()` part and I agree with that, but I don't understand why we need the `join()`.

vkarak: Ok, got it, I have a better suggestion. I will comment in the updated PR.

reframe/core/pipeline.py
```diff
@@ -925,7 +927,9 @@ def _process_hook_registry(cls):
 
         _pipeline_hooks = {}
         for stage, hooks in cls.pipeline_hooks().items():
-            stage_name = '_'.join(stage.split('_')[1:])
+            # Pop the stage pre_/post_ prefix
+            stage_name = re.match(r'p\w{2,3}_(\w+)', stage)[1]
```
vkarak, line 931: I'd avoid the fuss of regexes here. The following should cover also the cases where stage names have underscores.

```suggestion
            stage_name = stage.split('_', maxsplit=1)[1]
```

vkarak (changes requested):
Just a final minor comment.

vkarak (approved):
lgtm

PR #2083, opened 2021-07-19
[feat] New performance syntax

### LINKED ISSUE #2064: Explore new performance syntax
The idea is detaching the references from extracting a performance variable. We need to make it easier for users to just extract performance variables without the strict need for a reference to check against. This is particularly relevant with libraries and reusable tests, where a reference is never known if the test is system agnostic.

This syntax should be in line with the new sanity syntax introduced with #2007 (using decorators, rather than assigning variables).

A prototype implementation can be found [in this branch](https://github.com/jjotero/reframe/blob/feat/perf-syntax/) with the [generic gpu_burn](https://github.com/jjotero/reframe/blob/feat/perf-syntax/hpctestlib/microbenchmarks/gpu/gpu_burn/__init__.py), and the cscs-specific [gpu_burn](https://github.com/jjotero/reframe/blob/feat/perf-syntax/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py) checks.

### LINKED ISSUE #2113: Inserting a callable without the `__name__` attribute in crashes the metaclass.
See #2112 for the reproducer.

This can be just fixed by replacing that `v.__name__` for a `getattr(v, '__name__', None)` when checking if the callable is a final method.

jjotero:
This PR introduces a new syntax to capture and report performance data from a reframe check. In order to facilitate writing generic and composable tests, the `reference` variable is now entirely optional and each performance variable is extracted with a member function decorated with the `performance_function` decorator:
```python
import reframe as rfm
import re

@rfm.simple_test
    valid_systems = ['dom:login']
    executable = 'echo 40 bananas'
    v = variable(str)

    @sanity_function
    def assert_bananas(self): 
         
    @performance_function('bananas')
    def perf0(self):
        return sn.extractsingle(r'(\d+)\s*bananas', self.stdout, 1, int)

    @performance_function('bananas', perf_key='b-1')
    def perf1(self, blah=222):
        f = open(self.stdout.evaluate(), 'r')
        stdout = ''.join(f.readlines())
        f.close()
        return int(re.match(r'(\d+)\s*bananas', stdout)[1])
```
Reframe will pickup these functions marked as `@performance_functions` and run them all by default:
```bash
PERFORMANCE REPORT
Base
- dom:login
   - PrgEnv-gnu
      * num_tasks: 1
      * perf0: 40 bananas
      * b-1: 40 bananas
```
Note that the units are attached to the decorated function, and the name of the resulting performance variable can also be customised with the `perf_key` decorator argument. If any decorated performance function has more than one argument without a default value (the `self` placeholder), an error is raised:
```python
def foo(x, y=10):
    # cool

def bar(x, y):
   # not cool
```
So now, what if someone derives a test and wants to change the performance variables from the base test? We have the `perf_variables` attribute for that. During class instantiation, ReFrame will initialise `perf_variables` with all the functions decorated with the `performance_function` decorator. However, if one wants to modify this mapping, one can do so in a similarly to when dealing with the old `perf_patterns` as:
```python
@rfm.simple_test
class Derived(Base):
    @run_before('performance')
    def my_report(self):
        l = lambda self: sn.extractsingle(r'(\d+)\s*bananas', self.stdout, 1, int)
        self.perf_variables = {
                'a': self.perf0(),
                'b': sn.make_performance_function(l, 'more_bananas', self),
                'c': sn.make_performance_function(
                    sn.extractsingle(r'(\d+)\s*bananas', self.stdout, 1, int),
                    'other_bananas'
                )
        }
```
With this syntax, the functions that go into the `perf_variables` dictionary are guaranteed to be suitable performance functions, otherwise an error is raised. Note the addition of the `sn.make_performance_function` method, which allows to add more performance functions inline (it works with both a generic callable and a deferred expression). The performance report for this `Derived` test looks as follows:
```bash
PERFORMANCE REPORT
Base
- dom:login
   - PrgEnv-gnu
      * num_tasks: 1
      * a: 40 bananas
      * b: 40 more_bananas
      * c: 40 other_bananas
```

Closes #2064
Closes #2113


cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -24,30 +24,29 @@ class gpu_burn_check(GpuBurn):
     num_tasks = 0
     reference = {
         'dom:gpu': {
-            'perf': (4115, -0.10, None, 'Gflop/s'),
+            'min_perf': (4115, -0.10, None),
```
jgphpc, line 27:
```suggestion
            my_limits = "-0.10, None, 'Gflop/s'"
            ...
            'min_perf': (4115, my_limits),
```
Feel free to ignore

jgphpc: nevermind

victorusu (commented):
@jjotero, I really loved this PR! This syntax is very clear, extensible, and very powerful. 
But I have a question why is the `make_performance_function` a regression test method? Would it make sense to have it as an independent module that can be used by regression classes? I think that the module approach is more "modular".
And last but not least, I loved, really loved the bananas examples! I do not know if the `b-1` example was intentional, but I missed a [`b-2`](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwjMpPXBlvHxAhWx_7sIHX8KDmUQyCkwAHoECA0QAw&url=https%3A%2F%2Fwww.youtube.com%2Fwatch%3Fv%3DHHxQUw0Upfk&usg=AOvVaw2mip0UDm_FdtpjgG6vdxgd)!

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
jjotero: I see what you mean there. I guess that one can still have the units as a comment in there instead of being part of the reference tuple. At the moment setting the reference tuple the "old way" when using this new syntax would throw an error. I think it's a good thing to throw an error, because otherwise people would use a bit of a hybrid syntax, which might lead to errors if at some point we drop the current syntax

jjotero:
@victorusu I was not aware of that show! 😂
`self.make_performance_function` is implemented as a wrapper of the `@performance_function` decorator, which is not available from the instance. Where else are you thinking this `make_performance_function` method would be used if not in a regression test instance?

reframe/core/pipeline.py
```diff
@@ -1608,52 +1609,101 @@ def check_performance(self):
...
-                value = sn.evaluate(expr)
-                key = '%s:%s' % (self._current_partition.fullname, tag)
-                if key not in self.reference:
-                    raise SanityError(
-                        "tag `%s' not resolved in references for `%s'" %
-                        (tag, self._current_partition.fullname))
+            if hasattr(self, '_rfm_perf_report') or self._rfm_perf_fns:
+                if hasattr(self, 'perf_patterns'):
+                    raise ReframeSyntaxError(
+                        f"assigning a value to 'perf_patters' conflicts with ",
```
jgphpc, line 1617: typo

reframe/core/meta.py
```diff
@@ -252,6 +253,69 @@ def sanity_function(fn):
 
         namespace['sanity_function'] = sanity_function
         namespace['deferrable'] = deferrable
+
+        # Machinery to add performance variables
+        def performance_function(units, *, perf_key=None):
+            '''Decoreate a function to extract a performance variable.
```
jgphpc, line 259: typo

reframe/core/pipeline.py
```diff
@@ -1608,52 +1609,101 @@ def check_performance(self):
...
+                    if key not in self.reference:
+                        raise SanityError(
+                            "tag `%s' not resolved in references for `%s'" %
+                            (tag, self._current_partition.fullname))
+
+                    self._perfvalues[key] = (value, *self.reference[key])
+                    self._perf_logger.log_performance(logging.INFO, tag, value,
+                                                      *self.reference[key])
+
+            # Check the performace variables against their references.
```
jgphpc, line 1706: typo

jgphpc (approved):
lgtm
[gist](https://gist.github.com/jgphpc/1483c954053a32c2e2f4517a447026f7) to myself.

ekouts (approved):
Lgtm so far. After all agreeing on the syntax we should update the documentation

vkarak:
I've started looking into the new syntax. I like the concept of the `@performance_function` and the way it works. One of my first comments is that I'm generally against removing the units from the reference. When I read a test and I see the units I can immediately get an understanding of the measurements. Otherwise I would have to look into the various performance functions, which can be scattered over different files. I'm generally against on the recommendation of writing the unit as a comment, because very easily it'll become outdated. I'd vote to issue a warning if the unit in the perf function and the reference do not match. That'd be helpful for example when a library changes; if the unit of a performance function changes, then most likely the reference must be updated.

Regarding the `@performance_report` I'm still looking at it :-)

jjotero:
Changed the `@performance_report` decorator to `@performance_variables`.

vkarak:
@brandongc If I understand correctly your point, that could be an option, but we didn't want to touch the `perf_patterns` syntax for backward compatibility. Also, as a matter of style we believe the decorators feel a bit more natural and integrated to the language compared to something like this:

```python
perf_patterns = [PerfVar(...), PerfVar(...)]
```

reframe/core/pipeline.py
```diff
@@ -1655,52 +1656,111 @@ def check_performance(self):
...
+                        name, unit = var
+                        ref_tuple = (0, None, None, unit)
+                        self.reference.update({'*': {name: ref_tuple}})
+
+                # We first evaluate and log all performance values and then we
+                # check them against the reference. This way we always log them
+                # even if the don't meet the reference.
+                for tag, expr in self.perf_patterns.items():
+                    value = sn.evaluate(expr)
+                    key = '%s:%s' % (self._current_partition.fullname, tag)
```
vkarak, line 1753: We can convert this to an `f-string` as well as some others further down.

reframe/core/pipeline.py
```diff
@@ -1655,52 +1656,111 @@ def check_performance(self):
...
+                            if ref[3] != unit:
+                                raise ReframeSyntaxError(
+                                    f'units for {tag!r} in the reference '
+                                    f'{key!r} do not match those specified in '
+                                    f'the performance function ({unit})'
+                                )
+
+                            # Pop the unit from the ref tuple (redundant)
+                            ref = ref[:3]
+
```
vkarak, line 1710: Let's remove this line.

reframe/core/pipeline.py
```diff
@@ -1655,52 +1656,111 @@ def check_performance(self):
...
+                # Log the performance variables
+                self._setup_perf_logging()
+                for tag, expr in self.perf_patterns.items():
+                    try:
+                        value, unit = sn.evaluate(expr)
+                        unit = getattr(unit, '_rfm_perf_units', None)
+                        if not unit:
+                            raise TypeError('Trigger the TypeError below')
+
+                    except TypeError:
```
vkarak, line 1687: This pattern is a bit weird here. We get a  unit that actually may not contain a unit?

vkarak (changes requested):
I finished the review of the PR since my comments are not related to what we decide on the `@performance_variables` decorator.

reframe/core/namespaces.py
```diff
@@ -59,7 +59,7 @@ def __repr__(self):
 
     def _raise_namespace_clash(self, name):
         '''Raise an error if there is a namespace clash.'''
-        raise ValueError(
+        raise KeyError(
             f'{name!r} is already present in the current namespace'
```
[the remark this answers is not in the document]
jjotero: Yes, I was thinking though, that perhaps it makes sense to leave that one as a `KeyError` since this class just implements a base namespace and the users are never exposed to this low level namespace mechanisms. Thoughts?

jjotero (commented):
I have introduced the variable `perf_variables` of type `typ.Dict[str, _DeferredPerformanceExpression]`. A `_DeferredPerformanceExpression` is derived from the `_DeferredExpression` class with the addition that it has a member `unit` holding the performance units of each performance variable.

As a result of this internal reorg, I have moved the `make_performance_function` utility into the pipeline. This function is basically an external convert-constructor for the `_DeferredPerformanceExpression` class that can take either a callable with its arguments and the units, or just a deferred expression and the units. This could be integrated somehow into the typed descriptor, and it probably makes sense to think of this also in relation to #563 (which also needs some type of conversion function).

jjotero:
I've added the unit tests for all this new syntax. Also, I've moved the `make_performance_function` method from the pipeline into the `utility/sanity` module, since this method can be used with a free function. Now, since this `make_performance_function` returns a `_DeferredPerformanceExpression`, I think we should be consistent on how we name this function and all the other `deferrable` and so on. I'll try to think of a better name for this.

reframe/core/deferrable.py
```diff
@@ -56,17 +56,17 @@ def evaluate(self):
...
         fn_kwargs = {}
         for k, v in self._kwargs.items():
             fn_kwargs[k] = (
-                v.evaluate() if isinstance(v, type(self)) else v
+                v.evaluate() if isinstance(v, _DeferredExpression) else v
             )
 
         ret = self._fn(*fn_args, **fn_kwargs)
-        if isinstance(ret, type(self)):
+        while isinstance(ret, _DeferredExpression):
```
vkarak, line 69: Perhaps a comment here that we use `while` to capture recursive deferrable calls would be helpful.

cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
```diff
@@ -62,17 +61,31 @@ def set_gpu_arch(self):
...
+            # Get the references
+            ref, lt, ut = self.reference[key]
+
+            # Flag the slow nodes
+            for nid in nids:
+                try:
+                    val = self.perf(nid).evaluate()[0]
+                    sn.assert_reference(val, ref, lt, ut).evaluate()
+                except SanityError:
+                    perf_report[nid] = self.perf(nid)
```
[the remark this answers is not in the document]
vkarak: This sounds like a good idea.

reframe/utility/__init__.py
```diff
@@ -304,6 +304,27 @@ def attrs(obj):
     return ret
 
 
+def is_trivially_callable(fn, *, non_def_args=0):
```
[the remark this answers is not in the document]
vkarak: Yes, I figured this and it's useful, but I suggest to adapt a bit the documentation of the function because now it reads as "A trivially callable object is one that can be called without providing any additional arguments to its call method," which is a bit contradictory 😭

reframe/core/namespaces.py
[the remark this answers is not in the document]
vkarak: Agreed. This is not the final exception thrown to the user. Let's leave it as such.

reframe/core/pipeline.py
```diff
@@ -1655,52 +1656,111 @@ def check_performance(self):
...
-                if key not in self.reference:
-                    raise SanityError(
-                        "tag `%s' not resolved in references for `%s'" %
-                        (tag, self._current_partition.fullname))
+            if hasattr(self, '_rfm_perf_vars') or self._rfm_perf_fns:
+                if hasattr(self, 'perf_patterns'):
+                    raise ReframeSyntaxError(
+                        f"assigning a value to 'perf_pattenrs' conflicts ",
+                        f"with using the 'performance_variables' decorator ",
+                        f"(class {self.__class__.__qualname__})"
```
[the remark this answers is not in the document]
vkarak: > The only thing is that internally it would help if we have a single representation as we do now.

I don't remember what I exactly I meant with that, but I agree with the current design.

reframe/core/pipeline.py
```diff
@@ -1666,63 +1691,52 @@ def check_performance(self):
...
         with osext.change_dir(self._stagedir):
-            if hasattr(self, '_rfm_perf_vars') or self._rfm_perf_fns:
+            if self.perf_variables or self._rfm_perf_fns:
                 if hasattr(self, 'perf_patterns'):
                     raise ReframeSyntaxError(
                         f"assigning a value to 'perf_pattenrs' conflicts ",
-                        f"with using the 'performance_variables' decorator ",
+                        f"with using the 'performance_function' decorator ",
+                        f"or setting a value to 'perf_variables' "
                         f"(class {self.__class__.__qualname__})"
```
vkarak, line 1700: The name of the class is not needed here. It will be printed by the loader.

reframe/core/pipeline.py
```diff
@@ -1666,63 +1691,52 @@ def check_performance(self):
...
-                    except TypeError:
-                        raise ReframeSyntaxError(
-                            f'function assigned for performance variable '
-                            f'{tag!r} is not a performance function'
+                        value = expr.evaluate()
+                        unit = expr.unit
+                    except Exception as e:
+                        logging.getlogger().warning(
+                            f'skipping evaluation of performance variable '
+                            f'{tag!r} in test {self.name!r}: {e}'
```
vkarak, line 1712: Again, we don't need the test name here. The logger knows the context of the test, so it could log the test name regardless.

reframe/core/pipeline.py
```diff
@@ -1666,63 +1691,52 @@ def check_performance(self):
...
                             if ref[3] != unit:
-                                raise ReframeSyntaxError(
-                                    f'units for {tag!r} in the reference '
-                                    f'{key!r} do not match those specified in '
-                                    f'the performance function ({unit})'
+                                logging.getlogger().warning(
+                                    f'reference unit ({key!r}) for the '
+                                    f'performance variable {tag!r} '
+                                    f'does not match the unit specified '
+                                    f'in the performance function ({unit!r})'
```
vkarak, line 1729: Perhaps we need to say which one of the two will be ignored. We can simply add that "{unit!r} will be used"

unittests/test_pipeline.py
```diff
@@ -1249,6 +1249,134 @@ def extract_perf(patt, tag):
...
+
+        return val
+
+    sanity_file.write_text('result = success\n')
+    perf_file.write_text('perf1 = 1.0\n')
+
+    # Make the unit in the reference different from the performance function
+    perftest.reference = {
+        '*': {
+            'value1': (0, None, None, 'unit_')
```
vkarak, line 1338: I think that you could assert that the warning about the unit mismatch in `log_output`.

unittests/test_pipeline.py
```diff
@@ -1249,6 +1249,134 @@ def extract_perf(patt, tag):
...
+    perf_file.write_text('perf1 = 1.0\n')
+    perftest.perf_patterns = {}
+    with pytest.raises(ReframeSyntaxError):
+        _run_sanity(perftest, *dummy_gpu_exec_ctx)
+
+
+def test_perf_function_raises_exception(perftest, sanity_file,
+                                        perf_file, dummy_gpu_exec_ctx):
+    # The lambda takes a single arg, and the make_performance_function
+    # will inject the self argument too. This triggers an exception
```
vkarak, line 1367: Is this accurate?

reframe/core/meta.py
```diff
@@ -271,7 +261,7 @@ def sanity_function(fn):
         namespace['sanity_function'] = sanity_function
         namespace['deferrable'] = deferrable
 
-        # Machinery to add performance variables
+        # Machinery to add performance functions
         def performance_function(units, *, perf_key=None):
```
vkarak, line 265: We still need to document this separately in the rst file.

unittests/test_meta.py
```diff
@@ -254,3 +256,91 @@ class AllowFinalOverride(Base):
...
+    Base().assert_perf_fn_return()
+
+    # Test the performance function set has been built correctly
+    assert len(Base._rfm_perf_fns) == 2
+    perf_fns = {'perf_a', 'perf_b'}
+    for fn in Base._rfm_perf_fns:
+        assert fn._rfm_perf_key in perf_fns
+        perf_fns.remove(fn._rfm_perf_key)
+
+    assert not perf_fns
```
vkarak, line 303: I think the following covers everything:

```suggestion
    assert {fn._rfm_perf_key for fn in Base._rfm_perf_fns} == {'perf_a', 'perf_b'}
```

unittests/test_meta.py
```diff
@@ -254,3 +256,91 @@ class AllowFinalOverride(Base):
...
+            '''Perf function with custom key.'''
+
+    # Test the performance function set is correct with class inheritance
+    assert len(Derived._rfm_perf_fns) == 3
+    perf_fns = {'perf_b', 'perf_c', 'my_perf_fn'}
+    for fn in Derived._rfm_perf_fns:
+        assert fn._rfm_perf_key in perf_fns
+        perf_fns.remove(fn._rfm_perf_key)
+
+    assert not perf_fns
```
vkarak, line 326: Check my suggestion above.

unittests/test_meta.py
```diff
@@ -254,3 +256,91 @@ class AllowFinalOverride(Base):
...
+
+
+def test_performance_function_errors(MyMeta):
+    with pytest.raises(ReframeSyntaxError):
```
vkarak, line 330: I have the feeling that here we can throw a `TypeError` and the error would still be traced back to the right user line. Similarly for anything in the `performance_function` decorator.

hpctestlib/microbenchmarks/gpu/gpu_burn/__init__.py
```diff
@@ -96,27 +96,27 @@ def count_successful_burns(self):
...
 
-    def extract_perf(self, what, nid=None):
+    def _extract_perf_metric(self, metric, nid=None):
         '''Utility to extract performance metrics.'''
 
-        if what not in {'perf', 'temp'}:
+        if metric not in {'perf', 'temp'}:
             raise ValueError(
-                f"unsupported value in 'what' argument ({what}!r)"
+                f"unsupported value in 'metric' argument ({metric}!r)"
```
vkarak, line 104:
```suggestion
                f"unsupported value in 'metric' argument: {metric}!r"
```

vkarak (changes requested):

reframe/core/meta.py
```diff
@@ -82,7 +82,13 @@ def __setitem__(self, key, value):
                         'once in the class body'
                     ) from None
             elif hasattr(value, '_rfm_perf_key'):
-                self['_rfm_perf_fns'].add(value)
+                try:
+                    self['_rfm_perf_fns'][key] = value
+                except KeyError:
+                    raise ReframeSyntaxError(
+                        f'the performance function {key} has already been '
```
vkarak, line 89:
```suggestion
                        f'the performance function {key!r} has already been '
```

reframe/core/parameters.py
```diff
@@ -45,6 +46,16 @@ def filter_params(x):
                 return x
 
         self.values = tuple(values)
+
+        # Validate the filter_param argument
+        try:
+            valid = utils.is_trivially_callable(filter_params, non_def_args=1)
+        except TypeError:
+            raise TypeError('the provided parameter filter is not a callable')
```
vkarak, line 54:
```suggestion
            raise TypeError('the provided parameter filter is not a callable') from None
```

docs/deferrable_functions_reference.rst
```diff
@@ -48,6 +50,18 @@ Currently ReFrame provides three broad categories of deferrable functions:
    They include, but are not limited to, functions to iterate over regex matches in a file, extracting and converting values from regex matches, computing statistical information on series of data etc.
 
 
+Deferrable performance functions
```
vkarak, line 53: I would make this a subsection

docs/deferrable_functions_reference.rst
```diff
@@ -48,6 +50,18 @@ Currently ReFrame provides three broad categories of deferrable functions:
...
+Therefore, this kind of deferrable functions have an associated unit that can be used to interpret the return values from these functions.
+The unit of a deferrable performance function can be accessed through the public member :attr:`unit`.
+Regular deferrable functions can be promoted to deferrable performance functions using the :func:`~reframe.utility.sanity.make_performance_function` utility.
+Also, this utility allows to create performance functions directly from any callable.
+
+Deferrable function utilities
+-----------------------------
+
 Users can write their own deferrable functions as well.
 The page ":doc:`deferrables`" explains in detail how deferrable functions work and how users can write their own.
```
vkarak, line 66: This paragraph should probably be moved before the "performance functions."

docs/deferrable_functions_reference.rst
```diff
@@ -48,6 +50,18 @@ Currently ReFrame provides three broad categories of deferrable functions:
...
+Deferrable performance functions
+--------------------------------
+
+Deferrable performance functions are a special type of deferrable functions which are intended for measuring a given quantity.
+Therefore, this kind of deferrable functions have an associated unit that can be used to interpret the return values from these functions.
+The unit of a deferrable performance function can be accessed through the public member :attr:`unit`.
+Regular deferrable functions can be promoted to deferrable performance functions using the :func:`~reframe.utility.sanity.make_performance_function` utility.
+Also, this utility allows to create performance functions directly from any callable.
+
+Deferrable function utilities
```
vkarak, line 62:
```suggestion
List of deferrable function and utilities
```

docs/regression_test_api.rst
```diff
@@ -322,6 +322,21 @@ Built-in functions
...
+   The decorated function must take a single argument without a default value (i.e. ``self``) and any number of arguments with default values.
+   A test may decorate multiple member functions as performance functions, where each of the decorated functions must be provided with the units of the performance quantitites to be extracted from the test.
+   These performance units must be of type :class:`str`.
+   Any performance function may be overridden in a derived class and multiple bases may define their own performance functions.
+   In the event of a name conflict, the derived class will follow Python's `MRO <https://docs.python.org/3/library/stdtypes.html#class.__mro__>`_ to choose the appropriate performance function.
+   However, defining more than one performance function with the same name in the same class is disallowed.
+
+   The full set of performance functions of a regression test is stored under :attr:`~reframe.core.pipeline.RegerssionTest.perf_variables` as key-value pairs, where, by default, the key is the name of the decorated member function, and the value is the deferred performance function itself.
```
vkarak, line 337:
```suggestion
   The full set of performance functions of a regression test is stored under :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` as key-value pairs, where, by default, the key is the name of the decorated member function, and the value is the deferred performance function itself.
```

docs/regression_test_api.rst
```diff
@@ -322,6 +322,21 @@ Built-in functions
 
   .. versionadded:: 3.7.0
 
+.. py:decorator:: RegressionMixin.performance_function(unit, *, perf_key=None)
+
+   Decorate a member function as a performance function of the test.
+
+   This decorator converts the decorated method into a performance deferrable function (see :ref:`deferrable-functions`) whose evaluation is deferred to the performance stage of the regression test.
```
vkarak, line 329: We can have a more precise link here to the actual subsection.

reframe/core/deferrable.py
```diff
@@ -66,10 +67,15 @@ def evaluate(self):
...
         ret = self._fn(*fn_args, **fn_kwargs)
+
+        # Evaluate the return for as long as a deferred expression returns
+        # another deferred expression.
         while isinstance(ret, _DeferredExpression):
             ret = ret.evaluate()
 
+        # Cache the results for any subsequent evaluate calls.
         self._cached = (ret,)
+
```
vkarak, line 78: Remove this line.

unittests/test_meta.py
vkarak: > Changed to TypeError, but now I'm confused 😅. I thought the plan was to raise ReframeSyntaxError for everything that we impose and control.

If a programming error can be traced back to the user frame that triggered it, then ReFrame can avoid a stack trace automatically. In this case, it can be traced because it's a direct call chain from the decorator call. The problem has been with the variables and parameters, where we do several checks at a later point. In these cases, the user line information is lost and we can't trace back the error to the user code; the stack trace has no user frame. So to avoid printing a stack trace for such user errors, we invented the `ReframeSyntaxError`. That's the rule of thumb, I hope it makes sense...

unittests/test_pipeline.py
vkarak: You're right. You can't! I thought we were using the logger 😅

vkarak (changes requested):
Looks good! Just minor corrections and some fine tuning to the docs. I could also take care of some of those.

reframe/core/meta.py
vkarak: I'm fixing that now.

reframe/core/parameters.py
vkarak: I'm fixing that now.

docs/regression_test_api.rst
vkarak: I'm fixing that now.

reframe/core/pipeline.py
```diff
@@ -647,22 +647,36 @@ def pipeline_hooks(cls):
     #: :default: :class:`None`
     perf_patterns = variable(typ.Dict[str, _DeferredExpression], type(None))
 
-    #: Mapping with the performance variables of the test.
+    #: The performance variables associated with the test.
+    #:
+    #: In this context, a performance variable is a key-value pair, where the
+    #: key is the desired variable name and the value is the deferred
+    #: performance expression (i.e. the result of a :doc:`deferrable
```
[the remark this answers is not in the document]
vkarak: Here I suggest adding `.. _deferrable-perf-functions:` before the section in the sanity functions documentation. Then you can reference it anywhere with ``:ref:`deferrable-perf-functions` `` and the title of the section will be inserted. To avoid typing here, I would also change this to ``see :ref:`..` for more information`.

vkarak (approved):
Latm! Goes in! I just made a bit of fine tuning of the docs.

reframe/utility/sanity.py
```diff
@@ -904,15 +910,22 @@ def defer(x):
...
 
-def evaluate(expr):
+def evaluate(expr, cache=False):
     '''Evaluate a deferred expression.
 
     If ``expr`` is not a deferred expression, it will be returned as is.
+    If ``expr`` is a deferred expression and ``cache`` is ``True``, the
+    results of the deferred expression will be cached and subsequent calls
+    to ``evaluate`` on this deferred expression will simply return the
+    previously cached result.
```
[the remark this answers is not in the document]
vkarak: Yeah, it was too late yesterday night that I was reading it. Now that I reread it's clear also from this sentence :-) The additional note doesn't hurt though.

unittests/test_meta.py
vkarak: I don't know where would be the right place to be honest, because that's mostly a developer's documentation, isn't it?

PR #2084, opened 2021-07-20
PR for tests generalization for LAMMPS and AMBER

hurricane642:
This PR contain base for the `LAMMPS` and `AMBER` tests and their implementation for `cscs-checks` with updated syntax and unified style. I'll be happy your review and suggestions!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -141,7 +141,7 @@ stage('Cleanup') {
                 }
             }
         }
-        catchError(stageResult: 'FAILURE') {
+        catchError(buildResult: 'SUCCESS', stageResult: 'FAILURE') {
             parallel builds
         }
     }
@@ -163,8 +163,7 @@ stage('Cleanup Stale') {
             }
         }
     }
-
-    catchError(stageResult: 'FAILURE') {
+    catchError(buildResult: 'SUCCESS',  stageResult: 'FAILURE') {
         parallel builds
     }
 }
--- a/cscs-checks/apps/amber/amber/__init__.py
+++ b/cscs-checks/apps/amber/amber/__init__.py
@@ -1,0 +1,5 @@
+from .amber_base import (
+  AmberBaseCheck
+)
+
+__all__ = ["AmberBaseCheck"]
--- a/cscs-checks/apps/amber/amber/__init__.py
+++ b/cscs-checks/apps/amber/amber/__init__.py
@@ -1,5 +1,0 @@
-from .amber_base import (
-  AmberBaseCheck
-)
-
-__all__ = ["AmberBaseCheck"]
--- a/cscs-checks/apps/amber/amber/amber_base.py
+++ b/cscs-checks/apps/amber/amber/amber_base.py
@@ -22,15 +22,6 @@ class AmberBaseCheck(rfm.RunOnlyRegressionTest):
         'JAC_production_NVE',
     ])
 
-    @run_after('init')
-    def download_files(self):
-        self.prerun_cmds = [
-            # cannot use wget because it is not installed on eiger
-            f'curl -LJO https://github.com/victorusu/amber_benchmark_suite'
-            f'/raw/main/amber_16_benchmark_suite/PME/{self.benchmark}.tar.bz2',
-            f'tar xf {self.benchmark}.tar.bz2'
-        ]
-
     @run_after('setup')
     def set_executable_opts(self):
         self.executable_opts = ['-O',
--- a/cscs-checks/apps/amber/amber/amber_base.py
+++ b/cscs-checks/apps/amber/amber/amber_base.py
@@ -1,0 +1,66 @@
+# Copyright 2016-2021 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+
+
+class AmberBaseCheck(rfm.RunOnlyRegressionTest):
+    modules = ['Amber']
+    input_file = variable(str)
+    ouput_file = variable(str)
+    num_tasks_per_node = required
+    ener_ref = variable(typ.Dict[str, typ.Tuple[int, float]])
+    benchmark = parameter([
+        # NVE simulations
+        'Cellulose_production_NVE',
+        'FactorIX_production_NVE',
+        'JAC_production_NVE_4fs',
+        'JAC_production_NVE',
+    ])
+
+    @run_after('init')
+    def download_files(self):
+        self.prerun_cmds = [
+            # cannot use wget because it is not installed on eiger
+            f'curl -LJO https://github.com/victorusu/amber_benchmark_suite'
+            f'/raw/main/amber_16_benchmark_suite/PME/{self.benchmark}.tar.bz2',
+            f'tar xf {self.benchmark}.tar.bz2'
+        ]
+
+    @run_after('setup')
+    def set_executable_opts(self):
+        self.executable_opts = ['-O',
+                                '-i', self.input_file,
+                                '-o', self.output_file]
+        self.keep_files = [self.output_file]
+
+    @run_after('setup')
+    def set_sanity_patterns(self):
+        energy = sn.extractsingle(r' Etot\s+=\s+(?P<energy>\S+)',
+                                  self.output_file, 'energy', float, item=-2)
+        energy_reference = self.ener_ref[self.benchmark][0]
+        energy_diff = sn.abs(energy - energy_reference)
+        ref_ener_diff = sn.abs(self.ener_ref[self.benchmark][0] *
+                               self.ener_ref[self.benchmark][1])
+        self.sanity_patterns = sn.all([
+            sn.assert_found(r'Final Performance Info:', self.output_file),
+            sn.assert_lt(energy_diff, ref_ener_diff)
+        ])
+
+    @run_after('setup')
+    def set_generic_perf_references(self):
+        self.reference.update({'*': {
+            self.benchmark: (0, None, None, 'ns/day')
+        }})
+
+    @run_after('setup')
+    def set_perf_patterns(self):
+        self.perf_patterns = {
+            self.benchmark: sn.extractsingle(r'ns/day =\s+(?P<perf>\S+)',
+                                             self.output_file, 'perf',
+                                             float, item=1)
+        }
--- a/cscs-checks/apps/amber/amber_der.py
+++ b/cscs-checks/apps/amber/amber_der.py
@@ -91,6 +91,15 @@ class AmberGPUCheck(amber_base.AmberBaseCheck):
         self.reference = REFERENCE_GPU_PERFORMANCE
 
 
+    @run_after('init')
+    def download_files(self):
+        self.prerun_cmds = [
+            # cannot use wget because it is not installed on eiger
+            f'curl -LJO https://github.com/victorusu/amber_benchmark_suite'
+            f'/raw/main/amber_16_benchmark_suite/PME/{self.benchmark}.tar.bz2',
+            f'tar xf {self.benchmark}.tar.bz2'
+        ]
+
 
 @rfm.simple_test
 class AmberCPUCheck(amber_base.AmberBaseCheck):
@@ -109,6 +118,16 @@ class AmberCPUCheck(amber_base.AmberBaseCheck):
     input_file = 'mdin.CPU'
     output_file = 'amber.out'
     ener_ref = REFERENCE_ENERGY
+
+
+    @run_after('init')
+    def download_files(self):
+        self.prerun_cmds = [
+            # cannot use wget because it is not installed on eiger
+            f'curl -LJO https://github.com/victorusu/amber_benchmark_suite'
+            f'/raw/main/amber_16_benchmark_suite/PME/{self.benchmark}.tar.bz2',
+            f'tar xf {self.benchmark}.tar.bz2'
+        ]
 
     @run_after('init')
     def set_description(self):
--- a/cscs-checks/apps/amber/amber_der.py
+++ b/cscs-checks/apps/amber/amber_der.py
@@ -1,0 +1,154 @@
+# FIXME: Use tuples as dictionary keys as soon as
+# https://github.com/eth-cscs/reframe/issues/2022 is in
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+import sys
+sys.path.append('amber')
+import amber_base
+
+daint_gpu_performance = {
+    'Cellulose_production_NVE': (30.0, -0.05, None, 'ns/day'),
+    'FactorIX_production_NVE': (134.0, -0.05, None, 'ns/day'),
+    'JAC_production_NVE': (388.0, -0.05, None, 'ns/day'),
+    'JAC_production_NVE_4fs': (742, -0.05, None, 'ns/day'),
+}
+
+REFERENCE_GPU_PERFORMANCE = {
+    'daint:gpu': daint_gpu_performance,
+    'dom:gpu': daint_gpu_performance
+}
+
+daint_mc_performance_small = {
+    'Cellulose_production_NVE': (8.0, -0.30, None, 'ns/day'),
+    'FactorIX_production_NVE': (34.0, -0.30, None, 'ns/day'),
+    'JAC_production_NVE': (90.0, -0.30, None, 'ns/day'),
+    'JAC_production_NVE_4fs': (150.0, -0.30, None, 'ns/day'),
+}
+
+eiger_mc_performance_small = {
+    'Cellulose_production_NVE': (3.2, -0.30, None, 'ns/day'),
+    'FactorIX_production_NVE': (7.0, -0.30, None, 'ns/day'),
+    'JAC_production_NVE': (30.0, -0.30, None, 'ns/day'),
+    'JAC_production_NVE_4fs': (45.0, -0.30, None, 'ns/day'),
+}
+
+REFERENCE_CPU_PERFORMANCE_SMALL = {
+    'daint:mc': daint_mc_performance_small,
+    'dom:mc': daint_mc_performance_small,
+    'eiger:mc': eiger_mc_performance_small,
+    'pilatus:mc': eiger_mc_performance_small,
+}
+
+REFERENCE_CPU_PERFORMANCE_LARGE = {
+    'daint:mc': {
+        'Cellulose_production_NVE': (10.0, -0.30, None, 'ns/day'),
+        'FactorIX_production_NVE': (36.0, -0.30, None, 'ns/day'),
+        'JAC_production_NVE': (78.0, -0.30, None, 'ns/day'),
+        'JAC_production_NVE_4fs': (135.0, -0.30, None, 'ns/day'),
+    },
+    'eiger:mc': {
+        'Cellulose_production_NVE': (1.3, -0.30, None, 'ns/day'),
+        'FactorIX_production_NVE': (3.5, -0.30, None, 'ns/day'),
+        'JAC_production_NVE': (17.0, -0.30, None, 'ns/day'),
+        'JAC_production_NVE_4fs': (30.5, -0.30, None, 'ns/day'),
+    },
+}
+
+REFERENCE_ENERGY = {
+    # every system has a different reference energy and drift
+    'Cellulose_production_NVE': (-443246, 5.0E-05),
+    'FactorIX_production_NVE': (-234188, 1.0E-04),
+    'JAC_production_NVE_4fs': (-44810, 1.0E-03),
+    'JAC_production_NVE': (-58138, 5.0E-04),
+}
+
+
+@rfm.simple_test
+class AmberGPUCheck(amber_base.AmberBaseCheck):
+    strict_check = False
+    extra_resources = {
+            'switches': {
+                'num_switches': 1
+            }
+        }
+    input_file = 'mdin.GPU'
+    output_file = 'amber.out'
+    valid_systems = ['daint:gpu', 'dom:gpu']
+    valid_prog_environs = ['builtin']
+    maintainers = ['VH', 'SO']
+    executable = 'pmemd.cuda.MPI'
+    num_tasks = 1
+    num_gpus_per_node = 1
+    num_tasks_per_node = 1
+    ener_ref = REFERENCE_ENERGY
+    descr = f'Amber GPU check'
+    tags = {'maintenance', 'production', 'health'}
+
+    @run_after('setup')
+    def set_perf_reference(self):
+        self.reference = REFERENCE_GPU_PERFORMANCE
+
+
+
+@rfm.simple_test
+class AmberCPUCheck(amber_base.AmberBaseCheck):
+    strict_check = False
+    extra_resources = {
+            'switches': {
+                'num_switches': 1
+           }
+        }
+    tags = {'maintenance', 'production'}
+    scale = parameter(['small', 'large'])
+    valid_systems = ['daint:mc', 'eiger:mc']
+    maintainers = ['VH', 'SO']
+    valid_prog_environs = ['builtin']
+    executable = 'pmemd.MPI'
+    input_file = 'mdin.CPU'
+    output_file = 'amber.out'
+    ener_ref = REFERENCE_ENERGY
+
+    @run_after('init')
+    def set_description(self):
+        self.mydescr = f'Amber parallel {self.scale} CPU check'
+
+    @run_after('init')
+    def set_additional_systems(self):
+        if self.scale == 'small':
+            self.valid_systems += ['dom:mc', 'pilatus:mc']
+
+    @run_after('init')
+    def set_hierarchical_prgenvs(self):
+        if self.current_system.name in ['eiger', 'pilatus']:
+            self.valid_prog_environs = ['cpeIntel']
+
+    @run_after('setup')
+    def set_perf_reference(self):
+        if self.scale == 'small':
+            self.reference = REFERENCE_CPU_PERFORMANCE_SMALL
+        else:
+            self.reference = REFERENCE_CPU_PERFORMANCE_LARGE
+
+    @run_after('init')
+    def set_num_tasks_cray_xc(self):
+        if self.current_system.name in ['daint', 'dom']:
+            self.num_tasks_per_node = 36
+            if self.scale == 'small':
+                self.num_nodes = 6
+            else:
+                self.num_nodes = 16
+            self.num_tasks = self.num_nodes * self.num_tasks_per_node
+
+    @run_after('init')
+    def set_num_tasks_cray_shasta(self):
+        if self.current_system.name in ['eiger', 'pilatus']:
+            self.num_tasks_per_node = 128
+            if self.scale == 'small':
+                self.num_nodes = 4
+            else:
+                # there are too many processors, the large jobs cannot start
+                # need to decrease to just 8 nodes
+                self.num_nodes = 8
+            self.num_tasks = self.num_nodes * self.num_tasks_per_node
--- a/cscs-checks/apps/amber/amber_der.py
+++ b/cscs-checks/apps/amber/amber_der.py
@@ -4,9 +4,7 @@
 import reframe as rfm
 import reframe.utility.sanity as sn
 import reframe.utility.typecheck as typ
-import sys
-sys.path.append('amber')
-import amber_base
+from reframe.base_check import AmberBaseCheck
 
 daint_gpu_performance = {
     'Cellulose_production_NVE': (30.0, -0.05, None, 'ns/day'),
@@ -58,31 +56,49 @@ REFERENCE_CPU_PERFORMANCE_LARGE = {
 
 REFERENCE_ENERGY = {
     # every system has a different reference energy and drift
-    'Cellulose_production_NVE': (-443246, 5.0E-05),
-    'FactorIX_production_NVE': (-234188, 1.0E-04),
-    'JAC_production_NVE_4fs': (-44810, 1.0E-03),
-    'JAC_production_NVE': (-58138, 5.0E-04),
+    'Cellulose_production_NVE': (-443246.0, 5.0E-05),
+    'FactorIX_production_NVE': (-234188.0, 1.0E-04),
+    'JAC_production_NVE_4fs': (-44810.0, 1.0E-03),
+    'JAC_production_NVE': (-58138.0, 5.0E-04),
 }
 
-
-@rfm.simple_test
-class AmberGPUCheck(amber_base.AmberBaseCheck):
+class AmberCheck(AmberBaseCheck):
+    benchmark = parameter([
+        # NVE simulations
+        'Cellulose_production_NVE',
+        'FactorIX_production_NVE',
+        'JAC_production_NVE_4fs',
+        'JAC_production_NVE',
+    ])
     strict_check = False
     extra_resources = {
             'switches': {
                 'num_switches': 1
             }
         }
-    input_file = 'mdin.GPU'
+    ener_ref = REFERENCE_ENERGY
     output_file = 'amber.out'
+    maintainers = ['VH', 'SO']
+
+    @run_after('init')
+    def source_install(self):
+        self.prerun_cmds = [
+            # cannot use wget because it is not installed on eiger
+            f'curl -LJO https://github.com/victorusu/amber_benchmark_suite'
+            f'/raw/main/amber_16_benchmark_suite/PME/{self.benchmark}.tar.bz2',
+            f'tar xf {self.benchmark}.tar.bz2'
+        ]
+
+
+@rfm.simple_test
+class AmberGPUCheck(AmberCheck):
+    input_file = 'mdin.GPU'
     valid_systems = ['daint:gpu', 'dom:gpu']
     valid_prog_environs = ['builtin']
-    maintainers = ['VH', 'SO']
     executable = 'pmemd.cuda.MPI'
     num_tasks = 1
     num_gpus_per_node = 1
     num_tasks_per_node = 1
-    ener_ref = REFERENCE_ENERGY
     descr = f'Amber GPU check'
     tags = {'maintenance', 'production', 'health'}
 
@@ -91,43 +107,14 @@ class AmberGPUCheck(amber_base.AmberBaseCheck):
         self.reference = REFERENCE_GPU_PERFORMANCE
 
 
-    @run_after('init')
-    def download_files(self):
-        self.prerun_cmds = [
-            # cannot use wget because it is not installed on eiger
-            f'curl -LJO https://github.com/victorusu/amber_benchmark_suite'
-            f'/raw/main/amber_16_benchmark_suite/PME/{self.benchmark}.tar.bz2',
-            f'tar xf {self.benchmark}.tar.bz2'
-        ]
-
-
 @rfm.simple_test
-class AmberCPUCheck(amber_base.AmberBaseCheck):
-    strict_check = False
-    extra_resources = {
-            'switches': {
-                'num_switches': 1
-           }
-        }
+class AmberCPUCheck(AmberCheck):
     tags = {'maintenance', 'production'}
     scale = parameter(['small', 'large'])
     valid_systems = ['daint:mc', 'eiger:mc']
-    maintainers = ['VH', 'SO']
     valid_prog_environs = ['builtin']
     executable = 'pmemd.MPI'
     input_file = 'mdin.CPU'
-    output_file = 'amber.out'
-    ener_ref = REFERENCE_ENERGY
-
-
-    @run_after('init')
-    def download_files(self):
-        self.prerun_cmds = [
-            # cannot use wget because it is not installed on eiger
-            f'curl -LJO https://github.com/victorusu/amber_benchmark_suite'
-            f'/raw/main/amber_16_benchmark_suite/PME/{self.benchmark}.tar.bz2',
-            f'tar xf {self.benchmark}.tar.bz2'
-        ]
 
     @run_after('init')
     def set_description(self):
--- a/cscs-checks/apps/lammps/lammps_der.py
+++ b/cscs-checks/apps/lammps/lammps_der.py
@@ -1,0 +1,141 @@
+# Copyright 2016-2021 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import os
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+from reframe.base_check import LAMMPSBaseCheck
+
+
+REFERENCE_ENERGY = {
+    # every system has a different reference energy and drift
+    'small': (-4.6195, 6.0E-04),
+    'large': (-4.6195, 6.0E-04)
+}
+
+REFERENCE_GPU_PERFORMANCE = {
+    'maint': {
+        'small': {
+            'dom:gpu': {'perf': (3457, -0.10, None, 'timesteps/s')},
+            'daint:gpu': {'perf': (2524, -0.10, None, 'timesteps/s')}
+        },
+        'large': {
+            'daint:gpu': {'perf': (3832, -0.05, None, 'timesteps/s')}
+        }
+    },
+    'prod': {
+        'small': {
+            'dom:gpu': {'perf': (3132, -0.05, None, 'timesteps/s')},
+            'daint:gpu': {'perf': (2400, -0.40, None, 'timesteps/s')}
+        },
+        'large': {
+            'daint:gpu': {'perf': (3260, -0.50, None, 'timesteps/s')}
+        }
+    },
+}
+
+REFERENCE_CPU_PERFORMANCE = {
+    'prod': {
+        'small': {
+            'dom:mc': {'perf': (4394, -0.05, None, 'timesteps/s')},
+            'daint:mc': {'perf': (3824, -0.10, None, 'timesteps/s')},
+            'eiger:mc': {'perf': (4500, -0.10, None, 'timesteps/s')},
+            'pilatus:mc': {'perf': (5000, -0.10, None, 'timesteps/s')}
+        },
+        'large': {
+            'daint:mc': {'perf': (5310, -0.65, None, 'timesteps/s')},
+            'eiger:mc': {'perf': (6500, -0.10, None, 'timesteps/s')},
+            'pilatus:mc': {'perf': (7500, -0.10, None, 'timesteps/s')}
+        }
+    },
+}
+
+
+class LAMMPSCheck(LAMMPSBaseCheck):
+        strict_check = False
+        extra_resources = {
+            'switches': {
+                'num_switches': 1
+            }
+        }
+
+        tags = {'scs', 'external-resources'}
+        maintainers = ['TR', 'VH']
+
+        @run_after('init')
+        def source_install(self):
+            # Reset sources dir relative to the SCS apps prefix
+            self.sourcesdir = os.path.join(self.current_system.resourcesdir,
+                                       'LAMMPS')
+
+        @run_after('init')
+        def env_define(self):
+                if self.current_system.name in ['eiger', 'pilatus']:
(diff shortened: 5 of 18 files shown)


jjotero:
ok to test

jjotero:
Hi @hurricane642! One comment regarding the file structure. The base tests are meant to go into `hpctestlib/apps` and there you can have you Amber tests in `hpctestlib/apps/amber/__init__.py`. Then, there is no need to prepend the derived file with the `_der.py` and you can just leave the names as amber there (and the same for all the other apps). 

So when you import your base, the python import should look as `from hpctestlib.apps.amber import AmberBase`.

cscs-checks/apps/amber/amber.py
```diff
@@ -57,103 +54,68 @@
...
+            'num_switches': 1
+        }
+    }
+    ener_ref = REFERENCE_ENERGY
+    output_file = 'amber.out'
+    maintainers = ['VH', 'SO']
 
     @run_after('init')
-    def download_files(self):
+    def source_install(self):
```
jjotero, line 85: I think you're not installing anything here. These are just files for the checks themselves. So probably the previous name made more sense.

jjotero (changes requested):
A few more comments on the structure before delving into the test-specifics:

- Rename the `cscs-checks/apps/lammps/lammps` to `cscs-checks/apps/lammps/lammps_check.py` and then use git mv if you want to. Otherwise, we'd be losing all the git history.
- Delete the `__init__` files that just have a version ID in there.
- For each test, just replace the `__init__` with the `*_base.py` files. That intermediate step is not needed. Have a look into the other tests in the library to see how it's done there.

hurricane642:
Good evening!
Thank you a lot for your comments, I fix first two points, but didn't understand the last one. Can you explain slightly more detailed, what don't you like?

jjotero (changes requested):
I think it's better if each app is under its own directory instead of having all the python files in the same place. The reason behind this is that each test might have their own additional files that one might need for the tests, and then you don't want to have all those files mixed up in the same directory (please, have a look into `hpctestlib/microbenchmarks/gpu` and see how it's done in there). So in short, I would rename `hpctestlib/apps/amber.py` to `hpctestlib/apps/amber/__init__.py` (and the same for all the other apps).

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+
+__all__ = ["AmberBaseCheck"]
+
+
+class AmberBaseCheck(rfm.RunOnlyRegressionTest):
+    modules = ['Amber']
```
jjotero, line 14: Modules go into the derived class. What if other centres call their modules different?

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    input_file = variable(str)
+    output_file = variable(str)
+    num_tasks_per_node = required
+    ener_ref = variable(typ.Dict[str, typ.Tuple[float, float]])
```
jjotero, line 18: See the comment below for lammps. This is not needed here.

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+    extract = variable(typ.Dict[str, str])
```
jjotero, line 19: What's this variable for? It's not referenced anywhere in this base class?

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @run_after('setup')
+    def set_executable_opts(self):
+        self.executable_opts = ['-O',
+                                '-i', self.input_file,
+                                '-o', self.output_file]
+        self.keep_files = [self.output_file]
```
jjotero, line 26: This line has nothing to do with setting executable options, which is the name of this hook. You should move this somewhere else.

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @run_after('setup')
+    def set_generic_perf_references(self):
+        self.reference.update({'*': {
+            self.benchmark: (0, None, None, 'ns/day')
```
jjotero, line 31: Similarly to lammps, `self.benchmark` is not defined or declared anywhere in this class.

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+    @sanity_function
+    def set_sanity_patterns(self):
+        energy = sn.extractsingle(r' Etot\s+=\s+(?P<energy>\S+)',
+                                  self.output_file, 'energy', float, item=-2)
+        energy_reference = self.ener_ref[self.benchmark][0]
+        energy_diff = sn.abs(energy - energy_reference)
+        ref_ener_diff = sn.abs(self.ener_ref[self.benchmark][0] *
+                               self.ener_ref[self.benchmark][1])
```
jjotero, line 49: See comments below for the sanity function in lammps. These base classes do not need this `self.ener_ref` dictionary.

hpctestlib/apps/vasp/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+
+__all__ = ["VaspBaseCheck"]
+
+
+class VASPBaseCheck(rfm.RunOnlyRegressionTest):
+    modules = ['VASP']
```
jjotero, line 14: Same comment as in the other tests

hpctestlib/apps/vasp/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+    num_tasks_per_node = required
+    force_ref = variable(typ.Dict[str, typ.Tuple[float, float]])
```
jjotero, line 16: Same comment as in the other tests.

jjotero (changes requested):
I have just briefly gone through the base classes for now.

I think there are a few conceptual changes to be made to the base classes. Things like the `self.benchmark` parameter (which referenced, but not declared in the base classes) should not be enforced from the base class. If someone wants to parameterise all these tests, they're free to do so in the base class. But if someone doesn't they should not need to define anything such as `self.benchmark`, because there is only one of them. Also, (and this is for the derived classes) perhaps the name `variant` makes more sense than `benchmark`.

cscs-checks/apps/amber/amber_check.py
```diff
@@ -131,29 +98,41 @@ def set_perf_patterns(self):
                                              float, item=1)
         }
 
+    @run_after('init')
```
jjotero, line 101: Hooks must show up in chronological order. post init hooks run before the setup stage.

cscs-checks/apps/amber/amber_check.py
```diff
@@ -57,65 +54,35 @@
...
-        energy_diff = sn.abs(energy - energy_reference)
-        ref_ener_diff = sn.abs(self.ener_ref[self.benchmark][0] *
-                               self.ener_ref[self.benchmark][1])
-        self.sanity_patterns = sn.all([
-            sn.assert_found(r'Final Performance Info:', self.output_file),
-            sn.assert_lt(energy_diff, ref_ener_diff)
-        ])
+    run_after('init')(bind(ut.define_reference))
 
     @run_after('setup')
```
jjotero, line 87: You can merge this hook with the one below and make it `run_before('performance')`

cscs-checks/apps/amber/amber_check.py
```diff
@@ -57,65 +54,35 @@
...
-            'FactorIX_production_NVE': (-234188, 1.0E-04),
-            'JAC_production_NVE_4fs': (-44810, 1.0E-03),
-            'JAC_production_NVE': (-58138, 5.0E-04),
+    strict_check = False
+    extra_resources = {
+        'switches': {
+            'num_switches': 1
         }
+    }
+    references = REFERENCE_ENERGY
```
jjotero, line 81: You don't need to store this here since it's not used. Also, notice that we have already a variable `reference`, so having another one named as `references` is simply a no.

cscs-checks/apps/amber/amber_check.py
```diff
@@ -57,65 +54,35 @@
...
+    output_file = 'amber.out'
```
jjotero, line 82: It doesn't hurt to set this one in the base class instead.

hpctestlib/apps/utils/__init__.py
```diff
@@ -0,0 +1,9 @@
+import re
+import reframe as rfm
+
+__all__ = ['define_reference']
+
+
+def define_reference(test: rfm.RegressionTest):
+    test.reference_value = test.references[test.benchmark][0]
+    test.reference_difference = test.references[test.benchmark][1]
```
jjotero, line 9: I see what you're trying to do here. But the test attributes that you're setting here are test-specific, which means that they should not be part of a utils-style library. The fact that you might use the same names in different tests, doesn't mean that this has the same meaning in all the tests you use this. In any case, with the recommendations I made above, you no longer need this code.

jjotero (changes requested):
I have gone in detail through the Amber test only, and I think that's already enough for now :)

It's OK if you start working in other tests too, but do not add them into this PR, otherwise it will just be never-ending. Make another branch from this one and keep your work in these other tests in there. Also, I've noticed a few things with your commit messages ... please, read the contributing guidelines and be a bit more specific on those commit messages 😉

jjotero (changes requested):
The amber base class looks good! I see that you still have to address some of the comments in the derived class though. Once that's done, I'll start looking into other tests.

Please be more explicit with the commit messages. Do not use `sanity commit` as the message because that doesn't tell anything about what that commit did to the code.

hurricane642:
Hello! Thank got the review!
About the `sanity commit` - it's an indicator, that this commit is meaningless and contains only some formal elements (add/delete blank lines, work with indents, etc.). So in the final version of PR, these commits will be squashed with the important one.

jjotero:
> About the sanity commit - it's an indicator, that this commit is meaningless and contains only some formal elements (add/delete blank lines, work with indents, etc.). So in the final version of PR, these commits will be squashed with the important one.

Still not a good practice. The commit messages should also be intended for the PR reviewers. If you have a commit that just formats the text, say so (`Fix indents`, `Remove blank lines`). Whether these commits get squashed at the end or not, please, write the commit messages as indicated by the contributing guidelines.

jjotero (changes requested):
There're a couple of rather minor comments above still unresolved. I'll start reviewing some other test today.

hurricane642:
Ok, I understood, I'll keep that in mind

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -5,58 +5,125 @@
...
+    maintainers = ['LM']
+    tags = {'scs'}
+    strict_check = False
+    extra_resources = {
+        'switches': {
+            'num_switches': 1
+        }
+    }
+    scale = parameter(['small', 'large'])
+    benchmark = parameter(['prod', 'maint'])
```
victorusu, line 77: I think that the `benchmark` should not be pointing to these tags. I think that this is a site implementation detail. They should be moved to the derived class.

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -5,58 +5,125 @@
...
+        'maint': (199.6, None, 0.06, 's'),
+        'prod': (199.6, None, 0.06, 's')
+    }
+}
+
+
+class Cp2kCheck(Cp2k):
+    modules = ['CP2K']
+    executable = 'cp2k.psmp'
+    executable_opts = ['H2O-256.inp']
```
victorusu, line 67: The Amber tests have the input files as parameters. I think that this should also be the case here. I would introduce a `variant`, similarly to the Amber tests` and set it to `H2O-256`. I think that in a library it is important to know the tests that one is importing from.

victorusu: So, this should be moved to the base class.

cscs-checks/apps/cp2k/cp2k_check.py
victorusu: Nevermind. I have mistakenly confused this for the base class... 😬

cscs-checks/apps/cpmd/cpmd_check.py
```diff
@@ -5,60 +5,75 @@
...
+    maintainers = ['AJ', 'LM']
+    tags = {'production'}
+    valid_systems = ['daint:gpu']
+    descr = 'CPMD check (C4H6 metadynamics)'
+    num_tasks_per_node = 1
+    valid_prog_environs = ['builtin']
+    modules = ['CPMD']
+    executable = 'cpmd.x'
+    input_file = 'ana_c4h6.in'
+    readonly_files = ['ana_c4h6.in', 'C_MT_BLYP', 'H_MT_BLYP']
```
victorusu, line 38: The input file information should be part of the base class, no?
And the name of the input file should be a parameter, similar to the Amber case.

cscs-checks/apps/amber/amber_check.py
```diff
@@ -58,25 +59,30 @@
 }
 
 
-class AmberBaseCheck(rfm.RunOnlyRegressionTest):
+class AmberCheck(Amber):
+    modules = ['Amber']
     valid_prog_environs = ['builtin']
+
+    # NVE simulations
```
victorusu, line 66: I think that the energy and its variation should be part of the base class.

cscs-checks/apps/lammps/lammps_check.py
```diff
@@ -7,102 +7,169 @@
...
+    strict_check = False
+    extra_resources = {
+        'switches': {
+            'num_switches': 1
+        }
+    }
+
+    tags = {'scs', 'external-resources'}
+    maintainers = ['TR', 'VH']
+    energy_value = -4.6195
```
victorusu, line 92: I think that input-related info should be part of the base class.

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -132,46 +167,34 @@ def set_cpu_binding(self):
...
-                          for v in ['maint', 'prod']))
+@rfm.simple_test
 class Cp2kGpuCheck(Cp2kCheck):
-    def __init__(self, scale, variant):
-        super().__init__()
-        self.descr = 'CP2K GPU check (version: %s, %s)' % (scale, variant)
-        self.valid_systems = ['daint:gpu']
-        self.num_gpus_per_node = 1
-        if scale == 'small':
+    valid_systems = ['daint:gpu']
```
victorusu, line 172: I think that GPU/CPU "concept" tests should be part of the library and the tests should inherit from the appropriate base classes.

cscs-checks/apps/lammps/lammps_check.py
```diff
@@ -7,102 +7,169 @@
...
-            }
-        }
+    @run_after('init')
+    def set_tags(self):
+        self.tags |= {'maintenance' if self.benchmark == 'maint'
+                      else 'production'}
+
+
+@rfm.simple_test
+class LAMMPSGPUCheck(LAMMPSCheck):
```
victorusu, line 128: I think here again, that the CPU/GPU tests should be part of the library. Here, we should be inheriting from the appropriate base classes.

cscs-checks/apps/lammps/lammps_check.py
```diff
@@ -7,102 +7,169 @@
...
+    scale = parameter(['small', 'large'])
+    benchmark = parameter(['prod', 'maint'])
+    valid_systems = ['daint:gpu']
+    executable = 'lmp_mpi'
+    input_file = 'in.lj.gpu'
+    executable_opts = ['-sf gpu', '-pk gpu 1', '-in', input_file]
```
victorusu, line 134: If we have the CPU and GPU tests, then the `executable_opts` should be moved to the base class.

cscs-checks/apps/lammps/lammps_check.py
```diff
@@ -7,102 +7,169 @@
...
+                      else 'production'}
+
+
+@rfm.simple_test
+class LAMMPSGPUCheck(LAMMPSCheck):
+    scale = parameter(['small', 'large'])
+    benchmark = parameter(['prod', 'maint'])
+    valid_systems = ['daint:gpu']
+    executable = 'lmp_mpi'
+    input_file = 'in.lj.gpu'
```
victorusu, line 133: I think that input files should be parameters from the base class in the library.

cscs-checks/apps/lammps/lammps_check.py
```diff
@@ -7,102 +7,169 @@
...
-        }
-        self.reference = references[variant][scale]
-        self.tags |= {'maintenance' if variant == 'maint' else 'production'}
 
+@rfm.simple_test
+class LAMMPSCPUCheck(LAMMPSCheck):
+    scale = parameter(['small', 'large'])
+    benchmark = parameter(['prod'])
+    valid_systems = ['daint:mc', 'eiger:mc', 'pilatus:mc']
+    input_file = 'in.lj.cpu'
```
victorusu, line 161: Do you see? The input file is being defined again here. The previous GPU class also defines this.

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -5,51 +5,133 @@
...
+
+
+class QuantumESPRESSOCheck(QuantumESPRESSO):
     scale = parameter(['small', 'large'])
-    variant = parameter(['maint', 'prod'])
+    benchmark = parameter(['maint', 'prod'])
+
+    modules = ['QuantumESPRESSO']
+    executable = 'pw.x'
+    input_file = 'ausurf.in'
```
victorusu, line 70: The input file, I think should be moved to the library class as a parameter.

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -5,51 +5,133 @@
...
 class QuantumESPRESSOCpuCheck(QuantumESPRESSOCheck):
-    def __init__(self):
-        super().__init__()
+    valid_systems = ['daint:mc', 'eiger:mc', 'pilatus:mc']
+
+    @run_after('init')
+    def set_reference(self):
+        if self.scale == 'small':
+            self.reference = REFERENCE_CPU_PERFORMANCE_SMALL
+            self.energy_value = -11427.09017168
```
victorusu, line 119: This information is input-file dependent. Therefore, I think that this should move together to the base class.

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -5,51 +5,133 @@
...
+            self.energy_tolerance = 1E-06
+        else:
+            self.reference = REFERENCE_CPU_PERFORMANCE_LARGE
+            self.energy_value = -11427.09017152
```
victorusu, line 123: The energy value and tolerance for the GPU/CPU tests are the same because they are only driven by the input file.

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -134,53 +173,31 @@ def set_cpu_binding(self):
...
+    valid_systems = ['daint:gpu']
+    num_gpus_per_node = 1
+    num_tasks_per_node = 1
+    num_cpus_per_task = 12
+
+    @run_after('init')
+    def set_reference(self):
+        if self.scale == 'small':
+            self.reference = REFERENCE_GPU_PERFORMANCE_SMALL
+            self.energy_value = -11427.09017168
```
victorusu, line 185: These references should be defined in a base class for the GPU. @hurricane642, could please add a `#FIXME: check the tolerances and the reference difference between the CPU and GPU tests?` 

Side note: @jjotero We will need to review these numbers later. They seem a bit strange. Look at the tolerances. They are one order of magnitude tighter than the CPU ones. This is strange.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -5,45 +5,102 @@
...
+    modules = ['VASP']
+    maintainers = ['LM']
+    tags = {'scs'}
+    strict_check = False
+    extra_resources = {
+        'switches': {
+            'num_switches': 1
+        }
+    }
+    force_value = -.85026214E+03
```
victorusu, line 65: This is input file dependent and, I think they should be moved to the base class. IMHO, the input files should be parameters of the base class.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -5,45 +5,102 @@
...
 
-@rfm.parameterized_test(*([v] for v in ['maint', 'prod']))
+@rfm.simple_test
 class VASPCpuCheck(VASPCheck):
-    def __init__(self, variant):
-        super().__init__()
-        self.descr = f'VASP CPU check (variant: {variant})'
-        self.valid_systems = ['daint:mc', 'dom:mc', 'eiger:mc', 'pilatus:mc']
+    valid_systems = ['daint:mc', 'dom:mc', 'eiger:mc', 'pilatus:mc']
+    executable = 'vasp_std'
```
victorusu, line 98: The executable should be on the base class too.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -91,31 +135,24 @@ def set_cpu_binding(self):
...
 class VASPGpuCheck(VASPCheck):
-    def __init__(self, variant):
-        super().__init__()
-        self.descr = f'VASP GPU check (variant: {variant})'
-        self.valid_systems = ['daint:gpu', 'dom:gpu']
-        self.executable = 'vasp_gpu'
-        self.variables = {'CRAY_CUDA_MPS': '1'}
-        self.num_gpus_per_node = 1
+    valid_systems = ['daint:gpu', 'dom:gpu']
+    executable = 'vasp_gpu'
```
victorusu, line 141: We should have a CPU/GPU base class tests.

cscs-checks/apps/amber/amber_check.py
```diff
@@ -131,29 +101,33 @@ def set_perf_patterns(self):
...
+        }})
+
 
 @rfm.simple_test
-class AmberGPUCheck(AmberBaseCheck):
-    num_tasks = 1
-    num_tasks_per_node = 1
-    num_gpus_per_node = 1
+class amber_gpu_check(AmberCheck):
+    input_file = 'mdin.GPU'
```
victorusu, line 113: I think we should have CPU/GPU base classes.

victorusu:
@jjotero and @hurricane642, I think we should have a meeting about where we will split the tests...
IMHO, currently, the issue is that if we have a single base class, then one can specialize as much as one wants. However, the "usefulness" is not so great because one needs still to be an "expert" of the code to use it. For example, one has to define the executables (especially if they change between CPU and GPU), executable options, the reference energies, and the tolerances depending on the accelerator or the input files. I think that we should hold this information in the base classes. 

My opinion is to have input-file dependent information in the base class as parameters, similar to what we have done with the Amber tests. We could have the `variant` parameter, which states the actual scientific system being executed, and if one needs to let's say, exclude a given test, then one should filter that test out from the `variant` list.

My current opinion for CPU and GPU tests is currently the opposite though... 😬 
I think that we should have different base classes. The issue is that one has to inherit the test twice, one for the GPU and another for the CPU. 

Having two separate tests (not classes - I mean really mean test) is a "requirement" due to the fact that most of the programs require either a different `executable` or a different set of `executable_opts` and sometimes different energy and/or energy tolerances. And this is where I think that we have a problem because it requires "experts" to set those parameters. Furthermore, I think that it also fits well, at least for our center, on how we split/map physically the CPU and GPU tests into the system. Currently, we do it thru the virtual partitions. In this case, it is "easier" to set the valid virtual partitions for the class that inherits from the appropriate class and avoids to the set the `executable` and/or executable options based. Of course, this is a matter of style.

In this proposal, then the hierarchy of tests in the library would look like this:
```
# In the library
ClassA (input file dependent info)  <-- ClassB (CPU only test)
                                                           <-- ClassC (GPU only test)
```

Then, a centre  would import those tests as:
```python
class MyCPUTests(ClassB):
    valid_systems = ['daint:mc', 'dom:mc']
    # set the reference performance
    ...

class MyGPUTests(ClassC):
    # set the reference performance
    ...

# supporting a different accelator
class MyCustomAcceleratorTests(ClassA):
    valid_systems = ['daint:xx', 'dom:xx']
    # set the reference performance
    ...

...
```

Other than importing like this:
```python
class MyGPUTests(ClassA):
   
    @run_before('run')
    def set_gpu_executable_and_executable_opts(self):
        # requires "expert" knowledge
        ...
   
    @run_before('run')
    def set_input_files(self):
        # requires "expert" knowledge
        ...

    # set the reference performance
    ...

class MyCPUTests(ClassA):
    valid_systems = ['daint:mc', 'dom:mc']

    @run_before('run')
    def set_gpu_executable_and_executable_opts(self):
        # requires "expert" knowledge
        ...
   
    @run_before('run')
    def set_input_files(self):
        # requires "expert" knowledge
        ...
        
    # set the reference performance
    ...
...
```

One can also argue that having a parameter, e.g. `accelerator` to decide whether the test is CPU or GPU accelerate would be enough. This parameter would help separate and define the `executable` and `executable_opts`. But I argue that this will introduce complexity that is better handled by having dedicated classes. 

The classes in this style would look like:
```python
# In the library we would have
class (...):
    accelerator = parameter(['', 'gpu'])

class MyGPUTests(ClassA):
    # accelerator parameter that defines the executable and executable_opts "automagically"
    accelerator = ['gpu']
   
    # set the reference performance
    ...

class MyCPUTests(ClassA):
    valid_systems = ['daint:mc', 'dom:mc']
    # accelerator parameter that defines the executable and executable_opts "automagically"
    accelerator = ['']

    # set the reference performance
    ...
...
```

Please note that independent of having either different base classes for different accelerators or a dedicated parameter, we need to define some sort of convention per test. Because GPU on Amber means CUDA, but in other applications may mean AMD GPUs. So, this must be test-dependent.

jjotero:
@victorusu I think that you've nailed it by looking as these test as "scientific systems", as you say above. So I think that this is what the classes should represent, regardless of the code, compilation and such. I can see scenarios where one might decide to go for the class split, and others where it makes sense to keep them as a single class. So I'd say that we select approaches as we see fit on a case-by-case basis.

Onto the rationale 🙃!

- If the scientific system being tested is the same for both GPU, CPU and so on is the same, I think that they should stay in the same class. The tests in this case should check the health of some features from the code, and this meaning (I think) should not be split across different classes.
- If a test is focused on performance, and the test is built with one particular architecture in place, then I think that this should then be split.

However, this opinion might change in the next 5 minutes 😂

jjotero (changes requested):
This is moving in the right direction. I've added a few minor comments in the amber checks. I think that the amber base should be renamed to `NVE`. This is to be specific of the type of benchmark that class runs. Also (for some reason I couldn't comment on the code) the `modules` variable should be set in the derived class.

Now, the million dollar question here: _how do we deal with the GPU vs. CPU versions?_
As we discussed yesterday, the `input_file` and the `executable` depend on which of these platforms we are to run the test on, and these two variables must be set in the library level, because the user may not know (care) what they have to be set to. There're 2 options for this:

- Deriving the current base class into CPU- and GPU-specific classes. We would expose those two classes from the library as `NVE_CPU` and `NVE_GPU` (or something like that). The problem I've got with this is that once we "fork" the classes like this, there is no way back (users that want to the 2 versions of these `NVE` checks in CPU and GPU would have to inherit from 2 different classes all the time - i.e. annoying)
- Keeping all into one single class. Perhaps one of the coolest features of the ReFrame parameters is that you can cherry pick a subset of parameters inherited from a base class. Then, similarly to how we've done it so far with the `variant` parameter, we could add another one in the base class called `platform` which would look like this:
```python
    # Each platform parameter is a tuple with a tag, input file and executable.
    platform = parameter([
        ('cpu', 'mdin.CPU', 'pmemd.MPI'),
        ('gpu', 'mdin.GPU', 'pmemd.cuda.MPI'),
    ])

    @run_after('init')
    def unpack_platform_parameter(self):
        self.platform, self.input_file, self.executable = self.platform
```
With this users would just need to inherit from a single class and then set the `num_tasks` and so on as they see fit for each case. But, how about the users that just want a single `platform` out of all of the ones provided in here? This is where the `parameter` magic comes into place. If a user would like to have (for example) a CPU test only, they could do:
```python
def my_param_filter(inherited_params, match):
    '''Filter out variants by inspecting the parameter's first element.'''
    filtered_params = tuple()
    for p in inherited_params:
        if match in p[0]:
            filtered_params+=(p)
    return filtered_params

inherit_cpu_only = lambda p: my_param_filter(p, 'cpu')
inherit_gpu_only = lambda p: my_param_filter(p, 'gpu')

class nve_cpu(amber.NVE):
    platform = parameter(inherit_params=True, filter_params=inherit_cpu_only)
    ...

class nve_gpu(amber.NVE):
    platform = parameter(inherit_params=True, filter_params=inherit_gpu_only)
    ...
```
I like this second option a lot more :)

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+    '''
+
+    #: Amber output file.
+    #:
+    #: :default: : 'amber.out'
+    output_file = variable(str, value='amber.out')
+
+    #: Amber input file.
+    #:
+    #: :default: : 'amber.out'
```
jjotero, line 35:
```suggestion
    #: Amber input file. This file is set by the post-init hook
    #: :func:`unpack_platform_parameter`.
    #:
    #: :default: :class:`required`
```

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+    #: :default: :class:`required`
+    energy_value = variable(float)
+
+    #: Maximum deviation from the reference value of energy,
+    #: that is acceptable.
+    #:
+    #: :default: :class:`required`
+    energy_tolerance = variable(float)
+
+    executable_files = parameter([
```
jjotero, line 51: I would still name this `platform`. I think it's more descriptive.
```suggestion
    # Parameter pack containing the platform ID, input file and executable.
    platform = parameter([
```

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+        ('JAC_production_NVE_4fs', -44810.0, 1.0E-03),
+        ('JAC_production_NVE', -58138.0, 5.0E-04)
+    ])
+
+    num_tasks_per_node = required
+    executable = required
+
+    @run_after('init')
+    def unpack_platform_parameter(self):
+        self.platform, self.input_file, self.executable = self.executable_files
```
jjotero, line 69:
```suggestion
        self.platf_id, self.input_file, self.executable = self.platform
```

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+        ('FactorIX_production_NVE', -234188.0, 1.0E-04),
+        ('JAC_production_NVE_4fs', -44810.0, 1.0E-03),
+        ('JAC_production_NVE', -58138.0, 5.0E-04)
+    ])
+
+    num_tasks_per_node = required
+    executable = required
+
+    @run_after('init')
+    def unpack_platform_parameter(self):
```
jjotero, line 68:
```suggestion
    def unpack_platform_parameter(self):
        '''Set the executable and input file.'''
```

hpctestlib/apps/amber/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+    #:
+    #: :default: :class:`required`
+    energy_tolerance = variable(float)
+
+    executable_files = parameter([
+        ('cpu', 'mdin.CPU', 'pmemd.MPI'),
+        ('gpu', 'mdin.GPU', 'pmemd.cuda.MPI')
+    ])
+
+    # NVE simulations
```
jjotero, line 56:
```suggestion
    # NVE simulation parameter pack with the benchmark name, energy reference and energy tolerance for each case.
```

jjotero (changes requested):
Just cosmetic suggestions for the amber check.

With the current state of this check, I think that now it would make more sense to rename the `amber/__ini__.py` to `amber/nve.py` so then the import from the user code would look like 
```python
from hpctestlib.apps.amber.nve import NVE
```

What do you think?

hurricane642:
This makes sense in general, but I think it would be better to call the class `Amber_NVE`. In general, a user can test several packages in one program. And by this logic, we would have several classes with the same name` NVE`, which can cause problems.

cscs-checks/apps/cp2k/cp2k_check.py
jjotero: I agree. `executable` and `executable_opts` should go in the base class.

hpctestlib/apps/cp2k/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    (approximately the reference). The default assumption
+    is that CP2K is already installed on the device under test.
+    '''
+
+    #: Reference value of energy, that is used for the comparison
+    #: with the execution ouput on the sanity step. Final value of
+    #: energy should be approximately the same
+    #:
+    #: :default: :class:`required`
+    energy_value = variable(float)
```
jjotero, line 37: For consistency with amber:
```suggestion
    energy_reference = variable(float)
```

jjotero: For cp2k, we provide a single case, so we can set the entry references and tolerances directly here in the base class.

hpctestlib/apps/cp2k/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+# SPDX-License-Identifier: BSD-3-Clause
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+
+__all__ = ["Cp2k"]
+
+
+class Cp2k(rfm.RunOnlyRegressionTest):
```
jjotero, line 13: I now think that all the base checks related to the apps should have the name of the scientific system and not just cp2k and so on. Also, I now think that because this is not a generic cp2k class, the file should also be renamed from `__init__` to the benchmark name (like in the amber case, where the equivalent to this file is named `nve`). I guess this applies to the rest of the apps too.

cscs-checks/apps/cp2k/cp2k_check.py
jjotero: We must be consistent with the naming convention across the different apps. In amber, the `benchmark` parameter defines the benchmark name, and some other benchmark specific quantities (i.e. the scientific system). Extending this meaning of `benchmark` over to here, cp2k only has a single benchmark.

Probably makes sense to rename this parameter to something like `mode`?

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -5,58 +5,125 @@
...
+
+REFERENCE_GPU_PERFORMANCE_LARGE = {
+    'daint:mc': {
+        'maint': (199.6, None, 0.06, 's'),
+        'prod': (199.6, None, 0.06, 's')
+    }
+}
+
+
+class Cp2kCheck(Cp2k):
```
jjotero, line 64: There's only a single class deriving from this one, so you can merge them.

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -5,58 +5,125 @@
...
+    strict_check = False
+    extra_resources = {
+        'switches': {
+            'num_switches': 1
+        }
+    }
+    scale = parameter(['small', 'large'])
+    benchmark = parameter(['prod', 'maint'])
+    energy_value = -4404.2323
+    energy_tolerance = 1E-04
```
jjotero, line 79: These two go in the base

cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -5,58 +5,125 @@
...
 
+    @run_after('setup')
+    def set_perf_patterns(self):
         self.perf_patterns = {
-            'time': sn.extractsingle(r'^ CP2K(\s+[\d\.]+){4}\s+(?P<perf>\S+)',
-                                     self.stdout, 'perf', float)
+            self.benchmark: sn.extractsingle(
+                r'^ CP2K(\s+[\d\.]+){4}\s+(?P<perf>\S+)',
+                self.stdout, 'perf', float)
         }
```
jjotero, line 100: This should go in the base class

jjotero (changes requested):
Just went through the `cp2k` check. I now think that instead of naming the files in the library with `__init__`, we should name them with the relevant benchmark name. Then, potentially, others can come up with new benchmarks and add them under the same app directory. The same applies to the name of the base classes. I think they should be named after the benchmark and not after the application they're running.

hpctestlib/apps/amber/nve.py
```diff
@@ -0,0 +1,137 @@
...
+        '''Set the executable options for the Amber. Determine the
+        using of input and ouput files.
+        '''
+
+        self.executable_opts = ['-O',
+                                '-i', self.input_file,
+                                '-o', self.output_file]
+
+    @sanity_function
+    def set_sanity_patterns(self):
```
vkarak, line 126:
```suggestion
    def assert_energy_readout(self):
```

cscs-checks/apps/amber/amber_check.py
```diff
@@ -58,102 +59,55 @@
...
+
+def inherit_cpu_only(p):
+    return my_param_filter(p, 'cpu')
+
+
+def inherit_gpu_only(p):
+    return my_param_filter(p, 'gpu')
+
+
+class AmberCheck(Amber_NVE):
```
vkarak, line 79: What about this?

```suggestion
class AmberCheckCSCS(Amber_NVE):
```

cscs-checks/apps/amber/amber_check.py
```diff
@@ -58,102 +59,55 @@
...
+            filtered_params.append(p)
+    return tuple(filtered_params)
+
+
+def inherit_cpu_only(p):
+    return my_param_filter(p, 'cpu')
+
+
+def inherit_gpu_only(p):
+    return my_param_filter(p, 'gpu')
```
vkarak, line 76: I think that you can easily replace these with a single line lambda in the parameter definitions:

```python
    platform = parameter(
        inherit_params=True,
        filter_params=lambda params: p for p in params if p == 'gpu'
    )

File: cscs-checks/apps/amber/amber_check.py [hunk 46]
```diff
@@ -58,102 +59,55 @@
...
     num_tasks = 1
-    num_tasks_per_node = 1
     num_gpus_per_node = 1
-    valid_systems = ['daint:gpu', 'dom:gpu']
-    executable = 'pmemd.cuda.MPI'
-    input_file = 'mdin.GPU'
-    output_file = 'amber.out'
+    num_tasks_per_node = 1
     descr = f'Amber GPU check'
     tags = {'maintenance', 'production', 'health'}
```
vkarak, line 98: I would vote for adding some generic tags at the library level, too, and here we just augment them (or replace them if we like). What about `sciapp`, `chemistry` ?

cscs-checks/apps/amber/amber_check.py
```diff
@@ -197,3 +144,10 @@ def set_num_tasks_cray_shasta(self):
                 # need to decrease to just 8 nodes
                 self.num_nodes = 8
             self.num_tasks = self.num_nodes * self.num_tasks_per_node
+
+    @run_after('setup')
+    def set_perf_reference(self):
+        if self.scale == 'small':
+            self.reference = REFERENCE_CPU_PERFORMANCE_SMALL
```
vkarak, line 151: I'd vote for a single multi-level dictionary of references, so that we can get the tuple that we want with a single line here.

vkarak (changes requested):
I like how the Amber test is now shaped. My comments are for some implementation improvements, but do not change the design.

cscs-checks/apps/amber/amber_check.py
hurricane642: As far as I know, `for` and `if` statements cannot be part of `lambda` functions. I found a way to shorten all those awful functions a bit with a `filter`, but 2 still had to be kept. Maybe you know how else this can be improved? I tried using the `filter` directly in the `parameter`, but its syntax made it difficult.

cscs-checks/apps/amber/amber_check.py
hurricane642: I agree, it would make things easier. But then we would have to change the `set_perf_patterns` function for the new kind of dictionaries, respectively we would have to add a new artificial level of keys in `REFERENCE_GPU_PERFORMANCE` and a new parameter (`scale`) for the GPU test. I think the current version is more natural, don't you think?

jjotero:
Regarding the parameter filters, I'd suggest that we filter the tests like this
```python
def inherit_cpu_only(params):
    return tuple(filter(lambda p: 'cpu' in p[0], params))
def inherit_gpu_only(params):
    return tuple(filter(lambda p: 'gpu' in p[0], params))
```
This would still work if the library implements different cpu and gpu flavours, like `cpu+omp`, `gpu+cuda`, `gpu+hip` and so on.

cscs-checks/apps/amber/amber_check.py
```diff
@@ -57,103 +50,61 @@
...
-                                             float, item=1)
-        }
 
+    @run_before('run')
+    def set_perf_reference(self):
+        for key in list(self.REFERENCE_DICT):
+            if self.current_partition.fullname in key:
+                d = {self.current_partition.fullname:
+                     self.REFERENCE_DICT[key]}
+                self.reference = d
```
jjotero, line 84: I would say that we make this `set_perf_reference` a different hook for both the GPU and CPU cases, simply because their references above are compacted in a different format. So move this hook into the GPU-specific class, and then create a similar one in the GPU class where you also deal with the `small` & `large` scopes of the dictionary.

This also avoids having to store another reference to this `REFERENCE_DICT` in the class, which is a bit confusing.
```suggestion
        for key, val in REFERENCE_GPU_PERFORMANCE.items():
            if self.current_partition.fullname in key:
                self.reference = {'*': val}
                break
        else:
            raise ValueError(
                f'could not find a reference for the current '
                f'partition {self.current_partition.fullname!r}'
            )
```

cscs-checks/apps/amber/amber_check.py
```diff
@@ -5,41 +5,34 @@
...
+    },
+    ('eiger:mc', 'pilatus:mc'): {
+        'Cellulose_production_NVE': (3.2, -0.30, None, 'ns/day'),
+        'FactorIX_production_NVE': (7.0, -0.30, None, 'ns/day'),
+        'JAC_production_NVE': (30.0, -0.30, None, 'ns/day'),
+        'JAC_production_NVE_4fs': (45.0, -0.30, None, 'ns/day'),
+    },
 }
 
 REFERENCE_CPU_PERFORMANCE_LARGE = {
```
jjotero, line 38: The keys for this dictionary also must be made tuples.

cscs-checks/apps/amber/amber_check.py
```diff
@@ -197,3 +141,7 @@ def set_num_tasks_cray_shasta(self):
                 # need to decrease to just 8 nodes
                 self.num_nodes = 8
             self.num_tasks = self.num_nodes * self.num_tasks_per_node
+
+    @run_after('setup')
+    def set_reference_dict(self):
+        self.REFERENCE_DICT = REFERENCE_CPU_PERFORMANCE[self.scale]
```
jjotero, line 147:
```suggestion
    @run_before('performance')
    def set_perf_reference(self):
        for key, val in REFERENCE_CPU_PERFORMANCE[self.scale].items():
            if self.current_partition.fullname in key:
                self.reference = {'*': val}
                break
        else:
            raise ValueError(
                f'could not find a reference for the current '
                f'partition {self.current_partition.fullname!r}'
            )
```

jjotero: Note that the keys of `REFERENCE_CPU_PERFORMANCE_LARGE` also must be made tuples in order for this to work.

jjotero (changes requested):

cscs-checks/apps/amber/amber_check.py
hurricane642: Why? Thank to string `if self.current_partition.fullname in key:` we check either substring in the string (as in case `REFERENCE_CPU_PERFORMANCE_LARGE`) or element in a tuple

PR #2087, opened 2021-07-20
[feat] Improve handling of user programming errors

### LINKED ISSUE #1899: Issue more user-friendly errors on variable and parameter errors

vkarak:
This PR improves the handling of user programming errors in ReFrame. Generally, I can see three broad categories of user errors:

1. Standard Python programming errors, such as `NameError`, `AttributeError`, `ValueError` and `TypeError`.
2. ReFrame errors that are raised while a test is executed due to programming errors. Examples are the `ValueError`s raised by a `TypedField`.
3. ReFrame errors that are raised during the test class creation. These are all the errors related to the variable and parameter syntax, as well as all the built-ins.

Errors in (1) and (2) can be tracked back to user test line that triggered them, whereas errors in (3) cannot. So a user programming error can be defined as one of the error types in (1) and if the frame that triggered it is in the "user space." In these cases, we simply print the error message and its source context and continue by skipping the offending test. For the category (3), we can't do that, because we can't differentiate whether a `ValueError` or `AttributeError` is actually a user programming error or a bug. So for these type of errors, we raise `ReframeSyntaxError`, which is treated as a soft error, i.e., print the error message (without source context obviously, but showing the test that is being created) and skip the test. In all cases, the users can use `-v` to get the full stack trace.

This PR also deactivates and deprecates the `--ignore-check-conflicts`. It had practically no effect other than not stopping the loading of tests. Currently, if a duplicate test is loaded, it is simply ignored with a warning message.

Fixes #1899.


reframe/core/variables.py
```diff
@@ -47,7 +48,7 @@ def __init__(self, *args, **kwargs):
         self._default_value = kwargs.pop('value', Undefined)
 
         if not issubclass(field_type, fields.Field):
-            raise ValueError(
+            raise ReframeSyntaxError(
```
jjotero, line 51: Passing an invalid descriptor here as `fields` is not a syntax error. If I get a syntax error from this, as a user, I'd interpret that I have some syntax errors in this API call, but the API call is sound. Instead, the problem is just that the values that I'm passing have the wrong type. So actually, perhaps a `TypeError` would make even more sense here.

unittests/test_variables.py
```diff
@@ -180,7 +183,7 @@ def test_invalid_field():
     class Foo:
         '''An invalid descriptor'''
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ReframeSyntaxError):
```
jjotero, line 186: See comment above. I think this should be a `TypeError`

ekouts (approved):
Lgtm

reframe/core/variables.py
vkarak: I agree with that and I had doubts about this and a couple of other errors. The problem is, however, that if we raise any builtin exception, then we get a full stack trace and loading is stopped, because we assume that this is a framework error and not a user one, as it actually is.

reframe/core/decorators.py
```diff
@@ -51,22 +49,18 @@ def _instantiate_all():
             try:
                 if cls in mod.__rfm_skip_tests:
                     continue
-
```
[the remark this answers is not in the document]
vkarak: I removed the line because there is no block of code after the `if` at the same level.

reframe/core/variables.py
jjotero: To me, this seems something that should be addressed at the loader level, and not here. I also think hat we need some rationale to define what is considered a user error and what is not. Our `TypedField` descriptors also throw builtin errors when the user sets them to a wrong type. The only difference is that such errors get thrown later on.
In any case, the `field` argument here is not documented, so I still think this should be a `TypeError`.

teojgo:
@vakak should the documentation and the reference manual be also updated regarding the deprecation of `--ignore-check-conflicts`?

reframe/core/variables.py
vkarak: Ok, I thought it was documented. I can then change it back to `TypeError`. As of the user error, my proposal for the rational is in the description of this PR. And indeed it's the loader that handles them by issuing a warning. The problem with these errors is that you can't easily track their origin back to the user, especially with all the metaclass magic. So for the metaclass magic we need to issue a specific error that can help the loader differentiate.

vkarak:
@teojgo Good catch, thanks!

victorusu (approved):
`lgtm`

jjotero (approved):
I think we should create 2 issues from this PR then:

- Making the loader smarter and catching all the potential metaclass errors.
- Create the rationale in the docs.

reframe/core/variables.py
jjotero: Actually, what about a try/except block in the `load_from_file` method from the `RegressionCheckLoader`, and we catch all Exceptions there? We could then just raise a warning saying _"Could not load checks from file XXXX due to errors"_

vkarak:
I just realised that I need to update the tips and tricks tutorial about [debugging](http://localhost:8000/tutorial_tips_tricks.html#debugging). It even has an implicit reference to the `@parameterized_test` decorator 😬

vkarak:
@jjotero

> Making the loader smarter and catching all the potential metaclass errors.

What do you mean here? Currently, it catches what it needs to, since metaclasses throw `ReframeSyntaxError`. Do you mean that it should become smarter in order to allow metaclasses thrown anything?

> Create the rationale in the docs.

I think this should be an internal documentation. In the documentation ("Tips & Tricks" tutorial) I have the following now:

```
Generally, ReFrame will not print the full stack trace for user programming errors and will not block the test loading process. If a test has errors and cannot be loaded, an error message will be printed and the loading of the remaining tests will continue.
```

PR #2090, opened 2021-07-22
[packaging] Build ReFrame using `pyproject.toml` and `setup.cfg`

### LINKED ISSUE #1954: Move to a `pyproject.toml` based packaging
We should consider the transition to [PEP517](https://setuptools.readthedocs.io/en/latest/userguide/index.html#transition-to-pep517)

teojgo:
* This approach is consistent with PEP517, PEP518.

* Change the ci scripts to reflect the change.

Closes #1954


vkarak (approved):
Lgtm. I guess that the `./bootstrap.sh` is independent of this changes, right?

teojgo:
> Lgtm. I guess that the `./bootstrap.sh` is independent of this changes, right?

Yes I removed completely also from the deployment so that we catch everything.

PR #2091, opened 2021-07-23
[bugfix] Remove unnecessary file locks

### LINKED ISSUE #2088: ReFrame fails to write the topology json file if file locking is not supported on HOME
This is easily reproduced in eiger for example where `flock` throws an `Unknown error 524` error.

vkarak:
Fixes #2088.


teojgo (approved):
Checked on eiger/pilatus and works fine.

PR #2094, opened 2021-07-23
[bugfix] Do not emit `module unuse <path>` after loading modules with Lmod from a specific path

### LINKED ISSUE #2065: Revise the module loading feature of ReFrame
With Lmod, the command `module unuse <path>` unloads the loaded modules that belong to that path.

ReFrame loads modules doing like the following:
```
module use /path/to/module
module load module/version
module unuse /path/to/module
```

With Lmod it looks like that can't be done as the `module unuse` seems to unload all the modules in that path. Since ReFrame `module unuse`s  the path, it is like if the module was never loaded.

It can be reporduced on eiger with:
On eiger
```
$> module load cray-python
$> python --version
Python 3.8.5
$> module unuse /path/to/cray/python/dir

Inactive Modules:
  1) PrgEnv-cray     2) cce     3) cray-dsmml     4) cray-libsci     5) cray-mpich     6) cray-python     7) craype     8) perftools-base

$> python --version
Python 2.7.18
```

rsarm:
Closes #2065

### DIFF
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -1154,3 +1154,4 @@ class SpackImpl(ModulesSystemImpl):
 
     def emit_unload_instr(self, module):
         return [f'spack unload {module.fullname}']
+                                                                                                                            
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -876,17 +876,20 @@ class TMod4Impl(TModImpl):
 
         return super().conflicted_modules(module)
 
+    def _load_instr_collections(self, module):
+        cmds = [f'module restore {module}']
+
+        # Here we append module searchpath removal/addition commands
+        # since 'restore' discards previous module path manipulations
+        for op, mp in self._extra_module_paths:
+            operation = 'use' if op == '+' else 'unuse'
+            cmds += [f'module {operation} {mp}']
+
+        return cmds
+
     def emit_load_instr(self, module):
         if module.collection:
-            cmds = [f'module restore {module}']
-
-            # Here we append module searchpath removal/addition commands
-            # since 'restore' discards previous module path manipulations
-            for op, mp in self._extra_module_paths:
-                operation = 'use' if op == '+' else 'unuse'
-                cmds += [f'module {operation} {mp}']
-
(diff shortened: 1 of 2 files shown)


rsarm:
@jenkins-cscs retry daint

ekouts (approved):
lgtm

unittests/test_modules.py
```diff
@@ -236,16 +236,12 @@ def _emit_load_commands_tmod(modules_system):
     assert emit_cmds('m0') == ['module load m0']
 
 
-def _emit_load_commands_tmod4(modules_system):
-    emit_cmds = modules_system.emit_load_commands
+def _emit_load_commands_tmod4_lmod(modules_system, emit_cmds):
```
vkarak, line 239: I don't like that you pass here the `emit_cmds` as a list that you modify it in the function. Actually, you don't need that function at all. Since the lmod commands are those that don't have the final `unuse`, you should use that as your base function and then call that from the `tmod4` version of it, by adding to it the missing `unuse`.

reframe/core/modules.py
```diff
@@ -876,17 +876,20 @@ def conflicted_modules(self, module):
 
         return super().conflicted_modules(module)
 
-    def emit_load_instr(self, module):
-        if module.collection:
-            cmds = [f'module restore {module}']
+    def _load_instr_collections(self, module):
```
vkarak, line 879:
```suggestion
    def _emit_restore_instr(self, module):
```

reframe/core/modules.py
```diff
@@ -1004,6 +1007,18 @@ def unload_all(self):
...
+    def emit_load_instr(self, module):
+        if module.collection:
+            return self._load_instr_collections(module)
+
+        cmds = []
+        if module.path:
+            cmds.append(f'module use {module.path}')
+
+        cmds.append(f'module load {module.fullname}')
+
```
vkarak, line 1019: Remove this line.

vkarak (changes requested):

vkarak (approved):
lgtm

PR #2095, opened 2021-07-23
[feat] Use `RegressionTestMeta` in the build system classes

### LINKED ISSUE #2030: Make the `BuildSystem` classes use the machinery from the `variables`.

jjotero:
Closes #2030.


vkarak (approved):
lgtm

PR #2100, opened 2021-07-27
[bugfix] Make `test_cancel_(with_grace|term_ignore)` unit tests more robust

### LINKED ISSUE #2069: Make `test_cancel_term_ignore` unit test more robust
This unit tests fails occasionally in all CIs. We should have a look into it:

```
________________________ test_cancel_term_ignore[local] ________________________

minimal_job = <reframe.core.schedulers.local._LocalJob object at 0x2b4a75e3d780>
scheduler = <class 'reframe.core.schedulers.local.LocalJobScheduler'>
local_only = None

    def test_cancel_term_ignore(minimal_job, scheduler, local_only):
        # This test emulates a descendant process of the spawned job that
        # ignores the SIGTERM signal:
        #
        #   reframe --- local job script --- sleep_deeply.sh --- sleep
        #                                      (TERM IGN)
        #
        #  Since the "local job script" does not ignore SIGTERM, it will be
        #  terminated immediately after we cancel the job. However, the deeply
        #  spawned sleep will ignore it. We need to make sure that our
        #  implementation grants the sleep process a grace period and then
        #  kills it.
        minimal_job.time_limit = '1m'
        prepare_job(minimal_job,
                    command=os.path.join(test_util.TEST_RESOURCES_CHECKS,
                                         'src', 'sleep_deeply.sh'),
                    pre_run=[''],
                    post_run=[''],
                    prepare_cmds=[''])
        minimal_job.submit()
    
        # Stall a bit here to let the the spawned process start and install its
        # signal handler for SIGTERM
        time.sleep(1)
    
        t_grace = time.time()
        minimal_job.cancel()
        time.sleep(0.1)
        minimal_job.wait()
        t_grace = time.time() - t_grace
    
        # Read pid of spawned sleep
        with open(minimal_job.stdout) as fp:
            sleep_pid = int(fp.read())
    
        assert t_grace >= 2 and t_grace < 5
        assert minimal_job.state == 'FAILURE'
        assert minimal_job.signal == signal.SIGKILL
    
        # Verify that the spawned sleep is killed, too, but back off a bit in
        # order to allow the sleep process to wake up and get the signal
        time.sleep(0.1)
>       assert_process_died(sleep_pid)

unittests/test_schedulers.py:574: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

pid = 28671

    def assert_process_died(pid):
        try:
            os.kill(pid, 0)
            if os.getpid() == 1:
                # We are running in a container; so pid is likely a zombie; reap it
                if os.waitpid(pid, os.WNOHANG)[0] == 0:
                    pytest.fail(f'process {pid} is still alive')
            else:
>               pytest.fail(f'process {pid} is still alive')
E               Failed: process 28671 is still alive

unittests/test_schedulers.py:481: Failed
```

teojgo:
Fixes #2069

### DIFF
--- a/unittests/test_schedulers.py
+++ b/unittests/test_schedulers.py
@@ -123,6 +123,20 @@ def prepare_job(job, command='hostname',
         )
 
 
+def get_sleep_pid(job, attempts=3):
+    # Try reading the pid of spawned sleep, until a valid value is retrieved
+    for i in range(attempts):
+        try:
+            with open(job.stdout) as fp:
+                sleep_pid = int(fp.read())
+                return sleep_pid
+        except ValueError:
+            time.sleep(1)
+
+    pytest.fail(f'failed to retrieve the spawned sleep process pid after '
+                f'{attempts} attempts')
+
+
 def assert_job_script_sanity(job):
     '''Assert the sanity of the produced script file.'''
     with open(job.script_filename) as fp:
@@ -130,6 +144,20 @@ def assert_job_script_sanity(job):
                              fp.read())
         assert ['echo prepare', 'echo prerun', 'hostname',
                 'echo postrun'] == matches
+
+
+def get_sleep_pid(job, attempts=3):
+    # Try reading the pid of spawned sleep, until a valid value is retrieved
+    for i in range(attempts):
+        try:
+            with open(job.stdout) as fp:
+                sleep_pid = int(fp.read())
+                return sleep_pid
+        except ValueError:
+            time.sleep(1)
+
+    pytest.fail(f'failed to retrieve the spawned sleep process pid after '
+                f'{attempts} attempts')
 
 
 def _expected_sge_directives(job):
@@ -525,19 +553,7 @@ def test_cancel_with_grace(minimal_job, scheduler, local_only):
     # signal handler for SIGTERM
     time.sleep(1)
 
-    # Try reading the pid of spawned sleep, until a valid value is retrieved
-    for i in range(3):
-        try:
-            with open(minimal_job.stdout) as fp:
-                sleep_pid = int(fp.read())
-        except ValueError:
-            time.sleep(1)
-            continue
-        else:
-            break
-    else:
-        pytest.fail('failed to retrieve the spawned sleep process pid')
-
+    sleep_pid = get_sleep_pid(minimal_job)
     t_grace = time.time()
     minimal_job.cancel()
     time.sleep(0.1)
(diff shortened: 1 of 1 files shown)


unittests/test_schedulers.py
```diff
@@ -525,16 +525,25 @@ def test_cancel_with_grace(minimal_job, scheduler, local_only):
...
+    # Try reading the pid of spawned sleep, until a valid value is retrieved
+    for i in range(3):
+        try:
+            with open(minimal_job.stdout) as fp:
+                sleep_pid = int(fp.read())
+        except ValueError:
+            time.sleep(1)
+            continue
+        else:
+            break
```
jjotero, line 537: I think we don't need the `else` here since you've got the `continue` when you catch the exception.
```suggestion

        break
```

jjotero (commented):
Since this code shows up twice in here, would it make sense to make this a function that takes an`stdout` and returns the `sleep_pid`?

unittests/test_schedulers.py
teojgo: I need it in order to break out of the loop without raising the pytest failure in case the pid is retrieved

jjotero: Sure, but you'd still be breaking out of the loop without the `else`. If you move the `break` statement after the `try`/`except` block, it will only be executed when the `try` succeeds.

Alternatively, you can leave the `break` in that `else`, but then I'd suggest to remove the `continue` from the `except` since it is redundant here.

teojgo: Ok, it's clearer now, thanks.

unittests/test_schedulers.py
```diff
@@ -123,6 +123,20 @@ def prepare_job(job, command='hostname',
         )
 
 
+def get_sleep_pid(job, attempts=3):
+    # Try reading the pid of spawned sleep, until a valid value is retrieved
+    for i in range(attempts):
```
vkarak, line 128: Is there a reason for not trying indefinitely? There is no way that you will not get it, right?

unittests/test_schedulers.py
```diff
@@ -132,6 +146,20 @@ def assert_job_script_sanity(job):
                 'echo postrun'] == matches
 
 
+def get_sleep_pid(job, attempts=3):
```
vkarak, line 149: Why do you need this function twice?

vkarak (changes requested):

unittests/test_schedulers.py
teojgo: Oops, I did an extra copy paste

unittests/test_schedulers.py
teojgo: I don't try indefinetely to be on the safe side so that it does not run forever

unittests/test_schedulers.py
```diff
@@ -123,6 +123,20 @@ def prepare_job(job, command='hostname',
...
+    # Try reading the pid of spawned sleep, until a valid value is retrieved
+    for i in range(attempts):
+        try:
+            with open(job.stdout) as fp:
+                sleep_pid = int(fp.read())
+                return sleep_pid
+        except ValueError:
```
jjotero, line 133:
```suggestion
                return sleep_pid

        except ValueError:
```

jjotero (commented):
Just a very minor comment

jjotero (changes requested):
Looks good except for that `get_sleep_pid` function that is defined twice.

jjotero (approved):
lgtm

vkarak (approved):
I just did a couple of minor improvements.

PR #2102, opened 2021-07-28
[ci] Test run Spack and EasyBuild tutorial examples in a container

rsarm:


ci-scripts/dockerfiles/tutorials.dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN useradd -ms /bin/bash rfmuser
+
+USER rfmuser
+
+# Install Spack
+RUN git clone https://github.com/spack/spack ~/spack && \
+    cd ~/spack && \
+    git checkout releases/v${_SPACK_VER}
+
+RUN pip3 install easybuild==${_EB_VER}
```
jgphpc, line 21:
```suggestion
# Install Easybuild
RUN pip3 install easybuild==${_EB_VER}
```

ci-scripts/dockerfiles/tutorials.dockerfile
```diff
@@ -0,0 +1,36 @@
+#
+# Execute this from the top-level ReFrame source directory
+#
+
+
+FROM reframehpc/rfm-ci-base:lmod
```
jgphpc, line 6: Is `rfm-ci-base:lmod` built using [Lmod-base.dockerfile](https://github.com/eth-cscs/reframe/blob/master/ci-scripts/dockerfiles/Lmod-base.dockerfile) ?

Maybe out of scope here but it would be nice to have a pointer to the dockerfile used.
Nvidia is always pointing to the dockerfiles they are using. For example: https://hub.docker.com/r/nvidia/cuda -> Ubuntu 20.04

ci-scripts/dockerfiles/tutorials.dockerfile
```diff
@@ -0,0 +1,36 @@
+#
+# Execute this from the top-level ReFrame source directory
```
jgphpc, line 2:
```suggestion
# Execute this from the top-level ReFrame source directory
# sudo docker build -f ci-scripts/dockerfiles/tutorials.dockerfile -t rfm-tutorials:build_systems .
```

ci-scripts/dockerfiles/tutorials.dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ENV _SPACK_VER=0.16
+ENV _EB_VER=4.4.1
+
+# ReFrame user
+RUN useradd -ms /bin/bash rfmuser
+
+USER rfmuser
+
+# Install Spack
+RUN git clone https://github.com/spack/spack ~/spack && \
```
jgphpc, line 17: I get `==> Error: spack requires 'curl'. Make sure it is in your path.`
Do you get the same ?

jgphpc:
BZip2SpackCheck fails with: `sanity error: pattern 'Version 1.0.6' not found in 'rfm_BZip2SpackCheck_job.err'`

I understand that [spack_test.py](https://github.com/eth-cscs/reframe/blob/master/tutorials/build_systems/spack/spack_test.py) sets the version of bzip2 to load in [src/myenv/spack.yaml#L3](https://github.com/eth-cscs/reframe/blob/master/tutorials/build_systems/spack/src/myenv/spack.yaml#L3):
```
specs:
    - bzip2@1.0.6
```

Then `spack env activate -V -d myenv` should load what's needed but it looks like it is not found/loaded correctly.

- Should this extra line added by ReFrame  ?
```
spack load bzip2
```
(In comparison, [eb_test.py#L26](https://github.com/eth-cscs/reframe/blob/master/tutorials/build_systems/easybuild/eb_test.py#L26) adds bzip2 with: `self.modules = self.build_system.generated_modules`)

If not, I understand that [src/myenv/spack.yaml](https://github.com/eth-cscs/reframe/blob/master/tutorials/build_systems/spack/src/myenv/spack.yaml#L9) sets the modulepath with:
```
lmod: spack/share/spack/lmod
```
but in my case (with tutorials.dockerfile), `ls BZip2SpackCheck/spack/share/spack/` returns `modules`
Should it be `lmod` instead ?

What did I miss ?

rsarm:
@jenkins-cscs retry daint

ci-scripts/dockerfiles/tutorials.dockerfile
vkarak: @jgphpc I'd skip this comment, since the statement is self explanatory.

ci-scripts/dockerfiles/tutorials.dockerfile
vkarak: @jgphpc The docker files we have in Dockerhub are only for the CI. They are not meant for supplying ReFrame.

rsarm:
@jgphpc it was wip. Now the `spack load` is there.

vkarak:
@jgphpc Can you approve this PR?

jgphpc (approved):

tutorials/build_systems/spack/spack_test.py
```diff
@@ -19,6 +19,7 @@ class BZip2SpackCheck(rfm.RegressionTest):
     @run_before('compile')
     def setup_build_system(self):
         self.build_system.environment = 'myenv'
+        self.build_system.specs = ['bzip2@1.0.6']
```
vkarak, line 22: If we change the tutorial example, we should make sure that the text is up to date. Also note that now an environment is not required anymore.

vkarak (changes requested):

vkarak (approved):

PR #2107, opened 2021-08-02
[bugfix] Fix default `executable` assignment

### LINKED ISSUE #2106: Cannot make `executable` required.
If the `excutable` variable is `required` it will always be initialised by ReFrame in the pipeline's `__rfm_init__` method to the test's name. Whereas that is absolutely fine for either `RegressionTest` or `CompileOnlyRegressionTest`, this behaviour should not occur with `RunOnlyRegressionTest` classes, where the user should have the full power on requiring the `executable` variable.

Possible workaround: move this default assignment into the `compile` stage.

jjotero:
`RunOnlyRegressionTest` classes must be able to set the `executable` variable as `required`. However, this variable was set a default value during class instantiation. This assignment has now been moved into the `compile` stage.

closes #2106

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -355,7 +355,7 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
     #: :type: :class:`str`
     #: :default: :class:`required`
     #:
-    #: .. versionchanged:: 3.8
+    #: .. versionchanged:: 3.7.3
     #:    Default value changed from ``os.path.join('.', self.name)`` to
     #:    :class:`required`.
     executable = variable(str)
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -861,10 +861,6 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
         if not hasattr(self, 'descr'):
             self.descr = self.name
 
-        # Pass if the executable is a required variable.
-        if not hasattr(self, 'executable'):
-            self.executable = os.path.join('.', self.name)
-
         self._perfvalues = {}
 
         # Static directories of the regression check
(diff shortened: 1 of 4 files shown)


teojgo (changes requested):
Since the executable is required now, you should add an executable here: https://github.com/eth-cscs/reframe/blob/8a9ceeda69619d367820d5297f5018183d52a897/unittests/test_policies.py#L139 and maybe also the resources for the unittests.

teojgo:
@jjotero @vkarak Do you think that we should also mention that the `executable` is required for the `RunOnlyRegressionTest` in the tutorial?

jjotero:
@teojgo I think it could add that in the regression test API docs instead. Like a version changed box or something like that.

reframe/core/pipeline.py
```diff
@@ -349,7 +349,12 @@ def pipeline_hooks(cls):
     #: The name of the executable to be launched during the run phase.
     #:
     #: :type: :class:`str`
-    #: :default: ``os.path.join('.', self.name)``
+    #: :default: :class:`required`
+    #:
+    #: .. versionchanged:: 3.8
```
vkarak, line 354:
```suggestion
    #: .. versionchanged:: 3.7.3
```

vkarak (changes requested):
@jjotero @teojgo I think that a small note in the tutorial to alert the readers would be good.

teojgo (approved):

vkarak (approved):
Lgtm now. I've done a minor rephrase in the docs.

ireed:
Is there a way to use two self.executable values in a single test? For example, if i want to run two different programs, how can i do something like _self.executable1='prog1'_, _self.executable2='prog2'_ with both appearing in the same submission script? This is to avoid a hack such as _self.executable='prog1\nsrun prog2'_, which i do quite often, but is not portable.

teojgo:
> Is there a way to use two self.executable values in a single test? For example, if i want to run two different programs, how can i do something like _self.executable1='prog1'_, _self.executable2='prog2'_ with both appearing in the same submission script? This is to avoid a hack such as _self.executable='prog1\nsrun prog2'_, which i do quite often, but is not portable.

If you want to run to executables with the same srun options then you can put the executable commands inside a bash script `run.sh` and then set the executable as `'run.sh'`.

vkarak:
@ireed You may also have a look on how to create scripts with multiple parallel launcher commands: https://reframe-hpc.readthedocs.io/en/stable/tutorial_advanced.html#adding-more-parallel-launch-commands

PR #2114, opened 2021-08-09
[test]  Explicitly set the `executable` in the base class of the libsci_resolve test

teojgo:


cscs-checks/compile/libsci_resolve.py
```diff
@@ -18,7 +18,7 @@ class LibSciResolveBaseTest(rfm.CompileOnlyRegressionTest):
```
jjotero, line 18: I think that just setting the executable here should do the trick. No need to replace the `executable` by the `name` in the hook below
```suggestion
    executable = 'libsci.x'
```

jjotero (approved):
lgtm

PR #2117, opened 2021-08-10
[feat] Allow setting test variables from the command line

### LINKED ISSUE #563: Pass command line options to a regression test

vkarak:
This PR adds the possibility to set test variables from the command line. More specifically, users can now pass `-S foo=X` from the command line to set the variable `foo` to `X` in all loaded tests. To set multiple variables, users should pass the `-S` option multiple times.

Behind the scenes the framework sets the default value of a variable to whatever was passed from the command line. This happen at the variable injection time, just before the final test is constructed and returned to the frontend. This has some implications:

- The body of the class has executed before the variables are set from the command line. This means that in the following example `num_tasks` will be `1` regardless of the value passed from the command-line.
```python
class my_test(rfm.RegressionTest):
    foo = variable(int, value=1)
    num_tasks = foo
```
- The variables are set *before* any filtering in the tests is performed. This gives us some nice flexibility, since the `--skip-systme-check` and `--skip-prgenv-check` option could be simply implemented by overriding the `valid_systems` and `valid_prog_environs` using this mechanism. We could even support dynamic tags that would be set from the command line and the test could act upon them. ~~The downside is that the variable change will happen to *all* the loaded tests; there is no way to limit that, unless we support a syntax like this: `-S my_test.foo=3`, which would set `foo` only for test `my_test`.~~
- Users can prefix a variable with a test name, such as `-S my_test.foo=3`, in which case `foo` will only be set for `my_test`. Notice that wildcards are not accepted in such a syntax.
- Since the variables are set *before* the test is instantiated, this mechanism will have no effect if the variable is (re)set anywhere in the pipeline, including the `init` phase. This is, however, consistent with the default values behaviour, which is exactly what this mechanism leverages.

If a variable that is asked to be set from the command line is not declared in a test, it will be silently ignored.

One key aspect of this PR is to handle transparently the conversion from the string passed to the command line to the type of the variable. ~~For the moment, the default conversion function is pretty simple. If the field is a `TypedField` it tries to convert to each one of the supported types doing `typ(s)` until it succeeds. `None` are handled specially and if the string `'None'` is passed and the accepted type is `NoneType` it will be converted to `None`. Users when declaring variables are given the option to pass a custom conversion function, in which case it will be used instead of the default. If a conversion is not possible a `TypeError` will be raised and the test will be skipped.~~
The type conversion happens in the `TypedField` by passing the command-line value wrapped in a special object. When a `TypedField` is attempted to be set by such an object, the `TypedField` will try to implicitly convert the value to any of its supported types. If no conversion is successful, only then a `TypeError` will be thrown. The implicit type conversions happen by calling `typ(s)`, where `typ` is the type we want to convert to and `s` is the source string. All built-in types accept this type of conversion, but we have extended the type system used by the `TypedField`s to allow even conversion of iterables, as well as any user type using the same convention. If the field type is `List[int]` and the user passes the string `1,2` from the command line, this will be converted to the list `[1, 2]` automatically. Behind the scenes, this works by inspecting the call to the type constructor and delegating that to a user-defined cast method for the source type. In the case of strings, this method is `__rfm_cast_str__()`, but `str` can be replaced with any other type name. Users that want to use this type of conversion with their types in `TypedField`s, will have to use the `ConvertibleType` metaclass for their classes and define the `__rfm_cast_str(s)` method as a class method. For the container types, the type metaclasses that implement the type system define this method to take care of the conversion. Here is an example of a custom type that supports implicit conversions from string:

```python
    class X(metaclass=types.ConvertibleType):
        def __init__(self, x):
            self.x = x

        @classmethod
        def __rfm_cast_str__(cls, s):
            return X(int(s))

    assert X('3').x == 3
```

Note: the `__rfm_cast_<type>__` methods can only take a single argument and do not support keyword arguments.

~~This mechanism allows us to support conversions also for built-in types that they normally do not, such as the `NoneType`. We define the `Null` type that supports conversions from `'null'` to `None` and we register to this type the `NoneType`, so that `isinstance(None, Null)` is true.~~

Conversion to `None` are treated at the loader. The loader simply checks if the passed value is `@none` and in that case sets the variable to `None`, otherwise the conversion is delegated to the `TypedField`.

## ~~Limitations~~

- ~~Conversion of iterables is not supported, but we could add it.~~
- ~~The implicit conversion of `'null'` to `None` will fail to set a string-or-None field to `None`, since the types are tried in the order they are declared. If the field was defined as `None-or-string`, the conversion would be successful. To mitigate that, all relevant variables have been changed to `None-or-string`.~~
- ~~Currently all the fields that allow `None` values are of type `type(None)`. If we want to allow implicit conversions to `None` from the command line, we should update all these fields to `typ.Null`.~~

## Todos

- [x] Document the feature
- [x] Document the new functionality of the `typecheck` module
- [x] Support conversion of iterables
- [x] Support conversions to `None` properly
- [x] Support the `mytest.var = val` syntax

Fixes #563.


victorusu:
@vkarak, I (think that I can) understand the reasoning behind the use of `*args` and `**kwargs`. 
But I remember that in 2.x ReFrame era we had a bunch of those in the framework which made data flow a bit difficult to understand. I remember that I particularly suffered from things that were crossing the boundaries of the front end and backend. We had major code refactors and some design changes that made the need for the `*args` and `**kwargs` usage minimal throughout the framework. Now, the code is much easier to follow.

So, I wonder if the use of `*args` and `**kwargs` can be avoided for the sake of code understanding and readability. 

I am particularly concerned with the functions
```python
def _instantiate(cls, inst_args, *args, **kwargs):
         kwargs['_rfm_use_params'] = True
         ...
```
Can't we name the `**kwargs` to something that makes sense here?

The interface of these functions is a bit more difficult to understand. What is expected to be passed in the `*args` and `**kwargs` for the following functions?

```python
def load_from_module(self, module, *args, **kwargs):
def load_from_file(self, filename, force=False, *args, **kwargs):
def load_from_dir(self, dirname, recurse=False, force=False, *args, **kwargs):
def load_all(self, force=False, *args, **kwargs):
```

An example of where it became difficult to reason about the code is 
```python
def load_from_file(self, filename, force=False, *args, **kwargs):
         if not self._validate_source(filename):
             return []

         try:
             return self.load_from_module(
                 util.import_module_from_file(filename, force), *args, **kwargs)
    ...
```
As you can see, in order to know what is being passed to `load_from_file` and I need to look into `util.import_module_from_file(filename, force), *args, **kwargs)` and the chain continues. 

What do you think?

reframe/frontend/loader.py
```diff
@@ -115,7 +115,7 @@ def prefix(self):
     def recurse(self):
         return self._recurse
 
-    def load_from_module(self, module):
+    def load_from_module(self, module, *args, **kwargs):
```
jjotero, line 118: As mentioned in the previous comment, if we pass the `set_check_attributes` as an argument to the constructor of the loader, there is no need for all these `*args, **kwargs` here. The `load_from_module` could just use that attribute whenever needed.

reframe/frontend/loader.py
```diff
@@ -133,7 +133,7 @@ def load_from_module(self, module):
             getlogger().debug('No tests registered')
             return []
 
-        candidates = module._rfm_gettests()
+        candidates = module._rfm_gettests(*args, **kwargs)
```
jjotero, line 136: And this is the line that requires some updates on the test registry. The only reason why the argument forwarding goes on beyond this point, is because this function returns the instances of the class, and with its current implementation there is no way to retrieve the classes. So I think that instead of forwarding these arguments, we should just make the test registry a class that the loader understands and can access the regression test classes in the registry before explicitly calling an `instantiate_all` function. Hence, if we can access the classes, we can just set the variable defaults here with a simple `setattr`.

reframe/core/decorators.py
```diff
@@ -27,33 +27,33 @@
...
+    def _instantiate(cls, inst_args, *args, **kwargs):
+        kwargs['_rfm_use_params'] = True
+        if isinstance(inst_args, collections.abc.Sequence):
+            args += inst_args
+        elif isinstance(inst_args, collections.abc.Mapping):
+            kwargs.update(args)
+
+        return cls(*args, **kwargs)
+
+    def _instantiate_all(*args, **kwargs):
```
jjotero, line 46: Rather than setting this `_instantiate_all` function as the module's `_rfm_gettests`, I think it would make more sense to make this test registry a class such that the loader can access each of the registered classes in it. Then the `instantiate_all` method can be a public member function that the loader calls.

Also, that would solve the whole `inst_args`, `*args`, `**kwargs` issue here, which is rather complex otherwise.

reframe/core/meta.py
```diff
@@ -335,13 +335,12 @@ def __call__(cls, *args, **kwargs):
 
         # Intercept constructor arguments
         _rfm_use_params = kwargs.pop('_rfm_use_params', False)
-
+        _rfm_external_vals = kwargs.pop('_rfm_external_vals', {})
         obj = cls.__new__(cls, *args, **kwargs)
 
         # Insert the var & param spaces
-        cls._rfm_var_space.inject(obj, cls)
+        cls._rfm_var_space.inject(obj, cls, _rfm_external_vals)
```
jjotero, line 342: If we leave it to the loader to do just the `setattr` as mentioned above, there is no need for any of these changes in the metaclass and in the variables.

reframe/core/variables.py
```diff
@@ -34,27 +34,47 @@ class TestVar:
...
+
+                try:
+                    return typ(s)
+                except Exception:
+                    continue
+
+            raise TypeError(errmsg)
+
+        # Set the conversion function
+        self._conv_fn = conv_fn or _default_conv
```
jjotero, line 77: The variables just store the default value until it gets injected during class instantiation. This is completely transparent of the value's type, and anything type-related is just down to the descriptor. Conceptually, I think this conversion should not be here.

reframe/core/variables.py
```diff
@@ -34,27 +34,47 @@ class TestVar:
...
         field_type = kwargs.pop('field', fields.TypedField)
         self._default_value = kwargs.pop('value', Undefined)
-
         if not issubclass(field_type, fields.Field):
             raise TypeError(
                 f'field {field_type!r} is not derived from '
                 f'{fields.Field.__qualname__}'
             )
 
+        conv_fn = kwargs.pop('conv', None)
```
jjotero, line 55: If we leave the conversion to the descriptors, there is no need to add a conversion function on every variable declaration, which is kind of inconvenient.

jjotero (commented):
I've got a bit of a similar concern as @victorusu with respect to that very long argument forwarding. If we ignore for a second the type conversion issue, the metaclass already provides the mechanism to set the default values of a variable by simply doing `cls.var = 5`. The only reason why we're not able to do so right now, is because the test registry returns the tests already instantiated, which means it would be too late for the loader to change a variable's default value. Therefore, I think this test registry is the main "bottleneck", and I think this should be made a proper class that the loader can interact with. This would avoid that very long argument forwarding, which is very difficult to keep track of.

Regarding the type conversion, I don't think it should be handled at the variable level. The variables are type-agnostic and their sole function is to hold a default value until the class is instantiated and the value gets passed to the data descriptor, whatever that descriptor is. The descriptor is the one responsible for handling the types (if it cares about types at all), and I think that this should be the pace where the type conversion is addressed, and not in the variables.

Below I described an alternative approach, which I think would simplify things.

reframe/utility/typecheck.py
```diff
@@ -93,24 +93,35 @@
...
-class _TypeFactory(abc.ABCMeta):
-    def register_subtypes(cls):
-        for t in cls._subtypes:
-            cls.register(t)
+class Type(abc.ABCMeta):
+    def __call__(cls, *args, **kwargs):
+        if (len(args) == 1 and
+            not kwargs and isinstance(args[0], str) and
+            hasattr(cls, '__rfm_cast__')):
+            return cls.__rfm_cast__(*args)
```
jjotero, line 101: I think the argument interpretation should be left to the `__rfm_cast__` method.
```suggestion
        if hasattr(cls, '__rfm_cast__'):
            return cls.__rfm_cast__(*args, **kwargs)
```

reframe/utility/typecheck.py
```diff
@@ -129,15 +140,20 @@ def __getitem__(cls, elem_type):
             raise TypeError('invalid type specification for container type: '
                             'expected ContainerType[elem_type]')
 
-        ret = _ContainerType('%s[%s]' % (cls.__name__, elem_type.__name__),
-                             cls._bases, cls._namespace)
+        ret = _SequenceType('%s[%s]' % (cls.__name__, elem_type.__name__),
+                            cls._bases, cls._namespace)
         ret._elem_type = elem_type
-        ret.register_subtypes()
+        ret.register_container_type()
```
jjotero, line 146: Isn't this already called by the `__init__` method above?

reframe/utility/typecheck.py
```diff
@@ -129,15 +140,20 @@ def __getitem__(cls, elem_type):
...
         cls.register(ret)
         return ret
 
+    def __rfm_cast__(cls, s):
+        container_type = cls._type
+        elem_type = cls._elem_type
+        return container_type(elem_type(e) for e in s.split(','))
```
jjotero, line 153: I think the special treatment for certain arguments passed to the call method can go here. This would allow to extend the supported conversions on each container independently. If the `*args` & `**kwargs` do not match any implemented conversion, then just return `cls(*args, **kwargs)`.

reframe/utility/typecheck.py
```diff
@@ -93,24 +93,35 @@
...
 
 
-# Metaclasses that implement the isinstance logic for the different aggregate
-# types
+# Metaclasses that implement the isinstance logic for the different builtin
+# container types
 
-class _ContainerType(_TypeFactory):
+class _ContainerType(Type):
+    def register_container_type(cls):
```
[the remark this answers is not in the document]
vkarak: Yes, this used to be a bit more complex before when instead of `_type` I had `_subtypes` which was a tuple.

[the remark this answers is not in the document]
vkarak: I haven't put a comment eventually, because I don't think it's needed. I've now put a check that `_type` is defined and along with the name of the function, it should be self explaining now. I've also removed some redundant calls to this method elsewhere.

reframe/utility/typecheck.py
vkarak: I don't understand very well this. Is it something more general related to the comment above or specific to this call?

reframe/utility/typecheck.py
vkarak: The way you propose it here, `__rfm_cast__` will always supersede the `__init__` constructor, if present. Is this really what we want? This is why I have scoped it here. Also in the way it is implemented now as a `classmethod`, you can't simply create an instance as `C(...)`, because that would lead to an infinite recursion. You'd have to do call explicitly `C.__new__` and `C.__init__`.

vkarak: While I see the value in `__rfm_cast__` interpreting the arguments at will, I don't like that `__init__` has no meaning. I see this functioning as a dynamic dispatch mechanism. What about allowing users to define methods per type they want to convert from and have the dispatching happening here? For example, `__rfm_cast_<type>__` will have the signature `__rfm_cast_type__(x: type) -> type`. Honestly, I don't see a scenario where your conversion function would need more than one argument; and we call it "cast"...

reframe/utility/typecheck.py
vkarak: Yes, indeed. I removed it.

reframe/utility/typecheck.py
jjotero: I like this `__rfm_cast_<type>__` option. It's far cleaner than potentially having a fat "general" `__rfm_cast__`.

>  I don't see a scenario where your conversion function would need more than one argument; and we call it "cast" anyway...

Yeah, I also think that having this check here makes sense.

reframe/utility/typecheck.py
jjotero: Yeah, this was related to the comment above, so all good.

jjotero:
I have a minor comment regarding which entity should be the one that actually updates the variables with the values from the command line (this is just me thinking out loud). It feels a bit strange that the registry's `instantiate_all` function takes the external vars as an argument and causes a side effect on the classes present in the registry. If we were to call this `instantiate_all` function multiple times from the loader (for whatever reason) as
```python
registry.instantiate_all(extern_vars)
...
registry.instantiate_all()
```
the second call would also return the test instances with the variables set to the values in `extern_vars` even though this argument is not passed in this second call. We just instantiate this once, so it's not a big deal in any case, but it just feels weird to me.
Perhaps a more explicit way to do this is setting the variables from the loader. The registry is an iterable and exposes the classes to the loader. The downside of this is that we'd be looping over the classes twice ... 😅
Alternatively, the `instantiate_all` could also copy the classes before setting the value.

vkarak:
> [...] and causes a side effect on the classes present in the registry. 

Yes, I agree. The fact that the argument passed causes a side effects works against the idea of actually passing this argument. I will fix that.

> The downside of this is that we'd be looping over the classes twice ... 😅

I bet it won't bother @victorusu's crazy parameter exploration experiments. :-)

reframe/frontend/loader.py
```diff
@@ -114,6 +118,17 @@ def prefix(self):
...
+    def _set_defaults(self, test_registry):
+        for test in test_registry:
+            for name, val in self._external_vars.items():
+                if '.' in name:
+                    testname, varname = name.split('.', maxsplit=1)
+                else:
+                    testname, varname = test.__name__, name
+
+                if testname == test.__name__:
+                    setattr(test, varname, fields.make_convertible(val))
```
jjotero, line 130: Following the same argument as above, I think we might need to do something about this line. If a user sets a regular python attribute from the command line, this here would set it to a `_Convertible` type, which the user can't do much with, and probably get an error hard to debug.
So back to the metaclass (😭). What about moving the following lines into an independent function called `setvar` where `if name not in var_space: raise AttributeError(f'{name!} is not a variable')`
https://github.com/eth-cscs/reframe/blob/4b0cb19f72af8f4b9f954133a54240011bd510e8/reframe/core/meta.py#L422-L431
Then, the `__setattr__` method in the metaclass can just call the `setvar` method in that first `try/except` block and the rest is the same. That would allow us to be 100% sure that the users are setting variables and variables only from the command line.

jjotero (commented):
Just a couple of comments on the error checking side of things. 

Also, I'm a big fan of this new `_set_defaults` function in the loader, because it now gives us the room to make this loader account for any corner case we might encounter, such as the conversion of `None` :) 
I know that is now dealt with by the `Null` type in the `typecheck` utility, but this `_set_defaults` now allows us to interpret certain values specially. For example, we could say that in order to set a variable to `None`, users would have to set it as `-S x=<None>` (or something like that) and this `_set_defaults` would match that string and then just do a `test.setvar('x', None)` (and not a `test.setvar('x', fields.make_convertible(value))`). This would allow users to set to `None` from the command line without having to use the custom `Null` type.

reframe/frontend/loader.py
```diff
@@ -114,6 +118,17 @@ def prefix(self):
     def recurse(self):
         return self._recurse
 
+    def _set_defaults(self, test_registry):
+        for test in test_registry:
+            for name, val in self._external_vars.items():
+                if '.' in name:
+                    testname, varname = name.split('.', maxsplit=1)
```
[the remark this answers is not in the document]
vkarak: The `.` in the name, as we set it here, would not set an attribute in the nested object. It won't act as an operator. It will just create an attribute at the class named `f.a`. Of course this attribute will only be accessible through `__dict__`. In any case, the change that you propose below will cover that as well, so I will not explicitly look for a single `.`.

reframe/frontend/loader.py
vkarak: One little problem I'm facing here is this. When calling `setvar()` from `__setattr__()` in the metaclass, I'd like to differentiate the error thrown when a variable is not defined and the syntax error `cannot override variable descriptor`. The obvious way would be to raise an `AttributeError`, but I'm reluctant for two reasons: (a) the error is not an attribute error (b) I don't want another try/catch in the loader to convert the error to something friendlier. What I've gone for is to define a `VariableError` that derives from `ReframeSyntaxError`. Currently, I'm throwing the `VariableError` only here, and everything works fine, but would it make sense to throw it in other locations, too? However, if we go for "throw a VariableError for any variable error," then the `cannot override variable descriptor` isn't it a variable error as well? So we're back to the start that we can't differentiate the errors raised in `setvars()`. To avoid this I'm stating that `VariableError` is raised only when a reference to a variable fails, similarly to `AttributeError`. Are we happy with this?

vkarak: Hmm, I just hit a more fundamental problem. Check this:

```
./bin/reframe -c unittests/resources/checks_unlisted/externalvars.py -r -S external_x.foo=3 -S external_y.foo=2 -S foolist=3,4
```

`foolist` is defined only in the `external_y`, so if we choose to throw a hard error for trying to set `external_x`, then we practically abort loading the the whole file. I would argue, that we simply skip inexistent variables and not set them. I'm not sure if we even want to issue a warning, because it could lead to a flood of warnings without a big reason.

reframe/frontend/loader.py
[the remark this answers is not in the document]
jjotero: Yeah, all this is covered by the code below, so all good here from my side.

reframe/core/meta.py
```diff
@@ -417,19 +433,8 @@ class attribute. This behavior does not apply when the assigned value
...
-                if not hasattr(value, '__get__'):
-                    var_space[name].define(value)
-                    return
-                elif not var_space[name].field is value:
-                    desc = '.'.join([cls.__qualname__, name])
-                    raise ReframeSyntaxError(
-                        f'cannot override variable descriptor {desc!r}'
-                    )
-
+            cls.setvar(name, value)
```
jjotero, line 437: If no `AttributeError` is raised here, this function has to return because the variable's value has been set and there's nothing else to do here.
```suggestion
            cls.setvar(name, value)
            return
```

reframe/core/meta.py
```diff
@@ -417,19 +433,8 @@ class attribute. This behavior does not apply when the assigned value
...
         except AttributeError:
             pass
```
jjotero, line 439: Perhaps it's now worth replacing this `pass` by a comment here saying that `setvar` will raise an `AttributeError` when unsuccessful at setting the variable's value.

reframe/core/meta.py
jjotero: This still works fine because all the namespace clashing checks happen during class creation, and this is after that. So even if without this return here, the default `__setattr__` below is called, this attribute gets silently replaced by the variable's descriptor during the first instantiation.

jjotero: Actually, without this return here, the `__getattribute__` above might not retrieve the right value, since the variables only take precedence over regular attributes in the class `__dict__` after the first class instantiation. In any case, this value and the variable's default value would always be the same.

jjotero (changes requested):
Just a minor comment on the `__setattr__` in the metaclass. Other than that, all this looks ready for the docs to me!

reframe/core/meta.py
vkarak: You are right. Factoring out the the code to `setvar()` is not equivalent to the code before, because previously you were returning from the `__setattr__()` and now I'm only returning from `setvar()` 😬

reframe/core/meta.py
vkarak: I've replaced also the `pass` later on with a comment and added a couple more comments.

reframe/core/meta.py
vkarak: It needs another little twist apart from the `return`. We need to handle the case when the class variables are converted to their internal descriptors at injection time. Previously, this was falling through to the final `super().__setattr_(name, value)`, but by putting just the `return` here, we cut that fall through, causing everything to fail. We need an explicit `else` in the `setvar()` where we're checking whether we're actually injecting the descriptor or not and call the superclass' `__setattr__` explicitly.

jjotero: Ah true! I'd suggest that we do this a little different then. How about making `setvar()` return a `bool` indicating whether the variable was set or not? With this, the bit in `__setattr__()` would just be
```python
if cls.setvar(name, value):
    return
```
`setvar()` could also return `False` when injecting the descriptor, so that it would fall through to the base `__setattr__`.

reframe/core/meta.py
```diff
@@ -417,31 +437,22 @@ class attribute. This behavior does not apply when the assigned value
...
 
         # Catch attempts to override a test parameter
         try:
             param_space = super().__getattribute__('_rfm_param_space')
             if name in param_space.params:
                 raise ReframeSyntaxError(f'cannot override parameter {name!r}')
-
         except AttributeError:
-            pass
+            '''Test is not parametrized'''
```
jjotero, line 453: Just for consistency with the `__getattr__` above.
```suggestion
            '''Catch early access attempt to the parameter space.'''
```

reframe/core/meta.py
```diff
@@ -401,6 +401,26 @@ def __getattr__(cls, name):
...
+                desc = '.'.join([cls.__qualname__, name])
+                raise ReframeSyntaxError(
+                    f'cannot override variable descriptor {desc!r}'
+                )
+            else:
+                # We are injecting the underlying descriptor of the variable,
+                # so set it as a normal class attribute
+                super(cls).__setattr__(name, value)
+        else:
+            raise AttributeError(f'{cls.__name__!r} has no variable {name!r}')
```
jjotero, line 422: This is what I meant with my comment above.
```suggestion
        try:
            var_space = super().__getattribute__('_rfm_var_space')
            if name in var_space:
                if not hasattr(value, '__get__'):
                    var_space[name].define(value)
                    return True
                elif var_space[name].field is not value:
                    desc = '.'.join([cls.__qualname__, name])
                    raise ReframeSyntaxError(
                        f'cannot override variable descriptor {desc!r}'
                    )
        except AttributeError:
            '''Catch early access attempt to the variable space.'''
        finally:
            return False

```

jjotero: Then, the part in `__setattr__` is just
```python
if cls.setvar(name, value):
    return
```
and you don't have to suppress that exception in the loader when setting the variables from the command line

reframe/core/meta.py
vkarak: I've also commented properly the `setvar()` now for future reference.

reframe/core/meta.py
vkarak: One problem with this suggestion is that we can't return in `finally` and also raise an exception in some other code path. The exception will be suppressed. If we stick with `setvar()` returning boolean I can fix that.

jjotero: I guess we could also just move that `return False` after the `try/except`, but this also looks good to me.

reframe/core/meta.py
jjotero: Maybe we should add in there "change this at your own risk" 😂

jjotero:
Looks good!

reframe/core/buildsystems.py
```diff
@@ -236,7 +236,7 @@ class Make(BuildSystem):
     #:
     #: :type: :class:`str`
     #: :default: :class:`None`
-    makefile = variable(str, type(None), value=None)
+    makefile = variable(type(None), str, value=None)
```
vkarak, line 239: I will revert all these changes, since they are not needed now that the loader handles the `@none` special value.

reframe/frontend/loader.py
```diff
@@ -119,6 +119,10 @@ def recurse(self):
         return self._recurse
 
     def _set_defaults(self, test_registry):
+        if not test_registry:
+            return
```
jjotero, line 123: Do we need this?

jjotero (changes requested):

docs/manpage.rst
```diff
@@ -386,6 +386,61 @@ Options controlling ReFrame execution
...
+   - In the following test, :attr:`num_tasks` will have always the value ``1`` regardless of any command-line assignment of the variable :attr:`foo`:
+
+   .. code-block:: python
+
+      @rfm.simple_test
+      class my_test(rfm.RegressionTest):
+          foo = variable(int, value=1)
+          num_tasks = foo
+
+   - If the variable is set in any pipeline hook, the command line assignment will have an effect until the variable assignment in the pipeline hook is reached.
```
teojgo, line 425:
```suggestion
   - If the variable is subsequently set in a pipeline hook, the command line assignment will have an effect only until the variable assignment in the pipeline hook is reached.
```

docs/manpage.rst
```diff
@@ -386,6 +386,61 @@ Options controlling ReFrame execution
...
+     The variable will be then overwritten.
+   - The `test filtering <#test-filtering>`__ happens *after* a test is instantiated, so the only way to scope a variable assignment is to prefix it with the test class name.
+     However, this has some positive side effects:
+
+     - Doing ``-S valid_systems='*'`` and ``-S valid_prog_environs='*'`` is the equivalent of passing the :option:`--skip-system-check` and :option:`--skip-prgenv-check` options.
```
teojgo, line 430:
```suggestion
     - The command line options ``-S valid_systems='*'`` and ``-S valid_prog_environs='*'`` are equivalent to passing the :option:`--skip-system-check` and :option:`--skip-prgenv-check` options.
```

docs/manpage.rst
```diff
@@ -386,6 +386,61 @@ Options controlling ReFrame execution
...
+     - Users could alter the behavior of tests based on tag values that they pass from the command line, by changing the behavior of a test in a post-init hook based on the value of the :attr:`~reframe.core.pipeline.RegressionTest.tags` attribute.
+     - Users could force a test with required variables to run if they set these variables from the command line.
```
teojgo, line 432:
```suggestion
     - Users could force a test with required variables to run only if they set these variables from the command line.
```

docs/manpage.rst
```diff
@@ -386,6 +386,61 @@ Options controlling ReFrame execution
...
+       For example, the following test could only be run if invoked with ``-S num_tasks=<NUM>``:
```
teojgo, line 433:
```suggestion
       For example, the following test will run only when invoked with ``-S num_tasks=<NUM>``:
```

teojgo (changes requested):
Lgtm, I have some minor comments that you can take into account if you think they make sense.

reframe/frontend/loader.py
vkarak: Yes, if we load a file that has only `parameterized_test`s then the registry is empty. We do a similar thing here:

https://github.com/eth-cscs/reframe/blob/da18b68ab7cdcd2057d5b1eed1f77db2b3588631/reframe/frontend/loader.py#L144

docs/manpage.rst
```diff
@@ -386,6 +386,61 @@ Options controlling ReFrame execution
...
+
+   - Sequence types: ``-S seqvar=1,2,3,4``
+   - Mapping types: ``-S mapvar=a:1,b:2,c:3``
+
+   Conversions to arbitrary objects are also supported.
+   See :class:`~reframe.utility.typecheck.ConvertibleType` for more details.
+
+
+   The optional ``TEST.`` prefix refers to the test class name (*not* the test name).
+   If the test name is not explicitly set through the :attr:`~reframe.core.pipeline.RegressionTest.name` attribute or if the test is not parametrised, then the class name is also the test name.
```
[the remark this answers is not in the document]
vkarak: I am trying to see it from a user perspective that may not be so familiar with ReFrame tests. In such cases, it can be confusing for them the "test class name" vs the "test name," given also that tests are listed by name. I could make it the sentence a bit less specific here by saying sth like "generally, the test name is the same as the test class name" and refer to the `name` attribute, which btw needs to be expanded, because we don't mention anything about the default name.

docs/manpage.rst
vkarak: I prefer my version here.

docs/manpage.rst
vkarak: I'd probably change "Doing" to "Passing".

docs/manpage.rst
vkarak: I prefer the original text here.

docs/manpage.rst
vkarak: I want to stress that this is the only way to run it.

docs/manpage.rst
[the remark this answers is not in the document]
jjotero: My point is that there is no need for a reference to the `name` attribute at all. The first sentence says that `TEST` is the  class name, not the test name. And the second one explains the default values of the test `name` attribute, even though the previous sentence said that this is attribute has nothing to do with this command line option.

To remove the ambiguity, I'd simply change this to  _The optional ``TEST.`` prefix refers to the test class._

reframe/frontend/loader.py
jjotero: But when `not test_registry`, this `if` statement has the same effect as the `for` loop below, that's why I don't see the need for it.

vkarak: It is `None`. I could make the test more explicit.

docs/manpage.rst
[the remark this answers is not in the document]
vkarak: Ok, I removed completely the sentence with the reference to the `name` attribute.

jjotero (approved):
lgtm

teojgo (approved):

PR #2119, opened 2021-08-11
[feat] Refactor test registry

jjotero:
The test registry is now a class that the loader can interact with. This `TestRegistry` class has a public member function `instantiate_all` which is equivalent to `_rfm_gettests()` in the previous implementation. This should give the loader far more control over the test instantiation and simplify the implementation of #2117.

However, the `parameterized_test` decorator deals with the construction arguments in a different way and it's incompatible with this new registry. In any case, since this decorator is deprecated, we can keep the current test registry as a "legacy_registry" which is only used by tests decorated as `parameterized_test`.

Similarly, the deprecated `require_version` decorator also complicates things slightly, forcing the `TestRegistry` class to account for the tests to be skipped. All this extra logic can be trimmed out once the `require_version` decorator is dropped.


reframe/core/decorators.py
```diff
@@ -27,7 +27,84 @@
...
+    def create(cls, test, *args, **kwargs):
+        obj = cls()
+        obj.add(test, *args, **kwargs)
+        return obj
+
+    def add(self, test, *args, **kwargs):
+        if test in self._tests:
+            self._tests[test].append((args, kwargs))
+        else:
+            self._tests[test] = [(args, kwargs)]
```
vkarak, line 57: We could write this in a simpler way:

```suggestion
        self._tests.setdefault(test, [])
        self._tests[test].append((args, kwargs))
```

reframe/frontend/loader.py
```diff
@@ -129,16 +135,24 @@ def load_from_module(self, module):
...
-            )
-            return []
+        candidates = registry.instantiate_all() if registry else []
+        legacy_candidates = legacy_registry() if legacy_registry else []
+        for c in (candidates, legacy_candidates):
+            if not isinstance(c, collections.abc.Sequence):
+                getlogger().warning(
+                    f'Tests not registered correctly in {module.__name__!r}'
+                )
+                return []
```
vkarak, line 152: To be honest, I think that this is a leftover from the days that user had to provide the `_get_checks()` method that instantiated the tests, so we had to check that what they returned was a sequence. We can remove this completely.

vkarak (changes requested):
Lgtm. Just a couple of minor comments.

reframe/frontend/loader.py
jjotero: Yeah, I also noticed that this was not covered by the tests.

vkarak (approved):

PR #2122, opened 2021-08-11
[bugfix] Catch `TypeError` when `filter_params` function does not return a tuple

### LINKED ISSUE #2118: Wrong `filter_param` function in the `parameter` builtin can raise a `TypeError`
If the `filter_param` function passed to the `parameter` builtin does not return a tuple, it will raise a `TypeError`. This requirement should be relaxed to just require an iterable on return. If the return of the filtering function is not an iterable, a `ReframeSyntaxError` should be raised instead.

Currently, none of this is in the docs, so it should also be documented when this gets addressed.

Reproducer:
```python
    valid_systems = ['*']
    valid_prog_environs = ['*']
    executable 'echo bananas'

    p = parameter(range(10))

    @sanity_function
    def assert_bananas(self):

class Derived(Base):
    # If the filter_params function does not return a tuple, we get a TypeError
    p = parameter(inherit_params=True, filter_params=lambda x: 1)
```
This produces
```bash
./bin/reframe: skipping test file '/users/jjotero/reframe/tutorials/my_tests/test.py': type error: tutorials/my_tests/test.py:50: unsupported operand type(s) for +: 'int' and 'tuple' 
```

jjotero:
The requirements of this function have now been relaxed to just require an iterable in return. If this requirement is still not met, we now raise a `ReframeSyntaxError`.

Closes #2118


ekouts (approved):

docs/regression_test_api.rst
```diff
@@ -79,14 +80,33 @@ Built-in types
...
+     If no values are provided when creating a new parameter, the parameter is considered as *declared* but not *defined* (i.e. an abstract parameter).
+     On the other hand, if the parameter was defined in any of the base classes, the parameter may inherit the values provided in the parent class, and any values provided in the current class may effectively extend the values for this parameter.
+  :param inherit_params: If :obj:`True`, any parameter values defined in any base class will be inherited.
+     However, if this argument is set as :obj:`True` and the parameter does not exist in any of the parent parameter spaces, it has the same effect as setting this option to :obj:`False`.
   :param filter_params: Function to filter/modify the inherited parameter values that may have been provided in any of the parent parameter spaces.
      This function must accept a single argument, which will be passed as an iterable containing the inherited parameter values.
-     This only has an effect if used with ``inherit_params=True``.
+     This function must also return an iterable, and it will only have an effect if used with ``inherit_params=True``.
```
vkarak, line 109: "This function must accept a single iterable argument and return an iterable. It will be called with the inherited parameter values and it must return the filtered parameter set. This function has an effect only if `inherit_params=True`.

reframe/core/parameters.py
```diff
@@ -124,9 +124,17 @@ def extend(self, cls):
...
         for name, p in local_param_space.items():
-            self.params[name] = (
-                p.filter_params(self.params.get(name, ())) + p.values
-            )
+            try:
+                self.params[name] = (
+                    tuple(p.filter_params(self.params.get(name, ()))) +
+                    p.values
+                )
+            except TypeError:
```
vkarak, line 132: Here we are masking potential `TypeError` raised in the user callable. I'd rather use the `is_trivially_callable()` utility from #2083 and raise a separate `TypeError` (I don't think that `ReframeSyntaxError` is needed here, since the error is triggered immediately by the offending user line).

unittests/test_parameters.py
```diff
@@ -72,13 +72,26 @@ class MyTest(TwoParams):
...
 
 
+def test_wrong_filter():
+    with pytest.raises(ReframeSyntaxError):
+        class Foo(ExtendParams):
+            P1 = parameter(inherit_params=True, filter_params=lambda x: 1)
+
+    with pytest.raises(ReframeSyntaxError):
+        class Bar(ExtendParams):
+            P1 = parameter(inherit_params=True, filter_params=lambda x, y: [])
```
vkarak, line 92: Hmm, I don't think that we document that the returned iterable must not be empty, do we?

vkarak (changes requested):

reframe/core/parameters.py
jjotero: Good point. I've split this into multiple steps to be able to raise the appropriate error on each step. Regarding the use of the `is_trivially_callable()` utility, that can be added instead in the `TestParam`'s `__init__()` once #2083 is in.

unittests/test_parameters.py
jjotero: Oh, the issue here is with the two arguments in the filtering lambda, not the empty iterable on return.

jjotero: I've removed this unit tests for now until the `is_trivially_callable()` becomes available.

reframe/core/parameters.py
```diff
@@ -125,16 +125,17 @@ def extend(self, cls):
...
-                )
-            except TypeError:
-                raise ReframeSyntaxError(
-                    f"argument 'filter_param' must be a callable taking a "
-                    f"single argument that returns an iterable "
-                    f"(parameter {name!r})"
-                )
+                filt_vals = p.filter_params(self.params.get(name, ()))
+            except Exception as err:
+                raise err
```
vkarak, line 130:
```suggestion
            except Exception:
                raise
```

reframe/core/parameters.py
```diff
@@ -125,16 +125,17 @@ def extend(self, cls):
...
+            else:
+                try:
+                    self.params[name] = (tuple(filt_vals) + p.values)
+                except TypeError:
+                    raise ReframeSyntaxError(
+                        f"'filter_param' must return an iterable "
+                        f"(parameter {name!r})"
+                    )
```
vkarak, line 138:
```suggestion
                    ) from None
```

vkarak (changes requested):
Lgtm, just a couple of minor comments.

vkarak:
I am fixing those two little comments.

vkarak (approved):

PR #2123, opened 2021-08-11
[feat] Automatically detect location of Spack

### LINKED ISSUE #2097: Automatic handling of `SPACK_ROOT`?
The Spack build system implicitly requires the `SPACK_ROOT` environment variable to be defined, but this requirement isn't written down explicitly, besides the fact the variable is now mentioned in the new [Spack tutorial](https://reframe-hpc.readthedocs.io/en/v3.7.1/tutorial_build_automation.html#using-spack-to-build-the-test-code).

The problem I'm facing is that `SPACK_ROOT` may be defined in the login node but not in the compute node.  The workaround I'm using is to have in the tests
```py
    variables = {
        # ...
        'SPACK_ROOT': os.getenv('SPACK_ROOT'),
    }
```
I guess I may not the be only one in this situation.  Should there be a more automatic handling of `SPACK_ROOT` (maybe automatise what I'm manually doing in the tests)?

giordano:
This removes the need to set the `SPACK_ROOT` environment variable.  Fixes #2097.

~I'm not sure how to test this though as it requires a working Spack to generate
the script.~


giordano:
I guess the obvious question is: why don't you do
```py
cmds = ['. $(spack location --spack-root)/share/spack/setup-env.sh']
```
?  The answer is because I'm not sure that'd work on my system: the problem I was seeing in #2097 is that `SPACK_ROOT` was set in the login node, but not on the compute one, despite the fact it is in my `.bashrc`.  Likewise, `spack` would be in the `PATH` on the login node but not on the compute one.  However I'll need to do some more investigation about whether using `$(spack location --spack-root)` would work.  I think that not requiring to set the variable `SPACK_ROOT` would be good anyways though.

giordano:
Ok, I finally figured out what was my trouble in #2097: the job scripts had
```sh
#!/bin/bash
```
as shebang, so my init file wasn't loaded.  The solution was to use the [`RFM_USE_LOGIN_SHELL`](https://reframe-hpc.readthedocs.io/en/stable/manpage.html#envvar-RFM_USE_LOGIN_SHELL) environment variable.  With this, `SPACK_ROOT` is now set, and `spack` is in the path.

This pull request replaces `$SPACK_ROOT` with `spack location --spack-root`, since we require `spack` to be in the `PATH` anyways, but Spack doesn't require `SPACK_ROOT` to be set

teojgo:
Ok to test

docs/tutorial_build_automation.rst
```diff
@@ -135,7 +135,7 @@ Here is what ReFrame generates as a build script in this example:
 
 .. code:: bash
 
-   . $SPACK_ROOT/share/spack/setup-env.sh
+   . "$(spack location --spack-root)/share/spack/setup-env.sh"
```
vkarak, line 138: Do we need the quotes here? I would go without unless you have a better reason.

reframe/core/buildsystems.py
```diff
@@ -865,7 +865,7 @@ def emit_build_commands(self, environ):
         return ret
 
     def _env_activate_cmds(self):
-        cmds = ['. $SPACK_ROOT/share/spack/setup-env.sh']
+        cmds = ['. "$(spack location --spack-root)/share/spack/setup-env.sh"']
```
vkarak, line 868: Do we need the quotes here? I would go without unless you have a better reason.

vkarak (changes requested):
Lgtm. I'm not sure about whether we need the quotes or not.

reframe/core/buildsystems.py
giordano: To protect the expansion of `$(spack location --spack-root)` in case it contains whitespaces.  I'm actually very surprised double quotes aren't used anywhere in ReFrame :slightly_smiling_face:

giordano: For example, try to put
```sh
#!/bin/bash
. $(spack location --spack-root)/share/spack/setup-env.sh
```
in https://www.shellcheck.net/.  See also https://github.com/koalaman/shellcheck/wiki/SC2046

vkarak: Thanks for the inputs! I'll open an issue for shell checking all the generated scripts from ReFrame. 

> I'm actually very surprised double quotes aren't used anywhere in ReFrame 🙂

Generally, the concept has always been in ReFrame to generate as simple shell scripts as possible and have any more complicated logic in the framework. Actually, I think that this is the first time the framework emits a command substitution in the script :) Also, I wouldn't like to quote user commands that are meant to be inserted verbatim, such as  `prerun_cmds` etc.

vkarak (approved):

PR #2136, opened 2021-08-18
[bugfix] Do not crash if processor and/or device topology files are not properly formatted JSON files

### LINKED ISSUE #2110: ReFrame topology stack traces
I have recently pulled from master and I started to see this error
```
$ ~/reframe/bin/reframe -C ~/reframe/config/cscs.py -c ~/reframe/custom-reframe-checks/apps/namd/namd_large_check.py -r --performance-report -n NAMDCPUExplorationCheck_True_1_16_True_True_False_True_False_True_False
$HOME/reframe/bin/reframe: httpjson: could not connect to serverhttpjson-server:12345: [Errno -2] Name or service not known
$HOME/reframe/bin/reframe: could not initialize the httpjson handler; ignoring ...
  File "$HOME/reframe/bin/reframe", line 22, in <module>
    cli.main()
  File "$HOME/reframe/reframe/frontend/cli.py", line 611, in main
    autodetect.detect_topology()
  File "$HOME/reframe/reframe/frontend/autodetect.py", line 187, in detect_topology
    topo_file, _subschema('#/defs/processor_info')
  File "$HOME/reframe/reframe/frontend/autodetect.py", line 82, in _load_info
    return _validate_info(json.load(fp), schema)
  File "/usr/lib64/python3.6/json/__init__.py", line 299, in load
    parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
  File "/usr/lib64/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/lib64/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib64/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The solution was to 
```
rm -rf ~/.reframe
```

Apparently, if the topology file is empty, then one gets this error. @teojgo can confirm that.
We should check whether the JSON file is empty and take proper action (like regenerate the topology file) and/or catch the exception being thrown.

What do you think?

In order to reproduce the error, just create an empty topology file in the `~/reframe` folder.

ekouts:
Whenever the processor and device files exist but cannot be loaded (empty file, invalid syntax etc), Reframe just stops and throws a throwback. With this PR we ignore the `JSONDecodeError` and try to regenerate the file.
Closes #2110.

### DIFF
--- a/reframe/frontend/autodetect.py
+++ b/reframe/frontend/autodetect.py
@@ -190,8 +190,8 @@ def detect_topology():
                 found_procinfo = True
             except json.decoder.JSONDecodeError:
                 getlogger().debug(
-                f'> could not load {topo_file!r}; ignoring...'
-            )
+                    f'> could not load {topo_file!r}; ignoring...'
+                )
 
         if not found_devinfo and os.path.exists(dev_file):
             getlogger().debug(
@@ -204,8 +204,8 @@ def detect_topology():
                 found_devinfo = True
             except json.decoder.JSONDecodeError:
                 getlogger().debug(
-                f'> could not load {dev_file!r}; ignoring...'
-            )
+                    f'> could not load {dev_file!r}; ignoring...'
+                )
 
         if found_procinfo and found_devinfo:
             continue
--- a/reframe/frontend/autodetect.py
+++ b/reframe/frontend/autodetect.py
@@ -188,7 +188,10 @@ def detect_topology():
                     topo_file, _subschema('#/defs/processor_info')
                 )
                 found_procinfo = True
-            except json.decoder.JSONDecodeError:
+            except json.decoder.JSONDecodeError as e:
+                getlogger().debug(
+                    f'> error while loading the file: {e}'
+                )
                 getlogger().debug(
                     f'> could not load {topo_file!r}; ignoring...'
                 )
@@ -202,7 +205,10 @@ def detect_topology():
                     dev_file, _subschema('#/defs/devices')
                 )
                 found_devinfo = True
-            except json.decoder.JSONDecodeError:
+            except json.decoder.JSONDecodeError as e:
+                getlogger().debug(
+                    f'> error while loading the file: {e}'
+                )
                 getlogger().debug(
                     f'> could not load {dev_file!r}; ignoring...'
                 )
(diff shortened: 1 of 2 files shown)


reframe/frontend/autodetect.py
```diff
@@ -183,17 +183,29 @@ def detect_topology():
...
-            part.processor._info = _load_info(
-                topo_file, _subschema('#/defs/processor_info')
-            )
-            found_procinfo = True
+            try:
+                part.processor._info = _load_info(
+                    topo_file, _subschema('#/defs/processor_info')
+                )
+                found_procinfo = True
+            except json.decoder.JSONDecodeError:
```
vkarak, line 191:
```suggestion
            except json.decoder.JSONDecodeError as e:
```

And then we should print the error message from `e`, unless it's multiline and screws the line.

reframe/frontend/autodetect.py
```diff
@@ -183,17 +183,29 @@ def detect_topology():
...
                 f'> found devices file {dev_file!r}; loading...'
             )
-            part._devices = _load_info(dev_file, _subschema('#/defs/devices'))
-            found_devinfo = True
+            try:
+                part._devices = _load_info(
+                    dev_file, _subschema('#/defs/devices')
+                )
+                found_devinfo = True
+            except json.decoder.JSONDecodeError:
```
vkarak, line 205: See my comment above.

vkarak (changes requested):
I also think that we can add easily a unit test to test that the auto-detection is not broken if the topology files are not valid json files.

reframe/frontend/autodetect.py
ekouts: The output in verbose will be now:
```
> found topology file '/Users/ekoutsaniti/.reframe/topology/generic-default/processor.json'; loading...
> error while loading the file: Expecting value: line 1 column 1 (char 0)
> could not load '/Users/ekoutsaniti/.reframe/topology/generic-default/processor.json'; ignoring...
> no topology file found; auto-detecting...
```

teojgo (approved):

reframe/frontend/autodetect.py
vkarak: I've put everything in a single `debug()` call.

vkarak (approved):

PR #2140, opened 2021-08-19
[bugfix] Use partition's and system's environment when autodetecting the topology

### LINKED ISSUE #2115: Auto detection of remote partition topology fails with launcher not found error
When I try to auto-detect the processor topology of the remote partition, ReFrame fails to create JSON file. After looking into the verbose output, the job script it emits to detect the topology is as follows:

```
#!/bin/bash
#SBATCH --job-name="rfm-detect-job"
#SBATCH --ntasks=1
#SBATCH --output=rfm-detect-job.out
#SBATCH --error=rfm-detect-job.err
#SBATCH --partition=compute
#SBATCH --exclude=openhpc-compute-[8-15]

_onerror()
{
  exitcode=$?
  echo "-reframe: command \`$BASH_COMMAND' failed (exit code: $exitcode)"
  exit $exitcode
}

trap _onerror ERR

./bootstrap.sh
mpirun -np 1 ./bin/reframe --detect-host-topology=topo.json
```

Looks like ReFrame is not loading the modules defined in the system partition before calling `mpirun` to detect the topology. So, the job fails with error `line 19: mpirun: command not found`.  The problem will persist even if ReFrame loads the modules in the system partition and MPI is not there in the modules of system partition.

 A better solution could be to remove launcher from this job?

ekouts:
In this PR we include the partition's and system's modules in the topology autodetecting script.
- For local partitions we temporarily change the environment of Reframe to include both system's and partition's modules and variables.
- For remote partitions we temporarily change the environment to include system's modules and variables and add the partition's environment in the autodetecting script.

For this PR we also added support for addition in the `SequenceView` and `MappingView` classes.

Closes #2115.

### DIFF
--- a/reframe/frontend/autodetect.py
+++ b/reframe/frontend/autodetect.py
@@ -200,14 +200,14 @@ def detect_topology():
         if not found_procinfo:
             # No topology found, try to auto-detect it
             getlogger().debug(f'> no topology file found; auto-detecting...')
-            temp_modules = rt.system.preload_environ.modules
-            temp_vars = rt.system.preload_environ.variables
+            modules = list(rt.system.preload_environ.modules)
+            vars = dict(rt.system.preload_environ.variables.items())
             if _is_part_local(part):
-                temp_modules += part.local_env.modules
-                temp_vars += part.local_env.variables
+                modules += part.local_env.modules
+                vars.update(part.local_env.variables)
+
                 # Unconditionally detect the system for fully local partitions
-                with runtime.temp_environment(modules=temp_modules,
-                                              variables=temp_vars):
+                with runtime.temp_environment(modules=modules, variables=vars):
                     part.processor._info = cpuinfo()
 
                 _save_info(topo_file, part.processor.info)
--- a/reframe/frontend/autodetect.py
+++ b/reframe/frontend/autodetect.py
@@ -10,10 +10,10 @@ import shutil
 import tempfile
 
 import reframe as rfm
+import reframe.core.runtime as runtime
 import reframe.utility.osext as osext
 from reframe.core.exceptions import ConfigError
 from reframe.core.logging import getlogger
-from reframe.core.runtime import runtime
 from reframe.core.schedulers import Job
 from reframe.utility.cpuinfo import cpuinfo
 
@@ -60,7 +60,7 @@ class _copy_reframe:
 def _subschema(fragment):
     '''Create a configuration subschema.'''
 
-    full_schema = runtime().site_config.schema
+    full_schema = runtime.runtime().site_config.schema
     return {
         '$schema': full_schema['$schema'],
         'defs': full_schema['defs'],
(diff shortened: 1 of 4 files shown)


reframe/frontend/autodetect.py
```diff
@@ -201,12 +200,18 @@ def detect_topology():
         if not found_procinfo:
             # No topology found, try to auto-detect it
             getlogger().debug(f'> no topology file found; auto-detecting...')
+            temp_modules = rt.system.preload_environ.modules
             if _is_part_local(part):
+                temp_modules += part.local_env.modules
                 # Unconditionally detect the system for fully local partitions
-                part.processor._info = cpuinfo()
+                with runtime.temp_environment(modules=temp_modules):
```
vkarak, line 207: You should also pass any `variables` that are set here. Perhaps we could modify the `temp_environment` to recognise an environment argument as well.

vkarak (changes requested):
Can you also update the description of the PR? Now you load the system modules locally, which I think is the correct thing.

reframe/frontend/autodetect.py
ekouts: I have added the variables but I am not sure it's worth it to add the environment in `temp_environment`. In case we do add it, should we call it with either environment or modules+vars? Or should we be able to call it with all three and make a new environment based on those?

reframe/frontend/autodetect.py
```diff
@@ -201,15 +201,19 @@ def detect_topology():
             # No topology found, try to auto-detect it
             getlogger().debug(f'> no topology file found; auto-detecting...')
             temp_modules = rt.system.preload_environ.modules
+            temp_vars = rt.system.preload_environ.variables
             if _is_part_local(part):
                 temp_modules += part.local_env.modules
+                temp_vars += part.local_env.variables
```
vkarak, line 207: Leave a blank line after here.

reframe/utility/__init__.py
```diff
@@ -1473,3 +1473,11 @@ def __repr__(self):
...
     def __str__(self):
         return str(self.__mapping)
+
+    def __add__(self, other):
+        if not isinstance(other, type(self)):
+            return NotImplemented
+
+        new_mapping = self.__mapping.copy()
+        new_mapping.update(other.__mapping)
+        return MappingView(new_mapping)
```
vkarak, line 1483: I don't think that it's a right approach to support `__add__` either here or in the `SequenceView`. Here we're offering a view to something that is not related to the original container. This has come as a workaround for the flex allocation node, but I don't think it's the right way. I believe that these view should provide access to the underlying container with a `data()` method.

vkarak (changes requested):

vkarak (approved):

reframe/frontend/autodetect.py
vkarak: Let's leave it as of now. That's fine.

PR #2148, opened 2021-08-24
[test] Modernize Slurm test

rsarm:


cscs-checks/system/slurm/slurm.py
```diff
@@ -285,9 +286,12 @@ class slurm_response_check(rfm.RunOnlyRegressionTest):
...
-        self.sanity_patterns = sn.assert_eq(self.job.exitcode, 0)
+    @sanity_function
+    def assert_exitcode_zero(self):
+        return sn.assert_eq(self.job.exitcode, 0)
+
+    @run_before('performance')
+    def set_perf_patterns(self):
         self.perf_patterns = {
             'real_time': sn.extractsingle(r'real (?P<real_time>\S+)',
                                           self.stderr, 'real_time', float)
```
jjotero, line 297:
```suggestion
    @performance_function('s')
    def real_time(self):
        return sn.extractsingle(r'real (?P<real_time>\S+)', self.stderr,
                                'real_time', float)
```

cscs-checks/system/slurm/slurm.py
```diff
@@ -198,25 +200,27 @@ def set_memory_limit(self):
...
+    sourcepath = 'eatmemory_mpi.c'
+    tags.add('mem')
+    executable_opts = ['100%']
+
+    @sanity_function
+    def assert_found_oom(self):
+        return sn.assert_found(r'(oom-kill)|(Killed)',
+                               self.stderr)
+
+    @run_before('performance')
```
jjotero, line 215: This now can be written as follows
```python
    def set_references(self):
         no_limit = (0, None, None, 'GB')
         self.reference = {
            '*': {
                'max_cn_memory': no_limit,
                'max_allocated_memory': (
                    sn.getattr(self, 'reference_meminfo'), -0.05, None, 'GB'
                 ),
             }
         }

    @performance_function('GB')
    def max_cn_memory(self):
        return sn.getattr(self, 'reference_meminfo')

    @performance_function('GB')
    def max_allocated_memory(self):
         regex = (r'^Eating \d+ MB\/mpi \*\d+mpi = -\d+ MB memory from \/proc\/'
                  r'meminfo: total: \d+ GB, free: \d+ GB, avail: \d+ GB, using:'
                  r' (\d+) GB')
         return sn.max(sn.extractall(regex, self.stdout, 1, int))
```

cscs-checks/system/slurm/slurm.py
jjotero: Hmm, I think the reference of this test is not being used. Unless, I'm missing something, I think that one would have to transform that `reference` in to the right reference for each value of the `command` parameter.
```python
self.reference = {
    '*': self.reference[self.command],
}
```

jjotero (changes requested):

cscs-checks/system/slurm/slurm.py
```diff
@@ -121,55 +121,57 @@ def set_memory_limit(self):
...
-        self.valid_systems = ['daint:gpu', 'dom:gpu']
-        self.executable = 'scontrol show job ${SLURM_JOB_ID}'
-        self.sanity_patterns = sn.assert_found(
+    valid_systems = ['daint:gpu', 'dom:gpu']
+    executable = 'scontrol show job ${SLURM_JOB_ID}'
+
+    @sanity_function
+    def assert_found_resources(self):
+        return sn.assert_found(
             r'.*(TresPerNode|Gres)=.*gpu:1.*', self.stdout)
```
vkarak, line 142:
```suggestion
        return sn.assert_found(
            r'.*(TresPerNode|Gres)=.*gpu:1.*', self.stdout
        )
```

cscs-checks/system/slurm/slurm.py
```diff
@@ -198,26 +200,20 @@ def set_memory_limit(self):
...
+    tags.add('mem')
+    executable_opts = ['100%']
+
+    @sanity_function
+    def assert_found_oom(self):
+        return sn.assert_found(r'(oom-kill)|(Killed)',
+                               self.stderr)
+
+    @run_before('performance')
+    def set_perf_patterns(self):
```
vkarak, line 216:
```suggestion
    def set_reference(self):
```

vkarak: The `reference_meminfo()` needs not to be any or a deferrable property. It is accessed in a `performance_function` or a pre-performance hook, so you can simply make it a function that returns the reference value that you want.

cscs-checks/system/slurm/slurm.py
```diff
@@ -227,9 +223,18 @@ def __init__(self):
                 ),
             }
         }
-        # }}}
 
-    # {{{ hooks
+    @performance_function('GB')
+    def max_cn_memory(self):
+        return sn.getattr(self, 'reference_meminfo')
```
vkarak, line 229: I don't know why we need this? Is it just for recording the reference meminfo?

cscs-checks/system/slurm/slurm.py
```diff
@@ -244,10 +249,9 @@ def set_tasks(self):
         self.num_tasks_per_node = tasks_per_node[partname]
         self.num_tasks = self.num_tasks_per_node
         self.job.launcher.options = ['-u']
-    # }}}
 
     @property
-    @sn.sanity_function
+    @deferrable
```
vkarak, line 254: Check my comment above.

vkarak (changes requested):

cscs-checks/system/slurm/slurm.py
rsarm: I think so, but I'm not sure. Maybe it comes from a time where it wasn't possible to put the reference on the perflog? I think we should remove it. What do you think @victorusu?

vkarak (approved):

Issue #1987, opened 2021-05-26
Specifying requirements, rather than e.g. valid_systems, for better test portability


casparvl:
**The issue**

The problem I want to solve is that I want to write a test for GPUs, that is portable to other systems. By portable I mean: the end-user should only have to modify the ReFrame configuration file, but should not have to touch my test. 

**Current options**

Suppose we have a test that only runs on GPUs. Currently, I could define in my reframe config a partition with something like

```
{
  'name': 'gpu',
   ...
  'devices': [
    {
      'type': 'gpu',
      'num_devices': 2,
    },
  ],
},
```
and in my tests I could do something like 

```
valid_systems = ['my_system:gpu']
```

The issue I have with this is that this puts system specific information (namely: which systems are valid) in my ReFrame test. I feel that's bad practice in general: the test definition should contain as little system specific information as possible. This makes them more portable ánd more maintainable.

*Option 1*

Of course, I could tell everyone one who wants to use this test "you _have_ to specify a system named 'my_system' with a partition 'gpu'", but that's clearly an ugly solution: the ReFrame configuration file should desribe the system, it should not be defined by tests you want to run... (and: what if you have two GPU partitions? Only one can be named 'gpu' and would run this test...)

*Option 2*

@victorusu pointed me to this option.

I could have the ReFrame test check if the current partition has a gpu device, and if not, skip it:
```
@rfm.run_after('setup')
def skip_if_no_gpu(self):
  device_count = [ dev.num_devices for dev in self.current_partition.devices if dev.device_type == 'gpu' ]
  self.skip_if(device_count == 0)
```
This would work, but (as @victorusu  himself said) is quite ugly. It still creates the test as part of the DAG and then at runtime decides to skip it. Furthermore, I have to write logic in the test itself for this selection, which is less readable then 'only' specifying the requirements.

**Requested feature**

It would be great if a test could specify something like `partition_requirements` in the `__init__`, which are then matched to the partition's properties described in the ReFrame config file. The test should only be added to the execution graph for those partitions that meet the requirements. Note that this is actually very similar to how `valid_systems` works, but rather than specifying explicitely which partition is valid (which is system-specific), you specify a generic hardware requirement (which is test-specific, and thus has every 'right' to be in the test definition file). In fact, I think specifying the `partition_requirements` could completely replace the `valid_systems`, since before we would use `valid_systems` mostly because we _knew_ that a certain partition met certain requirements - and therefore included it in `valid_systems`.

[...]

PR #2166, opened 2021-08-27
[feat] Support for test fixtures

### LINKED ISSUE #1577: Support for ReFrame test fixtures

### LINKED ISSUE #1793: Make subspace selection from the parameter space easy
We would like to make it easier to select a subspace from the parameter space, for which the test is valid, without having to write too many `ifs`.

jjotero:
This PR introduces support for ReFrame test fixtures. In this context, a fixture is just another test that prepares some resources for the parent test. Test fixtures may contain other fixtures, and so on, where the parent fixture (or test) will only execute after the required fixtures have finished. Fixtures are implemented using the dependencies framework, so the resources generated by a given fixture are not cleared (i.e. its `cleanup` pipeline stage is not executed) until the last test that requires from that fixture completes.

#### Resource sharing
Fixtures enable multiple tests to share resources. The level of resource sharing can be controlled by the fixture `scope`. The valid `scope` values are `session`, `partition`, `environment` and `test`, where `test` is the `scope` default value:

- `session`: The resources are shared by the whole ReFrame session. This must not depend on the partition nor the programming environment, so these fixtures must be `RunOnlyRegressionTest`s.
- `partition`: Fixtures with this scope will only share resources with tests within the same partition. This fixture will run as many times as the number of partitions required by the parent tests. This fixtures must not depend on the programming environment either, which means that they must also be `RunOnlyRegressionTest`s.
- `environment`: The resources are only shared across the tests in a given partition-environment combination. The fixture class can be of any kind from the derived classes from `RegressionTest`.
- `test`: The resources are private to the parent test.

```python
class MyFixture(rfm.RunOnlyRegressionTest):
   '''Manage a resource'''

@rfm.simple_test
class TestA(rfm.RegressionTest):
    # These resources are shared across the full session
    foo = fixture(MyFixture, scope='session')

    # This will trigger another private evaluation of `MyFixture`
    bar = fixture(MyFixture, scope='test')
```

#### Fork-Join model
A fixture may be parameterised using the `parameter` builtin. By default, a parameterised fixture will parameterise its parent tests. This behaviour is referred to as `fork` and can be controlled with the `action` argument. Alternatively, the parent test may want to gain access to all the variants from a given fixture, which in that case, the `action` argument must be set to `join`:
```python
class MyFixture(rfm.RegerssionTest):
   p = parameter(range(2))

@rfm.simple_test
class MyForkedTest(rfm.RegressionTest):
    '''This test will run once per variant of it's fixtures.'''
    f = fixture(MyFixture, action='fork')

    @run_after('setup')
    def access_fixtures(self):
        '''Print the parameter value from the fixture f.'''
        print(self.f.p) 

@rfm.simple_test
class MyReductionTest(rfm.RegressionTest):
    '''This test gathers all the variants from ``MyFixture``'''
    f = fixture(MyFixture, action='join')

    @run_after('setup')
    def access_fixtures(self):
        '''Print the parameter value from each fixture variant.

        Here f is a list containing all the variants from MyFixture.
        '''
        for fixture in self.f:
            print(fixture.p) 
```

#### Fixture parameter space sub-selection
When a given fixture has a complex  or large parameter space, the parent test may filter out undesired points in the parameter space. This can be controlled through the `variants` argument:
```python
class MyComplexFixture(rfm.RegressionTest):
    # A fixture with 400 different variants.
    p0 = parameter(range(100))
    p1 = parameter(['a', 'b', 'c', 'd'])

@rfm.simple_test
    # Use only 10 variants from MyComplexFixture
    foo = fixture(MyComplexFixture, variants={'p0': lambda x: x<10, 'p1': lambda x: x=='d'})

    # Select the fixture variants by index
    bar = fixture(MyComplexFixture, variants=range(300,310))

```

#### Setting fixture variables
A parent test may set the default value of variables from the fixture class through the `variables` argument. Fixtures that set default variables will be interpreted as different from other fixtures that use the same class with its default values.
```python
class MyFixtyre(rfm.RunOnlyRegressionTest):
    '''Manage a resource'''
    my_variable = variable(int, value=10)
    ...

@rfm.simple_test
    # Set my_var in the fixture foo
    foo = fixture(MyFixture, scope='session', variables={'my_variable': 5})

    # Here bar is a different fixture and will also run despite that foo has a session scope.
    bar = fixture(MyComplexFixture, scope='session')

```

#### Other code improvements
Each variant of a class derived from the `RegressionTest` class has a unique index that points to a location in both the parameter and fixture spaces. The variant of a class is selected during class instantiation, where the variant number is passed as the `variant_num` argument to the class constructor.
The total number of variants of a class can be queried through the `cls.num_variants` property, and the full name of a given class variant can be computed with the class method `cls.fullname(variant_num=<the_given_variant_num>)`. The current variant number of an instance is available through the property `self.variant_num`, along the variant sub-components of the parameter and fixture spaces (`self.param_variant` and `self.fixture_variant`, respectively).

The above makes serialisation of large parameter spaces trivial:
```python
@rfm.simple_test
    p0 = parameter(range(1000))
    p1 = parameter(['a', 'b', 'c', 'd'])

    @run_after('init')
    def serialise_param_space_execution(self):
        if self.variant_num > 0:
            self.depend_on(type(self).fullname(self.variant_num-1)
```
Also, sub-selecting custom portions of the full variant space based on the test parameters is straightforward with the class method `cls.get_variant_nums`:
```python
# Get a list of variant numbers for the variants with p0 < 500 and p1 == 'a'
variants = MyTest.get_variant_nums(p0=lambda x: x < 500, p1=lambda x: x == 'a')
```

#### Notes
Docs and unit tests to follow after we settle on the implementation. The current state of the different docstrings is just a draft. Those will be revisited alongside the documentation.

Closes #1577 
Closes #1793


vkarak:
I like this PR a lot. It gives a whole new dimension on how to run and combine tests! The addition of a `join` action for the fixtures is also very powerful and goes a step forward compared to pytest fixtures for example. I have just a couple of comments:

We should mention at the documentation how the standard concepts of "setup" and "tear-down" can be implemented with ReFrame fixtures. The "setup" I see it as a series of `post-init` hooks, whereas the "tear-down" is a set of cleanup hooks. There is also an implicit "tear-down", since ReFrame cleans up automatically the resources of the fixtures if their dependents have finished successfully.

My second comment is about the indexing and how it is exposed to the users. I like that the parametrised tests are now indexed consistently, but as a user it has little value to address "test 123" if I don't know which exact test in the parametrisation space this corresponds to. The same goes for the example of serialisation, the `index - 1` test could be anything. If we want to allow users to use raw indexes, I think that we should make the test indexing transparent, i.e., we should document the way the indexes are generated.

I will later have a look into the implementation.

jjotero:
Regarding the indexing, users can query the actual parameter values doing `cls.param_space[<variant_num>]`, and this will return a dictionary with the value for each parameter for this given variant number. The only issue is that, with fixtures, the "global" variant number can be larger than the size of the parameter space. That mapping from "global" (or collapsed) index into the parameter+fixture components is done by the `cls._map_variant_num` function. I think it may make sense to make this one public.

Also, another thing to take into account is that the conditions passed to filter the parameter space are always inclusive, so users would need to call the `cls.get_variant_nums` function multiple times to build more complex sub-sets
```python
param_set = set()
param_set.update(MyTest.get_variant_nums(<a_set_of_conditions>))
param_set.update(MyTest.get_variant_nums(<another_set_of_conditions>))
```

vkarak:
Let me think a bit about it also from the test naming perspective, what would be a reasonable user API. What you have now, it totally makes sense, but it feels a bit too low level.

vkarak:
I'm playing with the example fixtures on a system with multiple partitions and multiple environments and I'm getting `could not resolve dependency` errors when trying to list a the test example with the fixtures. Here is my file:

```python
import reframe as rfm


class MyFixture(rfm.RunOnlyRegressionTest):
    '''Manage a resource'''


@rfm.simple_test
class TestA(rfm.RegressionTest):
    # These resources are shared across the full session
#    foo = fixture(MyFixture, scope='session')

    # This will trigger another private evaluation of `MyFixture`
    bar = fixture(MyFixture, scope='environment')

    valid_systems = ['*']
    valid_prog_environs = ['*']
```

And here is the command I'm trying with:

```
./bin/reframe -C config/cscs.py -c ../misc/reframe-tests/fixtures/example.py -l
```

And here's the errors and the output:

```
[ReFrame Setup]
  version:           3.9.0-dev.0+99a0b5f6
  command:           './bin/reframe -C config/cscs.py -c ../misc/reframe-tests/fixtures/example.py -l'
  settings file:     'config/cscs.py'
  check search path: (R) '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py'

./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:login', 'builtin') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:login', 'PrgEnv-cray') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:login', 'PrgEnv-gnu') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:login', 'PrgEnv-intel') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:login', 'PrgEnv-pgi') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:login', 'PrgEnv-nvidia') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:gpu', 'builtin') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-cray') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-gnu') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-intel') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-pgi') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-nvidia') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:mc', 'builtin') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:mc', 'PrgEnv-cray') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:mc', 'PrgEnv-gnu') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:mc', 'PrgEnv-intel') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:mc', 'PrgEnv-pgi') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:mc', 'PrgEnv-nvidia') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:jupyter_gpu', 'builtin') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: could not resolve dependency: ('TestA_MyFixture', 'dom:jupyter_mc', 'builtin') -> 'MyFixture_dom:jupyter_gpu_PrgEnv-intel'
./bin/reframe: skipping all dependent test cases
  - ('TestA_MyFixture', 'dom:jupyter_mc', 'builtin')
  - ('TestA_MyFixture', 'dom:gpu', 'builtin')
  - ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-cray')
  - ('TestA_MyFixture', 'dom:login', 'PrgEnv-nvidia')
  - ('TestA_MyFixture', 'dom:mc', 'builtin')
  - ('TestA_MyFixture', 'dom:login', 'PrgEnv-intel')
  - ('TestA_MyFixture', 'dom:login', 'builtin')
  - ('TestA_MyFixture', 'dom:mc', 'PrgEnv-cray')
  - ('TestA_MyFixture', 'dom:login', 'PrgEnv-pgi')
  - ('TestA_MyFixture', 'dom:login', 'PrgEnv-gnu')
  - ('TestA_MyFixture', 'dom:mc', 'PrgEnv-nvidia')
  - ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-nvidia')
  - ('TestA_MyFixture', 'dom:mc', 'PrgEnv-intel')
  - ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-intel')
  - ('TestA_MyFixture', 'dom:jupyter_gpu', 'builtin')
  - ('TestA_MyFixture', 'dom:login', 'PrgEnv-cray')
  - ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-pgi')
  - ('TestA_MyFixture', 'dom:mc', 'PrgEnv-pgi')
  - ('TestA_MyFixture', 'dom:mc', 'PrgEnv-gnu')
  - ('TestA_MyFixture', 'dom:gpu', 'PrgEnv-gnu')

[List of matched checks]
- MyFixture_dom:gpu_PrgEnv-nvidia (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:mc_PrgEnv-intel (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:mc_PrgEnv-cray (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:login_PrgEnv-intel (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:mc_PrgEnv-gnu (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:login_PrgEnv-gnu (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:login_PrgEnv-cray (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:gpu_PrgEnv-intel (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:gpu_builtin (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:mc_builtin (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:gpu_PrgEnv-gnu (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:jupyter_mc_builtin (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:gpu_PrgEnv-cray (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:gpu_PrgEnv-pgi (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:mc_PrgEnv-pgi (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:jupyter_gpu_builtin (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:login_builtin (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:mc_PrgEnv-nvidia (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:login_PrgEnv-nvidia (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
- MyFixture_dom:login_PrgEnv-pgi (found in '/users/karakasv/Devel/misc/reframe-tests/fixtures/example.py')
Found 20 check(s)

Log file(s) saved in '/users/karakasv/Devel/reframe/reframe.log', '/users/karakasv/Devel/reframe/reframe.out'
```

On another note regarding this PR, I think that we should not list fixtures as separate tests when passing the `-l` option. We could still show them, though, in the detailed listing with the `-L` option.

vkarak:
This happens only when scope is either `partition` or `environment`.

vkarak:
@jjotero I'm still getting the same errors, aren't you?

vkarak:
Also, I just noticed that sometimes I get the dependency error, some others not.

jjotero:
Yes, it turns out the bug I fixed yesterday was something else. I'll push the fix for your reproducer shortly.

vkarak:
> bomb-proof

Caution, I have a big 💣  

This works, but it fails with if you try to scope it with the `-p` option. It seems that fixtures are built for environments that are not filtered:

```
--system=daint:login -p builtin
```

```
./bin/reframe: could not resolve dependency: ('TestA_MyFixture_MyFixture_MyFixture_MyFixture', 'daint:login', 'builtin') -> 'MyFixture_daint:login_gnu'
./bin/reframe: skipping all dependent test cases
  - ('TestA_MyFixture_MyFixture_MyFixture_MyFixture', 'daint:login', 'builtin')
```

jjotero:
_In my defence, this is not a bug on the fixtures side per-se_ (🥲) The above error is a side effect of the sys/env filtering happening too late. I can see a workaround, a hack, and a proper fix to address this:
- The workaround: Simply document that filtering with `--system` and `-p` from the command line is disallowed when using fixtures (not ideal).
- The hack: Move the filtering into the `RegressionTestRegistry`. Then, we can simply set `valid_partitions` and `valid_prog_environs` to the filtered sys/env pair if the test supports it.
- The LTS fix: run the whole `valid_partitions`-`valid_prog_environs` matrix through the parameter space. As discussed, this might be the right path to also address #1987. This option would also simplify the `FixtureRegistry` implementation.

jjotero:
"_The bomb has been defused ..._"

The main issue with the above 💣is that `--system` and `-p` are treated differently by the framework. The runtime would only load the partition into `runtime.runtime().system.partitions` as specified by the `--system` option. However, regardless what we pass in `-p`, this runtime structure will contain all the environments per partition as specified in the config file. This means that there is no way for the fixture registry to check whether a given environment is a valid one or not, because that filtering is done much later by the framework.

Since those environments that do not match the patterns provided by `-p` are never used in the session, I have moved that "filtering" way ahead to when this `runtime` structure is constructed. Then each partition in the runtime structure will only have the environments that match the patters provided by `-p`. This makes this information available for the `FixtureRegistry`, which now always selects a valid partition and a valid environment.

jjotero:
I have added the new class method `cls.get_variant_info` that allows to access the "raw" variant information. See the gist below to play around with it. Also, the naming scheme of the fixtures has been improved a bit.

https://gist.github.com/jjotero/cf6a4051ce77d9d8cfb912aebee28e4d

vkarak:
I don't understand why `fullname()` is part of the metaclass which makes it eventually a test class method. The final name of the fixtures uses information from the object at the injection time (the valid prog. envs and the partitions). This practically restricts the possibilities we have with the `fullname` and also leads to inconsistent names: at the test level the name uses currently the variant id, whereas the fixture name uses the prog. env and partition from the scope (wherever relevant). Not having access to the fixture scope information (i.e., the exact environment and partition that is instantiated for) makes it difficult to support a richer naming syntax as the one I proposed in #1794. Whereas I could reproduce the test name proposed there, I got stuck when I tried to reproduce the fixture name that uses scope information. I believe that the `fullname()` must be an instance method that has access to all the information of either the test or the fixture.

Another comment is why do you index by name the fixtures in the `FixtureRegistry` and not by the variant number? That practically duplicates the naming logic.

vkarak:
I have already a version with the naming scheme in #1794, but I came across the following, which is present also in your branch. Suppose you have this test:

```python
class MyFixture(rfm.RunOnlyRegressionTest):
    '''Manage a resource'''
    p = parameter([1, 2])


@rfm.simple_test
    f = fixture(MyFixture, scope='test', action='join')
    t = fixture(MyFixture, scope='test')

    valid_systems = ['*']
    valid_prog_environs = ['*']
```

The produced dependencies for one variant of `TestA` (e.g.`TestA^MyFixture@0+MyFixture^`) are the following:

```
      - ('TestA^MyFixture@0+MyFixture^', 'generic:default', 'builtin') -> ('MyFixture_1~TestA^MyFixture@0+MyFixture^', 'generic:default', 'builtin')
      - ('TestA^MyFixture@0+MyFixture^', 'generic:default', 'builtin') -> ('MyFixture_1~TestA^MyFixture@0+MyFixture^', 'generic:default', 'builtin')
      - ('TestA^MyFixture@0+MyFixture^', 'generic:default', 'builtin') -> ('MyFixture_2~TestA^MyFixture@0+MyFixture^', 'generic:default', 'builtin')
```

Which means that one of the fixtures of `f` and `t` is shared. Is this intended? I saw the same with my naming scheme, but got back to your original branch to check if I have messed up with something. This is how it looks with my naming scheme:

```
    Dependencies (conceptual):
      TestA%t.p=2
        ^MyFixture%p=2~TestA%t.p=2
        ^MyFixture%p=1~TestA%t.p=2
        ^MyFixture%p=2~TestA%t.p=2

    Dependencies (actual):
      - ('TestA%t.p=2', 'generic:default', 'builtin') -> ('MyFixture%p=2~TestA%t.p=2', 'generic:default', 'builtin')
      - ('TestA%t.p=2', 'generic:default', 'builtin') -> ('MyFixture%p=1~TestA%t.p=2', 'generic:default', 'builtin')
      - ('TestA%t.p=2', 'generic:default', 'builtin') -> ('MyFixture%p=2~TestA%t.p=2', 'generic:default', 'builtin')
```

The first `MyFixture%p=2~TestA%t.p=2` is coming from `t` and the second one from `f`. Checking in the registry's `add()` method it seems that indeed both of these fixtures are registered with the same name.

jjotero:
Yes, that's correct. Both fixtures are the same because the final fixture name does not take into account the `action` argument. However, the scope set here is `'tests'` and this is the same test that is simply using the fixture twice. So from the scope point of view, this behaviour is correct.

jjotero:
> Another comment is why do you index by name the fixtures in the FixtureRegistry and not by the variant number? That practically duplicates the naming logic.

One is allowed to register the same fixture multiple times with different scopes, which in this case they make their way in to the name. Also, a fixture can be registered multiple times setting variables from the fixture directive, and this will also modify their name. All these examples would have the same variant, but different name. Perhaps referring to this as the `'name'` is not very accurate of what it represents. This is just a unique string that I can use to build and fetch the dependencies later on.

vkarak:
> Yes, that's correct. Both fixtures are the same because the final fixture name does not take into account the action argument. However, the scope set here is 'tests' and this is the same test that is simply using the fixture twice. So from the scope point of view, this behaviour is correct.

I'm fine with this behaviour, but I think that this is worth pointing it out in the documentation, because I was expecting that `f` and `t` are different instances of `MyFixture` that would be executed separately. This means that `MyFixture%p=2~TestA%t.p=2` will be executed only once and reused two times.

PR #2168, opened 2021-08-30
[bugfix] Fix incomplete processor info object when processor topology is auto-detected

ekouts:
When we detect automatically the topology we used to set directly only the `_info` attribute of the processor of the partition. As a result the other attributes of the `ProcessorType` were not set and we would get for example `self.current_partition.processor.num_cpus` as `None` while `self.current_partition.processor.info` is a non empty dictionary. The other attributes were only set in the `_init_`: https://github.com/eth-cscs/reframe/blob/34ee3d0bc91f7c9f5e56db1c6b46a49e8e5ee623/reframe/core/systems.py#L38

Same for the devices, we should set `self.current_partition._devices` to a list of `DeviceType` objects and not dictionaries.

### DIFF
--- a/reframe/core/systems.py
+++ b/reframe/core/systems.py
@@ -15,9 +15,9 @@ from reframe.core.modules import ModulesSystem
 
 class ProcessorType(jsonext.JSONSerializable):
     '''A representation of a processor inside ReFrame.
-    You can access all the keys of the processor dictionary in configuration
-    as direct properties. You can see a full list
-    `here <config_reference.html#processor-info>`__.
+
+    You can access all the keys of the `processor configuration object
+    <config_reference.html#processor-info>`__.
 
     .. versionadded:: 3.5.0
 
@@ -111,9 +111,9 @@ class ProcessorType(jsonext.JSONSerializable):
 
 class DeviceType(jsonext.JSONSerializable):
     '''A representation of a device inside ReFrame.
-    You can access all the keys of the device dictionary in configuration
-    as direct properties. You can see a full list
-    `here <config_reference.html#device-info>`__.
+
+    You can access all the keys of the `device configuration object
+    <config_reference.html#device-info>`__.
 
     .. versionadded:: 3.5.0
 
--- a/reframe/core/systems.py
+++ b/reframe/core/systems.py
@@ -23,20 +23,33 @@ class ProcessorType(jsonext.JSONSerializable):
 
     '''
 
+    __slots__ = (
+        '__info', 'arch', 'num_cpus', 'num_cpus_per_core',
+        'num_cpus_per_socket', 'num_sockets', 'topology'
+    )
+
+    def __deepcopy__(self, memo):
+        # This is a read-only object; simply return ourself
+        return self
+
     def __init__(self, processor_info):
-        self._arch = None
-        self._num_cpus = None
-        self._num_cpus_per_core = None
-        self._num_cpus_per_socket = None
-        self._num_sockets = None
-        self._topology = None
-        self._info = processor_info
+        self.__info = processor_info
 
-        if not processor_info:
-            return
+    def __getattr__(self, name):
+        if name in self.__slots__:
+            return self.__info.get(name, None)
+        else:
+            raise AttributeError(
+                f'{type(self).__qualname__!r} object has no attribute {name!r}'
+            )
 
-        for key, val in processor_info.items():
-            setattr(self, f'_{key}', val)
+    def __setattr__(self, name, value):
+        if name in self.__slots__:
+            raise AttributeError(
+                f'attribute {name!r} is not writeable'
+            )
+        else:
+            super().__setattr__(name, value)
 
     @property
     def info(self):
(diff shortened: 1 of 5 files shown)


unittests/test_autodetect.py
```diff
@@ -54,7 +54,15 @@ def test_autotect(exec_ctx):
     detect_topology()
     part = runtime().system.partitions[0]
     assert part.processor.info == cpuinfo()
-    assert part.devices == [{'type': 'gpu', 'arch': 'a100', 'num_devices': 8}]
+    if part.processor.info:
+        assert part.processor.num_cpus == part.processor.info['num_cpus']
```
vkarak, line 58: Leave a blank line after this one.

vkarak (changes requested):
This is a serious one! Thanks for catching it @ekouts! I would suggest and alternative design altogether, though, for both the `ProcessorType` and the `DeviceType`. I would just make them simple wrappers of their underlying dictionary. All the properties should simply do `return self._processor_info.get('propname', None)` and both `ProcessorType` and `DeviceType` should provide an `update(info)` method that would update the underlying dictionary.

ekouts:
So just to make sure I get this right:
> All the properties should simply do return self._processor_info.get('propname', None)

I will change the `__getattr__` in order to try to get all the properties from the dictionary. Should we allow certain attribute names or leave the user ask for anything and just return `None`? It might be a small problem for typos if the user doesn't know he is just putting the wrong name.

> both ProcessorType and DeviceType should provide an update(info) method that would update the underlying dictionary.

Should this `update` work like a dictionary update, so it will be "additive" or overwrite the old dictionary? I guess we want the second one, right?

vkarak:
@ekouts You raise good points:

> I will change the __getattr__ in order to try to get all the properties from the dictionary.

That's a good idea.

> Should we allow certain attribute names or leave the user ask for anything and just return None? It might be a small problem for typos if the user doesn't know he is just putting the wrong name.

I would vote for returning `None` only for the known properties (those defined in the docs/json schema). For the rest, you should delegate to `super().__getattr__()` so as to get an `AttributeError`. To achieve this, the class must keep a list of the known properties.

> Should __getattr__ handle the attributes like num_cores_per_socket that are inferred (this one is self.num_cores // self._num_sockets) or they will stay as they are now?

I would keep them as properties that do `self.known_property`.

> We also would not be getting the properties in the documentation, although the names are mostly self explanatory I guess.

We should point to the configuration description for the standard properties and document here only the derived ones.

> Should this update work like a dictionary update, so it will be "additive" or overwrite the old dictionary? I guess we want the second one, right?

What if works in an additive way but the target dictionary (the one that comes from the conf file) takes precedence? If the inserted/autodetected value is already there, then we simply use the existing one. Something like a `setdefault()` for the whole dictionary.

ekouts:
> What if works in an additive way but the target dictionary (the one that comes from the conf file) takes precedence? If the inserted/autodetected value is already there, then we simply use the existing one. Something like a setdefault() for the whole dictionary.

Now if the processor dictionary in the configuration is not an empty dictionary in the configuration we don't try the auto-detection at all. For me it's better (and obviously much easier) to keep this behavior since you might prefer to have little "reliable" information than ReFrame's guess. But if you think it's worth it I can try that.

vkarak:
Ok, let's keep it as is for the moment and bring it up for discussion.

reframe/core/systems.py
```diff
@@ -15,6 +15,9 @@
 
 class ProcessorType(jsonext.JSONSerializable):
     '''A representation of a processor inside ReFrame.
```
vkarak, line 17: Leave a blank line after the title line.

reframe/core/systems.py
```diff
@@ -15,6 +15,9 @@
...
+    You can access all the keys of the processor dictionary in configuration
+    as direct properties. You can see a full list
+    `here <config_reference.html#processor-info>`__.
```
vkarak, line 20:
```suggestion
    You can access all the keys of the `processor configuration object <config_reference.html#processor-info>`__.
```

reframe/core/systems.py
```diff
@@ -142,6 +111,9 @@ def num_cores_per_numa_node(self):
 
 class DeviceType(jsonext.JSONSerializable):
     '''A representation of a device inside ReFrame.
+    You can access all the keys of the device dictionary in configuration
+    as direct properties. You can see a full list
+    `here <config_reference.html#device-info>`__.
```
vkarak, line 116: See my comments above.

reframe/core/systems.py
```diff
@@ -150,17 +122,31 @@ class DeviceType(jsonext.JSONSerializable):
...
-        for key, val in device_info.items():
-            setattr(self, f'_{key}', val)
+    def __setattr__(self, name, value):
+        if name in self.__slots__:
+            raise AttributeError(
+                f'attribute {name!r} is not writeable'
+            )
+        else:
+            # Let Python data model raise the AttributeError here
+            super().__setattr__(name, value)
```
vkarak, line 149: I think that we can factor out all these to a `_ReadOnlyInfo` object that both `ProcessorType` and `DeviceType` will inherit from. I will also rename these final objects to `ProcessorInfo` and `DeviceInfo`. "Type" can be deceiving, since it implies a `type` object.

unittests/test_autodetect.py
```diff
@@ -56,6 +56,7 @@ def test_autotect(exec_ctx):
     assert part.processor.info == cpuinfo()
     if part.processor.info:
         assert part.processor.num_cpus == part.processor.info['num_cpus']
+
     assert len(part.devices) == 1
     assert part.devices[0].info == {
         'type': 'gpu',
```
vkarak, line 62: I think that we should add a couple of more tests to verify the immutability of both the `ProcessorType` and `DeviceType`.

vkarak (changes requested):

ekouts:
@jenkins-cscs retry all

vkarak (approved):

vkarak:
I changed a bit the implementation, because the `__slots__`, as we had them, didn't have an effect, since the `ProcessorInfo` was inheriting a `__dict__` from the `JSONSerializable`. So I introduced the `__slots__` to that one and then made the encoding/decoding to understand also objects that define `__slots__`. The problem I had afterwards, though, was that the `ProcessorInfo` could not be restored, since the logged attributes were not writeable when the object was being restored (by `--restore-session`). So eventually, I just used the `__slots__` to store only the internal `_info` object, so that this will be the only attribute to be dumped and read from the JSON representation. I used another class attribute, `_known_attrs` to treat specially the known attributes that we used to have in `__slots__`.

PR #2177, opened 2021-09-09
[testlib] Add generic CPMD check for the cscs-supported apps into hpctestlib/apps

hurricane642:
This is a continuation of #2084, devoted exclusively to the CPMD test. A new file nve.py has been added, introducing the main class for the CPMD test, as well as the CSCS tests inherited from it.


vkarak:
This still needs substantial work, so I'm closing the PR due to time constraints. It will be added to the library at a later time.

PR #2181, opened 2021-09-09
[test] Add generic SPARK check for the cscs-supported apps into hpctestlib/apps

hurricane642:
This is a continuation of the work on the test library. This PR work has been done with SPARK tests. A new file base_check.py has been added, introducing the main class for the SPARK test, as well as the CSCS tests inherited from it.


teojgo:
Ok to test

hpctestlib/apps/spark/base_check.py
```diff
@@ -0,0 +1,55 @@
...
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import math
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+
+
+class Spark_BaseCheck(rfm.RunOnlyRegressionTest, pin_prefix=True):
```
jjotero, line 12:
```suggestion
class ComputePi(rfm.RunOnlyRegressionTest, pin_prefix=True):
```

cscs-checks/apps/spark/spark_check.py
```diff
@@ -3,32 +3,20 @@
...
 # SPDX-License-Identifier: BSD-3-Clause
 
-import math
-
 import reframe as rfm
-import reframe.utility.sanity as sn
-from reframe.core.backends import getlauncher
 
+from reframe.core.backends import getlauncher
+from hpctestlib.apps.spark.base_check import Spark_BaseCheck
```
jjotero, line 9:
```suggestion
from hpctestlib.apps.spark.compute_pi import ComputePi
```

cscs-checks/apps/spark/spark_check.py
```diff
@@ -3,32 +3,20 @@
...
-        self.prerun_cmds = ['start-all.sh']
-        self.postrun_cmds = ['stop-all.sh']
-        self.num_tasks = 3
-        self.num_tasks_per_node = 1
-        pi_value = sn.extractsingle(r'Pi is roughly\s+(?P<pi>\S+)',
-                                    self.stdout, 'pi', float)
-        self.sanity_patterns = sn.assert_lt(sn.abs(pi_value - math.pi), 0.01)
-        self.maintainers = ['TM', 'RS']
-        self.tags = {'production'}
+class SparkCheck(Spark_BaseCheck):
```
jjotero, line 12:
```suggestion
class SparkCheck(ComputePi):
```

jjotero (commented):
I'm not convinced that having a name such as `base_check` or so is the right approach. I think the name should be more explicit of what the test does, which in this case is computing pi. For this reason, I'd suggest that we move this "base check" to `hpctestlib/apps/spark/compute_pi/__init__.py` change the class names as suggested below. This would allow other people to introduce other spark checks for whatever other case they want.

teojgo:
@jenkins-cscs retry dom

PR #2183, opened 2021-09-09
[testlib] Add PyTorch library test and base CSCS test on top of it

hurricane642:
This is a continuation of the work on the test library. This PR work has been done with PyTorch tests. A new file base_check.py has been added, introducing the main class for the PyTorch test, as well as the CSCS tests inherited from it.


hpctestlib/ml/pytorch/horovod.py
```diff
@@ -0,0 +1,112 @@
...
+    framework for TensorFlow, Keras, PyTorch, and Apache
+    MXNet. The goal of Horovod is to make distributed deep
+    learning fast and easy to use (see github.com/horovod/horovod).
+
+    This test tests the performance of Pytorch and Horovod using
+    two classic deep learning models. It checks whether learning is
+    performed to the end. The default assumption
+    is that Pytorch and Horovod is already installed on the device
+    under test.
+    '''
```
vkarak, line 28: Can you update the docs similarly to the Tensorflow test?

cscs-checks/apps/pytorch/pytorch_horovod_check.py
```diff
@@ -3,82 +3,54 @@
...
-                          for mpi_task in [32, 8, 1]
-                          for model in ['inception_v3', 'resnet50']])
-class PytorchHorovodTest(rfm.RunOnlyRegressionTest):
-    def __init__(self, model, mpi_task):
-        self.descr = 'Distributed training with Pytorch and Horovod'
-        self.valid_systems = ['daint:gpu']
-        if mpi_task < 20:
+
+@rfm.simple_test
+class cscs_tensorflow_horovod_check(pytorch_cnn_check):
```
vkarak, line 13:
```suggestion
class cscs_pytorch_horovod_check(pytorch_cnn_check):
```

cscs-checks/apps/pytorch/pytorch_horovod_check.py
```diff
@@ -3,82 +3,54 @@
...
+    num_nodes = parameter([8, 32, 1])
```
vkarak, line 14:
```suggestion
    num_nodes = parameter([1, 8, 32])
```

cscs-checks/apps/pytorch/pytorch_horovod_check.py
```diff
@@ -3,82 +3,54 @@
...
 
-        self.valid_prog_environs = ['builtin']
-        self.modules = ['PyTorch']
-        self.num_tasks_per_node = 1
-        self.num_cpus_per_task = 12
-        self.num_tasks = mpi_task
-        batch_size = 64
+    @run_before('run')
+    def setup_run(self):
+        proc = self.current_partition.processor
```
vkarak, line 32:
```suggestion
        self.skip_if_no_procinfo()
        proc = self.current_partition.processor
```

vkarak: And, if not present, add the same to the Tf2 test.

cscs-checks/apps/pytorch/pytorch_horovod_check.py
```diff
@@ -3,82 +3,54 @@
...
-                'throughput_per_job': (ref_per_job, -0.1, None, 'images/s'),
+    # @run_before('performance')
+    # def set_performance(self):
+        ref_per_gpu_sm60 = 131 if self.model == 'inception_v3' else 201
+        ref_total_sm60 = ref_per_gpu_sm60 * self.num_nodes
+        allref = {
+            'sm_60': {
+                'throughput_total': (ref_total_sm60, -0.05, None, 'images/s'),
+                'throughput_iteration': (ref_per_gpu_sm60, -0.05, None,
+                                         'images/s')
```
vkarak, line 50: This is not so readable. I would add the `allref` dictionary at the class level, along with separate references for each benchmark. As of the `throughput_total` I would compute it on-the-fly here and add it to the final `reference` dictionary.

vkarak (changes requested):

cscs-checks/apps/pytorch/pytorch_horovod_check.py
rsarm: Here the idea is to create the `allref` dictionary at the class level with only the `'throughput_iteration'` key? and then to define the final reference on the `setup_run` function with something like following?
```
self.reference = {
                '*': {
                    **self.allref['sm_60'][self.model],
                    'throughput_total': (ref_total_sm60, -0.05, None, 'images/s'),
                }
            }
```
where `ref_total_sm60` is computed from the value of `'throughput_iteration'` set on `allref` times `self.num_nodes`?

cscs-checks/apps/pytorch/pytorch_horovod_check.py
```diff
@@ -10,16 +10,26 @@
...
+        'sm_60': {
+            'inception_v3': {
+                'throughput_per_gpu': (131, -0.05, None, 'images/s'),
+            },
+            'resnet50': {
+                'throughput_per_gpu': (201, -0.05, None, 'images/s'),
+            }
+        }
+    }
+    model_name = parameter(allref['sm_60'].keys())
```
vkarak, line 32: Nice trick, but we loose a bit in readability, I think...

cscs-checks/apps/pytorch/pytorch_horovod_check.py
```diff
@@ -38,19 +49,12 @@ def setup_run(self):
...
-        ref_total_sm60 = ref_per_gpu_sm60 * self.num_nodes
-        allref = {
-            'sm_60': {
-                'throughput_total': (ref_total_sm60, -0.05, None, 'images/s'),
-                'throughput_iteration': (ref_per_gpu_sm60, -0.05, None,
-                                         'images/s')
-            }
-        }
+        ref_per_gpu = self.allref['sm_60'][self.model]['throughput_per_gpu'][0]
+        ref_total = ref_per_gpu * self.num_nodes
```
vkarak, line 53: This should be within the `suppress` block below.

vkarak (changes requested):
Just a couple of minor comments still.

vkarak (approved):

vkarak:
ok to test

PR #2186, opened 2021-09-09
[testlib] Add GROMACS benchmark tests

hurricane642:
~~This is a continuation of the work on the test library. This PR work has been done with GROMACS tests. A new file base_check.py has been added, introducing the main class for the GROMACS test, as well as the CSCS tests inherited from it.~~


cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -98,7 +185,7 @@ def setup_filtering_criteria(self):
             self.valid_systems = []
 
         # Setup prog env. filtering
-        if self.current_system.name in ('eiger', 'pilatus'):
+        if self.current_system.name in ['eiger', 'pilatus']:
```
vkarak, line 188: The tuple should have been fine here.

hpctestlib/sciapps/gromacs/checks.py
```diff
@@ -146,5 +146,6 @@ def assert_energy_readout(self):
         energy_diff = sn.abs(energy - self.energy_ref)
         return sn.all([
             sn.assert_found('Finished mdrun', 'md.log'),
-            sn.assert_lt(energy_diff, self.energy_tol)
+            sn.assert_reference(energy, self.__energy_ref,
+                                -self.__energy_tol, self.__energy_tol)
```
vkarak, line 150: I don't understand the `__energy_ref` here. This should have failed actually...

cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -119,6 +206,11 @@ def setup_run(self):
         if self.current_partition.fullname in ('daint:gpu', 'dom:gpu'):
             arch = 'sm_60'
 
+        try:
+            found = self.allref[self.num_nodes][arch][self.bench_name]
+        except KeyError:
+            self.skip("Test configuration is not supported")
```
vkarak, line 212:
```suggestion
            self.skip('Test configuration is not supported')
```

Also a bit more info on the unsupported configuration would help.

vkarak (changes requested):

vkarak (approved):

vkarak:
ok to test

cscs-checks/apps/gromacs/gromacs_check.py
```diff
@@ -3,139 +3,223 @@
...
-            'prod': {
-                'small': {
-                    'dom:gpu': {'perf': (37.0, -0.05, None, 'ns/day')},
-                    'daint:gpu': {'perf': (35.0, -0.10, None, 'ns/day')}
-                },
-                'large': {
-                    'daint:gpu': {'perf': (63.0, -0.20, None, 'ns/day')}
-                }
+    }
+    executable_opts += ['-dlb yes', '-ntomp 1', '-npme -1']
```
victorusu, line 20: This line is not being executed...
The generated job script contains only the base class options:
```
srun gmx_mpi mdrun -nb gpu -s benchmark.tpr
```
It should be 
```
srun gmx_mpi mdrun -dlb yes -ntomp 1 -npme 0 -s herflat.tpr
```

vkarak: Oh, that's because #2268 !

victorusu: I will make it a post init hook with a FIXME note... When we fix #2268, then we revert to this version.
What do you think?

vkarak: It wouldn't work either because of #2268. As I have it now it won't work in any case, because the executable opts are set in a hook. I shouldn't have updated there. But there is a workaround if in the library we use `+=`. Can you try it?

vkarak: And generally, I think that libraries should always use `+=`...

hpctestlib/sciapps/gromacs/checks.py
```diff
@@ -25,6 +25,12 @@ class gromacs_check(rfm.RunOnlyRegressionTest):
 
     '''
 
+    #: The version of Horovod to use.
+    #:
+    #: :type: :class:`str`
+    #: :default: ``'v0.21.0'``
```
vkarak, line 31:
```suggestion
    #: :default: ``'1.0.0'``
```

hpctestlib/sciapps/gromacs/checks.py
```diff
@@ -56,10 +63,9 @@ def prepare_test(self):
         self.__bench, self.__nrg_ref, self.__nrg_tol = self.benchmark_info
         self.descr = f'GROMACS {self.__bench} benchmark (NB: {self.nb_impl})'
         self.prerun_cmds = [
-            f'curl -LJO https://github.com/victorusu/GROMACS_Benchmark_Suite/raw/main/{self.__bench}/benchmark.tpr'  # noqa: E501
+            f'curl -LJO https://github.com/victorusu/GROMACS_Benchmark_Suite/raw/{self.benchmark_version}/{self.__bench}/benchmark.tpr' # noqa: E501
         ]
-        self.executable_opts = ['mdrun', '-nb', self.nb_impl,
-                                '-s benchmark.tpr']
+        self.executable_opts += ['-nb', self.nb_impl, '-s benchmark.tpr']
```
vkarak, line 68: Now that #2273 and #2287 are merged, we don't need the `+=` here.

vkarak (changes requested):
Just a couple of minor comments.

vkarak:
@victorusu We will also need to document the test properly. I will address the PR comments, but can you expand a bit on what the actual tests. An important thing is to point to the input files.

vkarak (approved):
I've done some fixes in the documentation and now it's ready to be merged.

vkarak:
@jenkins-cscs retry daint

PR #2191, opened 2021-09-13
[test] Add generic ICON check for the cscs-supported apps into hpctestlib/apps

hurricane642:
This is a continuation of the work on the test library. This PR work has been done with ICON tests. A new file base_check.py has been added, introducing the main class for the ICON test, as well as the CSCS tests inherited from it.
Problems with this test in general 2. First, the original test is broken, so there is no way to check the functionality of the written test. Second, I did not find any normal documentation for this framework, as well as a description of the test, so perhaps the documentation of the test leaves much to be desired.

### DIFF
--- a/cscs-checks/apps/icon/rrtmgp_check.py
+++ b/cscs-checks/apps/icon/rrtmgp_check.py
@@ -6,40 +6,21 @@
 import os
 
 import reframe as rfm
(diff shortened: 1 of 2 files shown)


vkarak:
@hurricane642 I don't think that this test is relevant anymore. So I wouldn't invest in it.

PR #2194, opened 2021-09-16
[feat] Execute build phase asynchronously

### LINKED ISSUE #1617: Execute build phase asynchronously

### LINKED ISSUE #1708: Improve the execution of the test pipeline in the execution policies
In order to scale the design and support extensions such as #1617, we will need to redesign how a test transitions from one pipeline phase to the next. The idea is to use buffers in between and the event handlers will be responsible to moving the tasks for queue to queue. Also more events will be needed. All these can also be named in an homogeneous way, so that we could dynamically call the right callbacks and perhaps even move things automatically from queue to queue.

### LINKED ISSUE #2109: All `CompileOnlyRegressionTest` before `RunOnlyRegressionTest`
Hi,
consider the following:

```python
import reframe as rfm
from reframe.utility import udeps
from reframe.core.decorators import run_after, run_before

@rfm.simple_test
class ABuildTest(rfm.CompileOnlyRegressionTest):
    descr = "A Build"
    valid_systems = ["hpc:local"]
    valid_prog_environs = ["*"]
    build_system = "Spack"

    @run_before("compile")
    def setup_build_system(self):
        self.build_system.specs = ["fio"]
        self.build_system.install_tree = "$SPACK_ROOT/opt/spack"

    @run_before("sanity")
    def set_sanity_patterns(self):
        self.sanity_patterns = sn.assert_found("Updating", self.stdout)


@rfm.simple_test
class ALRunTest(rfm.RunOnlyRegressionTest):
    descr = "A Run"
    valid_systems = ["hpc:hpc"]
    valid_prog_environs = ["*"]
    case = parameter([0, 1])
    executable = "date"

    @run_after("init")
    def set_depends_on(self):
        self.depends_on("ABuildTest", udeps.fully)

    @sanity_function
    def assert_passed(self):
        return True


@rfm.simple_test
class BBuildTest(rfm.CompileOnlyRegressionTest):
    descr = "B Build"
    valid_systems = ["hpc:local"]
    valid_prog_environs = ["*"]
    build_system = "Spack"

    @run_before("compile")
    def setup_build_system(self):
        self.build_system.specs = ["fio"]
        self.build_system.install_tree = "$SPACK_ROOT/opt/spack"

    @run_before("sanity")
    def set_sanity_patterns(self):
        self.sanity_patterns = sn.assert_found("Updating", self.stdout)


@rfm.simple_test
class BLRunTest(rfm.RunOnlyRegressionTest):
    descr = "B Run"
    valid_systems = ["hpc:hpc"]
    valid_prog_environs = ["*"]
    case = parameter([2, 3, 4])
    executable = "date"

    @run_after("init")
    def set_depends_on(self):
        self.depends_on("BBuildTest", udeps.fully)

    @sanity_function
    def assert_passed(self):
        return True
```

As you can see there are 2 applications (`A` and `B`) and each one is divided into `CompileOnlyRegressionTest` and `RunOnlyRegressionTest`.

Even if the dependencies are correclty defined:
```console
$ ./reframe/bin/reframe --checkpath deps --list -vv
...
Full test DAG:
  ('ABuildTest', 'hpc:local', 'gcc') -> []
  ('ALRunTest_0', 'hpc:hpc', 'gcc') -> [('ABuildTest', 'hpc:local', 'gcc')]
  ('ALRunTest_1', 'hpc:hpc', 'gcc') -> [('ABuildTest', 'hpc:local', 'gcc')]
  ('BBuildTest', 'hpc:local', 'gcc') -> []
  ('BLRunTest_2', 'hpc:hpc', 'gcc') -> [('BBuildTest', 'hpc:local', 'gcc')]
  ('BLRunTest_3', 'hpc:hpc', 'gcc') -> [('BBuildTest', 'hpc:local', 'gcc')]
  ('BLRunTest_4', 'hpc:hpc', 'gcc') -> [('BBuildTest', 'hpc:local', 'gcc')]
...
```
On running (`./reframe/bin/reframe --checkpath deps --run -v`) it wil run `ABuildTest` first then `BBuildTest` and then all the `ALRunTest_*`/`BLRunTest_*` in parallel.
Is there a particular reason why `reFrame` is running all the `CompileOnlyRegressionTest` first and then `RunOnlyRegressionTest`?

Thanks

ekouts:
This PR refactors the policies of ReFrame, and especially how the asynchronous policy is managed. Additionally we allow the asynchronous execution of the build phase.

The main ideas of the PR are:
- In the policy we keep a set of all the "current_tasks".
- All of the tests start from the `wait` stage.
- After all tests are processed (and are in the `current_tasks` set) we go into a busy loop in which:
  - We poll all the schedulers but without immediately advancing any test.
  - We check all the tasks of the `current_tasks` set (green stages) and try to advance them to the next stage.
- Retired tasks (the tasks that have successfully finished but haven't cleaned up their resources yet) are managed differently because of the dependencies. A test is not cleaned up until all tests that are depending on it have successfully finished. If a test (or any of the tests that depend on it) fail the stage directory will not be "cleaned up" and consequently the ReFrame execution might finish without emptying the `retired` stack.

reframe_policies

With the async building a new issue was created. By default ReFrame will build the test locally (and not in the partition). So the `max_jobs` limit of the partition might not be good enough for the "forced" local tests (the ones that have the attribute `local` to `True`) + the building jobs. For this reason we added a new limit (current name is `rfm_max_jobs` but might change) for these jobs.

Finally there is also to possibility to not advance all tests in every loop but we can set a time limit (to avoid for example that all tests in the beginning are serially doing some "setup work", like conflict module resolution or hook that perform heavy operations) without submitting jobs to the schedulers. I would recommend we set arbitrarily the time limit to maybe 10 seconds(?). This would make it more relevant in the beginning, where the tests are setting up, doing submissions etc and not so much after the jobs are submitted and we poll for them.

Closes #1708.
Closes #1617.
Closes #2109.


reframe/frontend/executors/policies.py
```diff
@@ -424,32 +458,53 @@ def runcase(self, case):
...
+                               f'{partname!r}')
             forced_local_jobs, part_jobs = split_jobs(
                 self._running_tasks[partname]
             )
             part.scheduler.poll(*part_jobs)
             self.local_scheduler.poll(*forced_local_jobs)
 
             # Trigger notifications for finished jobs
-            for t in self._running_tasks[partname]:
+            for t in self._running_tasks[partname][:]:
```
vkarak, line 491: Why do you need a copy here?

reframe/frontend/executors/policies.py
```diff
@@ -227,6 +233,9 @@ def __init__(self):
         # Index tasks by test cases
         self._task_index = {}
 
+        # All currently building tasks per partition
+        self._building_tasks = {}
```
vkarak, line 237:
```suggestion
        self._build_tasks = {}
```

reframe/frontend/executors/policies.py
```diff
@@ -227,6 +233,9 @@ def __init__(self):
...
+
         # All currently running tasks per partition
```
vkarak, line 239: Is this now accurate? I think that the running tasks are all those in the run phase.

reframe/frontend/executors/policies.py
```diff
@@ -394,20 +428,20 @@ def runcase(self, case):
...
+            #         f'{partition.max_jobs} job(s)'
+            #     )
+            #     self._poll_tasks()
+
+            # if len(self._running_tasks[partname]) < partition.max_jobs:
+            #     # Task was put in _ready_tasks during setup
+            #     self._ready_tasks[partname].pop()
+            #     self._reschedule(task)
+            # else:
+            #     self.printer.status('HOLD', task.check.info(), just='right')
```
vkarak, line 444: Why did you comment that out?

reframe/frontend/executors/policies.py
```diff
@@ -424,32 +458,53 @@ def runcase(self, case):
     def _poll_tasks(self):
         '''Update the counts of running checks per partition.'''
 
-        def split_jobs(tasks):
+        def split_jobs(tasks, build_split=False, build_jobs=False):
```
vkarak, line 461: Why do you need to differentiate the build jobs here?

reframe/frontend/executors/policies.py
```diff
@@ -424,32 +458,53 @@ def runcase(self, case):
...
+                               f'{partname!r}')
+            forced_local_jobs, part_jobs = split_jobs(
+                self._building_tasks[partname], build_split=True,
+                build_jobs=True
+            )
+            part.scheduler.poll(*part_jobs)
+            self.local_scheduler.poll(*forced_local_jobs)
+
+            # Trigger notifications for finished jobs
+            for t in self._building_tasks[partname][:]:
```
vkarak, line 505: Why the copy here?

reframe/frontend/executors/policies.py
```diff
@@ -499,7 +554,9 @@ def _failall(self, cause):
     def _reschedule(self, task):
```
vkarak, line 554:
```suggestion
    def _reschedule_compile(self, task):
```

reframe/frontend/executors/policies.py
```diff
@@ -499,7 +554,9 @@ def _failall(self, cause):
...
         getlogger().debug2(f'Scheduling test case {task.testcase} for running')
```
vkarak, line 555: The log message here is no more accurate.

unittests/test_policies.py
```diff
@@ -722,8 +728,8 @@ def test_kbd_interrupt_in_wait_with_limited_concurrency(
     runner, _ = async_runner
     with pytest.raises(KeyboardInterrupt):
         runner.runall(make_cases([
-            KeyboardInterruptCheck(), SleepCheck(10),
-            SleepCheck(10), SleepCheck(10)
+            SleepCheck(10), SleepCheck(10),
+            KeyboardInterruptCheck(), SleepCheck(10)
```
vkarak, line 732: Which unit test was failing without this change and how?

vkarak (changes requested):

reframe/core/pipeline.py
```diff
@@ -1557,6 +1557,36 @@ def run(self):
...
+           You may not override this method directly unless you are in
+           special test. See `here
+           <migration_2_to_3.html#force-override-a-pipeline-method>`__ for
+           more details.
+
+
+           .. versionchanged:: 3.4
+              Overriding this method directly in no longer allowed. See `here
+              <migration_2_to_3.html#force-override-a-pipeline-method>`__ for
+              more details.
```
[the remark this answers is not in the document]
ekouts: Just the `versionchanged` or the warning as well?

reframe/frontend/executors/policies.py
ekouts: I am not sure why I put two extra arguments here, I think just by mistake. I will keep one since `split_jobs` needs to know if it needs to return a list of build_jobs or the run jobs.

unittests/test_policies.py
ekouts: The unittest was failing is `test_kbd_interrupt_in_wait_with_limited_concurrency`. This is because we set the max jobs allowed to 2 and the order of the processing of the tests is different. Refrane would pick up two `SleepCheck` tests, and when they are over it picks up the other `SleepCheck` and `KeyboardInterruptCheck`. Finally only one would be aborted this way.
This is the error before changing:
```
=============================================================================== FAILURES ================================================================================
__________________________________________________________ test_kbd_interrupt_in_wait_with_limited_concurrency __________________________________________________________

async_runner = (<reframe.frontend.executors.Runner object at 0x7fce9e40fa90>, <unittests.test_policies._TaskEventMonitor object at 0x7fce9e40f4d0>)
make_cases = <function make_cases.<locals>._make_cases at 0x7fce9e4c8e60>, make_exec_ctx = <function make_exec_ctx.<locals>._make_exec_ctx at 0x7fce9e5a7290>

    def test_kbd_interrupt_in_wait_with_limited_concurrency(
            async_runner, make_cases, make_exec_ctx
    ):
        # The general idea for this test is to allow enough time for all the
        # four checks to be submitted and at the same time we need the
        # KeyboardInterruptCheck to finish first (the corresponding wait should
        # trigger the failure), so as to make the framework kill the remaining
        # three.
        make_exec_ctx(options=max_jobs_opts(2))
        runner, _ = async_runner
        with pytest.raises(KeyboardInterrupt):
            runner.runall(make_cases([
                KeyboardInterruptCheck(), SleepCheck(10),
                SleepCheck(10), SleepCheck(10)
            ]))
    
>       assert_interrupted_run(runner)

unittests/test_policies.py:735: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

runner = <reframe.frontend.executors.Runner object at 0x7fce9e40fa90>

    def assert_interrupted_run(runner):
        assert 4 == runner.stats.num_cases()
        assert_runall(runner)
        assert 1 == len(runner.stats.failed())
>       assert 3 == len(runner.stats.aborted())
E       assert 3 == 1
E        +  where 1 = len([<reframe.frontend.executors.RegressionTask object at 0x7fce9e1d6750>])
E        +    where [<reframe.frontend.executors.RegressionTask object at 0x7fce9e1d6750>] = <bound method TestStats.aborted of <reframe.frontend.statistics.TestStats object at 0x7fce9e4186d0>>()
E        +      where <bound method TestStats.aborted of <reframe.frontend.statistics.TestStats object at 0x7fce9e4186d0>> = <reframe.frontend.statistics.TestStats object at 0x7fce9e4186d0>.aborted
E        +        where <reframe.frontend.statistics.TestStats object at 0x7fce9e4186d0> = <reframe.frontend.executors.Runner object at 0x7fce9e40fa90>.stats

unittests/test_policies.py:695: AssertionError
======================================================================== short test summary info ========================================================================
FAILED unittests/test_policies.py::test_kbd_interrupt_in_wait_with_limited_concurrency - assert 3 == 1
========================================================================== 1 failed in 12.72s ===========================================================================
```

reframe/frontend/executors/policies.py
ekouts: The problem is that it complicated the logic a bit - and unnecessarily in my opinion. For me it's simpler if we simply setup the test in the `runcase` and I commented this out to test it like this first. I am having a look if I can make it work with scheduling tasks as soon as possible.

reframe/core/pipeline.py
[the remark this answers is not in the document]
vkarak: Also the reference to the migration from 2 to 3.

unittests/test_policies.py
vkarak: I think that this happened because you removed the logic of submitting tests from the the `runcase()`.

reframe/frontend/executors/policies.py
vkarak: I think that we should schedule tasks as soon as possible. Imagine you have thousands of tests that need to load modules. By default the setup will try to track any conflicts by interacting with the module system, an operation that can be slow, so practically you'd have to pay this cost for all the tests before even actually trying to execute any test. Besides, the fact that we select to reschedule when we reach the partition concurrency limit is just a choice that has its root in reducing the poll ratio of the old, poll-hungry mechanism. We could poll at each an every call of `runcase()` which would also allow us to finalise tests as soon as possible, too. So I wouldn't eliminate this logic from here.

ekouts:
After our conversation last week, I was updating this PR so that we have two job limits per partition, `max_jobs` for the `run` phase and `max_build_jobs` for the `build_phase`, but I am not sure this is what we actually want. If we have tests that are building remotely then we probably want to follow the normal partition limit even on the `build` phase in this case.

So maybe the solution could be that we use the `forced_local_max_jobs` limit:
- for the `build` phase, when `build_locally=True` or `local=True`
- for the `run` phase when `local=True` 

Otherwise we check the `max_jobs` limit. Not sure which limit to follow when the partition is local? Follow the same logic as with normal partition?

Another idea would be to have a "universal" limit for *forced* local jobs. In the end what we care about is the sum of the local jobs building, even if they come from different partitions so we could set that on the general configuration and not per partition.

@vkarak, @teojgo do you have an opinion on what the limits should be?

ekouts:
Personally I prefer the last idea, of one universal local limit, since we treat local jobs specially anyway. I am not sure if it should count all the local jobs or just the "forced" ones (by `local=True` or `build_locally=True`).

teojgo:
> After our conversation last week, I was updating this PR so that we have two job limits per partition, `max_jobs` for the `run` phase and `max_build_jobs` for the `build_phase`, but I am not sure this is what we actually want. If we have tests that are building remotely then we probably want to follow the normal partition limit even on the `build` phase in this case.
> 
> So maybe the solution could be that we use the `forced_local_max_jobs` limit:
> 
> * for the `build` phase, when `build_locally=True` or `local=True`
> * for the `run` phase when `local=True`
> 
> Otherwise we check the `max_jobs` limit. Not sure which limit to follow when the partition is local? Follow the same logic as with normal partition?
> 
> Another idea would be to have a "universal" limit for _forced_ local jobs. In the end what we care about is the sum of the local jobs building, even if they come from different partitions so we could set that on the general configuration and not per partition.
> 
> @vkarak, @teojgo do you have an opinion on what the limits should be?

We could somehow associate it to the number of cores of the host system in case it is local, since we have the information from the autodetection.

ekouts:
But my question is not so much the number we will set but what kind of limits we want to have. Two per partition (and if yes what would be the second `max_build_jobs` or `forced_local_max_jobs`) or one `local_max_jobs` for the whole system?

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
-        self._partitions.add(partition)
+    def advance_ready_to_compile(self, task):
+        if task.check.local or task.check.build_locally:
+            if len(self._local_scheduler_tasks) < self._rfm_max_jobs:
+                try:
+                    task.compile()
+                    task.policy_stage = 'compiling'
+                    self._local_scheduler_tasks.add(task)
+                except TaskExit:
+                    self._current_tasks.remove(task)
```
vkarak, line 416: The following pattern comes up again and again, so you should consider it wrapping it in a function:

```python
try:
    task.stage()
except TaskExit:
   self._current_tasks.remove(task)
```

What about perhaps something like `self._exec_stage(task.compile)`?

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
-            (check.name, partition.fullname, environ.name)
-        )
+        partname = task.check.current_partition.fullname
+        if len(self._scheduler_tasks[partname]) < self._max_jobs[partname]:
+            try:
+                task.compile()
+                task.policy_stage = 'compiling'
+                self._scheduler_tasks[partname].add(task)
+            except TaskExit:
+                self._current_tasks.remove(task)
```
vkarak, line 429: This is almost identical to the above part. Couldn't we have a specially named pseudo-partition (just for the data structures of the policies) so that we avoid this duplication?

Also is there a reason that you bump only the local tests and not both in case you can?

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
+
+            if isinstance(task.check, RunOnlyRegressionTest):
+                try:
+                    task.compile()
+                    task.compile_wait()
+                except TaskExit:
+                    # Run and run_wait are no-ops for
+                    # CompileOnlyRegressionTest. This shouldn't fail.
+                    self._current_tasks.remove(task)
+                    return 1
```
vkarak, line 391: See my comment below about this pattern.

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
-                    break
+    # TODO all this prints have to obviously leave from here...
+    def on_task_setup(self, task):
+        # print(task.check.name, 'setup')
+        pass
 
-                self._reschedule(task)
-                num_rescheduled += 1
+    def on_task_run(self, task):
+        if isinstance(task.check, RunOnlyRegressionTest):
```
vkarak, line 569: Check my suggestion for output.

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
+                                         'total'])
+        msg = f'{task.check.info()} [{timings}]'
+        self.printer.status('OK', msg, just='right')
+        timings = task.pipeline_timings(['setup',
+                                         'compile_complete',
+                                         'run_complete',
+                                         'sanity',
+                                         'performance',
+                                         'total'])
+        getlogger().verbose(f'==> {timings}')
```
vkarak, line 633: Isn't this almost identical to the above?

reframe/schemas/config.json
```diff
@@ -237,6 +237,7 @@
                     "stagedir": {"type": "string"},
                     "outputdir": {"type": "string"},
                     "resourcesdir": {"type": "string"},
+                    "rfm_max_jobs": {"type": "number"},
```
vkarak, line 240: We should find a better name.

unittests/test_policies.py
```diff
@@ -591,9 +597,10 @@ def _read_timestamps(tasks):
 
 def test_concurrency_unlimited(async_runner, make_cases, make_exec_ctx):
     num_checks = 3
-    make_exec_ctx(options=max_jobs_opts(num_checks))
+    # make_exec_ctx(options=max_jobs_opts(num_checks))
```
vkarak, line 600: Why did you comment these out?

vkarak (changes requested):
I like the flow of the new implementation, but there is quite some duplication which we should eliminate before merging. Also an extensive comment explaining the key idea of the implementation would be more than helpful for other maintainers.

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
-                f'maximum number of failures ({self.max_failures}) reached'
-            )
+        # Set partition-based counters, if not set already
+        self._scheduler_tasks.setdefault(partition.fullname, set())
+        self._max_jobs.setdefault(partition.fullname, partition.max_jobs)
+        task = RegressionTask(case, self.task_listeners)
+        self._task_index[case] = task
+        self.stats.add_task(task)
+        self.printer.info(
+            f'==> added {check.name} on {partition.fullname} '
```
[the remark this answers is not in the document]
ekouts: Okay, should I print the `RUN` message in the run phase? Like this for build-only tests the message will appear almost when the test will be finishing. Or should I print it in the build phase for all the tests?

unittests/test_policies.py
ekouts: It has no effect, since the tests run as "forced" locals and this is why I had to set manually `runner.policy._rfm_max_jobs`. Should I remove or change the test so it runs in a local partition?

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
+        jobs = []
+        for t in self._local_scheduler_tasks:
+            if t.policy_stage == 'compiling':
+                jobs.append(t.check.build_job)
+            elif t.policy_stage == 'running':
+                jobs.append(t.check.job)
+
+        self.local_scheduler.poll(*jobs)
+
+    def advance_all(self, tasks, timeout=None):
```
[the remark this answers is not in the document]
ekouts: I added a figure and some explanation in the description of the PR, let me know if you think it's more clear now.

reframe/frontend/executors/policies.py
ekouts: > Also is there a reason that you bump only the local tests and not both in case you can?

I don't understand that comment, I am always trying to bump the test first but I am checking the appropriate job limit.

> Couldn't we have a specially named pseudo-partition (just for the data structures of the policies) so that we avoid this duplication?

This is a bit silly but I wanted to avoid finding a name that will not clash without a real partition. I guess we can put two underscores in the beginning and be safe.

reframe/frontend/executors/policies.py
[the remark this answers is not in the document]
vkarak: No, when start running the test pipeline: `RUN` here means "run the test." I doesn't matter how many phases it has.

reframe/frontend/executors/policies.py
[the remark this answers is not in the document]
vkarak: I've seen the description of the PR and is very good; I meant to add a comment in the code...

reframe/frontend/executors/policies.py
vkarak: > I don't understand that comment, I am always trying to bump the test first but I am checking the appropriate job limit.

I mean that if you have local tests waiting, you will only bump those and return. Eventually, you will bump them all, but wouldn't it make sense to bump both a local and a normal test at once? Just asking.

> This is a bit silly but I wanted to avoid finding a name that will not clash with a real partition. I guess we can put 1 or 2 underscores in the beginning and be safe.

I think the `_rfm` namespace is de facto reserved for ReFrame :-D So `_rfm_local` should be ok.

reframe/frontend/executors/policies.py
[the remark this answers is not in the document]
ekouts: ah all right, I will add some comments to the code as well 👍

unittests/test_policies.py
vkarak: Hmm, you're right. I wonder how this worked before? Also I think that you should not set it directly on policy but do it through the configuration with the `make_exec_ctx`.

ekouts: > Also I think that you should not set it directly on policy but do it through the configuration with the make_exec_ctx.

Not sure what you mean. I set the `_rfm_max_jobs` to different values (1, 3, `num_checks`) in three different tests. Should I make 3 different fixtures or is there a way to pass an argument?

reframe/frontend/executors/policies.py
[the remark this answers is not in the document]
ekouts: Okay, just to make sure, you mean after they setup?

reframe/frontend/executors/policies.py
```diff
@@ -221,337 +235,404 @@ def exit(self):
...
+        t_init = time.time()
+        num_prog = 0
+
+        # progress might remove the tasks that retire or fail
+        for t in list(tasks):
```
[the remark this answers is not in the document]
ekouts: Just to make sure, you mean to pop them and and add them back in in the end? Tests in general are removed only in case of errors or when they are retired. I will have to change the `_advance_{stage}` methods to return something extra to know whether it should be added back or not.

reframe/frontend/executors/policies.py
```diff
@@ -326,52 +338,56 @@ def exit(self):
...
-            num_prog += method(t)
+            bump_state = getattr(self, f'advance_{t._current_stage}')
+            num_progressed += bump_state(t)
             t_elapsed = time.time() - t_init
-            if timeout and t_elapsed > timeout and num_prog:
+            if timeout and t_elapsed > timeout and num_progressed:
                 break
 
-    def advance_wait(self, task):
+        getlogger().debug2(f"Bumped {num_progressed} test(s).")
```
vkarak, line 374:
```suggestion
        getlogger().debug2(f'Bumped {num_progressed} test(s)')
```

reframe/frontend/executors/policies.py
```diff
@@ -98,8 +96,8 @@ def runcase(self, case):
         check, partition, environ = case
 
         self.printer.status(
-            'RUN', '%s on %s using %s' %
-            (check.name, partition.fullname, environ.name)
+            'RUN',
+            f'{check.name} on {partition.fullname} using {environ.name}'
```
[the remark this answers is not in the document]
vkarak: Actually, using `check.info()` should make it forward compatible; so better use that.

reframe/frontend/executors/policies.py
```diff
@@ -326,52 +338,56 @@ def exit(self):
...
                 self._current_tasks.remove(task)
                 return 1
-
         elif self.deps_succeeded(task):
             try:
+                self.printer.status(
+                    'RUN', f'{task.check.name} on '
+                    f'{task.testcase.partition.fullname} using '
+                    f'{task.testcase.environ.name}'
+                )
```
vkarak, line 390: See my comment above about using `check.info()`.

reframe/frontend/executors/policies.py
```diff
@@ -405,25 +417,14 @@ def advance_wait(self, task):
...
-                return 1
-            else:
-                return 0
-
-        partname = task.check.current_partition.fullname
+    def advance_setup(self, task):
+        partname = (
+            '_rfm_local' if task.check.local or task.check.build_locally
+            else task.check.current_partition.fullname
+        )
```
vkarak, line 424: Make that a utility function at the top of the file:

```suggestion
        partname = _get_partition_name(task.check)
```

unittests/test_policies.py
```diff
@@ -595,12 +595,13 @@ def _read_timestamps(tasks):
...
-def test_concurrency_unlimited(async_runner, make_cases, make_exec_ctx):
+def test_concurrency_unlimited(async_runner, make_cases):
     num_checks = 3
-    # make_exec_ctx(options=max_jobs_opts(num_checks))
 
     runner, monitor = async_runner
-    runner.policy._rfm_max_jobs = num_checks
+    runner.policy._max_jobs = {
+        '_rfm_local': num_checks
+    }
```
vkarak, line 604: You should still do it with the `make_exec_ctx` except that the `max_jobs_opts()` should also set the `max_local_jobs` option.

vkarak (changes requested):
We might still need updating the docs on how reframe executes tests. I don't think that we mention there that the build phase is also asynchronous. And perhaps we need to test this PR at scale, have you done that already?

ekouts:
> And perhaps we need to test this PR at scale, have you done that already?

I had run the cscs-checks many times before on Piz Daint, but after making the last changes I will run it again to make sure that no test is failing because of the policies.

reframe/frontend/executors/policies.py
```diff
@@ -326,52 +338,56 @@ def exit(self):
...
             part.scheduler.poll(*jobs)
 
         jobs = []
-        for t in self._local_scheduler_tasks:
-            if t.policy_stage == 'compiling':
+        for t in self._scheduler_tasks['_rfm_local']:
+            if t._current_stage == 'compile':
                 jobs.append(t.check.build_job)
-            elif t.policy_stage == 'running':
+            elif t._current_stage == 'run':
```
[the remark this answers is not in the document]
ekouts: Actually this is a change I made last week and I think we don't need to keep another variable for the "pipeline stage". I think we can use the `_current_stage` that is updated by the safe_call (I made a property but I forgot to use it)

reframe/frontend/executors/policies.py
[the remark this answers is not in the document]
ekouts: Just a small detail. `check.info()` has the full information (partition and environment after the setup. And I think it's not nice to print the message after the setup, in case of failure. Is there a way to get this from self._case at this point? https://github.com/eth-cscs/reframe/blob/master/reframe/core/pipeline.py#L1188

reframe/frontend/executors/policies.py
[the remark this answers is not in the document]
vkarak: Yes, exactly. But I tried that and it was hanging for some reason. Let's leave it as is for now.

unittests/test_policies.py
vkarak: It doesn't work with the `make_exec_ctx`, because the runtime is changed before the asynchronous execution policy is created and the `max_loca_jobs` is read during the construction of the policy, so it gets the default value of 8.

reframe/schemas/config.json
```diff
@@ -237,6 +237,8 @@
                     "stagedir": {"type": "string"},
                     "outputdir": {"type": "string"},
                     "resourcesdir": {"type": "string"},
+                    "policy_timeout": {"type": "number"},
+                    "max_local_jobs": {"type": "number"},
```
vkarak, line 241: The options should be sorted alphabetically everywhere in the schema.

reframe/schemas/config.json
```diff
@@ -237,6 +237,8 @@
                     "stagedir": {"type": "string"},
                     "outputdir": {"type": "string"},
                     "resourcesdir": {"type": "string"},
+                    "policy_timeout": {"type": "number"},
```
vkarak, line 240: This should be a `general` option, not a system-specific one.

reframe/frontend/executors/policies.py
```diff
@@ -219,339 +232,362 @@ def exit(self):
...
 
         self._pollctl = _PollController()
 
         # Index tasks by test cases
         self._task_index = {}
 
-        # All currently running tasks per partition
-        self._running_tasks = {}
+        # A set of all the current tasks
+        self._current_tasks = set()
```
vkarak, line 288: I changed that to an `OrderedSet`. The unit tests assume that the tasks will be advanced in the order they are passed to `runall()`. This is the only way to test different corner case scenarios with faulty tests.

vkarak:
I don't think that we can merge this PR right away. I believe it needs further testing on large scale scenarios and I don't want to rush it in, because it reimplements completely the asynchronous execution policy. For example, I am not sure the `pipeline_timeout` (formerly, `policy_timeout` works correctly). Generally, Implementation-wise, I am fine.

vkarak:
We will also need to change the version annotations in the documentation for things added or changed by this PR to mention 3.10 instead of 3.9.3.

docs/pipeline.rst
```diff
@@ -133,6 +134,67 @@ When the `concurrency limit <config_reference.html#.systems[].partitions[].max_j
...
+The following table show in which context each pipeline stage executes:
+
+.. table::
+   :align: center
+
+   ============== =================
+   Pipeline Stage Execution Context
+   ============== =================
+   *Setup*        ReFrame
+   *Compile*      ReFrame if :attr:`~reframe.core.pipeline.RegressionTest.build_locally` is :obj:`True`, partition otherwise.
```
ekouts, line 181:
```suggestion
   *Compile*      ReFrame if :attr:`~reframe.core.pipeline.RegressionTest.build_locally` or :attr:`~reframe.core.pipeline.RegressionTest.local` is :obj:`True`, partition otherwise.
```

ekouts: Actually it should be:
```suggestion
   *Compile*      ReFrame if :attr:`~reframe.core.pipeline.RegressionTest.build_locally` or :attr:`~reframe.core.pipeline.RegressionTest.local` is :obj:`True` or if :option:`--force-local` is passed, partition otherwise.
```
I will push the fix.

docs/config_reference.rst
```diff
@@ -1301,6 +1301,16 @@ General Configuration
   Timeout value in seconds used when checking if a git repository exists.
 
 
+.. js:attribute:: .general[].pipeline_statistics
+
+   A boolean value indicating whether we want to collect statistic information for the execution of the asynchronous pipeline.
```
vkarak, line 1306:
```suggestion
   Dump pipeline progress for the asynchronous execution policy.
   This is option is meant for debug purposes only.
```

You should also mention the file that it generates.

Generally, what we do here is not statistics, but rather dumping the progress of the tests in the pipeline.

docs/config_reference.rst
```diff
@@ -1301,6 +1301,16 @@ General Configuration
   Timeout value in seconds used when checking if a git repository exists.
 
 
+.. js:attribute:: .general[].pipeline_statistics
```
vkarak, line 1304:
```suggestion
.. js:attribute:: .general[].dump_pipeline_progress
```

reframe/frontend/executors/policies.py
```diff
@@ -285,24 +324,48 @@ def runcase(self, case):
...
+                if timeout is not None:
+                    timeout = float(timeout)
+
                 self._advance_all(self._current_tasks, timeout)
+                if self._pipeline_statistics:
+                    num_retired = len(self._retired_tasks)
+
                 _cleanup_all(self._retired_tasks, not self.keep_stage_files)
+                if self._pipeline_statistics:
+                    new_num_retired = len(self._retired_tasks)
```
vkarak, line 352: Leave a blank line before the comment block.

vkarak, line 352:
```suggestion
                    num_retired_actual = num_retired - len(self._retired_tasks)
```

vkarak (changes requested):
Just a couple of minor changes still.

reframe/frontend/executors/policies.py
```diff
@@ -349,12 +349,13 @@ def exit(self):
 
                 _cleanup_all(self._retired_tasks, not self.keep_stage_files)
                 if self._pipeline_statistics:
-                    new_num_retired = len(self._retired_tasks)
+                    num_retired_actual = len(self._retired_tasks)
```
vkarak, line 352: This is not the `num_retired_actual`. You should do the subtraction here instead.

vkarak (changes requested):
Just a minor comment.

reframe/frontend/executors/policies.py
vkarak: I fixed that.

vkarak (approved):

PR #2201, opened 2021-09-24
[bugfix] Fix `--restore-session` failure when ReFrame root directory is not in `sys.path[0]`

### LINKED ISSUE #2135: Errors when ReFrame root directory is not first path in `sys.path`
Hello,

The context here is when I use ReFrame CI integration capabilities, I am getting errors in the dependent jobs. When we use `--restore-session` flag for dependent job, ReFrame is detecting modules from filenames [here](https://github.com/eth-cscs/reframe/blob/bb779ce73a4f081aef1b6e0c20c127061bec24d1/reframe/utility/__init__.py#L84). 

I am not sure of the rabbit hole details of why it is being done, but the issue I have noticed is that on that stated line, we implicitly assume `sys.path[0]` if ReFrame root directory. If it is not, ReFrame will have problems finding the modules. For example, in this [failed CI pipeline](https://gitlab.com/ska-telescope/sdp/ska-sdp-benchmark-tests/-/jobs/1512352442#L87), I printed the `sys.path` and we can see that ReFrame root directory is not the first path.

This is the directory structure of my repository:
```
.
├── apps
├── config
├── docs
├── helloworld
├── modules
├── output
├── perflogs
├── reframe
├── spack
└── topologies
```

And I usually launch tests from root of the repo using:
`reframe/bin/reframe -C <config> -c <test> -r`

rsarm:
Close #2135

### DIFF
--- a/reframe/utility/__init__.py
+++ b/reframe/utility/__init__.py
@@ -17,6 +17,8 @@ import sys
 import types
 import weakref
 
+import reframe
+
 from collections import UserDict
 from . import typecheck as typ
 
@@ -81,9 +83,7 @@ def import_module_from_file(filename, force=False):
         filename = os.path.join(filename, '__init__.py')
 
     # Express filename relative to reframe
-    rfm_root = os.path.realpath(os.path.join(os.path.dirname(__file__),
-                                '../..'))
-    rel_filename = os.path.relpath(filename, rfm_root)
+    rel_filename = os.path.relpath(filename, reframe.INSTALL_PREFIX)
     module_name = _get_module_name(rel_filename)
     if rel_filename.startswith('..'):
         # We cannot use the standard Python import mechanism here, because the
--- a/reframe/utility/__init__.py
+++ b/reframe/utility/__init__.py
@@ -81,7 +81,10 @@ def import_module_from_file(filename, force=False):
         filename = os.path.join(filename, '__init__.py')
 
     # Express filename relative to reframe
-    rel_filename = os.path.relpath(filename, sys.path[0])
+    rfm_root = os.path.realpath(os.path.join(os.path.dirname(__file__),
+                                '../..'))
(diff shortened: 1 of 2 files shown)


rsarm:
@mahendrapaipur could you have a look?

mahendrapaipuri:
@rsarm Looks good. Just a minor suggestion. Do you think moving 
```
rfm_root = os.path.realpath(os.path.join(os.path.dirname(__file__), '../..'))
```
into a new function like `get_rfm_root` makes sense? This makes it more modular and reusable if we ever need ReFrame root directory in further developments.

teojgo (approved):

reframe/utility/__init__.py
```diff
@@ -81,7 +81,9 @@ def import_module_from_file(filename, force=False):
         filename = os.path.join(filename, '__init__.py')
 
     # Express filename relative to reframe
-    rel_filename = os.path.relpath(filename, sys.path[0])
+    rfm_root = os.path.realpath(os.path.join(os.path.dirname(__file__),
+                                '../..'))
```
vkarak, line 85: Isn't that equivalent to this?

```suggestion
    rfm_root = os.path.realpath(reframe.INSTALL_PREFIX)
```

And another question is that it's not obvious to me the purpose of `os.path.realpath` here.

vkarak (changes requested):

reframe/utility/__init__.py
rsarm: The `os.path.realpath` was to get the resulting path without the `../../`

rsarm:
What about replacing by [here](https://github.com/eth-cscs/reframe/blob/9804e1cace17229a656ad7cb9b28069cee2437cf/reframe/core/exceptions.py#L308 ) `reframe.INSTALL_PREFIX` too?

vkarak (approved):

PR #2225, opened 2021-10-16
[feat] Add configuration parameter to control the timeout of the `git clone` commands

### LINKED ISSUE #2218: Add configuration parameter to control `git clone` timeouts in `sourcesdir`
Based on the discussion in the community [Slack](https://reframetalk.slack.com/archives/CF8ENLUDB/p1634035016088000).

mahendrapaipuri:
Fixes #2218

### DIFF
--- a/JenkinsfileDeploy
+++ b/JenkinsfileDeploy
@@ -2,7 +2,7 @@ def tag = env.TAG_NAME
 def version = tag.substring(1)                  // Remove the first character of the tag which is 'v'
 def machines = '[dom daint eiger pilatus tsa]'  // To be included in the PR title
 def reviewersPR = 'teojgo,vkarak'               // The Github usernames requested for review
-def ebTemplate = '3.7.2'                        // The eb recipe of ReFrame to use as template
+def ebTemplate = '3.8.3'                        // The eb recipe of ReFrame to use as template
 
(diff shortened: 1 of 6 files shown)


mahendrapaipuri:
@jjotero Yes, your suggestion sounds good and more generic. Why can't we do something more simpler like:

```
if timeout is not None:
    timeout = max(timeout, rt.runtime().get_option('general/0/min_timeout'))
```
As we define a default value for `min_timeout` in config schema, it will never raise an exception. Am I missing something here? 

Also, in the slack conversion I had with @vkarak he told
"Because the utility package is meant to be independent, not knowing anything about the runtime, including the configuration."
This solution will bring runtime configuration into utility package. Would it have any side effects?

jjotero:
@mahendrapaipuri your option requires the runtime to be present in order to call this `run_command` function. If the runtime is not setup, you get a `ReframeFatalError` when you call `rt.runtime()`. And as you said, since this is a utility function, it should never require the runtime to be present. However, with what I suggested, the runtime being present is entirely optional.

In any case, circular imports prevent us from doing this (😭), so I guess the `git_timeout` option looks like the best solution. After all, cloning a repo on a reframe test is a very common thing to do, so I can see valid reasons to keep it as a config option.

PR #2256, opened 2021-10-29
[doc] Update README

vkarak:
I updated a bit the README to make it more precise regarding the new super powerful capabilities of the framework 😂

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,16 +14,14 @@
 
 # ReFrame in a Nutshell
 
-ReFrame is a framework for writing regression tests for HPC systems.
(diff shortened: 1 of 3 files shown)


teojgo:
@vkarak do you think we should update the short description in the `setup.cfg`?

ekouts (approved):

vkarak:
@teojgo Yes, I will do that, thanks!

vkarak:
I've also updated the intro page of the documentation accordingly as well as the `setup.cfg` as indicated here.

PR #2257, opened 2021-11-01
[doc] Add tutorial about fixtures

### LINKED ISSUE #2241: Rewrite dependencies tutorial example using fixtures
We can keep the current tutorial, but add a new one, implementing the same set of tests, but using fixtures.

vkarak:
The tutorial rewrites the OSU benchmarks using fixtures and introduces the users into the concept. I have also added a couple of figures that help visualize the fixtures in this example. 

Fixes #2241.


docs/tutorial_fixtures.rst
```diff
@@ -0,0 +1,275 @@
...
+
+A test defines a fixture using the :func:`~reframe.core.pipeline.RegressionMixin.fixture` builtin and assigns it a name by assigning the return value of the builtin to a test variable, here ``osu_binaries``.
+This name will be used later to access the resource managed by the fixture.
+
+As stated previously, a fixture is another full-fledged ReFrame test, here the :class:`build_osu_benchmarks` which will take care of building the OSU benchmarks.
+Each fixture is associated with a scope.
+This practically indicates at which level a fixture is shared with other tests.
+There are four fixture scopes, which are listed below in decreasing order of generality:
+
+- ``session``: A fixture with this scope will be executed once per a ReFrame run session and will be shared across the whole run.
```
ekouts, line 33:
```suggestion
- ``session``: A fixture with this scope will be executed once per ReFrame run session and will be shared across the whole run.
```
I think the "a" is not correct syntax, but I don't have a strong opinion.

docs/tutorial_fixtures.rst
```diff
@@ -0,0 +1,275 @@
...
+
+Let's inspect the last fixture, the :class:`fetch_osu_benchmarks`:
+
+.. literalinclude:: ../tutorials/fixtures/osu_benchmarks.py
+   :lines: 11-22
+   :emphasize-lines: 8
+
+There is nothing special to this test -- it is just an ordinary test -- except that we force it to execute locally by setting its :attr:`~reframe.core.pipeline.RegressionTest.local` variable.
+The reason for that is that a fixture at session scope can execute with any partition/environment combination, so ReFrame could have to spawn a job in case it has chosen a remote partition to launch this fixture on.
+For this reason, we simply force it execute locally regardless of the chosen partition.
```
ekouts, line 81:
```suggestion
For this reason, we simply force it to execute locally regardless of the chosen partition.
```

docs/tutorial_fixtures.rst
```diff
@@ -0,0 +1,275 @@
...
+
+ReFrame will generate 4 fixtures for this test based on the partition and environment configurations for the current system.
+The following figure shows the generated fixtures as well as their conceptual dependencies.
+
+.. figure:: _static/img/fixtures-conceptual-deps.svg
+  :align: center
+
+  :sub:`Expanded fixtures and dependencies for the OSU benchmarks example.`
+
+Notice how the :class:`build_osu_benchmarks` fixture is populated three times, for each partition and environment combination, and the :class:`fetch_osu_benchmarks` is generated only once.
```
ekouts, line 129:
```suggestion
Notice how the :class:`build_osu_benchmarks` fixture is populated three times, one for each partition and environment combination, and the :class:`fetch_osu_benchmarks` is generated only once.
```
You can also leave as it was if you prefer it.

ekouts (approved):
Lgtm, I have some small suggestion only.

docs/tutorial_fixtures.rst
vkarak: You're correct.

docs/tutorial_fixtures.rst
vkarak: Perhaps `once` is better.

PR #2265, opened 2021-11-12
[feat] Add metadata to environment definitions in the configuration

### LINKED ISSUE #1973: Add metadata to environments definition
Towards the goal of shareable and reusable tests, I think it would be beneficial to allow metadata in environment definitions. Such metadata could include things such as compiler family, MPI implementation etc. This would allow tests to differentiate their behaviour based on this metadata instead of setting the `valid_prog_environs`, which could be set rather generically. This is very similar to what we have done with the processor information in #1773.

### LINKED ISSUE #2230: Enhancement: Custom user fields in environments
Enhancement request for ReFrame: add a user 'extras' field to the environments definition.

User story:
Whilst working to integrate ReFrame to Spack I have been trying to have ReFrame feed compiler data to Spack. 

In Spack we have the notation `compiler_name@compiler_version` - e.g. `gcc@10.3.0`.

I would like to somehow encode this information within an environment - so that a platform can have multiple environments which represent different Spack compiler environments.

Currently we have an 'extras' field in partitions, which we can embed extra information in - I wish this to be extended to environments.

Example:

```
   'environments': [
        {
            'name': 'gcc7',
            'cc': 'gcc',
            'cxx': 'g++',
            'ftn': 'gfortran',
            'extras': {
                  'spack_compiler_name': 'gcc',
                  'spack_compiler_version': '7.3.0'
            }
        },
        {
            'name': 'gcc10',
            'cc': 'gcc',
            'cxx': 'g++',
            'ftn': 'gfortran',,
            'extras': {
                  'spack_compiler_name': 'gcc',
                  'spack_compiler_version': '10.3.0'
            }
        },
        {
            'name': 'acfl21',
            'cc': 'armclang',
            'cxx': 'armclang++',
            'ftn': 'armflang',,
            'extras': {
                  'spack_compiler_name': 'arm',
                  'spack_compiler_version': '21.1.0.1069'
            }
        }
    ],
```

That way a simple test to check if the compiler is available could look something like this:

```

@rfm.simple_test
class SpackEnvTest(rfm.RunOnlyRegressionTest):

    valid_prog_environs = ['*']
    valid_systems = ['*']

    executable = 'spack'
    executable_opts = ['compiler list']

    @run_after('setup')
    def update_env(self):
        if self.current_environ:
            self.spack_compiler_name = current_environ.extras['spack_compiler_name']
            self.spack_compiler_version = current_environ.extras['spack_compiler_name']

    @run_before('run')
    def replace_launcher(self):
        self.job.launcher = getlauncher('local')()

    @sanity_function
    def validate_compiler(self):
        compiler_str = '{0}@{1}'.format(self.spack_compiler_name, self.spack_compiler_version)
        return sn.assert_found(compiler_str, self.stdout, 'compiler not found: \n{0}\n  * In list:\n{1}')
```

We run `spack compiler list` and query the output for the string of the compiler we are looking for, thus ensuring that the compiler is available before Spack builds any software using the environment compiler.

vkarak:
This is the complete analogous to the `extras` in the definition of the partitions. Users can now define an `extras` attribute in their environment configuration, which allows them to attach arbitrary properties to environments. These are accessible inside a test through the `extras` attribute of the `current_environ`.

Fixes #1973.
Fixes #2230.


reframe/core/systems.py
```diff
@@ -365,13 +365,11 @@ def devices(self):
...
     def extras(self):
-        '''User defined attributes of the system.
-
-        By default, it is an empty dictionary.
+        '''User defined properties defined in the configuration.
 
         .. versionadded:: 3.5.0
 
-        :type: :class:`object`
+        :type: :class:`Dict[str, object]`
```
ekouts, line 372: Is this correct? Now in the schema of the configuration we only check that is a valid object, right? Same for the `extras` of the environments.

teojgo (approved):

ekouts (approved):

reframe/core/systems.py
vkarak: Yes, object properties in JSON are strings, whereas their values can be anything. And JSON objects are translated to Python dicts.

vkarak:
@jenkins-cscs daint

rsarm (approved):

PR #2267, opened 2021-11-13
[feat] Add a utility function to skip tests that have no topology information

vkarak:
This is quite useful with the topology autodetection, since it should be called the the first time that the topology is requested in a test. Otherwise, users would have to write a post-setup hook doing the same thing. I am not adding for the moment an equivalent function for the device autodetection, because I would like to first see the usage scenarios in practice.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -2218,8 +2218,10 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
 
         proc = self.current_partition.processor
         pname = self.current_partition.fullname
-        self.skip_if(not proc.info,
-                     f'no topology information found for partition {pname!r}')
+        if msg is None:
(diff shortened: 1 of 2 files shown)


reframe/core/pipeline.py
```diff
@@ -2203,6 +2203,24 @@ def skip_if(self, cond, msg=None):
...
+
+        .. versionadded:: 3.9.1
+        '''
+        if not self.current_partition:
+            return
+
+        proc = self.current_partition.processor
+        pname = self.current_partition.fullname
+        self.skip_if(not proc.info,
+                     f'no topology information found for partition {pname!r}')
```
teojgo, line 2222:
```suggestion
                     msg or f'no topology information found for partition {pname!r}')
```

teojgo (changes requested):

teojgo (approved):

ekouts (approved):

PR #2272, opened 2021-11-18
[bugfix] Fix incorrect regex for `gpu_bandwidth_check` perf results

akesandgren:
The string tested on is multiline, so must not match stuff with *


teojgo:
Ok to test

victorusu:
I have removed the `import sys` that the unit test was complaining.
After the unit tests pass, `lgtm`

akesandgren:
Ah sorry, leftover from debugging this.

And now with fix for the D2D regex too which I forgot the first time around.

vkarak (approved):

PR #2273, opened 2021-11-19
[feat] Define the execution order of hooks and implement it properly

### LINKED ISSUE #2268: The order of execution of inherited hooks is currently undocumented and probably wrong
In the following example, we do not document in which order the hooks of the `setup` stage will execute:

```python
    @run_after('setup')
    def x(self):
        print('x')

class TestB(TestA):
    @run_after('setup')
    def y(self):
        print('y')
```

The order of execution is `TestB.y`, `TestA.x`, but this is not documented. This is quite important to be documented, because it affects test library design. Also I think that the order of execution is wrong, from a test library's perspective. If `TestA` is a library and sets some variable in a stage hook, I want to be able to use the value of that variable in my derived test.

vkarak:
As indicated in #2268 the order of execution of inherited hooks was not documented and it was also wrong. This PR solves this issue by imposing a reverse MRO execution order, so that the hooks of the most specialized class are executed last. In the part, hooks were executed starting from the derived classes going up to the bases, but without following strict MRO order.

Fixes #2268.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -197,11 +197,15 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
 
     @classmethod
     def pipeline_hooks(cls):
+        # The hook registry stores hooks in MRO order, but we want to attach
+        # the hooks in reverse order so that the more specialized hooks (those
+        # at the beginning of the MRO list) will be executed last
+
         ret = {}
         for hook in cls._rfm_hook_registry:
             for stage in hook.stages:
                 try:
-                    ret[stage].append(hook.fn)
+                    ret[stage].insert(0, hook.fn)
                 except KeyError:
                     ret[stage] = [hook.fn]
 
--- a/docs/regression_test_api.rst
+++ b/docs/regression_test_api.rst
@@ -346,7 +346,7 @@ In essence, these builtins exert control over the test creation, and they allow 
     @rfm.simple_test
     class TestC(rfm.RegressionTest):
         # Parameterize TestC for each ParamFix variant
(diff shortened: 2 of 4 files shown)


reframe/core/pipeline.py
```diff
@@ -197,11 +197,15 @@ def disable_hook(self, hook_name):
...
+        # The hook registry stores hooks in MRO order, but we want to attach
+        # the hooks in reverse order so that the more specialized hooks (those
+        # at the beginning of the MRO list) will be executed last
+
         ret = {}
         for hook in cls._rfm_hook_registry:
             for stage in hook.stages:
                 try:
-                    ret[stage].append(hook.fn)
+                    ret[stage].insert(0, hook.fn)
```
teojgo, line 208: Do you think that you can use a `deque` from `collections` and the `appendleft` method?

teojgo (changes requested):

reframe/core/pipeline.py
vkarak: Good point, the difference is humongous in performance (that's the numbers for 100K inserts):

```
Time for list.insert(0, x) = 2.1753318309783936
Time for dequeue.appendleft(x) = 0.014408111572265625
```

vkarak: The `deque` starts to pay off after 100 hooks, given that at the end we would need to convert the `deque`s to lists in order to return a dictionary of the stages and their lists of hooks. I don't expect to have a test with so many hooks, but rather many tests with few hooks, which would mean that the `deque` solution might eventually be slower for this case.

teojgo (approved):

PR #2281, opened 2021-11-23
[cscs] Remove `PrgEnv-pgi` and `PrgEnv-nvidia` workaround from Dom config

teojgo:


config/cscs.py
```diff
@@ -1129,14 +1126,23 @@
...
         },
+        {
+            'name': 'PrgEnv-nvidia',
+            'target_systems': [
+                'dom'
+            ],
+            'modules': [
+                'PrgEnv-nvidia',
+            ]
+        },
```
vkarak, line 1145: The only reason that these were so sparsely populated was because they were auto-generated from the old config. We should not follow that scheme and become more compact when adding new things or editing existing entries.

vkarak (changes requested):

config/cscs.py
teojgo: Do you mean changing the format to put arrays on one-line

vkarak: Yes, I can do that for this entry.

PR #2282, opened 2021-11-23
[cscs] Make CSCS config more compact

vkarak:
The config was very sparsely written since it was auto-generated from the old config when we transitioned to ReFrame 3.0. I've now made it much more compact by removing redundant new lines.


teojgo (approved):

vkarak:
@jenkins-cscs retry daint

PR #2286, opened 2021-11-24
[test] Update paraview_check for new release (5.9 and 5.10) and for new Mesa libraries

jfavre:
I removed the check on Vendor because it is redundant anyway with the check for llvmpipe


jfavre:
addresses https://jira.cscs.ch/browse/AVAILAPPS-12

vkarak (approved):

vkarak:
@jfavre The test fails on Eiger and Pilatus. Is this expected?

jfavre:
I cannot reproduce the failure of the test on eiger and pilatus, They work fine for me:
reframe -c paraview_check.py -n ParaViewCheck --system pilatus:mc -r

reframe -c paraview_check.py -n ParaViewCheck --system eiger:mc -r

vkarak:
@jfavre It doesn't pass for me in a fresh login. I guess that you have a `ParaView` module already available in your `MODULEPATH`. The test passes if I load `cpeCray` first by doing `-S modules=cpeCray,ParaView`. I think that the `valid_prog_environs` should be `['cpeCray']` on Eiger/Pilatus and not `builtin`. Check the other applications.

cscs-checks/apps/paraview/paraview_check.py
```diff
@@ -9,32 +9,28 @@
...
+    num_tasks = 12
+    num_tasks_per_node = 12
+    modules = ['ParaView']
+    executable = 'pvbatch'
+    executable_opts = ['coloredSphere.py']
+    maintainers = ['JF', 'TM']
+    tags = {'scs', 'production'}
 
-        if self.current_system.name in ['eiger', 'pilatus']:
+    @run_after('setup')
```
vkarak, line 23: Here is the bug. This should be a post-init hook:

```suggestion
    @run_after('init')
```

The reason for that is that after setup is too late for test filtering; the test has already started running (you can use `skip()` there, but it's much simpler if you just change the hook.

vkarak (changes requested):

vkarak (approved):

jfavre (commented):
looks good. thanks for clarifying

PR #2295, opened 2021-11-26
[doc] Use documentation tags for literalinclude

### LINKED ISSUE #2270: Revisit the `literalinclude` directives in the documentation
The `start-at/start`-before and `end-at/end-before` options should be used so that we don't rely on actual line numbers when including an other file in the documentation.

teojgo:
Fixes #2270

### DIFF
--- a/docs/tutorial_basics.rst
+++ b/docs/tutorial_basics.rst
@@ -40,7 +40,8 @@ Here is its C version:
 
 .. literalinclude:: ../tutorials/basics/hello/src/hello.c
    :language: c
-   :lines: 6-
+   :start-after: // rfmdocstart: helloworld
+   :end-before: // rfmdocend: helloworld
 
 
 And here is the ReFrame version of it:
@@ -51,7 +52,8 @@ And here is the ReFrame version of it:
 
 
 .. literalinclude:: ../tutorials/basics/hello/hello1.py
-   :lines: 6-
+   :start-after: # rfmdocstart: hellotest
+   :end-before: # rfmdocend: hellotest
 
 
 Regression tests in ReFrame are specially decorated classes that ultimately derive from :class:`~reframe.core.pipeline.RegressionTest`.
@@ -228,7 +230,8 @@ ReFrame allows you to avoid this in several ways but the most compact is to defi
 
 
 .. literalinclude:: ../tutorials/basics/hello/hello2.py
-   :lines: 6-
+   :start-after: # rfmdocstart: hellomultilang
+   :end-before: # rfmdocend: hellomultilang
 
 
 This test extends the ``hello1.py`` test by defining the ``lang`` parameter with the :py:func:`~reframe.core.pipeline.RegressionMixin.parameter` built-in.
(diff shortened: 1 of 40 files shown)


docs/tutorial_basics.rst
```diff
@@ -252,7 +255,8 @@ This is exactly what we want to do here, and we know that the test sources are n
 Hence, we move the ``sourcepath`` assignment into a pre-compile hook.
 
 .. literalinclude:: ../tutorials/basics/hello/hello2.py
-   :lines: 17-19
+   :start-after: # rdmdocstart: set_sourcepath
+   :end-before: # rdmdocend: set_sourcepath
```
ekouts, line 259:
```suggestion
   :start-after: # rfmdocstart: set_sourcepath
   :end-before: # rfmdocend: set_sourcepath
```
I assume this is by mistake. If not, it's really not important, it can stay as it is.

tutorials/basics/hello/hello2.py
```diff
@@ -14,10 +15,13 @@ class HelloMultiLangTest(rfm.RegressionTest):
     valid_systems = ['*']
     valid_prog_environs = ['*']
 
+    # rdmdocstart: set_sourcepath
     @run_before('compile')
     def set_sourcepath(self):
         self.sourcepath = f'hello.{self.lang}'
+    # rdmdocend: set_sourcepath
```
ekouts, line 22:
```suggestion
    # rfmdocstart: set_sourcepath
    @run_before('compile')
    def set_sourcepath(self):
        self.sourcepath = f'hello.{self.lang}'
    # rfmdocend: set_sourcepath
```
If you change it `docs/tutorial_basics.rst` you should change it here as well.

ekouts (approved):

docs/tutorial_basics.rst
teojgo: Yeah, thanks for noticing. `f` is next to `d` on the keyboard :)

vkarak (changes requested):
Just a couple of minor changes.

teojgo:
@jenkins-cscs retry all

tutorials/basics/stream/stream1.py
```diff
@@ -13,7 +13,8 @@ class StreamTest(rfm.RegressionTest):
     valid_systems = ['*']
     valid_prog_environs = ['gnu']
     prebuild_cmds = [
-        'wget http://www.cs.virginia.edu/stream/FTP/Code/stream.c',
+        'wget https://raw.githubusercontent.com/jeffhammond/STREAM/'
+        'master/stream.c'
```
vkarak, line 17: I would put that in a single line and add a `noqa: E501` annotation, because it's less confusing to read.

vkarak (changes requested):

vkarak (approved):

vkarak:
I will merge this, but perhaps we should consider including the `stream.c` directly in the benchmark's source instead of downloading.

PR #2309, opened 2021-12-01
[bugfix] Fix setting boolean variables from the command line

### LINKED ISSUE #2303: cannot disable exclusive_access from cmd line
I have a test that sets variable `exclusive_access` to True:
```
@rfm.simple_test
class BLASTest(rfm.RunOnlyRegressionTest):
    exclusive_access = True
    ...
```
According to the docs:
>Variable assignments passed from the command line happen before the test is instantiated and is the exact equivalent of assigning a new value to the variable at the end of the test class body. 

However, trying to override this from the cmd line does not work for me, ReFrame still adds the Slurm `--exclusive` flag:
```
reframe --verbose --run --checkpath blas.py --system hydra:skylake --prgenv foss-2020a --setvar BLASTest.exclusive_access=False --setvar num_cpus_per_task=4
```
I've tried setting it to `0`, `False`, `@none`, but nothing seems to work.
Am I doing something wrong? What's the right way to set this to False?
In any case, it does *not* show a warning that the variable is not set, so it's doing something with it.

teojgo:
Fixes #2303 

- [x] Add documentation for the Bool type


reframe/frontend/loader.py
```diff
@@ -135,6 +135,10 @@ def _set_defaults(self, test_registry):
                     # Treat special values
                     if val == '@none':
                         val = None
+                    elif val == '@True':
```
vkarak, line 138:
```suggestion
                    elif val == '@true':
```

reframe/frontend/loader.py
```diff
@@ -135,6 +135,10 @@ def _set_defaults(self, test_registry):
...
+                        val = True
+                    elif val == '@False':
```
vkarak, line 140:
```suggestion
                    elif val == '@false':
```

docs/manpage.rst
```diff
@@ -387,6 +387,7 @@ Options controlling ReFrame execution
    ReFrame will try to convert ``VAL`` to the type of the variable.
    If it does not succeed, a warning will be issued and the variable will not be set.
    ``VAL`` can take the special value ``@none`` to denote that the variable must be set to :obj:`None`.
+   Furthermore, ``VAL`` can take one of the special ``@True`` and ``@False`` values according to which the variable is set to :class:`True` and :class:`False` respectively.
```
vkarak, line 390: Combine all these:

```suggestion
   ``VAL`` can take one of the following special values: 
    - ``@none``: ...
    - ``@true``: ...
    - ``@false``: ...
```

We will also need to point out that if somebody passes any other value to a boolean variable this will be interpreted as true, because ReFrame tries to convert implicitly the given string to boolean, which is a valid operation in Python.

vkarak (changes requested):
I see what was the problem and what you propose here is the only solution. An alternative would be to define a `Bool` `ConvertibleType` with a custom conversion function and change all fields with type `bool` to `Bool`, which we wanted to avoid with `None`s in any case, so there is no meaning in doing that.

docs/manpage.rst
```diff
@@ -442,10 +448,13 @@ Options controlling ReFrame execution
         class my_test(rfm.RegressionTest):
             num_tasks = required
 
+   .. note::
+      Trying to set a boolean variable using a value different from ``@true`` or ``@false`` will be interpreted as :class:`True` since ReFrame will try to implicitly convert the given string to a boolean, which is a valid Python operation.
```
vkarak, line 452: Do we need this note twice?

docs/manpage.rst
```diff
@@ -386,8 +386,14 @@ Options controlling ReFrame execution
...
    If it does not succeed, a warning will be issued and the variable will not be set.
-   ``VAL`` can take the special value ``@none`` to denote that the variable must be set to :obj:`None`.
-   Furthermore, ``VAL`` can take one of the special ``@True`` and ``@False`` values according to which the variable is set to :class:`True` and :class:`False` respectively.
+   ``VAL`` can take one of the following special values:
+
+   - ``@none``: sets the variable to :obj:`None`
+   - ``@true``: sets the variable to :class:`True`
+   - ``@false``: sets the variable to :class:`False`
+
+   .. note::
```
vkarak, line 395: Perhaps make it a warning?

```suggestion
   .. warning::
```

vkarak (commented):
Just a couple of very minor comments.

victorusu (approved):
`lgtm`

vkarak (changes requested):
I think that @smoors is right in his comment in #2303. Supporting conversions to bool it's so natural and not giving an error when an obvious conversion is not supported is error-prone. Therefore, I suggest dropping the `@true` and `@false` special values and implement a `BoolType` that will support the conversions that we want. `@none` is not the same, because (a) there is no implicit conversion to `None` and (b) it's a special value anyway.

victorusu:
@vkarak, then the solution is this one? https://github.com/eth-cscs/reframe/pull/2309#pullrequestreview-822559350

vkarak:
> @vkarak, then the solution is this one? [#2309 (review)](https://github.com/eth-cscs/reframe/pull/2309#pullrequestreview-822559350)

Exactly! There is an exact example in the unit tests on how to do this. And in fact this is what we do to convert the aggregate types.

reframe/utility/typecheck.py
```diff
@@ -342,6 +342,35 @@ def __rfm_cast_str__(cls, s):
         return s
 
 
+class Bool(metaclass=_BuiltinType):
+    _type = bool
+
+    def __init__(self, val):
+        self._val = val
```
vkarak, line 349:
```suggestion
        self.__val = val
```

reframe/utility/typecheck.py
```diff
@@ -342,6 +342,35 @@ def __rfm_cast_str__(cls, s):
         return s
 
 
+class Bool(metaclass=_BuiltinType):
+    _type = bool
```
vkarak, line 346: Restrict further the type:

```suggestion
    _type = bool
    __slots__ = ('__val',)
```

unittests/resources/checks_unlisted/externalvars.py
```diff
@@ -8,7 +8,7 @@ class external_x(rfm.RunOnlyRegressionTest):
     valid_systems = ['*']
     valid_prog_environs = ['*']
     foo = variable(int, value=1)
-    ham = variable(bool, value=False)
+    ham = variable(Bool, value=False)
```
vkarak, line 11: How did this work?

```suggestion
    ham = variable(typ.Bool, value=False)
```

unittests/test_typecheck.py
```diff
@@ -18,6 +18,30 @@ def assert_type_hierarchy(builtin_type, ctype):
     assert not issubclass(ctype[int], ctype[ctype[int]])
 
 
+def test_bool_type():
+    t = True
+    f = False
+    assert isinstance(t, types.Bool)
+    assert isinstance(f, types.Bool)
```
vkarak, line 25:
```suggestion
    assert isinstance(True, types.Bool)
    assert isinstance(False, types.Bool)
```

unittests/test_typecheck.py
```diff
@@ -18,6 +18,30 @@ def assert_type_hierarchy(builtin_type, ctype):
...
+    with pytest.raises(TypeError):
+        types.Bool('foo')
+
+    # Test for boolean conversion
+    assert types.Bool('True')
+    assert types.Bool('YEs')
+    assert types.Bool('1')
+    assert not types.Bool('False')
+    assert not types.Bool('No')
+    assert not types.Bool('0')
```
vkarak, line 38: You need also to check initialisation from a `bool`.

unittests/test_typecheck.py
```diff
@@ -18,6 +18,30 @@ def assert_type_hierarchy(builtin_type, ctype):
...
+
+    # Test hash functions
+    assert hash(types.Bool('True')) == hash(types.Bool('YES'))
+    assert hash(types.Bool('False')) == hash(types.Bool('No'))
```
vkarak, line 42: You don't check for equality properties and you need to check that `Bool` and `bool` types can be used interchangeably:

```python
assert hash(types.Bool(1)) == hash(True)
assert hash(types.Bool(0)) == hash(False)
assert types.Bool(1) == True
assert True == types.Bool(1)
assert types.Bool(0) == False
assert False == types.Bool(0)
```

reframe/utility/typecheck.py
```diff
@@ -342,6 +342,35 @@ def __rfm_cast_str__(cls, s):
         return s
 
 
+class Bool(metaclass=_BuiltinType):
```
vkarak, line 345: We need to document this properly for the users.

vkarak (changes requested):
The implementation looks fine to me. I have some requests for fine tuning and adding documentation.

reframe/utility/typecheck.py
teojgo: Do you think I should document also for the use with external variables?

vkarak: Yes, we should mention that if you want to define a boolean variable that can be properly set from the command line, you should define it as `Bool`.

unittests/resources/checks_unlisted/externalvars.py
teojgo: I did not work actually, I need to add one more assertion to make sure that the tests are actually loaded without any problem with the declaration of the variables.

unittests/test_typecheck.py
```diff
@@ -33,13 +31,23 @@ def test_bool_type():
...
     assert hash(types.Bool('True')) == hash(types.Bool('YES'))
     assert hash(types.Bool('False')) == hash(types.Bool('No'))
+    assert hash(types.Bool(1)) == hash(True)
+    assert hash(types.Bool(0)) == hash(False)
+
+    # Test for equality
+    assert types.Bool(1) == True    # noqa: E712
+    assert True == types.Bool(1)    # noqa: E712
+    assert types.Bool(0) == False   # noqa: E712
+    assert False == types.Bool(0)   # noqa: E712
```
vkarak, line 50: Hmm, PEP8 proposes to write such comparisons as `if x is True` or `if x`. The first will not work for bool types, whereas the second one will. I don't know how to treat this actually, except than suggesting to use the second form.

teojgo: But isn't this already used in the tests for boolean conversion?

vkarak:
I changed a bit the implementation of `Bool`. We should not keep the value, because this is supposed to be the type and used for type checking and any implicit conversions. So it simply takes care of the conversion and returns a normal `bool`. Since the builtin `bool` type is registered as a subclass of `Bool`, any assignment of a builtin `bool` to a `Bool` variable will be accepted. Assignments from string will use the implicit conversion that we defined and all others will fail, as they would in any case on a `bool` variable. Also, for consistency with `@none` which is case sensitive, I made the conversion also case sensitive, accepting only lowercase. Finally, I updated all the boolean `RegressionTest` variables to use the `Bool` type.

vkarak (approved):

PR #2314, opened 2021-12-03
[feat] Add a command line option that allows decreasing the verbosity level of the output

### LINKED ISSUE #2283: Add a "quiet" option that would decrease the verbosity level
Currently, you can only increase the verbosity level from the command line.

rsarm:
Closes #2283


victorusu:
@rsarm, `lgtm`. I wonder if it makes sense to add an additional test that has move `q`s than `-v`, in order to check if it stays at `0` level...

vkarak (changes requested):
I'm not sure that this is exactly what we want. If you pass `-q` without having set the verbosity level elsewhere the output does not become quieter. I think the following should happen:

1. Allow `genera/verbose` to accept negative values. This automatically makes it equivalent to passing the `-q` option, in practice.
2. Rename `PrettyPrinter.inc_verbosity()` to `PrettyPrinter.adjust_verbosity()` and let it accept negative values and properly adjust decrease the logging level (I don't think that anything particular should change in that function, but just make sure it works with negative values).
3. The `options.quiet` must be subtracted from whatever you get from the `general/0/verbose` and pass the result to `adjust_verbosity()`.

vkarak (approved):
I fixed the PR and added documentation.

PR #2316, opened 2021-12-06
[testlib] Append to `executable_opts` in Gromacs library test

victorusu:
The `executable_opts` passed by the tests are overwritten by the library-defined ones. e.g.
For the CSCS checks, instead of generating these command-line options for GROMACs
```
gmx_mpi mdrun -dlb yes -ntomp 1 -npme -1 -nb cpu -s benchmark.tpr   # correct set
```
one gets
```
gmx_mpi mdrun -nb cpu -s benchmark.tpr  # wrong set
```

This PR fixes this issue.


vkarak (approved):
I'm not sure that this is a bug of the library or of the test that uses the library. Generally, we should answer the question: what assumptions should the library make and what the test?

PR #2329, opened 2021-12-07
[bugfix] Use conformant job names for Torque backend

### LINKED ISSUE #2275: torque failure
Trying to run the tutorial hello world job on torque results in an error:
```
Specified job names must begin with a letter (""rfm_HelloTest_job"" does not)!
Specified job names must be less than 16 characters (""rfm_HelloTest_job"" is 19)!
```
Specifically, that first error (the second error can be circumvented by renaming the class HelloTest to HT). I have attached my config file, error text, and log files.
[settings_and_error_text.tar.gz](https://github.com/eth-cscs/reframe/files/7572715/settings_and_error_text.tar.gz)

rsarm:
closes #2275

### DIFF
--- a/reframe/core/schedulers/pbs.py
+++ b/reframe/core/schedulers/pbs.py
@@ -114,7 +114,7 @@ class PbsJobScheduler(sched.JobScheduler):
 
     def emit_preamble(self, job):
         preamble = [
-            self._format_option('-N "%s"' % job.name),
+            self._format_option('-N %s' % job.name),
             self._format_option('-o %s' % job.stdout),
             self._format_option('-e %s' % job.stderr),
         ]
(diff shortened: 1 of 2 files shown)


rsarm:
From the documentation of `qsub`:
```
-N name   Define the name of the batch job.

          The  qsub −N option shall accept a value for the name option-argument that is a string of up to 15 alphanumeric characters
          in the portable character set (see the Base Definitions volume of POSIX.1‐2008, Section 6.1, Portable Character Set) where
          the first character is alphabetic.
```

vkarak (changes requested):
Thanks @rsarm! I would suggest to also truncate the name to the size required by the man page.

vkarak (approved):

PR #2343, opened 2021-12-14
gpu_burn.cu: Change code to time only the compute part. WIP

akesandgren:
Accumulate total time for finished reps,
this results in a more correct Performance value.
Read device temp values after each repetition and save maxtemp.
Change warmup to also run the compare phase, time them and print the time taken.
Duration must be larger than that to get performance data, also warn if the
duration was too short to finish a single repetition.


teojgo:
ok to test

akesandgren:
Hmmm, something might be fishy with this change. I get ridiculous performance numbers on our K80's.

akesandgren:
This is the wrong approach, closing

PR #2345, opened 2021-12-14
[testlib] Time the warmup phase and print minimum required duration in the GPU burn test

akesandgren:
- Calculates the time it takes to run compare+getErrors and removes that
   during Performance calculation.
 - Read device temp values after each repetition and save maxtemp.
 - Error out if duration was too short and it didn't complete a single
   repetition.
 - Use high_resolution_clock instead of system_clock


akesandgren:
This is a better rewrite than my previous attempt.

victorusu:
Ok to test

vkarak (approved):

vkarak:
@jenkins-cscs retry daint

PR #2349, opened 2021-12-15
[test] Update syntax of CPMD check

rsarm:


cscs-checks/apps/cpmd/cpmd_check.py
```diff
@@ -7,47 +7,36 @@
...
-            'stdout.txt', 'result', float)
-        energy_reference = 25.81
-        energy_diff = sn.abs(energy - energy_reference)
-        self.sanity_patterns = sn.assert_lt(energy_diff, 0.26)
-        self.perf_patterns = {
-            'time': sn.extractsingle(r'^ cpmd(\s+[\d\.]+){3}\s+(?P<perf>\S+)',
-                                     'stdout.txt', 'perf', float)
-        }
-        if scale == 'small':
+        if self.scale == 'small':
```
teojgo, line 39: Include the references in the first if block

teojgo (changes requested):

teojgo (approved):

cscs-checks/apps/cpmd/cpmd_check.py
```diff
@@ -7,47 +7,32 @@
...
-            'stdout.txt', 'result', float)
-        energy_reference = 25.81
-        energy_diff = sn.abs(energy - energy_reference)
-        self.sanity_patterns = sn.assert_lt(energy_diff, 0.26)
-        self.perf_patterns = {
-            'time': sn.extractsingle(r'^ cpmd(\s+[\d\.]+){3}\s+(?P<perf>\S+)',
-                                     'stdout.txt', 'perf', float)
-        }
-        if scale == 'small':
             self.reference = {
```
vkarak, line 36: I would rather add a reference dictionary at the class level with all the references and select here the valid set of references based on the variant. See how we've done it in other tests with `allref`.

vkarak (changes requested):
I have a minor comment. In the future, as soon as we have the new naming scheme we should consider replace the `scale` parameter with a `num_node`.

vkarak (approved):

PR #2350, opened 2021-12-15
[test] Update syntax of CP2K check

rsarm:


cscs-checks/apps/cp2k/cp2k_check.py
```diff
@@ -132,27 +134,28 @@ def set_cpu_binding(self):
...
         else:
             self.num_tasks = 96
 
-        self.num_tasks_per_node = 6
-        self.num_cpus_per_task = 2
-        self.variables = {
-            'CRAY_CUDA_MPS': '1',
-            'OMP_NUM_THREADS': str(self.num_cpus_per_task)
-        }
         references = {
```
teojgo, line 159: Put this in the class body

teojgo (changes requested):

teojgo (approved):

vkarak (approved):
Generally, the test needs further refinement, such as making the CPU and GPU variants a parameter, rather than use two different classes. I'm approving so as to eliminate the deprecation warnings at least.

PR #2351, opened 2021-12-16
[test ] Update syntax of VASP check

rsarm:


ekouts (approved):

vkarak (approved):
Lgtm. In the future, we should consider using a single test (without inheritance) and an additional parameter.

PR #2372, opened 2022-01-06
[testlib] Add support for build both GPU dgemm+sgemm executables

ebirn:
Motivation: We have a bunch of RTX6000 devices at our site, which have a large difference in float vs double performance. For others like V100, P100 the TF/s difference is roughly 2x. 

The changes to get the sgemm executable compiled:
- template the dgemm gpu test code to allow execution of both Dgemm and Sgemm.
- the makefiles are modified to define the codepaths through defines (float or double) and produce 2 executables

Note: I was not able to test the makefile.hip , so at this point I'm just assuming it will work

Question: if this is to be accepted, should the 'dgemm' directory be renamed, along with the Dgemm base test?


vkarak:
ok to test

vkarak (approved):
I will merge this PR as is. For the moment, I would wait for adapting the test to build both dgemm and sgemm because this tree is a bit old and uses some early practices of writing library tests. We should generally revamp the corresponding reframe test.

PR #2375, opened 2022-01-11
[test] Update CSCS ipcmagic test (JupyterLab)

lucamar:
I had to add the module `JuliaExtensions` (or `Julia`) explicitly, although it is a dependency of `jupyterlab`, otherwise I was getting the error message `Tcl command execution failed` when running the command `reframe -n cscs_ipcmagic_check -t prod -r` on Dom (module `jupyterlab` built in my sandbox for the time being).

### DIFF
--- a/cscs-checks/apps/jupyter/check_ipcmagic.py
+++ b/cscs-checks/apps/jupyter/check_ipcmagic.py
@@ -12,13 +12,12 @@ from hpctestlib.interactive.jupyter.ipcmagic import ipcmagic_check
 class cscs_ipcmagic_check(ipcmagic_check):
     valid_systems = ['daint:gpu', 'dom:gpu']
     valid_prog_environs = ['builtin']
+    modules = ['Horovod', 'jupyterlab']
 
     @run_after('init')
-    def set_modules(self):
+    def modules_workaround(self):
         if self.current_system.name == 'dom':
             self.modules = ['Horovod', 'JuliaExtensions', 'jupyterlab']
-        else:
-            self.modules = ['Horovod', 'jupyterlab']
 
     maintainers = ['RS', 'TR']
     tags = {'production'}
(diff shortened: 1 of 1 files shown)


cscs-checks/apps/jupyter/check_ipcmagic.py
```diff
@@ -12,7 +12,14 @@
 class cscs_ipcmagic_check(ipcmagic_check):
     valid_systems = ['daint:gpu', 'dom:gpu']
     valid_prog_environs = ['builtin']
-    modules = ['jupyterlab', 'Horovod']
+
+    @run_after('init')
+    def set_modules(self):
+        if self.current_system.name == 'dom':
+            self.modules = ['Horovod', 'JuliaExtensions', 'jupyterlab']
```
vkarak, line 19: Is this a workaround or is this how it is supposed to work?

lucamar: It is a workaround for Dom, where ReFrame cannot load the module`jupyterlab` without loading `Julia` or `JuliaExtensions` first: see for instance [the results of last daily run on Dom](https://jenkins.cscs.ch/blue/organizations/jenkins/reframe-dom-production-daily/detail/reframe-dom-production-daily/1475/pipeline) with the error reported by `cscs_ipcmagic_check`:
```
  * Reason: environ error: could not execute module operation: command 'modulecmd python show jupyterlab' failed with exit code 0:
```
This does not happen when you load the module manually (I reported a similar issue with ReFrame for `LAMMPS` in #2348).

vkarak: I would then suggest to write this part a bit differently to mark that it is a workaround:

```python
@run_after('init')
def modules_workaround(self):
        self.modules = ['Horovod', 'JuliaExtensions', 'jupyterlab']
```

And in the class body leave `modules = ['...']` as it is supposed to work.

lucamar: Thanks @vkarak, I have implemented your suggestion

lucamar:
@jenkins-cscs retry dom

vkarak (approved):

PR #2376, opened 2022-01-12
[bugfix] Make Slurm polling aware of Heterogeneous Jobs

teojgo:


reframe/core/schedulers/slurm.py
```diff
@@ -90,12 +90,16 @@ class SlurmJobScheduler(sched.JobScheduler):
     # standard job state polling using sacct.
     SACCT_SQUEUE_RATIO = 10
 
-    # This matches the format for both normal jobs as well as job arrays.
+    # This matches the format for both normal and heterogeneous jobs,
+    # as well as job arrays.
+    # For heterogeneous jobs, the job_id has the following format:
+    # <het_job_id>+<het_job_offset>
+    # See (`Heterogeneous Job Support<https://slurm.schedmd.com/heterogeneous_jobs.html>`__)  # noqa: E501
```
vkarak, line 97: No need for rst comments inside the code; just put the link.

vkarak (commented):
Is there a way we can add a unit test for this?

teojgo:
> Is there a way we can add a unit test for this?

Probably yes, like we do with https://github.com/eth-cscs/reframe/blob/f8dbad6e76ae9975473b99f888e98a52cbb43644/unittests/test_schedulers.py#L347

vkarak (approved):

PR #2377, opened 2022-01-12
[test] Set time limit for tests on Alps

lucamar:
Set a time limit of 10 minutes for all checks on Alps.

### DIFF
--- a/config/cscs.py
+++ b/config/cscs.py
@@ -560,6 +560,7 @@ site_configuration = {
                 {
                     'name': 'login',
                     'scheduler': 'local',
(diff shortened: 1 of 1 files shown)


vkarak (commented):
If we are sure that this does not make tests to fail, I'm fine with this.

lucamar:
Thanks @vkarak: the whole production test suite takes 16 minutes on Pilatus, therefore we should be safe with 10 minutes time limit for individual checks. This limit is already enforced on Dom and Piz Daint and it increases the chance to run sooner.

vkarak (approved):

PR #2383, opened 2022-01-17
[bugfix] Fix and enhance LSF support

### LINKED ISSUE #2389: Add Support for "Standard" Spectrum LSF Scheduler
The following issue was about introducing support for Spectrum LSF as a scheduler into ReFrame:

https://github.com/eth-cscs/reframe/issues/1055

but it was actually focused on a Spectrum LSF variant which is rather specific to LLNL Sierra system, where LSF is used together with IBM Cluster System Management (CSM).

On regular HPC platforms, Spectrum LSF is not used in conjunction with CSM, which makes the current implementation inadequate as the submission command / several options are different.

Would it be possible to add support for "standard" Spectrum LSF into ReFrame?

MarkLTurner:
Fixes:

-  LSF jobs need to be submitted with a system call. ```BSUB``` takes the content of a file, not the name.

Features:

- Add some important missing options to the LSF scheduler (e.g. access params, exclusivity, num cores per node etc.) that otherwise could only be forced into the emitted job script through the ```-J``` option to the ReFrame executable.
- Follow the structure in slurm.py wherein preamble options are optional.

Fixes #2389.


reframe/core/schedulers/lsf.py
```diff
@@ -30,45 +32,57 @@ def __init__(self):
...
 
         # add job time limit in minutes
         if job.time_limit is not None:
             preamble.append(
-                self._format_option(f'-W {int(job.time_limit // 60)}')
+                f'{self._prefix} -W {int(job.time_limit // 60)}'
             )
 
+        for opt in job.sched_access:
+            preamble.append('%s %s' % (self._prefix, opt))
```
vkarak, line 62:
```suggestion
            preamble.append(f'{self._prefix} {opt}')
```

reframe/core/schedulers/lsf.py
```diff
@@ -30,45 +32,57 @@ def __init__(self):
...
 
     def submit(self, job):
-        cmd = f'bsub {job.script_filename}'
-        completed = _run_strict(cmd, timeout=self._submit_timeout)
+        with open(job.script_filename, 'r') as f:
+            completed = subprocess.run(
+                args='bsub', stdin=f, capture_output=True)
         jobid_match = re.search(r'^Job <(?P<jobid>\S+)> is submitted',
-                                completed.stdout)
+                                completed.stdout.decode('utf-8'))
```
vkarak, line 82: Why do you need to decode the string here?

unittests/test_schedulers.py
```diff
@@ -136,13 +136,17 @@ def assert_job_script_sanity(job):
...
+        num_tasks_per_node * job.num_cpus_per_task)
     return set([
         f'#BSUB -J testjob',
         f'#BSUB -o {job.stdout}',
         f'#BSUB -e {job.stderr}',
-        f'#BSUB -nnodes {num_nodes}',
+        f'#BSUB -n {num_tasks}',
         f'#BSUB -W {int(job.time_limit // 60)}',
+        f'#BSUB -R "span[ptile={ptile}]"',
+        '#BSUB -x',
```
vkarak, line 149: Just for symmetry:

```suggestion
        f'#BSUB -x',
```

unittests/test_schedulers.py
```diff
@@ -136,13 +136,17 @@ def assert_job_script_sanity(job):
 def _expected_lsf_directives(job):
     num_tasks = job.num_tasks or 1
     num_tasks_per_node = job.num_tasks_per_node or 1
-    num_nodes = int(num_tasks // num_tasks_per_node)
+    ptile = min(
+        num_tasks * job.num_cpus_per_task,
+        num_tasks_per_node * job.num_cpus_per_task)
```
vkarak, line 141:
```suggestion
    ptile = min(
        num_tasks * job.num_cpus_per_task,
        num_tasks_per_node * job.num_cpus_per_task
    )
```

vkarak (changes requested):
Thanks @MarkLTurner for the PR and the enhancements! Afaics the key difference with the current version submitted by @rngoodner and your version here is that `bsub` needs to read the batch script from stdin instead of getting it as a file and what you implement here seems to be the standard LSF behaviour. If this PR breaks with LLNL's variant of LSF, we should probably consider separating the the non-standard variant registering it with a different name. @rngoodner would you have time to try this PR and check if it breaks for you?

reframe/core/schedulers/lsf.py
MarkLTurner: Without calling decode my jobs crash immediately with the following error:

> Traceback (most recent call last):
  File "/lustre/scafellpike/local/HT03340/hxt01/mxt16-hxt01/reframe/reframe/frontend/executors/__init__.py", line 267, in _safe_call
    return fn(*args, **kwargs)
  File "/lustre/scafellpike/local/HT03340/hxt01/mxt16-hxt01/reframe/reframe/core/hooks.py", line 80, in _fn
    func(obj, *args, **kwargs)
  File "/lustre/scafellpike/local/HT03340/hxt01/mxt16-hxt01/reframe/reframe/core/pipeline.py", line 2292, in run
    super().run()
  File "/lustre/scafellpike/local/HT03340/hxt01/mxt16-hxt01/reframe/reframe/core/pipeline.py", line 1680, in run
    self._job.submit()
  File "/lustre/scafellpike/local/HT03340/hxt01/mxt16-hxt01/reframe/reframe/core/schedulers/__init__.py", line 425, in submit
    return self.scheduler.submit(self)
  File "/lustre/scafellpike/local/HT03340/hxt01/mxt16-hxt01/reframe/reframe/core/schedulers/lsf.py", line 81, in submit
    jobid_match = re.search(r'^Job <(?P<jobid>\S+)> is submitted',
  File "/lustre/scafellpike/local/apps/python3/3.8.0/lib/python3.8/re.py", line 199, in search
    return _compile(pattern, flags).search(string)
TypeError: cannot use a string pattern on a bytes-like object

MarkLTurner:
Hi @vkarak, thank you for the review (and the great project!). I wasn't aware of the LLNL variant - as you say the existing code was probably designed around that. I am more than happy to split things up as you see fit.

reframe/core/schedulers/lsf.py
```diff
@@ -30,45 +32,57 @@ def __init__(self):
...
+            preamble.append(f'{self._prefix} -x')
 
         return preamble
 
     def submit(self, job):
-        cmd = f'bsub {job.script_filename}'
-        completed = _run_strict(cmd, timeout=self._submit_timeout)
+        with open(job.script_filename, 'r') as f:
+            completed = subprocess.run(
+                args='bsub', stdin=f, capture_output=True)
```
vkarak, line 80: I've done a PR to your fork, so that you can write this as follows and avoid the `decode()` later on:

```suggestion
        with open(job.script_filename, 'r') as fp:
            completed = _run_strict('bsub', stdin=fp)
```

vkarak (changes requested):
I've done a small suggestion so as to avoid the `decode()`.

rngoodner:
> Thanks @MarkLTurner for the PR and the enhancements! Afaics the key difference with the current version submitted by @rngoodner and your version here is that `bsub` needs to read the batch script from stdin instead of getting it as a file and what you implement here seems to be the standard LSF behaviour. If this PR breaks with LLNL's variant of LSF, we should probably consider separating the the non-standard variant registering it with a different name. @rngoodner would you have time to try this PR and check if it breaks for you?

I haven't looked into what this branch is doing, but a quick run with my current tests fails on this branch with:

```
...
 * Reason: type error: unsupported operand type(s) for *: 'int' and 'NoneType'
...
 File "reframe/reframe/core/schedulers/lsf.py", line 49, in emit_preamble
    min(job.num_tasks * job.num_cpus_per_task,
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

vkarak:
Thanks @rngoodner for trying it out. I think that the PR is currently more fundamentally broken. Also the unit tests are failing with the same error even without really submitting a job. @MarkLTurner could you have a look and fix the PR so that the unit tests pass?

vkarak (commented):
Lgtm, just a couple of style comments only, which I can fix. @rngoodner could you check now if that works in your case?

MarkLTurner:
The previous version of lsf.py allowed `-nnodes` to be emitted in the preamble. It's possible that @rngoodner needs this in the code still and maybe I shouldn't have removed it.

The reason why I didn't is that I see errors on some systems with this option:

> When LSB_CSM_JOBS is not set to Y and -csm is not set then other csm options are not allowed. Job not submitted.

Not sure I fully follow this, but there is a mention of this issue here: https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=options-nnodes

vkarak:
Hi @MarkLTurner, I was waiting for @rngoodner in case this was breaking on his system. From the various documentations that I checked, what you implement here seems to be the standard behaviour, so I plan to merge it. One little thing still, that I just discovered, is that we need an entry in documentation [here](https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#systems-.partitions-.scheduler) about the LSF scheduler support.

vkarak:
I also think that even if the `bsub` command as it is now does ot work for @rngoodner, it should be easy to just extend this standard LSF scheduler and adapt to that specific behaviour.

rngoodner:
My current tests fail with this branch:

```
--- stderr ---
You cannot specify -R/-M/-n/LSB_DEFAULT_RESREQ when CSM Easy Mode job submission is enabled. Job not submitted.
```

I'm not entirely sure what this means off-hand...

rngoodner:
I dug into the generated job submission scripts and it looks like the main difference is the use of `-n` instead of `-nnodes`.

new (not working):

```
#!/bin/bash
#BSUB -J rfm_FiestaTest_job
#BSUB -o rfm_FiestaTest_job.out
#BSUB -e rfm_FiestaTest_job.err
#BSUB -n 4
#BSUB -W 15
#BSUB -q pbatch
module load gcc/8.3.1
module load lrun/2020.03.05
module load spectrum-mpi/rolling-release
module load cuda/11.1.1
module load cmake/3.14.5
module load hdf5-parallel/1.10.4
lrun -N 4 -T 1 -M "-gpu" ./build/fiesta /usr/WS1/goodner2/rfm_fiesta/3D_Expansion_small_gpu-aware/fiesta.lua --kokkos-num-devices=1
```


old (working):

```
#!/bin/bash
#BSUB -J rfm_FiestaTest_job
#BSUB -o rfm_FiestaTest_job.out
#BSUB -e rfm_FiestaTest_job.err
#BSUB -nnodes 4
#BSUB -W 15
#BSUB -q pbatch
cd /usr/WS1/goodner2/rfm_fiesta/stage/lassen/default/gnu/FiestaTest
module load gcc/8.3.1
module load lrun/2020.03.05
module load spectrum-mpi/rolling-release
module load cuda/11.1.1
module load cmake/3.14.5
module load hdf5-parallel/1.10.4
lrun -N 4 -T 1 -M "-gpu" ./build/fiesta /usr/WS1/goodner2/rfm_fiesta/3D_Expansion_small_gpu-aware/fiesta.lua --kokkos-num-devices=1
```

vkarak:
I think we can easily control that through scheduler-specific configuration parameter as we do with some settings from the Slurm backend. I can update the PR accordingly. Another question I have to @rngoodner is the following: the current LSF backend submits jobs with `bsub < job_script` whereas yours was doing `bsub job_script`. Is this a problem? If yes, we could add another configuration variable to control also this.

rngoodner:
I like this more general approach. My original implementation was very minimal. It looks like my tests still fail though. I am getting `-nnodes` now, but the conflicting `-n` is also being put into the job script.

```
#!/bin/bash
#BSUB -J rfm_FiestaTest_job
#BSUB -o rfm_FiestaTest_job.out
#BSUB -e rfm_FiestaTest_job.err
#BSUB -n 4
#BSUB -nnodes 4
#BSUB -W 15
#BSUB -q pbatch
module load gcc/8.3.1
module load lrun/2020.03.05
module load spectrum-mpi/rolling-release
module load cuda/11.1.1
module load cmake/3.14.5
module load hdf5-parallel/1.10.4
lrun -N 4 -T 1 -M "-gpu" ./build/fiesta /usr/WS1/goodner2/rfm_fiesta/3D_Expansion_small_gpu-aware/fiesta.lua --kokkos-num-devices=1
```

Here is some of my test, which seems to be similar to your test case:

```
import reframe as rfm
import os

LAUNCHDIR = os.getcwd()

@rfm.simple_test
class FiestaTest(rfm.RegressionTest):
    valid_systems = ['lassen','xena']
    valid_prog_environs = ['gnu']
    sourcesdir = '../fiesta/'
    sourcepath = '.'
    executable = './build/fiesta'
    executable_opts = [f'{LAUNCHDIR}/3D_Expansion_small_gpu-aware/fiesta.lua', '--kokkos-num-devices=1']
    build_system = 'CMake'
    num_tasks = 4
    num_tasks_per_node = 1
    time_limit = '0d0h15m0s'
    reference = { TRUNCATED HERE
```

vkarak:
My two cents here. By default ReFrame sets all scheduler-specific test fields to `None` except `num_tasks` which is set to 1. The reason is for that is to allow backends to interpret `None` as "not specified by the user". We use that in the Slurm backend and simply do not emit anything, because we have seen strange behaviours in the past. What I could suggest for the LSF backend is to take advantage of that trick: 

1. If in my test I specify only `num_tasks`, it means that I don't care about the nodes, so `-n` should be the natural thing to emit. This is also compatible with the Slurm backend, which sets only the num tasks requirement and nothing else.
2. If I also specify `num_tasks_per_node`, it means that i do care about the nodes, so `-nnodes` should be emitted instead of `-n`.
3. If I specify the `num_cpus_per_task`, then the `ptile` directive should be emitted.

I believe this covers both cases.

MarkLTurner:
Thank you both for the helpful feedback. I think this implements the strategy above and would be surprised if the `-n` option was still emitted in your jobs @rngoodner.

vkarak (changes requested):
Just to be complete, I would add a couple of unit tests more that will check that the correct options are emitted in the job script. I could do that for you.

rngoodner:
I just pulled the last batch of changes and my tests pass now =)

MarkLTurner:
Hi @vkarak, that would be great. 

Thank you both for your patience with this.

MarkLTurner:
Hmm, there is still a problem with this when `ptile` is greater than the number of slots per node. So if the test specifies,

```
self.num_tasks = 64
self.num_cpus_per_task = 1
```

This yields:

```
#BSUB -n 64
#BSUB -R "span[ptile=64]"
```
Meaning each slot is overloaded which is not the intended behaviour.

My original implementation didn't do this by making use of `job.num_tasks_per_node` as follows:

```
num_physical_cores = min(
    job.num_tasks * job.num_cpus_per_task,
    job.num_tasks_per_node * job.num_cpus_per_task
)
```

This is no longer possible as we use `num_tasks_per_node` to emit `-nnodes` for Ryan's use case. Is there another way to allow users of ReFrame users to specify the number of cores they want on each node?

reframe/core/schedulers/lsf.py
```diff
@@ -30,45 +31,60 @@ def __init__(self):
...
         ]
 
+        if job.num_tasks_per_node is not None:
+            num_nodes = job.num_tasks // job.num_tasks_per_node
+            preamble.append(self._format_option(num_nodes, '-nnodes {0}'))
+        else:
+            preamble.append(self._format_option(job.num_tasks, '-n {0}'))
+
+        if job.num_cpus_per_task is not None:
+            num_physical_cores = job.num_tasks * job.num_cpus_per_task
```
vkarak, line 54: Following up on my other comment at the PR, I think that you cannot make an assumption about the physical cores here. I would just use the `job.num_cpus_per_task` here directly as an argument to `ptile`.

Trying to fix the unit tests I came across with the oversubscription problem that you described. The `fake_job` in unit tests defines this:

```python
    ret.num_tasks = 16
    ret.num_tasks_per_node = 2
    ret.num_tasks_per_core = 1
    ret.num_tasks_per_socket = 1
    ret.num_cpus_per_task = 18
```

And that assumed 18*16 = 288 tasks, which as a writer of a test setting the `num_cpus_per_task = 18` i wouldn't expect.

vkarak:
@MarkLTurner I have just made a PR to your branch that removes the assumption about physical cores and adds more unit tests. After that, I think, that we can merge this PR.

vkarak (approved):

PR #2384, opened 2022-01-17
[test] Modernize QuantumEspresso test

teojgo:


lucamar (approved):
Thanks @teojgo

cscs-checks/apps/quantumespresso/quantumespresso_check.py
```diff
@@ -10,47 +10,57 @@
...
-        self.executable = 'pw.x'
-        self.executable_opts = ['-in', 'ausurf.in',
-                                '-pd', '.true.']
+        self.tags |= {
+            'maintenance' if self.variant == 'maint' else 'production'
+        }
 
-        self.sanity_patterns = sn.all([
+    @sanity_function
+    def assert_simulation_successful(self):
```
vkarak, line 37:
```suggestion
    def assert_simulation_success(self):
```

vkarak (commented):
Would it make sense to parametrize it on the `cpu` vs `gpu` variants as well instead of having to additional classes?

teojgo:
> Would it make sense to parametrize it on the `cpu` vs `gpu` variants as well instead of having to additional classes?

It would need more if blocks, that's why I left it as is, since it already contains some.

vkarak (approved):

Issue #2336, opened 2021-12-10
using a test file named zlib.py never runs the test


ebryerwork:
If the test file name is 'zlib.py', the test will not run; in the log will appear 
```
[2021-12-10T10:39:08] verbose: reframe: Loaded 0 test(s)
[2021-12-10T10:39:08] verbose: reframe: Generated 0 test case(s)
```
If the test file is named anything else, it runs. I run ReFrame with the command `~/reframe/bin/reframe --run --purge-env --config-file ~/swst-icds-test/settings.py --checkpath ~/swst-icds-test/zlib.py -v`. I have attached the test file and my settings file.
[zlib_test.tar.gz](https://github.com/eth-cscs/reframe/files/7693902/zlib_test.tar.gz)

ebryerwork:
Naming the file lxml.py does the same thing. The test won't run. It seems that naming the test file the same as a module that reframe is using is not a good idea. This is an edge case of course.

vkarak:
I confirm that this is a bug and it can be very easily reproduced. Thanks for catching!

```
cp unittests/resources/checks/hellocheck.py unittests/resources/checks/zlib.py
./bin/reframe -c  unittests/resources/checks/zlib.py -l
```

```
Looking for tests in '/.../zlib.py'
Validating '/.../zlib.py': OK
No tests registered
Loaded 0 test(s)
Generated 0 test case(s)
Filtering test cases(s) by name: 0 remaining
Filtering test cases(s) by tags: 0 remaining
Filtering test cases(s) by other attributes: 0 remaining
Building and validating the full test DAG
Full test DAG:
  <empty>
Final number of test cases: 0
[List of matched checks]

Found 0 check(s)
```

rsarm:
The problem it's happening here:
https://github.com/eth-cscs/reframe/blob/4485ad2f9d92795010037f9fa7dab7edd6de0e17/reframe/utility/__init__.py#L58-L59
At this point, the modules from external are already in `sys.modules` and the function will return instead of continuing to import the test with the same name.

The issue depends on where the test is with respect to reframe's install directory since in order to enter in that function it has to pass by a `if rel_filename.startswith('..')`:
https://github.com/eth-cscs/reframe/blob/4485ad2f9d92795010037f9fa7dab7edd6de0e17/reframe/utility/__init__.py#L88

PR #2390, opened 2022-01-19
[bugfix] Avoid conflicts with already loaded Python modules when loading ReFrame tests

rsarm:
Closes https://github.com/eth-cscs/reframe/issues/2336


ekouts (approved):

reframe/utility/__init__.py
```diff
@@ -53,8 +54,12 @@ def _get_module_name(filename):
         return barename.replace(os.sep, '.')
 
 
-def _do_import_module_from_file(filename, module_name=None):
+def _do_import_module_from_file(filename, module_name=None, unique_name=False):
     module_name = module_name or _get_module_name(filename)
+    if unique_name:
+        module_hash = sha256(filename.encode('utf-8')).hexdigest()[:8]
+        module_name = f'rfm_{module_name}_{module_hash}'
```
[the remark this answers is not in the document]
vkarak: This function provides the mechanism to assign custom a name to the loaded module already, so I think it should not be the one changing it. This name mangling code should go in the `import_module_from_file()` instead.

reframe/utility/__init__.py
```diff
@@ -69,7 +74,7 @@ def _do_import_module_from_file(filename, module_name=None):
     return module
 
 
-def import_module_from_file(filename, force=False):
+def import_module_from_file(filename, force=False, unique_name=False):
```
vkarak, line 77: I am giving second thought as of whether we need an option here or use the proposed behaviour always. This function allows you to load a module in a non-standard way, i.e., outside a package, and it tries to create proper integration with `sys.modules` and do proper caching. So I believe the option is not needed and we always do the hashing.

unittests/test_utility.py
```diff
@@ -488,6 +490,16 @@ def test_import_from_file_load_unknown_path():
...
 
 
+def test_import_from_file_load_unknown_path_unique_name():
+    try:
+        util.import_module_from_file('/foo', unique_name=True)
+        pytest.fail()
+    except ImportError as e:
+        module_hash = sha256('/foo'.encode('utf-8')).hexdigest()[:8]
+        assert f'rfm_foo_{module_hash}' == e.name
+        assert '/foo' == e.path
```
vkarak, line 500: The way the test name is constructed it should be irrelevant for the user. We need to test the behaviour of this mechanism. I would suggest the following:

1. Create a temporary Python file in `tmp_dir` and name it `os.py` and define a variable.
2. Try to load it with this function using its absolute path.
3. The test should load and return correctly our  `os.py` and not the already imported one (instead of `os.py` you could name it as any module that we have loaded already)

The same approach should be take with the a dummy valid test, so that we check that the loader loads it. Currently, this is not the case because the loader does not set `unique=True`, so the bug that this PR tries to solve is likely still there ;-)

vkarak (changes requested):

reframe/utility/__init__.py
```diff
@@ -93,7 +89,10 @@ def import_module_from_file(filename, force=False, unique_name=False):
     if rel_filename.startswith('..'):
         # We cannot use the standard Python import mechanism here, because the
         # module to import is outside the top-level package
```
vkarak, line 91: We should extend a bit this comment to explain why we need to do the hashing.

reframe/utility/__init__.py
```diff
@@ -93,7 +89,10 @@ def import_module_from_file(filename, force=False, unique_name=False):
...
-        return _do_import_module_from_file(filename, module_name, unique_name)
+        module_hash = sha256(filename.encode('utf-8')).hexdigest()[:8]
+        module_name = f'rfm_{module_name}_{module_hash}'
+
```
vkarak, line 94: Remove this blank line.

unittests/test_loader.py
```diff
@@ -73,6 +89,20 @@ def test_load_fixtures(loader):
     assert 5 == len(tests)
 
 
+def test_existing_module_name(loader, tmp_path):
+    print(type(tmp_path))
```
vkarak, line 93: You probably forgot this.

unittests/test_utility.py
```diff
@@ -481,23 +479,15 @@ def test_import_from_file_load_abspath():
...
-    try:
-        util.import_module_from_file('/foo', unique_name=True)
-        pytest.fail()
-    except ImportError as e:
-        module_hash = sha256('/foo'.encode('utf-8')).hexdigest()[:8]
-        assert f'rfm_foo_{module_hash}' == e.name
-        assert '/foo' == e.path
+    module = util.import_module_from_file(test_file)
+    assert module.var == 1
+    assert not hasattr(module, 'abc')
```
vkarak, line 489: I would probably use `path` instead of `abc` here.

vkarak (changes requested):
Lgtm. Just a couple of minor comments still.

vkarak (approved):

PR #2392, opened 2022-01-20
[bugfix] Automatically convert numeric configuration parameters when set from the command line or the environment

### LINKED ISSUE #2246: Numeric configuration parameters that are set from environment variables are kept as strings in the internal configuration
Whereas if the same variable is loaded from the configuration file it will be converted to a number according to the configuration schema. This was the direct cause of #2237.

ekouts:
Fixes #2246.


teojgo:
Do you think that we could do something more generic, like adding a `type` option? https://docs.python.org/3/library/argparse.html#type

vkarak:
I agree with @teojgo, I was also thinking of something more generic.

reframe/frontend/argparse.py
```diff
@@ -82,7 +82,7 @@ def __getattr__(self, name):
         if name not in self.__option_map:
             return ret
 
-        envvar, _, action = self.__option_map[name]
+        envvar, _, action, cast_type = self.__option_map[name]
```
vkarak, line 85:
```suggestion
        envvar, _, action, arg_type = self.__option_map[name]
```

reframe/frontend/argparse.py
```diff
@@ -99,6 +99,14 @@ def __getattr__(self, name):
...
                             f'environment variable {envvar!r} not a boolean'
                         ) from None
+                elif action == 'store' and cast_type != str:
+                    try:
+                        ret = cast_type(ret)
+                    except ValueError:
+                        raise ValueError(
+                            f'environment variable {envvar!r} not a '
+                            f'{cast_type}'
+                        ) from None
```
vkarak, line 109:
```suggestion
                    except ValueError as err:
                        raise ValueError(
                            f'cannot convert environment variable {envvar!r} to {arg_type.__name__!r}'
                        ) from err
```

reframe/frontend/argparse.py
```diff
@@ -107,7 +115,7 @@ def update_config(self, site_config):
         namespace'''
         errors = []
         for option, spec in self.__option_map.items():
-            _, confvar, action = spec
+            _, confvar, action, _ = spec
```
vkarak, line 118:
```suggestion
            confvar, action = spec[1:3]
```

vkarak (changes requested):
I think there are more options that need conversion, e.g., `max_retries`. And we need a unit test. Also what happens if `type` is passed and no environment variable is associated with the option? Does the conversion still happen?

ekouts:
@vkarak I added the type to `max-retries` and `maxfail`. Let me know if I missed any other option. I also extended a bit the argparger unittests.
> Also what happens if type is passed and no environment variable is associated with the option? Does the conversion still happen?

No, it will not try to do any conversion if it the value is `None`.

reframe/frontend/cli.py
```diff
@@ -1175,24 +1175,14 @@ def module_unuse(*paths):
                 parsed_job_options.append(f'--{optstr} {valstr}')
 
         exec_policy.sched_options = parsed_job_options
-        try:
-            max_retries = int(options.max_retries)
-        except ValueError:
-            raise errors.ConfigError(
-                f'--max-retries is not a valid integer: {max_retries}'
-            ) from None
+        max_retries = int(options.max_retries)
```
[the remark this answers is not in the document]
ekouts: Sorry, you are right. I just changed the indentation and I didn't realize I was re-casting it.

vkarak (approved):

PR #2399, opened 2022-01-26
[test] Fix hostname printing in ipcmagic test

teojgo:


rsarm (approved):

hpctestlib/interactive/jupyter/ipcmagic/src/tf-hvd-sgd-ipc-tf2.py
```diff
@@ -31,7 +31,8 @@
 
 
 print(os.popen("ps -u $USER | grep ip").read())
-socket.gethostname()
+hostname = socket.gethostname()
+print('Running on node:', hostname)
```
vkarak, line 35: Is there a way to write the test avoiding this? It's not a big deal, since IPCMagic is a CSCS-specific thing.

teojgo: Here I made this more explicit so that we don't get duplicate matches for each node in case of job reports. @rsarm do you think we should do it differently?

rsarm: > Is there a way to write the test avoiding this? It's not a big deal, since IPCMagic is a CSCS-specific thing.

@vkarak  which part? checking the hostnames?

vkarak: I would generally try to avoid modifying the code of benchmarks just to fit us on the matching, but since we have written this, it's fine.

PR #2402, opened 2022-01-28
[test] CPU memory latency test: added eiger and pilatus, removed tave

sebkelle1:


cscs-checks/microbenchmarks/cpu/latency/latency.py
```diff
@@ -68,8 +74,6 @@ class CPULatencyTest(rfm.RegressionTest):
     def set_modules(self):
         if self.current_system.name in {'daint', 'dom'}:
             self.modules = ['craype-hugepages1G']
```
lucamar, line 76:
```suggestion
        if self.current_system.name in {'daint', 'dom', 'eiger', 'pilatus'}:
            self.modules = ['craype-hugepages1G']
```

lucamar: Thanks a lot @sebkelle1, it would help a lot having this additional test when looking for issues on suspect nodes. Do you also need to add `eiger` and `pilatus` here?

sebkelle1: there's no effect on `eiger` or `pilatus`, also not with 8M pages

lucamar (approved):
Thanks @sebkelle1

vkarak (approved):

vkarak:
@jenkins-cscs retry daint pilatus

PR #2403, opened 2022-01-31
[feat] Add convenience function to check if a test variables is defined

vkarak:
This PR adds a convenience function to check if a variable is defined. This is particularly useful in library tests, when you would like to set a default in a pipeline hook and you should only do that if the user has not defined another value with the `-S` option.

### DIFF
--- a/docs/regression_test_api.rst
+++ b/docs/regression_test_api.rst
@@ -670,6 +670,19 @@ Therefore, classes that derive from the base :class:`~reframe.core.pipeline.Regr
 .. automethod:: reframe.core.pipeline.RegressionMixin.variant_name
 
 
+---------------------
+Class-level utilities
+---------------------
+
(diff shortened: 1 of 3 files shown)


vkarak:
Do you think that it would make sense to add also a `set_var_default()` convenience method? Users could write:

```python
type(self).set_var_default('foo , 1)
```

instead of

```python
if not type(self).is_var_defined('foo'):
    self.foo = 1
```

vkarak:
Perhaps, we should better expose this functionality at the test level, so that users avoid the `type(self)` thing.

vkarak:
I have hidden the metaclass stuff and I exposed a new function at the level of `RegressionTest`. Now users can simply do `set_var_default('foo', 10)` in their test.

victorusu (approved):
`lgtm`

Issue #2360, opened 2021-12-26
Runtime not loading correct partitions when fixtures and/or deps are used


mahendrapaipuri:
Looks like `reframe.core.runtime.runtime()` not loading correct partition when either fixtures and/or dependencies are used. A minimal reproducible is as follows:

```
class DownloadTest(rfm.RunOnlyRegressionTest):
 
    def __init__(self):
        self.local = True
        self.sourcesdir = None
        self.executable = 'wget'
        self.executable_opts = ['https://gist.githubusercontent.com/ramsey/11072524/raw/62dace10c306381445a0110538097d5c02227f2d/hello1.c']

    @sanity_function
    def pass_sanity(self):
        return True
        
@rfm.simple_test
class BuildTest(rfm.CompileOnlyRegressionTest):
    
    # Share resource from fixture
    src_file = fixture(DownloadTest, scope='session')

    def __init__(self):
        self.valid_prog_environs = [
            'builtin',
        ]
        self.valid_systems = [
            'generic:mypartition-1', 
            'generic:mypartition-2',
            'generic:mypartition-3',
        ]
    
    @run_before('compile')
    def set_sourcedir(self):
        """Set source path based on dependencies"""
        self.sourcesdir = self.src_file.stagedir
        self.sourcepath = 'hello1.c'

    @run_before('compile')
    def set_build_system_attrs(self):
        """Set build directory and config options"""
        self.build_system = 'SingleSource'

    @sanity_function
    def pass_sanity(self):
        return True

@rfm.simple_test
class RunTest(rfm.RunOnlyRegressionTest):
    descr = 'Test to read the created file'
    valid_systems = [
        'generic:mypartition-1', 
        'generic:mypartition-2',
        'generic:mypartition-3',
    ]
    valid_prog_environs = ['builtin']
    
    @run_after('init')
    def set_dependencies(self):
        """Set dependencies of the test"""
        self.depends_on('BuildTest', udeps.by_part)

    @require_deps
    def set_executable(self, BuildTest):
        """Set executable path and executable"""
        self.sourcesdir = os.path.join(BuildTest().stagedir)
        self.executable = './BuildTest'

    @sanity_function
    def pass_sanity(self):
        return True
```

And `reframe_config.py` is defined as follows:

```
site_configuration = {
    'systems': [
        {
            'name': 'generic',
            'descr': 'Generic system. When no system is matched, it falls back to generic system',
            'hostnames': ['.*'],
            'modules_system': 'nomod',
[...]

PR #2414, opened 2022-02-02
[bugfix] Fix `--show-config` for `null` config options

### LINKED ISSUE #2388: Partition's `time_limit` does not define a default value in the config schema
```
./bin/reframe --show-config=systems/0/partitions/@default/time_limit
```

```
./bin/reframe: no such configuration parameter found: systems/0/partitions/@default/time_limit
```

teojgo:
Fixes #2388

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -769,8 +769,11 @@ def main():
         if config_param == 'all':
             printer.info(str(rt.site_config))
         else:
-            value = rt.get_option(config_param)
-            if value is None:
+            # Create a unique value to check against configuration parameters with
+            # value `None` and invalid ones
+            default = time.time()
+            value = rt.get_option(config_param, default)
+            if value == default:
                 printer.error(
                     f'no such configuration parameter found: {config_param}'
                 )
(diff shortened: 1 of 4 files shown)


reframe/frontend/cli.py
```diff
@@ -769,8 +769,11 @@ def restrict_logging():
         if config_param == 'all':
             printer.info(str(rt.site_config))
         else:
-            value = rt.get_option(config_param)
-            if value is None:
+            # Create a unique value to differentiate between configuration
+            # parameters with value `None` and invalid ones
+            default = time.time()
+            value = rt.get_option(config_param, default)
+            if value == default:
```
vkarak, line 776: Nice trick, but I would rather use another type of unique value:

```suggestion
            default = {'token'}
            value = rt.get_option(config_param, default)
            if value is default:
```

vkarak (changes requested):
Just a minor adaptation. 

@mahendrapaipuri Could you check if this one fixes also #2360? I did a quick test with `--show-config` and seems to be ok.

mahendrapaipuri:
@vkarak Sure. I would test this patch over the weekend.

vkarak (approved):

PR #2418, opened 2022-02-04
[doc] Use `compile` as the name of the build pipeline stage

### LINKED ISSUE #2317: Use consistent naming of pipeline stages across the documentation
The "compile" stage is mentioned as "build" in the [description](https://reframe-hpc.readthedocs.io/en/stable/pipeline.html#the-regression-test-pipeline) of the pipeline, but there is no stage named as such.

vkarak:
Fixes #2317.


ekouts (approved):
@vkarak Going over the doc page again I think I agree with you that there is no need for something else, it looks good

teojgo (approved):

PR #2421, opened 2022-02-07
[testlib] Add OSU benchmarks in the test library and adapt CSCS checks accordingly

rsarm:


rsarm:
I'm having troubles passing the version of the benchmark from the cscs test to the `fetch_osu_benchmark` fixture on the hpclib. I'm creating variables `version = variable(str, value='5.6.2')` and passing them to the fixtures with `{'version': version}`. The problem is that in the body of the class, where the fixture is defined, the value of `version` is the default one (from the definition of the variable) and not the one passed through `variables`, which only will take effect after init right?

vkarak (changes requested):
I think that the benchmarks, especially on the CSCS side are more complex than they should. Also, follow the Amber and Gromacs examples for writing the library tests. That means, you should not name the base file as `__init__.py` and the library tests must also be proper tests that can be run from the command line.

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -6,109 +6,52 @@
...
-class AllreduceTest(rfm.RegressionTest):
+class allreduce_check(osu_latency):
     variant = parameter(['small'], ['large'])
+    ctrl_msg_size = 8
+    perf_msg_size = 8
+    executable = 'osu_allreduce'
+    num_tasks_per_node = 1
+    num_gpus_per_node  = 1
+    osu_binaries = fixture(build_osu_benchmarks, scope='environment',
+                           variants=cpu_build_variant)
```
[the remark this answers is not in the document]
rsarm: Without specifying the variant, the build is going to run for the options `['cpu', 'cuda', 'rocm', 'openacc']`:
```
[ RUN      ] build_osu_benchmarks %build_type=openacc ~dom:gpu+PrgEnv-gnu @dom:gpu+PrgEnv-gnu
[ RUN      ] build_osu_benchmarks %build_type=rocm ~dom:gpu+PrgEnv-gnu @dom:gpu+PrgEnv-gnu
[ RUN      ] build_osu_benchmarks %build_type=cuda ~dom:gpu+PrgEnv-gnu @dom:gpu+PrgEnv-gnu
[ RUN      ] build_osu_benchmarks %build_type=cpu ~dom:gpu+PrgEnv-gnu @dom:gpu+PrgEnv-gnu
```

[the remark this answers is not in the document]
vkarak: Yes, but you could filter out the tests in a post-init hook by setting the `valid_systems` accordingly.

vkarak:
@rsarm I think you haven't add the new file, but the placement need to change as well. You should be able to simply import the tests from `hpctestlib.microbenchmarks.mpi.osu` and not `hpctestlib.microbenchmarks.mpi.osu.mpi`.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,163 @@
...
+            f'tar xzf {tarball}',
+            f'cd {self.build_prefix}'
+        ]
+        self.build_system.config_opts = [f'--enable-{self.build_type}']
+        self.build_system.make_opts = ['-C', 'mpi']
+        self.build_system.max_concurrency = 8
+
+    @sanity_function
+    def validate_build(self):
+        return True
```
vkarak, line 67: Perhaps a comment here as of why we seem not to validate here would help the reader understand better.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,163 @@
...
+
+    #: Performance message size
+    #:
+    #: This value is used for the performance checks
+    #:
+    #: :type: :class:`int`
+    #: :default: ``8``
+    perf_msg_size = variable(int, value=8)
+
+    #: Accelerator device type
```
vkarak, line 106: This is not accurate.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,163 @@
...
+    #: :default: ``8``
+    message_size = variable(int, value=8)
+
+    #: Performance message size
+    #:
+    #: This value is used for the performance checks
+    #:
+    #: :type: :class:`int`
+    #: :default: ``8``
+    perf_msg_size = variable(int, value=8)
```
vkarak, line 104: I don't think that we need this.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,163 @@
...
+
+    #: Accelerator device type
+    #:
+    #: Use accelerator device buffers, i.e cuda, openacc or rocm
+    #:
+    #: :type: :class:`str`
+    #: :default: ``None``
+    device_buffers = variable(str, type(None), value=None)
+
+    executables = {
```
vkarak, line 114:
```suggestion
    microbenchmarks = {
```

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,163 @@
...
+        if benchmark_type == 'pt2pt':
+            self.executable_opts += ['D', 'D']
+
+    @sanity_function
+    def validate_test(self):
+        return sn.assert_found(rf'^{self.message_size}', self.stdout)
+
+
+@rfm.simple_test
+class osu_bandwidth(osu_benchmark_test_base):
```
vkarak, line 147:
```suggestion
class osu_bandwidth_check(osu_benchmark_test_base):
```

Same

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,163 @@
...
+    @performance_function('MB/s', perf_key='bw')
+    def set_performance_patterns(self):
```
vkarak, line 149:
```suggestion
    @performance_function('MB/s')
    def bandwidth(self):
```

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,163 @@
...
+        """Bandwidth for the message size `message_size`."""
+
+        return sn.extractsingle(rf'^{self.message_size}\s+(?P<bw>\S+)',
+                                self.stdout, 'bw', float)
+
+
+@rfm.simple_test
+class osu_latency(osu_benchmark_test_base):
+    @performance_function('us', perf_key='latency')
+    def set_performance_patterns(self):
```
vkarak, line 159:
```suggestion
    @performance_function('us')
    def latency(self):
```

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -161,45 +93,76 @@ def set_performance_patterns(self):
...
 
 
-class P2PBaseTest(rfm.RegressionTest):
-    exclusive_access = True
-    strict_check = False
-    num_tasks = 2
+@rfm.simple_test
+class alltoall_check(osu_latency):
+    message_size = 8
+    executable = 'osu_alltoall'
```
vkarak, line 99: The various benchmarks must be part of the library. Here should go *only* the CSCS-specific parts.

hpctestlib/microbenchmarks/mpi/osu.py
vkarak: This is not currently a "final" test because it's not complete. It's missing the `executable`. The idea is that anybody can run the tests from the library directly from the command line.

vkarak (changes requested):

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
rsarm: My idea was to do not have the executables or job parameters like the number of task on the lib side. I thought the lib tests as having no meaning except that one is for latency tests and the other one for bandwidth tests. Depending of what's set in the tests that inherit from them, they could be an alltoall test over multiple nodes or a point to point test that uses two nodes.

We could put defaults on the lib side and make them point to point tests. What do you think?

vkarak: For P2P tests, `num_tasks = 2`, we know that. For collectives, we can make `num_tasks` required. The idea is that as a user I should be able to run the the benchmark from the command line without having to write a derived test.

rsarm:
The lib tests can be run like this:
```
RFM_COMPACT_TEST_NAMES=true ./bin/reframe -C config/cscs.py -r -c hpctestlib/microbenchmarks/mpi/osu.py -n 'osu_cpu_latency_pt2pt.osu_binaries.build_type=cuda' -S valid_systems='dom:gpu' -S valid_prog_environs=PrgEnv-gnu -S osu_cpu_latency_pt2pt.device_buffers=cuda
```

vkarak:
Actually, I managed to run the test on Dom as follows:

```
./bin/reframe -C config/cscs.py -r -c hpctestlib/microbenchmarks/mpi/osu.py -n 'osu_cpu_latency_pt2pt.osu_binaries.build_type=cuda' -S valid_systems='dom:gpu' -S valid_prog_environs=PrgEnv-gnu -S osu_cpu_latency_pt2pt.device_buffers=cuda -S osu_cpu_latency_pt2pt.variables=MPICH_RDMA_ENABLED_CUDA:1 -M 'PrgEnv-gnu: PrgEnv-gnu craype-accel-nvidia60 gcc/9.3.0' -r
```

I don't like that we have to use the `-M` option, but that's due to #2477. Ideally, we would replace that with `-S build_osu_benchmarks.modules=craype-accel-nvidia60,gcc/9.3.0`.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,168 @@
...
+class osu_cpu_bandwidth_pt2pt(osu_bandwidth):
+    @run_after('init')
+    def set_job_options(self):
+        self.executable = 'osu_latency'
+        self.num_tasks = 2
+        self.num_tasks_per_node = 1
+
+
+@rfm.simple_test
+class osu_allreduce(osu_latency):
```
vkarak, line 163:
```suggestion
class osu_allreduce_check(osu_latency):
```

vkarak: Perhaps we should also differentiate with latency and b/w tests.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,168 @@
...
+class osu_bandwidth(osu_benchmark_test_base):
+    @performance_function('MB/s', perf_key='bw')
+    def bandwidth(self):
+        """Bandwidth for the message size `message_size`."""
+
+        return sn.extractsingle(rf'^{self.message_size}\s+(?P<bw>\S+)',
+                                self.stdout, 'bw', float)
+
+
+class osu_latency(osu_benchmark_test_base):
```
vkarak, line 135: The *only* difference of the `osu_latency` vs. `osu_bandwidth` is the `message_size`. The performance function should be moved to the base class and *not* use named capturing groups. There is no need for that and these should become as simple as the following:

```python
class osu_latency(osu_benchmark_test_base):
    message_size = 8

class osu_bandwidth(osu_benchmark_test_base):
    message_size = 1024*1024
```

I am also inclined to change the `osu_benchmark_test_base` to `osu_benchmark`.

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -6,109 +6,49 @@
...
 
 @rfm.simple_test
-class AllreduceTest(rfm.RegressionTest):
-    variant = parameter(['small'], ['large'])
-    strict_check = False
+class allreduce_check(osu_latency):
+    number_of_nodes = parameter([6], [16])
+    message_size = 8
+    executable = 'osu_allreduce'
+    num_tasks_per_node = 1
```
vkarak, line 45: There is no need to reset this here.

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -6,109 +6,49 @@
...
+cuda_build_variant = build_osu_benchmarks_gpu.get_variant_nums(
+    build_type='cuda'
+)
 
 
 @rfm.simple_test
-class AllreduceTest(rfm.RegressionTest):
-    variant = parameter(['small'], ['large'])
-    strict_check = False
+class allreduce_check(osu_latency):
```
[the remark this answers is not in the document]
vkarak: And, similarly, you should prefix all CSCS site-specific checks with `cscs_`.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -0,0 +1,168 @@
...
+class osu_benchmark_test_base(rfm.RunOnlyRegressionTest):
+    '''Base class of OSU benchmarks runtime tests'''
+
+    #: Number of warmup iterations
+    #:
+    #: This value is passed to the excutable through the -x option.
+    #:
+    #: :type: :class:`int`
+    #: :default: ``1000``
+    num_warmup_iters = variable(int, value=1000)
```
vkarak, line 66: Aren't these a bit too many for warm up iterations?

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -38,11 +37,9 @@ def set_modules(self):
 
 
 @rfm.simple_test
-class allreduce_check(osu_latency):
-    number_of_nodes = parameter([6], [16])
-    message_size = 8
+class cscs_allreduce_check(osu_alltoall_check):
+    num_nodes = parameter([6], [16])
```
vkarak, line 41:
```suggestion
    num_nodes = parameter([6, 16])
```

cscs-checks/microbenchmarks/mpi/osu/osu_tests.py
```diff
@@ -57,17 +54,17 @@ class allreduce_check(osu_latency):
...
     @run_before('run')
     def set_num_tasks(self):
-        self.num_tasks = self.number_of_nodes
+        self.num_tasks = self.num_nodes
 
     @run_before('performance')
     def set_performance_patterns(self):
-        if self.number_of_nodes == 6:
-            self.reference = {
+        allref = {
```
vkarak, line 66: Move this in the class body.

hpctestlib/microbenchmarks/mpi/osu.py
vkarak: Correct point and I guess you look into different columns. I overlooked that.

vkarak: Correct point; I think you can still reuse the function. I'll have a look at it.

hpctestlib/microbenchmarks/mpi/osu.py
```diff
@@ -112,7 +116,7 @@ def set_executable(self):
         self.executable_opts = ['-m', f'{max_message_size}',
                                 '-x', f'{self.num_warmup_iters}',
                                 '-i', f'{self.num_iters}']
-        if self.device_buffers  and f'{self.device_buffers}' != 'cpu':
+        if f'{self.device_buffers}' != 'cpu':
```
vkarak, line 119:
```suggestion
        if self.device_buffers != 'cpu':
```

vkarak (changes requested):
I will also do further fine tuning pass.

vkarak:
We still need to update the docs for the test library version and then it'll be ready for merging.

rsarm:
@vkarak I gave it a try, it looks good to me. The ci failure was only a performance error.

vkarak:
@jenkins-cscs retry daint dom

PR #2424, opened 2022-02-10
[bugfix] Do not set up Spack shell support

giordano:
I have two-fold motivations for this:

1. I'm trying to use reframe with spack build system on a system where activating shell support seems to screw something up
2. not enabling shell support lets us skip the line https://github.com/eth-cscs/reframe/blob/f996fabf25a35f3bb2a5b7090dd0ea4bf327bace/reframe/core/buildsystems.py#L872 which is very useful when using https://github.com/eth-cscs/spack-batteries-included/ with the single squashfs file (there is no material `$(spack location --spack-root)/share/spack/setup-env.sh` file on disk, so this line would fail).  Also, that line caused problems in the past (see #2097, now solved), and not activating the shell support just makes things more transparent.

As an added bonus, this PR has a negative net diff :slightly_smiling_face:

### DIFF
--- a/docs/tutorial_build_automation.rst
+++ b/docs/tutorial_build_automation.rst
@@ -120,12 +120,10 @@ Here is what ReFrame generates as a build script for this example:
 
 .. code:: bash
 
-   . "$(spack location --spack-root)/share/spack/setup-env.sh"
    spack env create -d rfm_spack_env
-   spack env activate -V -d rfm_spack_env
-   spack config add "config:install_tree:root:opt/spack"
-   spack add bzip2@1.0.6
-   spack install
+   spack -e rfm_spack_env config add "config:install_tree:root:opt/spack"
+   spack -e rfm_spack_env add bzip2@1.0.6
+   spack -e rfm_spack_env install
 
 As you might have noticed ReFrame expects that Spack is already installed on the system.
 The packages specified in the environment and the tests will be installed in the test's stage directory, where the environment is copied before building.
@@ -155,10 +153,8 @@ Finally, here is the generated run script that ReFrame uses to run the test, onc
 .. code-block:: bash
 
    #!/bin/bash
-   . "$(spack location --spack-root)/share/spack/setup-env.sh"
    spack env create -d rfm_spack_env
-   spack env activate -V -d rfm_spack_env
-   spack load bzip2@1.0.6
+   eval `spack -e rfm_spack_env load --sh bzip2@1.0.6`
    bzip2 --help
 
 From this point on, sanity and performance checking are exactly identical to any other ReFrame test.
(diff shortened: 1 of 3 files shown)


docs/tutorial_build_automation.rst
```diff
@@ -155,10 +153,8 @@ Finally, here is the generated run script that ReFrame uses to run the test, onc
 .. code-block:: bash
 
    #!/bin/bash
-   . "$(spack location --spack-root)/share/spack/setup-env.sh"
    spack env create -d rfm_spack_env
-   spack env activate -V -d rfm_spack_env
-   spack load bzip2@1.0.6
+   eval `spack -e rfm_spack_env load --sh bzip2@1.0.6`
```
haampie, line 157: I think this is reasonable; either you use `spack env activate`, or you use `spack load`, but not both, cause `spack env activate --sh` also returns shell-environment variables changes for the spack-environment root specs and their transient run/link type dependencies.

so, it's either `eval "$(spack -e [env name] load --sh [spec name])"` where you use the environment to pick the just-built bzip2, or assuming you only have one root spec in the environment, you do `eval "$(spack env activate --sh [env name])"`

haampie: You can also put `concretize: together` instead of the default `concretize: separately` in the environment, and then you can be sure that `spack env activate --sh ...` always works (cant have multiple bzip2's in one env then)

haampie: not sure if you can change the value of `spack:concretize: ...` from the command line though :(

victorusu (approved):
`lgtm`

vkarak:
ok to test

reframe/core/buildsystems.py
```diff
@@ -850,38 +850,37 @@ def __init__(self):
...
+        self._auto_env = True
+        return [f'spack env create -d {self.environment}']
 
     def prepare_cmds(self):
-        cmds = self._env_activate_cmds()
+        cmds = self._create_env_cmds()
         if self.specs and self.emit_load_cmds:
-            cmds.append('spack load ' + ' '.join(s for s in self.specs))
+            cmds.append(f'eval `spack -e {self.environment} load --sh ' +
+                        ' '.join(s for s in self.specs) + '`')
```
vkarak, line 883:
```suggestion
                        ' '.join(self.specs) + '`')
```

docs/tutorial_build_automation.rst
vkarak: Users can pass a custom environment through the `self.build_system.environment` option and reframe will not auto-generate the environment if they do.

vkarak (changes requested):
Lgtm, except a small code fix.

vkarak (approved):

PR #2428, opened 2022-02-15
[feat] Support for loggable variables, parameters and properties of tests

### LINKED ISSUE #2380: Control which test fields can be logged or not
Sending arbitrary test attributes to Logstash through the `httpjson` handler can be problematic if two tests define the same attribute differently. If the test A sends attribute `x` as an integer and this is indexed as such by Elastic, then if test B sends it as a `string` in the same index, the whole log record will be ignored. For this reason, I propose to provide a finer control in the framework as of which attributes can be logged or not. More specifically, we could add a `loggable` argument to the `variable()` definition that would control whether this attribute is going to be logged or not. By default, this should be set to `False`. Additionally, we should have a way to define loggable properties. 

This feature should simplify also the logging implementation.

vkarak:
This PR adds the `loggable` argument to the `variable` and `parameter` interface to allow you to mark a test variable or parameter as loggable. By default, no test attribute (variable or parameter) is loggable. This contrasts with the current default behaviour where everything is loggable, which can lead to problems, such as the one described in #2380. Test properties can also be logged by decorating them with either the `@loggable` or `@loggable_as` decorators. A loggable attribute will become a log record attribute with the name `check_NAME`, where `NAME` is the name of the variable, parameter or property.

Almost all previously loggable attributes of `RegressionTest` and `Job` are also marked explicitly as loggable, except any attributes containing deferred expressions, such as the `sanity_patterns` and the `perf_patterns`. This PR breaks the logging of custom attributes that are not marked as loggable.

This PR does not set the loggable attributes in the various library tests. A separate PR will follow for that.

Fixes #2380.


victorusu (approved):
`lgtm`

ekouts (approved):

PR #2434, opened 2022-02-18
[test] Update application tests and remove production/maintenance variants

vkarak:
I have updated the reference numbers for several applications and benchmarks based on the average and 95th percentiles over a representative month of the past. I have also removed the distinction between production and maintenance variants, since this is not used. This might imply that we will have to update some dashboards as soon as this is merged.


lucamar (approved):
Thanks @vkarak, we might also need to review the reference values of Piz Daint a while after the upgrade

vkarak:
> Thanks @vkarak, we might also need to review the reference values of Piz Daint a while after the upgrade

Yes, I have now a dashboard with monthly averages and percentiles, so it should be easier to do so.

PR #2436, opened 2022-02-18
[bugfix] Enforce override of hooks with the same name

### LINKED ISSUE #2397: Strange behaviour when redefining hook and assigning it to a different stage
The following code leads to the strange behaviour that the second `check_x()` overrides the first (as expected) but the hooks are not overridden, leaving the second `check_x()` executing as a post-init hook. We should either disable redefinitions of the same hook in the same class or have the second definition *and* hook override completely the first one.

```python
@simple_test
class my_test(RegressionTest):
    x = variable(int, value=10)

    @run_after('init')
    def check_x(self):
        '''Test will not be listed or run if filtered here'''
        self.skip_if(self.x < 3, 'x is lower than 3')

    @run_after('setup')
    def check_x(self):
        '''Test will be skipped during runtime'''
        self.skip_if(self.x < 5, 'x is lower than 5')
```

teojgo:
Fixes #2397


reframe/core/hooks.py
```diff
@@ -74,6 +74,7 @@ def select_hooks(obj, kind):
 
         @functools.wraps(func)
         def _fn(obj, *args, **kwargs):
+
```
vkarak, line 77: Remove that line.

reframe/core/hooks.py
```diff
@@ -151,7 +152,10 @@ def add(self, v):
         '''
 
         if hasattr(v, '_rfm_attach'):
-            self.__hooks.add(Hook(v))
+            h = Hook(v)
```
vkarak, line 155: Add a blank line after this one.

unittests/test_pipeline.py
```diff
@@ -831,6 +831,21 @@ def y(self):
     assert test.foo == 10
 
 
+def test_overriden_hook_different_stages(HelloTest, local_exec_ctx):
+    @test_util.custom_prefix('unittests/resources/checks')
+    class MyTest(HelloTest):
+        @run_after('init')
+        def check(self):
```
vkarak, line 838: I would probably use another name for this one.

vkarak (changes requested):
Lgtm, just a couple of minor style comments, which I can fix.

vkarak (approved):

vkarak:
@jenkins-cscs retry daint

PR #2444, opened 2022-02-22
[feat] Introduce environment variables for controlling system auto-detection

### LINKED ISSUE #2440: Allow different ways of specifying a hostname command through the configuration
On #2216 the hostname detection was changed from `socket.gethostname()` to `socket.getfqdn()`. For some systems that's not convenient. We discussed internally the possibility of giving the option of specifying one of them through the configuration.

rsarm:
Closes #2440

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1316,19 +1316,6 @@ General Configuration
    .. versionadded:: 3.9.0
 
 
-.. js:attribute:: .general[].hostname_cmd
-
-   :required: No
-   :default: ``'hostname'``
-
-   Command to determine the systems name.
-   There are three values available: ``'hostname'``, ``fqdn`` and ``'xthostname'``.
-   The values ``'hostname'`` and ``'fqdn'`` can be specified to detect the system using Python's ``socket.gethostname()`` or ``socket.getfqdn()`` respectively.
-   With the ``'xthostname'`` option, ReFrame first tries to obtain the hostname from ``/etc/xthostname``, which provides the unqualified machine name in Cray systems.
-   If this cannot be found, ReFrame falls back to the ``'hostname'`` option.
-
-   .. versionadded:: 3.11.0
-
 .. js:attribute:: .general[].git_timeout
 
   :required: No
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1315,6 +1315,19 @@ General Configuration
    .. versionadded:: 3.9.0
 
 
+.. js:attribute:: .general[].hostname_cmd
+
+   :required: No
+   :default: ``'hostname'``
+
+   Command to determine the systems name.
+   There are three values available: ``'hostname'``, ``fqdn`` and ``'xthostname'``.
+   The values ``'hostname'`` and ``'fqdn'`` can be specified to detect the system using Python's ``socket.gethostname()`` or ``socket.getfqdn()`` respectively.
+   With the ``'xthostname'`` option, ReFrame first tries to obtain the hostname from ``/etc/xthostname``, which provides the unqualified machine name in Cray systems.
+   If this cannot be found, ReFrame falls back to the ``'hostname'`` option.
+
+   .. versionadded:: 3.11.0
+
 .. js:attribute:: .general[].git_timeout
 
   :required: No
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -232,6 +232,7 @@ System Partition Configuration
    - ``squeue``: Jobs will be launched using the `Slurm <https://www.schedmd.com/>`__ scheduler.
      This backend does not rely on job accounting to retrieve job statuses, but ReFrame does its best to query the job state as reliably as possible.
    - ``torque``: Jobs will be launched using the `Torque <https://en.wikipedia.org/wiki/TORQUE>`__ scheduler.
+   - ``lsf``: Jobs will be launched using the `LSF <https://www.ibm.com/docs/en/spectrum-lsf/10.1.0?topic=lsf-session-scheduler>`__ scheduler.
 
    .. versionadded:: 3.7.2
       Support for the SGE scheduler is added.
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -240,6 +240,9 @@ System Partition Configuration
    .. versionadded:: 3.8.2
       Support for the OAR scheduler is added.
 
+   .. versionadded:: 3.11.0
+      Support for the LSF scheduler is added.
+
    .. note::
 
       The way that multiple node jobs are submitted using the SGE scheduler can be very site-specific.
@@ -873,7 +876,7 @@ All logging handlers share the following set of common attributes:
 .. versionadded:: 3.4.2
    Allow arbitrary job attributes to be logged.
 
-.. versionchanged:: 3.10.2
+.. versionchanged:: 3.11.0
    Limit the number of attributes that can be logged. User attributes or properties must be explicitly marked as "loggable" in order to be selectable for logging.
 
 
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -236,12 +236,12 @@ Picking a System Configuration
 As discussed previously, ReFrame's configuration file can store the configurations for multiple systems.
 When launched, ReFrame will pick the first matching configuration and load it.
 
-There are three ways in which the system name can be determined.
-They can be specified in the configuration through the ``hostname_cmd`` option which can take one of the three values ``'hostname'``, ``fqdn`` or ``'xthostname'``.
-When ``hostname_cmd`` is set to ``'hostname'``, which is the default value, ReFrame will detect the system using Python's ``socket.gethostname()``.
-Similarly, with ``'hostname'``, ReFrame will use ``socket.getfqdn()``.
-With ``'xthostname'``, ReFrame first tries to obtain the hostname from ``/etc/xthostname``, which provides the unqualified *machine name* in Cray systems.
-If this cannot be found, ReFrame falls back to the ``'hostname'`` option.
+The way in which the system name is determined can be controlled by the environment variables ``RFM_AUTODETECT_METHOD``, ``RFM_AUTODETECT_FQDN`` and ``RFM_AUTODETECT_XTHOSTNAME``.
+``RFM_AUTODETECT_METHOD`` is a string that specifies the method used to detect the system name.
+Its default value is ``'hostname'``, which currently is the only supported option.
+The host name will be determined by ``socket.gethostname()``, unless the boolean ``RFM_AUTODETECT_FQDN`` is set, which make ReFrame use ``socket.getfqdn()`` instead.
+The last option, ``RFM_AUTODETECT_XTHOSTNAME`` is a boolean and when set, ReFrame first tries to obtain the hostname from ``/etc/xthostname``, which provides the unqualified *machine name* in Cray systems.
+If this cannot be found, ReFrame falls back to the one of the two options set by ``RFM_AUTODETECT_FQDN``.
 
 Having retrieved the hostname, ReFrame goes through all the systems in its configuration and tries to match the hostname against any of the patterns defined in each system's ``hostnames`` property.
 The detection process stops at the first match found, and that system's configuration is selected.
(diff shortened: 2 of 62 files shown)


reframe/utility/__init__.py
```diff
@@ -24,6 +25,23 @@
 from . import typecheck as typ
 
 
+def get_hostname_cmd(hostname_cmd, logger=None):
+    if hostname_cmd == 'xthostname':
```
ekouts, line 29: This is a tiny thing, but maybe `xthostnamex` is a better name than `xthostname`, since it is based on the name of the cray file?

ekouts (changes requested):
The implementation looks fine to me, but we should document the option in `docs/config_reference.rst` and `docs/manpage.rst`.

reframe/utility/__init__.py
rsarm: no, sorry, I was checking the behavior in case the file doesn't exist. I pushed it by mistake with the misspelled name.

docs/config_reference.rst
```diff
@@ -1316,6 +1316,19 @@ General Configuration
    .. versionadded:: 3.9.0
 
 
+.. js:attribute:: .general[].hostname_cmd
+
+   :required: No
+   :default: ``'hostname'``
+
+   Command to determine the systems name.
```
ekouts, line 1324:
```suggestion
   Command to determine the system's name.
```

docs/configure.rst
```diff
@@ -235,9 +235,14 @@ Picking a System Configuration
...
 As discussed previously, ReFrame's configuration file can store the configurations for multiple systems.
 When launched, ReFrame will pick the first matching configuration and load it.
-This process is performed as follows:
-ReFrame first tries to obtain the hostname from ``/etc/xthostname``, which provides the unqualified *machine name* in Cray systems.
-If this cannot be found, the hostname will be obtained from the standard ``hostname`` command.
+
+There are three ways in which the system name can be determined.
+They can be specified in the configuration through the ``hostname_cmd`` option which can take one of the three values ``'hostname'``, ``fqdn`` or ``'xthostname'``.
+When ``hostname_cmd`` is set to ``'hostname'``, which is the default value, ReFrame will detect the system using Python's ``socket.gethostname()``.
+Similarly, with ``'hostname'``, ReFrame will use ``socket.getfqdn()``.
```
ekouts, line 242:
```suggestion
Similarly, with ``'fqdn'``, ReFrame will use ``socket.getfqdn()``.
```

ekouts (changes requested):

vkarak (changes requested):
I don't think that controlling the method to retrieve the hostname can be part of the configuration. ReFrame uses the hostname to navigate through the configuration and eventually pick the right system subconfig. It's the exactly equivalent of passing `--system`: this cannot be part of the configuration. This means that you would need an environment variable to control the method, such as `RFM_SYSTEM_AUTODETECT_METH=fqdn`. Now the problem with that is that whatever we do we will be introducing a breaking behaviour. The current auto-detection works as follows:

1. Search for `xthostname`
2. If not found, use `fqdn` to retrieve the hostname (in the past we have been using `hostname`).

I agree that the auto-detection process should be controlled, but assuming that we use an environment variable as the above, what would be the default value? If it's `fqdn` we would break the auto-detection in Cray systems, if it's anything else, we would break the current behaviour. There are also aspects in auto-detection when you run in the cloud. See [this](https://github.com/JonShelley/reframe/blob/52e4ff0473703ab0b6f7f6002b6b003bf5df6050/reframe/core/config.py#L279-L298) for example, where they construct the system name from the instance information retrieved by a REST call. To support this, we would need to generalise the auto-detection mechanism a bit differently than how this PR tries to do it.

So perhaps we could simplify this PR by introducing three environment variables controlling the auto-detection though hostname:

```
RFM_AUTODETECT_METHOD=hostname
RFM_AUTODETECT_XTHOSTNAME=1   # (default)
RFM_AUTODETECT_FQDN=1 # default
```

The `hostname` autodetection method is the only supported method currently, but the other two variables control whether the `xthostname` will also be tried and whether the fqdn will be used instead of the `hostname`.

Does it make sense?

reframe/schemas/config.json
```diff
@@ -456,8 +456,11 @@
                     "colorize": {"type": "boolean"},
                     "compact_test_names": {"type": "boolean"},
                     "git_timeout": {"type": "number"},
+                    "hostname_cmd": {
+                        "type": "string",
+                        "enum": ["fqdn", "hostname", "xthostname"]
+                    },
```
ekouts, line 462: This should be removed

reframe/frontend/cli.py
```diff
@@ -640,11 +659,22 @@ def restrict_logging():
...
+                            else True),
+    }
+
     # First configure logging with our generic configuration so as to be able
     # to print pretty messages; logging will be reconfigured by user's
     # configuration later
     site_config = config.load_config(
-        os.path.join(reframe.INSTALL_PREFIX, 'reframe/core/settings.py')
+        os.path.join(reframe.INSTALL_PREFIX, 'reframe/core/settings.py'),
+        autodetect_opts=autodetect_opts
```
ekouts, line 677: I would leave as before since with `None` the method `load_config` will get the defaults anyway.
```suggestion
        os.path.join(reframe.INSTALL_PREFIX, 'reframe/core/settings.py')
```

ekouts (changes requested):

docs/configure.rst
```diff
@@ -235,9 +235,14 @@ Picking a System Configuration
...
 As discussed previously, ReFrame's configuration file can store the configurations for multiple systems.
 When launched, ReFrame will pick the first matching configuration and load it.
-This process is performed as follows:
-ReFrame first tries to obtain the hostname from ``/etc/xthostname``, which provides the unqualified *machine name* in Cray systems.
-If this cannot be found, the hostname will be obtained from the standard ``hostname`` command.
+
+The way in which the system name is determined can be controlled by the environment variables ``RFM_AUTODETECT_METHOD``, ``RFM_AUTODETECT_FQDN`` and ``RFM_AUTODETECT_XTHOSTNAME``.
+``RFM_AUTODETECT_METHOD`` is a string that specifies the method used to detect the system name.
+Its default value is ``'hostname'``, which currently is the only supported option.
+The host name will be determined by ``socket.gethostname()``, unless the boolean ``RFM_AUTODETECT_FQDN`` is set, which make ReFrame use ``socket.getfqdn()`` instead.
```
ekouts, line 242:
```suggestion
The host name will be determined by ``socket.gethostname()``, unless the boolean ``RFM_AUTODETECT_FQDN`` is set, which will make ReFrame use ``socket.getfqdn()`` instead.
```

reframe/core/config.py
```diff
@@ -67,6 +66,11 @@ def __init__(self, site_config, filename):
         self._subconfigs = {}
         self._local_system = None
         self._sticky_options = {}
+        self.autodetect_opts = {
```
ekouts, line 69: It's a small thing, but since you are not using this outside the `_SiteConfig` object I would name it `self._autodetect_opts`. Fine if you decide to leave it as it.

ekouts (approved):
I still have some small comments but in general lgtm.

docs/manpage.rst
```diff
@@ -1023,6 +1022,58 @@ Here is an alphabetical list of the environment variables recognized by ReFrame:
...
+      Associated configuration parameter N/A
+      ================================== ==================
+
+
+   .. versionadded:: 3.12.0
+
+
+.. envvar:: RFM_AUTODETECT_FQDN
+
+   Use FQDN as host name.
```
vkarak, line 1047:
```suggestion
   Use the fully qualified domain name as the host name.
```

docs/manpage.rst
```diff
@@ -1023,6 +1022,58 @@ Here is an alphabetical list of the environment variables recognized by ReFrame:
...
+      Associated configuration parameter N/A
+      ================================== ==================
+
+
+   .. versionadded:: 3.12.0
+
+
+.. envvar:: RFM_AUTODETECT_XTHOSTNAME
+
+   Use Cray's xthostname file to find the host name.
```
vkarak, line 1063:
```suggestion
   If present, use Cray's `/etc/xthostname` file to retrieve the current system's name.
```

reframe/core/config.py
```diff
@@ -67,6 +66,11 @@ def __init__(self, site_config, filename):
...
+            'autodetect_method': 'hostname',
+            'autodetect_xthostname': True,
+            'autodetect_fqdn': True,
```
vkarak, line 72: I don't like very much that we set defaults here as well as in the cli. I would vote to have defaults only in the cli.

reframe/frontend/cli.py
```diff
@@ -523,6 +523,32 @@ def main():
...
+        envvar='RFM_AUTODETECT_METHOD',
+        action='store',
+        help='Method to detect the system'
+    )
+    argparser.add_argument(
+        dest='autodetect_fqdn',
+        envvar='RFM_AUTODETECT_FQDN',
+        action='store_true',
+        help='Use FQDN as host name'
+    )
```
vkarak, line 544: I'd suggest to add a default value here, too.

reframe/frontend/cli.py
```diff
@@ -679,19 +708,25 @@ def restrict_logging():
         )
         sys.exit(0)
 
+    if autodetect_opts['autodetect_method'] not in ['hostname']:
+        printer.error('unknown autodetect method '
+                      f"`{options.autodetect_method}': Exiting...")
```
vkarak, line 713: This error should be raised by the `load_config()`.

reframe/frontend/cli.py
```diff
@@ -877,7 +912,7 @@ def print_infoline(param, value):
         'cmdline': ' '.join(sys.argv),
         'config_file': rt.site_config.filename,
         'data_version': runreport.DATA_VERSION,
-        'hostname': socket.getfqdn(),
+        'hostname': get_hostname_cmd(autodetect_opts),
```
vkarak, line 915: I would probably use here `socket.gethostname()` and add some debug messages in the autodetection process, since the hostname here is just informational.

reframe/utility/__init__.py
```diff
@@ -24,6 +25,29 @@
 from . import typecheck as typ
 
 
+def get_hostname_cmd(autodetect_opts, logger=None):
```
vkarak, line 28: Since I don't think that this is needed to be called by the cli (see my comment there), it's better if this functionality is embedded directly in the auto-detection code inside the `_SiteConfig`.

vkarak: Also, there is generally no need to pass the logger here. `getlogger()` should always return a valid logger object, even if that it doesn't log anything.

reframe/utility/__init__.py
```diff
@@ -24,6 +25,29 @@
...
+    if autodetect_opts['autodetect_xthostname']:
+        if os.path.exists('/etc/xthostname'):
+            # Get the cluster name on Cray systems
+            with open('/etc/xthostname') as fp:
+                return fp.read()
+
+        elif logger:
+            logger.debug("Did not find '/etc/xthostname')")
```
vkarak, line 36: I would rather have log messages, such as `trying '/etc/xthostname' ...` etc.

reframe/core/config.py
```diff
@@ -625,7 +625,7 @@ def _find_config_file():
     return None
 
 
-def load_config(filename=None):
+def load_config(filename=None, autodetect_opts=None):
```
vkarak, line 628: Instead of passing a dictionary of options here, I would rather add an argument about the method and then the rest of keyword arguments would be the autodetection options.

```suggestion
def load_config(filename=None, autodetect_method='hostname', **autodetect_opts):
```

vkarak (changes requested):

reframe/frontend/cli.py
rsarm: I think with the variables non-associated to command line options, the defaults set through the argparser are ignored. That's why I'm setting the defaults later on the cli.

vkarak: Ok, you are right.

reframe/utility/__init__.py
rsarm: Since it's a `utility` function, I didn't want to import `reframe.core` function to get a logger.

vkarak:
I'm updating a bit the implementation: I have made our `ArgumentParser` understand the `default` argument when we are treating options that are only set from the environment. I think that was a bug. Then I'm refactoring a bit the auto-detection. I don't think that the `gethostname_cmd()` should be a utility function. I am moving its functionality inside the configuration code. Also, there is no need to pass the auto-detection options at the creation of the configuration, since the auto-detection happens in `select_subconfig`. I will add a method to set the auto-detection method in `_SiteConfig` and this one will always be called from the cli based on the command line options.

vkarak (approved):

PR #2454, opened 2022-03-01
[test] Update VASP checks and reference values

lucamar:
I provide updated checks and reference values for VASP on Piz Daint after the upgrade: we will still need to monitor the performance over the next weeks, since I could not retrieve the data for the multicore check on Kibana.

### DIFF
--- a/cscs-checks/apps/vasp/vasp_check.py
+++ b/cscs-checks/apps/vasp/vasp_check.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import reframe as rfm
-import reframe.utility as util
 import reframe.utility.sanity as sn
 
 
--- a/cscs-checks/apps/vasp/vasp_check.py
+++ b/cscs-checks/apps/vasp/vasp_check.py
@@ -4,34 +4,79 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import reframe as rfm
+import reframe.utility as util
 import reframe.utility.sanity as sn
 
 
+@rfm.simple_test
 class VASPCheck(rfm.RunOnlyRegressionTest):
-    scale = parameter(['small', 'large'])
-    variant = parameter(['maint', 'prod'])
+    descr = f'VASP check '
     modules = ['VASP']
     executable = 'vasp_std'
-    keep_files = ['OUTCAR']
     extra_resources = {
         'switches': {
             'num_switches': 1
         }
     }
+    keep_files = ['OUTCAR']
     strict_check = False
+    use_multithreading = False
+    tags = {'maintenance', 'production'}
     maintainers = ['LM']
-    tags = {'scs'}
+
+    num_nodes = parameter([6, 16], loggable=True)
+    references = {
+        6: {
+            'sm_60': {
+                'dom:gpu': {'time': (56.0, None, 0.10, 's')},
+                'daint:gpu': {'time': (65.0, None, 0.15, 's')},
+            },
+            'broadwell': {
+                'dom:mc': {'time': (58.0, None, 0.10, 's')},
+                'daint:mc': {'time': (65.0, None, 0.15, 's')},
+            },
+            'zen2': {
+                'eiger:mc': {'time': (100.0, None, 0.10, 's')},
+                'pilatus:mc': {'time': (100.0, None, 0.10, 's')},
+            },
+        },
+        16: {
+            'sm_60': {
+                'daint:gpu': {'time': (55.0, None, 0.15, 's')},
+            },
+            'broadwell': {
+                'daint:mc': {'time': (55.0, None, 0.15, 's')},
+            },
+            'zen2': {
+                'eiger:mc': {'time': (100.0, None, 0.10, 's')},
+                'pilatus:mc': {'time': (100.0, None, 0.10, 's')}
+            }
+        }
+    }
+
 
     @run_after('init')
-    def setup_by_system(self):
+    def setup_system_filtering(self):
+        self.descr += f' ({self.num_nodes} node(s))'
+
+        # setup system filter
+        valid_systems = {
+            6: ['daint:gpu', 'daint:mc', 'dom:gpu', 'dom:mc',
+                'eiger:mc', 'pilatus:mc'],
+            16: ['daint:gpu', 'daint:mc', 'eiger:mc']
+        }
+
+        try:
+            self.valid_systems = valid_systems[self.num_nodes]
+        except KeyError:
+            self.valid_systems = []
+
+        # setup programming environment filter
         if self.current_system.name in ['eiger', 'pilatus']:
             self.valid_prog_environs = ['cpeIntel']
         else:
             self.valid_prog_environs = ['builtin']
 
-        self.tags |= {
-            'maintenance' if self.variant == 'maint' else 'production'
-        }
 
     @sanity_function
     def assert_reference(self):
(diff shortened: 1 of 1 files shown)


lucamar:
I could not retrieve data for the multicore checks, because they have not been executed, as far as this warning is concerned: 
```
[](https://jenkins.cscs.ch/blue/organizations/jenkins/reframe-daint-production-daily/detail/reframe-daint-production-daily/1937/pipeline#step-15-log-23)./bin/reframe: skipping test 'VASPCpuCheck': attribute error: cscs-checks/apps/vasp/vasp_check.py:77: 'VASPCpuCheck' object has no attribute 'references_by_variant'

        self.reference = self.references_by_variant[self.variant]
```
This pull request should fix the warning, which is also occurring for LAMMPS.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -8,16 +8,19 @@
 
 
 class VASPCheck(rfm.RunOnlyRegressionTest):
-    keep_files = ['OUTCAR']
-    maintainers = ['LM']
-    tags = {'scs'}
-    strict_check = False
+    scale = parameter(['small', 'large'])
+    variant = parameter(['maint', 'prod'])
```
vkarak, line 12: Do we need to have `maint` and `prod` variants? I have removed it from other tests, because we have not been practically updating the references for the `maint` variants.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -8,16 +8,19 @@
 
 
 class VASPCheck(rfm.RunOnlyRegressionTest):
-    keep_files = ['OUTCAR']
-    maintainers = ['LM']
-    tags = {'scs'}
-    strict_check = False
+    scale = parameter(['small', 'large'])
```
vkarak, line 11: I would be more explicit here and name this parameter as `num_nodes`. Check how we do it with the Gromacs test. The problem with this approach is that the variant has different meaning for different systems which can be hard to follow.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -8,16 +8,19 @@
...
+    executable = 'vasp_std'
+    keep_files = ['OUTCAR']
     extra_resources = {
         'switches': {
             'num_switches': 1
         }
     }
+    strict_check = False
+    maintainers = ['LM']
+    tags = {'scs'}
```
vkarak, line 23: Perhaps this tag has no meaning anymore... :-(

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -41,41 +48,52 @@ def time(self):
 
 @rfm.simple_test
 class VASPCpuCheck(VASPCheck):
```
vkarak, line 50: Since you are changing the tests, I would rather make the `cpu` and `gpu` versions of the test a parameter.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -41,41 +48,52 @@ def time(self):
...
+                'daint:mc': {'time': (65.0, None, 0.15, 's')},
+                'eiger:mc': {'time': (100.0, None, 0.10, 's')},
+                'pilatus:mc': {'time': (100.0, None, 0.10, 's')}
+            },
+            'large': {
+                'daint:mc': {'time': (55.0, None, 0.15, 's')},
+                'eiger:mc': {'time': (100.0, None, 0.10, 's')},
+                'pilatus:mc': {'time': (100.0, None, 0.10, 's')}
+            }
+        }
```
vkarak, line 95: I prefer to define the references in a dictionary in the class body. The reason is that you have the significant information of the test right at the beginning.

vkarak (changes requested):

cscs-checks/apps/vasp/vasp_check.py
lucamar: I have removed the two variants and kept `tags = {'maintenance', 'production'}`

cscs-checks/apps/vasp/vasp_check.py
lucamar: Tag removed

cscs-checks/apps/vasp/vasp_check.py
lucamar: I defined the references in the class body at the beginning (using the auto-detected architectures as well)

cscs-checks/apps/vasp/vasp_check.py
lucamar: I have changed the checks to be selected based on the architecture: initially I used `constraint = parameter(['gpu', 'mc'])`, but it seems redundant. However now both `gpu` and `mc` are executed and it seems that I cannot exclude the first with the command line option `--cpu-only`.

cscs-checks/apps/vasp/vasp_check.py
lucamar: I set `num_nodes = parameter([6, 16], loggable=True)`

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
-
-    @sanity_function
-    def assert_reference(self):
-        force = sn.extractsingle(r'1 F=\s+(?P<result>\S+)',
-                                 self.stdout, 'result', float)
-        return sn.assert_reference(force, -.85026214E+03, -1e-5, 1e-5)
+    }
 
     @performance_function('s')
     def time(self):
```
vkarak, line 57: I would probably call that `elapsed_time`.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
-                'eiger:mc': {'time': (100.0, None, 0.10, 's')},
-                'pilatus:mc': {'time': (100.0, None, 0.10, 's')}
-            },
-            'large': {
-                'daint:mc': {'time': (55.0, None, 0.15, 's')},
-                'eiger:mc': {'time': (100.0, None, 0.10, 's')},
-                'pilatus:mc': {'time': (100.0, None, 0.10, 's')}
-            }
+    def setup_system_filtering(self):
+        self.descr += f' ({self.num_nodes} node(s))'
```
vkarak, line 71: I would probably set the description here or in a another post-init hook.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
-    def set_task_distribution(self):
-        self.job.options = ['--distribution=block:block']
-
-    @run_before('run')
-    def set_cpu_binding(self):
-        self.job.launcher.options = ['--cpu-bind=cores']
+        try:
+            self.valid_systems = valid_systems[self.num_nodes]
+        except KeyError:
+            self.valid_systems = []
```
vkarak, line 83: Perhaps it's better if we use `skip_if` here with a message. This will silently ignore the test:

```suggestion
        skip_if(self.num_nodes not in valid_systems, f'no valid systems found for {self.num_nodes}(s)')
        self.valid_systems = valid_systems[self.num_nodes]
```

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
-            'small': {
-                'dom:gpu': {'time': (65.0, None, 0.15, 's')},
-                'daint:gpu': {'time': (65.0, None, 0.15, 's')}
-            },
-            'large': {
-                'daint:gpu': {'time': (55.0, None, 0.15, 's')}
+    @run_before('run')
+    def setup_run(self):
+        # set auto-detected architecture
+        proc = self.current_partition.processor
```
vkarak, line 95: You need to add a call to `skip_if_no_procinfo()` here, so as to skip the test is the processor topology info is not available.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
+        arch = proc.arch
```
vkarak, line 96: Leave a blank line after this one.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
+        elif arch == 'zen2':
+            self.num_tasks_per_node = 4
+            self.variables = {
+                'MPICH_OFI_STARTUP_CONNECT': '1'
             }
+
+        # common setup for every architecture
+        self.job.launcher.options = ['--cpu-bind=cores']
+        self.job.options = ['--distribution=block:block']
+        self.num_cpus_per_task = int(proc.num_cores / self.num_tasks_per_node)
```
vkarak, line 121:
```suggestion
        self.num_cpus_per_task = proc.num_cores // self.num_tasks_per_node
```

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
+        self.num_tasks = self.num_nodes * self.num_tasks_per_node
+        self.variables = {
```
vkarak, line 123: Here you override the `variables` for the `zen2` nodes.

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,130 +7,124 @@
...
+            self.skip(f'Configuration with {self.num_nodes} node(s) '
+                      f'is not supported on {arch!r}')
+
+        # custom settings for each architecture
+        if arch == 'sm_60':
+            self.num_tasks_per_node = 1
+        elif arch == 'broadwell':
+            self.num_tasks_per_node = 2
+        elif arch == 'zen2':
+            self.num_tasks_per_node = 4
```
vkarak, line 113: It seems that this is based on the number of sockets. You can get this information with `proc.num_sockets`.

vkarak (changes requested):

cscs-checks/apps/vasp/vasp_check.py
vkarak: You need to set `num_gpus_per_node` in order for the `--cpu-only` and `--gpu-only` options to take effect; if I'm not mistaken.

lucamar: I have added `num_gpus_per_node` but it does not seem to be taken into account by `--cpu-only` or `--gpu-only`: anyway one can always use `--system daint:gpu` or `--system daint:mc` to select the architecture directly.

cscs-checks/apps/vasp/vasp_check.py
lucamar: I have modified the performance function name accordingly.

cscs-checks/apps/vasp/vasp_check.py
lucamar: I have mode the setting of the description there as suggested.

cscs-checks/apps/vasp/vasp_check.py
lucamar: I have added the call to `self.skip_if` as suggested

cscs-checks/apps/vasp/vasp_check.py
lucamar: Done

cscs-checks/apps/vasp/vasp_check.py
lucamar: Changed in the commit as suggested (I included all the suggestions in a single commit)

cscs-checks/apps/vasp/vasp_check.py
lucamar: Modified with `self.variables += ...`, thanks

cscs-checks/apps/vasp/vasp_check.py
lucamar: I have used `self.num_tasks_per_node = proc.num_sockets` as suggested, thanks

cscs-checks/apps/vasp/vasp_check.py
lucamar: Added the call to `self.skip_if_no_procinfo()` as suggested

cscs-checks/apps/vasp/vasp_check.py
```diff
@@ -7,105 +7,121 @@
...
+            'OMP_PROC_BIND': 'close'
+        }
 
+        # custom settings for selected architectures
+        if arch == 'sm_60':
+            self.num_gpus_per_node = 1
+        elif arch == 'zen2':
+            self.variables += {
+                'MPICH_OFI_STARTUP_CONNECT': '1'
+            }
```
vkarak, line 124:
```suggestion
            self.variables.update({
                'MPICH_OFI_STARTUP_CONNECT': '1'
            })
```

The `+=` does not work on dictionaries. This path has not been executed by the CI, because it is temporarily disabled for Eiger/Pilatus.

vkarak (changes requested):
Lgtm except a small bug updating the `variables`.

cscs-checks/apps/vasp/vasp_check.py
vkarak: This is strange because I just checked the filter and it indeed checks whether the `num_gpus_per_node` is greater than zero:

https://github.com/eth-cscs/reframe/blob/7b27cbe664fa50bea964202db8b0cfb61f278288/reframe/frontend/filters.py#L70-L81

And here:

https://github.com/eth-cscs/reframe/blob/7b27cbe664fa50bea964202db8b0cfb61f278288/reframe/frontend/cli.py#L954-L957

vkarak: Yes, the problem is the following. Test filtering from the command line happens after the test is initialized and *before* the test starts executing. Here you set the `num_gpus_per_node` in a pre-run hook. This is too late for the `--cpu-only` and `--gpu-only` options to take it into account. By the time of the filtering, `num_gpus_per_node` is zero, therefore they seem to have no effect. You have to set `num_gpus_per_node` in a post-init hook to be able to filter correctly.

lucamar: Thanks for your note: I set `num_gpus_per_node` in `@run_before('run')` based on `gromacs/gromacs_check.py`, where it has no effect either. I don't know how to set it in `@run_after('init')`, since `current_partition` is not yet defined there.

cscs-checks/apps/vasp/vasp_check.py
lucamar: Thanks for pointing it out!

cscs-checks/apps/vasp/vasp_check.py
vkarak: You are right about the Gromacs check and we shouldn't have been setting it there. It's easily fixable in the Gromacs check because we can differentiate the type of the test using the `nb_impl` parameter and set the `num_gpus_per_node` in a post-init hook. But in the VASP test here, I see that you run both versions (GPU and CPU) identically? We could do the same trick as with Gromacs for VASP and add a parameter to differentiate the test. I think that's a better solution than relying on the software stack to do the differentiation; the test has quite a much of implicit information if running on Daint/Dom implies running the GPU version. What do you think?

lucamar: Yes, the two  checks of VASP are using the same input files. I can define `constraint = parameter(['gpu', 'mc'])`, although it is redundant, since one can choose it anyway with the command line option `--system daint:gpu` or `--system daint:mc`.

vkarak: That's the downside in this case, that the test is not different for each variant. We can keep it as is, if we are fine with the `num_gpus_per_node` filtering limitation.

lucamar: I think it is fine to leave it as it is, since we have still the option to select the `gpu` or `mc` checks if necessary. I would remove `num_gpus_per_node` since it is not having any effect in the pre-run hook, though

vkarak (approved):

PR #2458, opened 2022-03-03
[feat] Add new command line option to distribute single node jobs on multiple cluster nodes

### LINKED ISSUE #2334: Support an alternative flexible allocation scheme that would submit unique copies of the same test on different nodes
This is different from the current flexible tests, which submits one test on multiple nodes. The problem with the current setup is that you probably need to change the benchmark you are running in order to include node information, which also makes much tougher the sanity checking as well as performance checking that would allow us to identify bad nodes.

ekouts:
In this PRs there two new features for the slurm and local partitions:

## 1. The job object includes the `pin_nodes` attribute. 

```python
    ...
    # Pin nodes for compilation jobs
    @run_before('compile')
    def pin_compilation_node(self):
        if self.local or self.build_locally:
            return

        self.build_job.pin_nodes = ['nid004']

    # Pin nodes for run jobs
    @run_before('run')
    def pin_run_node(self):
        if self.local or self.build_locally:
            return

        self.job.pin_nodes = ['nid001', 'nid002']

```

The attribute has an effect only on Slurm. It will pass `--nodelist={}` on the job scripts with an abbreviated string of the nodes.
On the other schedulers it is ignored.

## 2. The user can pass the argument `--distribute` and parameterize all the tests from the cli. 

By default ReFrame will parameterize all the selected tests to all the nodes of the valid partitions in the requested state, taking into account the cli job options.

- ReFrame will first filter all the tests based on the filtering options the user passes (`-n`, `--system` etc) and then it will dynamically create dynamic parameterized tests on each partition for all the nodes.
- We can choose the state of nodes that we want to take into account. By passing `--distribute=idle` we get all nodes in state `idle`. To get all nodes of the partition we need to pass `--distribute=all` .

Closes #2334 .


reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
...
+            # make sure that it will not be overwriten by a parent
+            # post-init hook
+            def _rfm_distributed_set_valid_sys_env(obj):
+                obj.valid_systems = [t._partition.fullname]
+                obj.valid_prog_environs = [t._environ.name]
+
+            class BaseTest(t.check.__class__):
+                _rfm_nodelist = builtins.parameter(node_map[t._partition.fullname])
+                valid_systems = [t._partition.fullname]
+                valid_prog_environs = [t._environ.name]
```
vkarak, line 221: You don't need a base class here. Just set everything in `make_tests` body argument. For `_rfm_nodelist` you can use `$nid` as the argument name. This should work, because internally attributes are set with `setattr`.

reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
...
+
+            nc = make_test(
+                f'__D_{t._partition.name}_{t._environ.name}_{basename}',
```
vkarak, line 224:
```suggestion
                f'_D_{basename}_{t._partition.fullname.replace(":", "_"}',
```

vkarak: And generally you should not access the partition from the test case like that but rather:

```python
check, partition, environ = tc
```

reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
     return curr_verbosity - quiesce
 
 
+def distribute_tests(testcases, skip_system_check, skip_prgenv_check,
+                     node_map):
+    temporary_registry = None
+    new_checks = []
+    for t in testcases:
```
vkarak, line 195:
```suggestion
    for tc in testcases:
```

reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
...
+
+            nc = make_test(
+                f'__D_{t._partition.name}_{t._environ.name}_{basename}',
+                (BaseTest, ),
+                {},
+                methods=[
+                    builtins.run_before('run')(_rfm_distributed_set_run_nodes),
+                    builtins.run_before('compile')(_rfm_distributed_set_build_nodes),
+                    # TODO this hook is not working properly
+                    # builtins.run_after('init')(_rfm_distributed_set_valid_sys_env),
```
vkarak, line 231: This probably doesn't work because the function is reused and its closure changes. You should try having a function that returns a new hook function each time.

reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
...
+                ]
+            )
+            # We have to set the prefix manually
+            nc._rfm_dynamic_test_prefix = t.check.prefix
```
vkarak, line 235: This should work I think:

```suggestion
            nc._rfm_custom_prefix = t.check.prefix
```

reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
...
+
+            for i in range(nc.num_variants):
+                # Check if this variant should be instantiated
+                var_info = copy.deepcopy(nc.get_variant_info(i, recurse=True))
+                var_info['params'].pop('_rfm_nodelist')
+                if var_info == original_var_info:
+                    if temporary_registry is None:
+                        temporary_registry = TestRegistry.create(nc, variant_num=i)
+                    else:
+                        temporary_registry.add(nc, variant_num=i)
```
vkarak, line 245: I think that this can be further simplified:

```python
# outside the main loop
test_registry = TestRegistry()

# inside the loop
for i in range(nc.num_variants):
    test_registry.add(nc, variant_num=i)
```

reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
...
+
+    if temporary_registry:
+        new_checks = temporary_registry.instantiate_all()
+        return generate_testcases(new_checks, skip_system_check,
+                                  skip_prgenv_check)
+    else:
+        return []
```
vkarak, line 252: You don't need the if/else here, just use the first branch only.

reframe/core/pipeline.py
```diff
@@ -861,9 +861,12 @@ def __new__(cls, *args, **kwargs):
...
                 except AttributeError:
-                    prefix = os.path.abspath(
-                        os.path.dirname(inspect.getfile(cls))
-                    )
+                    try:
+                        prefix = cls._rfm_dynamic_test_prefix
+                    except AttributeError:
+                        prefix = os.path.abspath(
+                            os.path.dirname(inspect.getfile(cls))
+                        )
```
vkarak, line 869: You don't need this; using `_rfm_custom_prefix` should be enough.

vkarak (changes requested):
And we should think of moving the `distribute_tests()` to a different file, so that we can test it properly.

reframe/frontend/cli.py
```diff
@@ -183,6 +188,69 @@ def calc_verbosity(site_config, quiesce):
...
+                ]
+            )
+            # We have to set the prefix manually
+            nc._rfm_dynamic_test_prefix = t.check.prefix
+
+            for i in range(nc.num_variants):
+                # Check if this variant should be instantiated
+                var_info = copy.deepcopy(nc.get_variant_info(i, recurse=True))
+                var_info['params'].pop('_rfm_nodelist')
+                if var_info == original_var_info:
```
ekouts, line 241: @vkarak I finally remembered why I needed this... In case we have a parameterized test and we want to "distribute" one testcase of it with `-n TestFoo@1` we don't want to register all the variants, but only the ones that have already been selected.

vkarak: Still I would suggest to do it in a different way without having to rely on `get_variant_info()` or deep copies. I think you need to split the loop in two:

1. Loop over the test cases and collect the test classes and their variants in a dictionary. You can get the variant number of a test through the test's  `variant_num` attribute.
2. Loop over that dictionary and register only the variants that you have found.

Finally, instantiate all the tests in the registry.

unittests/test_distribute_tests.py
```diff
@@ -15,38 +15,40 @@
...
     )
 
     testcases = list(testcases)
     assert len(testcases) == 4
-    count = sum(map(lambda x : x._partition.fullname == 'sys0:p0', testcases))
+    count = sum(map(lambda x: x._partition.fullname == 'sys0:p0', testcases))
     assert count == 2
-    count = sum(map(lambda x : x._partition.fullname == 'sys0:p1', testcases))
+    count = sum(map(lambda x: x._partition.fullname == 'sys0:p1', testcases))
     assert count == 2
```
vkarak, line 37: Why do you need this? Isn't this something done already by the `test_cli.py`?

unittests/test_distribute_tests.py
```diff
@@ -15,38 +15,40 @@
...
 
     node_map = {
         'sys0:p0': ['n1', 'n2'],
         'sys0:p1': ['n3']
     }
-    new_cases = distribute_tests(testcases, False, False, node_map)
+    new_cases = distribute_tests(testcases, node_map)
     assert len(new_cases) == 6
-    count = sum(map(lambda x : x._partition.fullname == 'sys0:p0', new_cases))
+    count = sum(map(lambda x: x._partition.fullname == 'sys0:p0', new_cases))
```
vkarak, line 45: Don't access private fields here and generally:

```suggestion
    count = sum(map(lambda x: x[1].fullname == 'sys0:p0', new_cases))
```

unittests/test_distribute_tests.py
```diff
@@ -15,38 +15,40 @@
...
     assert count == 4
-    count = sum(map(lambda x : x._partition.fullname == 'sys0:p1', new_cases))
+    count = sum(map(lambda x: x._partition.fullname == 'sys0:p1', new_cases))
     assert count == 2
     for c in new_cases:
+        nodes = getattr(c.check, '$nid')
         if c._partition.fullname == 'sys0:p0':
-            assert c.check._rfm_nodelist in ('n1', 'n2')
+            assert nodes in ('n1', 'n2')
```
vkarak, line 52: Perhaps you mean this?

```suggestion
            assert nodes == ['n1', 'n2']
```

docs/manpage.rst
```diff
@@ -379,6 +379,22 @@ Options controlling ReFrame execution
 
    .. versionadded:: 3.2
 
+.. option:: --distribute=STATE
```
vkarak, line 382:
```suggestion
.. option:: --distribute=NODESTATE
```

docs/manpage.rst
```diff
@@ -379,6 +379,22 @@ Options controlling ReFrame execution
...
+
+   Parameterize and run tests on all the nodes of their partitions.
```
vkarak, line 384: First, you need to leave a blank line after the first line, so that it serves as a short description of the option. Second, I think you should change this to something more general and elaborate later:

```suggestion
   Distribute the tests on all the nodes in state ``NODESTATE`` in their respective valid partitions.
```

docs/manpage.rst
```diff
@@ -379,6 +379,22 @@ Options controlling ReFrame execution
...
+   The attribute ``pin_nodes`` will be set in these tests in order to be submitted on the selected nodes.
```
vkarak, line 385: I don't think that this is necessary here. It is more important how the mechanism works, i.e., it parametrises the selected test with the selected nodes on each of their partitions and that it uses the `$nid` parameter. Think of what the user would like to read when trying this option. The user will see a different name of her test and a strange parameter. So you need to explain those. The `pin_nodes` is an implementation detail.

vkarak (changes requested):
The implementation is in good shape except for a bit of fine tuning. One thing we still need to understand is where to put the `distribute_tests()`. It should not be in the cli. Also I am not sure if both the `test_cli` tests and those in `test_distribute_tests` are needed.

PS: I still need to run the PR though.

reframe/frontend/cli.py
```diff
@@ -232,8 +230,8 @@ def _fn(obj):
             f'_D_{basename}_{partition.fullname.replace(":", "_")}',
```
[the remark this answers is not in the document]
ekouts: I agree that it looks better but the underscore in the beginning was meant to not clash with user's tests, so that it will not overwrite the directories etc in stage/output. To be honest I don't expect normal tests to finish with `_D_` so I think it's fine to do as you suggested.

unittests/test_distribute_tests.py
ekouts: No, it is one of nodes in every case.

ekouts:
> One thing we still need to understand is where to put the distribute_tests()

I moved `distribute_tests` and `getallnodes` to a different file `distribute_tests.py` in the frontend. What do you think?

vkarak (changes requested):
Just some minor things still in the code + the following that I discovered by running it.

1. You need to define a formatting function for the `$nid` parameter so that it is formatted in a more friendly way. Practically convert the list into an abbreviated host list.
2. I don't think that we do need the `_D_` prefix/suffix. Since we do change the name based on the partition's name.
3. The `--distribute` option should assume a default (perhaps `idle`) if no argument is passed.
4. When I run this `--distribute=idle -J nodelist='nid0000[1-3]'` the test is not parametrised on just the three nodes I pass, but on all the idle nodes of the system.

Other than that, this feature is beautiful!

unittests/test_distribute_tests.py
```diff
@@ -49,6 +49,6 @@ def test_distribute_testcases(loader, default_exec_ctx):
     for c in new_cases:
         nodes = getattr(c.check, '$nid')
         if c._partition.fullname == 'sys0:p0':
-            assert nodes in ('n1', 'n2')
+            assert nodes in (['n1'], ['n2'])
```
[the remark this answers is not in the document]
ekouts: It should be parameterized in only one of the two. At this point the parameter has one of the two values `['n1']` or `['n2']`.

unittests/test_distribute_tests.py
```diff
@@ -49,6 +49,6 @@ def test_distribute_testcases(loader, default_exec_ctx):
     for c in new_cases:
         nodes = getattr(c.check, '$nid')
         if c._partition.fullname == 'sys0:p0':
```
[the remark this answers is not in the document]
ekouts: Didn't realize that 😛 I will use it

ekouts:
> Just some minor things still in the code + the following that I discovered by running it.
> 
>     1. You need to define a formatting function for the `$nid` parameter so that it is formatted in a more friendly way. Practically convert the list into an abbreviated host list.
> 
>     2. I don't think that we do need the `_D_` prefix/suffix. Since we do change the name based on the partition's name.
> 
>     3. The `--distribute` option should assume a default (perhaps `idle`) if no argument is passed.
> 
>     4. When I run this `--distribute=idle -J nodelist='nid0000[1-3]'` the test is not parametrised on just the three nodes I pass, but on all the idle nodes of the system.
> 
> 
> Other than that, this feature is beautiful!

1. Done
2. Should I completely remove it or put it in the end? Because in a comment in the review you suggest to put `_D_` in the end.
3. Done
4. Strange.. This must be a bug in the filtering function of slurm because I had tried with `-J w=nids` and it is working.

reframe/frontend/statistics.py
```diff
@@ -227,23 +227,26 @@ def print_failure_report(self, printer):
                 f"  * Node list: {util.nodelist_abbrev(r['nodelist'])}"
             )
             job_type = 'local' if r['scheduler'] == 'local' else 'batch job'
-            jobid = r['jobid']
```
ekouts, line 230: I removed this because it's unused.

unittests/test_distribute_tests.py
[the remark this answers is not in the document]
vkarak: I don't understand this. Shouldn't the first time return `['n1']` and the second `['n2']`. Currently, it seems to return `['n2']` both times. This is how I was suggesting to write this check:

```python
    def sys0p0_nodes():
        for nodelist in (['n1'], ['n2']):
            yield nodelist

    nodelist_iter = sys0p0_nodes()
    for tc in new_cases:
        print(type(tc.check).param_space['$nid'])
        nodes = getattr(tc.check, '$nid')
        if tc.partition.fullname == 'sys0:p0':
            assert nodes == next(nodelist_iter)
        else:
            assert nodes == ['n3']
```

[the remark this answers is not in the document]
ekouts: Ok, sorry, now I understand what you mean. `new_cases` has 4 testcases in `p0`, two for each environment, that's why the first two cases have `['n2']`. I pushed a change according to your suggestion.

vkarak:
> Should I completely remove it or put it in the end? Because in a comment in the review you suggest to put _D_ in the end.

I would probably remove it completely.

reframe/frontend/statistics.py
```diff
@@ -227,23 +227,26 @@ def print_failure_report(self, printer):
...
                          f"{r['dependencies_actual']}")
             printer.info(f"  * Maintainers: {r['maintainers']}")
             printer.info(f"  * Failing phase: {r['fail_phase']}")
-            if rt.runtime().get_option('general/0/compact_test_names'):
-                cls = r['display_name'].split(' ')[0]
-                variant = r['unique_name'].replace(cls, '').replace('_', '@')
-                nameoptarg = cls + variant
-            else:
-                nameoptarg = r['unique_name']
+            if not distribute_run:
```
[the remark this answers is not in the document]
vkarak: I agree, it's too complicated and rather scheduler specific.

vkarak (changes requested):
Just a minor suggestion still.

vkarak:
> Strange.. This must be a bug in the filtering function of slurm because I had tried with -J w=nids and it is working.

Indeed it works with `-w` but not with `nodelist`. The same happens with the `-J reservation`. There's something probably wrong with the long options. I need to check.

vkarak (approved):
Looks good to me now!

PR #2460, opened 2022-03-04
[feat] Add configuredir variable to the Autotools buildsystem

jacwah:
I am writing a test suite using Reframe for a project with the following structure (simplified):

```
src/
data/
configure
reframe/checks.py
```

The project is built using Autotools.

At first I simply set `sourcesdir = '..'`, but this did not work well in practice since copying the whole source tree for each build can be quite expensive. Especially on a file system with limited quota.

With the change in this PR, I am able to do an out of source build without copying the entire source tree:

```
self.build_system.configuredir = os.path.join(self.prefix, '..')
```

Using this method, the build output is generated in the stage directory while the source files are read from the original source directory.

Please let me know what you think!

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -597,6 +597,15 @@ class Autotools(ConfigureBasedBuildSystem):
     3. Issue ``make`` to compile the code.
     '''
 
+    #: The directory of the configure script.
+    #:
+    #: This can be changed to do an out of source build without copying the
+    #: entire source tree.
+    #:
+    #: :type: :class:`str`
+    #: :default: ``"."``
+    configuredir = variable(str, value='.')
+
     def emit_build_commands(self, environ):
         prepare_cmd = []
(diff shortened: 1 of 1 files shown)


reframe/core/buildsystems.py
```diff
@@ -597,6 +597,15 @@ class Autotools(ConfigureBasedBuildSystem):
     3. Issue ``make`` to compile the code.
     '''
 
+    #: The directory of the configure script.
+    #:
+    #: This can be changed to do an out of source build without copying the
+    #: entire source tree.
+    #:
+    #: :type: :class:`str`
+    #: :default: ``"."``
```
vkarak, line 606:
```suggestion
    #: :default: ``'.'``
```

vkarak (approved):
Hi @jacwah thanks for your PR. I think it makes sense. Generally, to avoid copying large input files, you should also consider using the `readonly_files` test attribute.

PR #2461, opened 2022-03-04
Revert "[config] Use new naming scheme in CSCS config"

vkarak:
Reverts eth-cscs/reframe#2450

The problem is that it breaks our performance dashboards if we don't set fields as loggable to all the tests (not just the libraries) due to KQL's limitation to match correctly the `check_display_name`. I've tried also Lucene in to match the display name but couldn't succeed. So I'm reverting this change for the moment.

### DIFF
--- a/config/cscs-ci.py
+++ b/config/cscs-ci.py
@@ -205,8 +205,10 @@ site_configuration = {
     ],
     'general': [
         {
-            'check_search_path': ['checks/'],
-            'check_search_recursive': True
+            'check_search_path': [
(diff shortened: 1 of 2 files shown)


ekouts (approved):

vkarak:
@ekouts Thanks for approving. It may eventually not be needed, because there is an alternative way to create the graphs based on the `display_name`, which is also more flexible. I will consider this first, before merging this one.

victorusu (approved):
`lgtm`

PR #2463, opened 2022-03-05
[feat] Extend the public API of `Job`

### LINKED ISSUE #2315: Document all of the `Job` attributes
Now that since a while we have two types of jobs (build and run), we should give the opportunity to the users to modify the `build_job` directly, since the corresponding `RegressionTest` attributes affect only the run job.

See discussion in #2304.

vkarak:
This PR exposes several `Job` attributes publicly by practically documenting them. It also uses the `variable` builtin to define them and also refactors a bit the arguments passed to the constructor of a `Job`. Now, the arguments that are passed to it are only those that must be set from the frontend either directly or indirectly through the configuration.

Fixes #2315.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -1767,7 +1767,7 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
             f'systems/0/partitions/@{self.current_partition.name}/time_limit')
         )
         self.job.max_pending_time = self.max_pending_time
-        self.exclusive_access = self.exclusive_access
+        self.job.exclusive_access = self.exclusive_access
         exec_cmd = [self.job.launcher.run_command(self.job),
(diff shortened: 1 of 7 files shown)


reframe/core/pipeline.py
```diff
@@ -1768,6 +1766,8 @@ def run(self):
         self.job.time_limit = (self.time_limit or rt.runtime().get_option(
             f'systems/0/partitions/@{self.current_partition.name}/time_limit')
         )
+        self.job.max_pending_time = self.max_pending_time
+        self.exclusive_access = self.exclusive_access
```
teojgo, line 1770: Do you mean `self.job.exclusive_access`?

teojgo (changes requested):

ekouts (approved):

teojgo (approved):

PR #2469, opened 2022-03-11
[test] Update CPMD check

lucamar:
I provide an updated check for CPMD (based on #2454 for VASP), with a modified Slurm setup yielding improved performance for both `gpu` and `mc` architecture (note that the current check only runs on the `gpu` constraint of Piz Daint):
- I am setting the number of nodes for the smaller test case to 6, as we do for the other application performance checks
- the current test case does not scale beyond 72 MPI tasks and it is outdated (see warning about `XC_DRIVER IN &DFT`)

### DIFF
--- a/cscs-checks/apps/cpmd/cpmd_check.py
+++ b/cscs-checks/apps/cpmd/cpmd_check.py
@@ -93,6 +93,9 @@ class CPMDCheck(rfm.RunOnlyRegressionTest):
         # common setup for every architecture
         self.job.launcher.options = ['--cpu-bind=cores']
         self.job.options = ['--distribution=block:block']
+        # FIXME: the current test case does not scale beyond 72 MPI tasks,
+        # so the last node in 16-nodes jobs will be used only partially. 
+        # The test case needs to be updated (warning about XC_DRIVER IN &DFT)
         self.num_tasks = 72
 
         try:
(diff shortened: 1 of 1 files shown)


cscs-checks/apps/cpmd/cpmd_check.py
```diff
@@ -3,80 +3,104 @@
...
+    @performance_function('s')
+    def elapsed_time(self):
+        return sn.extractsingle(r'^ cpmd(\s+[\d\.]+){3}\s+(?P<time>\S+)',
+                                self.stdout, 'time', float)
 
     @sanity_function
     def assert_energy_diff(self):
-        #  OpenMP version of CPMD segfaults
-        #  self.variables = { 'OMP_NUM_THREADS' : '8' }
+        # OpenMP version of CPMD segfaults
```
vkarak, line 54: Do we need this comment here?

cscs-checks/apps/cpmd/cpmd_check.py
```diff
@@ -3,80 +3,104 @@
...
+            self.variables = {
+                'CRAY_CUDA_MPS': '1'
+            }
+        else:
+            arch = proc.arch
+
+        # common setup for every architecture
+        self.job.launcher.options = ['--cpu-bind=cores']
+        self.job.options = ['--distribution=block:block']
+        self.num_tasks = 72
```
vkarak, line 97: Don't you need to set `num_tasks_per_node`?

cscs-checks/apps/cpmd/cpmd_check.py
lucamar: We can remove it, it was more or less a reminder to build and test again the OpenMP executable

vkarak (approved):

PR #2479, opened 2022-03-18
[feat] Extend the syntax of `valid_systems` and `valid_prog_environs`

### LINKED ISSUE #1987: Specifying requirements, rather than e.g. valid_systems, for better test portability

### LINKED ISSUE #2312: `--skip-system-check` and `--skip-prgenv-check` don't work with fixtures
The following crashes:

```
./bin/reframe -c tutorials/fixtures/osu_benchmarks.py --skip-system-check --skip-prgenv-check -n osu_latency_test -r
```

```
  File "/users/karakasv/Devel/reframe/reframe/frontend/executors/__init__.py", line 267, in _safe_call
    return fn(*args, **kwargs)
  File "/users/karakasv/Devel/reframe/reframe/core/hooks.py", line 80, in _fn
    func(obj, *args, **kwargs)
  File "/users/karakasv/Devel/reframe/reframe/core/pipeline.py", line 2265, in setup
    self._resolve_fixtures()
  File "/users/karakasv/Devel/reframe/reframe/core/pipeline.py", line 1325, in _resolve_fixtures
    deps = deps[0]
IndexError: list index out of range
```

However it works with `-S valid_systems='*' -S valid_prog_environs='*'`.

vkarak:
This PR introduces a `features` attribute in both the environment and system partition configuration and extends the syntax of `valid_systems` and `valid_prog_environs` so that system and environment filtering happens based on features and properties rather than on system and environment names only.

## Defining features and custom properties in partitions and environments

 Currently, you can define custom properties in a system partition of environment through the `extras` property. A new property `features` is added in both that you can use for defining boolean features. Examples:

```python
            'partitions': [
                {
                    'name': 'gpu',
                    ...
                    'features': ['cuda', 'mpi'],
                    'extras': {
                        'gpu_arch': 'a100'
                    },
                ...
            ]
```

```python
    'environments': [
        {
            'name': 'PrgEnv-gnu',
            'modules': ['PrgEnv-gnu'],
            'cc': 'gcc',
            'cxx': 'g++',
            'ftn': 'gfortran',
            'features': ['cxx14'],
            'extras': {
                'foo': 1,
                'bar': 'x'
            },
        },
    ],
```

## Selecting partitions/environment with features and/or extras

Both the `valid_systems` and `valid_prog_environs` *entries* now accept the following syntax:

- `+feat`: the partition or environment has `feat`. If this appears in `valid_systems`, the framework will also pick the test if the current partition defines a resource named `feat`.
- `-feat`: the partition or environment do not have `feat`.
- `%key=val`: the partition or environment define the property `key` with value `val` in their `extras`. `val` will be converted to the type of the corresponding key in the partition or environment extras. If such conversion is not possible, there is no match.

Features can be either ANDed or ORed in the following way:

- Multiple entries in `valid_systems` or `valid_prog_environs` are ORed. For example, `valid_systems = ['+feat1', '+feat2']` is interpreted as "any partition that defines either `feat1` or `feat2` is valid."
- Multiple feature and/or key/value properties defined in the same entry are ANDed. For example, `valid_systems = ['+feat1 +feat2 %x=1']` will select all partitions that define `feat1` and `feat2` and have the `x=1` in their extras.

The current syntax remains valid, but you can't mix them in a single entry. For example, the following are invalid:

```python
valid_prog_environs = ['foo +x']  # invalid
valid_systems = ['+y sys']  # invalid
```

Generally, the `valid_systems` and `valid_prog_environs` syntax is now validated more strictly when you set them, so if the syntax is not correct, ReFrame will not allow you to set them.

## Implementation details

The way the valid system and environment combinations are populated is completely different now. In the past it was just an external product of `valid_systems` and `valid_prog_environs` values filtered by the list of supported environments at each partition. With the introduction of selectable features and properties, however, an environment might define a feature only on its instantiation for a specific partition (because it sets the `target_systems`). That means expanding the feature set to environment names and then taking the external product to generate the combinations as before is insufficient. For this reason, now the system/partition combinations are expanded together. This is taken care of by the new `reframe.core.runtime.valid_sysenv_comb()` function which returns a dictionary of partitions and their valid environments for the current system. This is used also by the fixtures in order to pick a valid partition/environment combination based on their scope. Since this function relies on the values of `valid_systems` and `valid_prog_environs` to generate the valid combinations, the `--skip-system-check` and `--skip-prgenv-check` mechanisms have changed fundamentally. These options now set the `valid_systems` and `valid_prog_environs` to `['*']` after the object is created and *before* it has its fixtures injected. The reason for that is that before injecting and registering the test's fixtures, the `valid_systems` and `valid_prog_environs` must have their final value, in order to generate the right combinations.

## Future extensions for discussion

These are a couple of extensions that we consider for the future.

### Support relational operators in key/value comparisons

This is about supporting entries such as `%x>10`, which would mean select all systems/environment that define `x` greater than 10.

The complication with this feature is that we would have to do implicit type conversions from the values in `valid_systems` etc. as well as those defined in `extras`.

### Allow selecting tests on partition processor properties

This would allow us to have `valid_systems = ['%proc.arch=broadwell']` without having to define the same property in `extras`. This could feature be combined nicely with the above one and have entries such as `valid_systems = ['%proc.num_cores>=4']`.

### Utilities to extend valid systems/environment entries

It might be common to add more constraints to the valid systems/environments specification of a (library) test that you inherit. Adding an ORed constraint is pretty straightforward, since you simply have to extend `valid_systems` or `valid_prog_environs` with `+=`, but if you want to AND a new constraint you will have to add it to the various entries or just to some of them. So a shortcut to do that could be convenient. An example is the following:

```python
valid_systems.add_constraint(r'.*gpu.*', '+v100')
```

This would mean add `+v100` to all the `valid_systems` entries that have `gpu` in them. The above syntax implies that `valid_systems` becomes a list-like object, that behaves exactly like a list but with some additional functionality. Alternatively we could have the `add_constraint` as a free function:

```python
valid_systems = add_constraint(valid_systems, r'*.gpu.*', '+v100')
```

## Todos

- [x] Update documentation
- [x] Support implicit type conversion in property values comparisons

Fixes #1987.
Fixes #2312.


ekouts (changes requested):
There is an issue with the fixtures. When the `scope='session'` this simple test fails:
```
import reframe as rfm


class Fix(rfm.RunOnlyRegressionTest):
    executable = 'hostname'

    @sanity_function
    def san(self):
        return True


@rfm.simple_test
    # f = fixture(Fix, scope='test')
    f = fixture(Fix, scope='session')

    executable = 'hostname'
    valid_systems = ['*']
    valid_prog_environs = ['+feat']

    @sanity_function
    def san(self):
        return True
```
with this error:
```
  * Reason: index error: list index out of range
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/frontend/executors/__init__.py", line 293, in _safe_call
    return fn(*args, **kwargs)
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/core/hooks.py", line 80, in _fn
    func(obj, *args, **kwargs)
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/core/pipeline.py", line 2556, in setup
    self._resolve_fixtures()
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/core/pipeline.py", line 1602, in _resolve_fixtures
    deps = deps[0]
IndexError: list index out of range
```
When `scope='test'` it runs normally. If I set the partition to `'+cpu'` it fails for both cases.

Another thing is that I don't like very much that features and extras are simple strings (so empty spaces or "special characters" like `+` etc are allowed in the configuration, but we cannot somehow select systems with this syntax. If the feature is called `+plus`, we cannot set the valid systems with `++plus`. I think we should either restrict the configuration (if possible) or extend the syntax of the valid systems. Or at least document that this case will not work. What do you think?

unittests/test_cli.py
```diff
@@ -764,6 +764,7 @@ def test_failure_stats(run_reframe):
         checkpath=['unittests/resources/checks/frontend_checks.py'],
         more_options=['-n', 'SanityFailureCheck', '--failure-stats']
     )
+    print(stdout)
```
ekouts, line 767: Remove this line

ekouts (approved):
Overall it looks good to me. We can restrict further the configuration schema, but it's also fine as it is.

vkarak:
> Overall it looks good to me. We can restrict further the configuration schema, but it's also fine as it is.

I'm working on that an I have an updated schema. I only need to check if it is possible to impose a pattern to the properties of a JSON object. I will push today.

reframe/schemas/config.json
```diff
@@ -362,10 +361,15 @@
...
+                    "extras": {
+                        "type": "object",
+                        "propertyNames": {
+                            "pattern": "^[a-zA-Z_][a-zA-Z0-9_]*$"
+                        }
+                    },
                     "features": {
                         "type": "array",
-                        "items": {"type": "string"}
+                        "items": {"$ref": "#/defs/alphanum_ext_string"}
```
ekouts, line 372: I think we want the same pattern as the extras, right? Do we want to allow `-` in features?

vkarak: That's probably an omission.

vkarak: > I think we want the same pattern as the extras, right?

No, extras' value can be non-strings. Only the keys must follow a certain pattern.

 > Do we want to allow `-` in features?

Correct point, we can't. I will change it to simple alphanum.

ekouts: Yes, I didn't write it properly but this is what I meant. To set the same pattern to extras' keys and feautures. I agree that extras' value should remain object.

vkarak: I explicitly restricted it a bit more, so as to a valid Python identifier in case we need to create easily an object out of it in the future.

unittests/test_fixtures.py
```diff
@@ -319,9 +324,26 @@ def _fixture_wrapper(**kwargs):
...
+
+        return _X()
+
+    return _make_test
+
+
 def test_fixture_registry_all(ctx_sys, simple_fixture):
     '''Test with all valid partition and environments available.'''
 
+    class MyTest(rfm.RegressionTest):
```
teojgo, line 342: You can use the `simple_test` fixture here

reframe/core/runtime.py
```diff
@@ -267,6 +267,113 @@ def is_env_loaded(environ):
...
+            for subspec in spec.split(' '):
+                if subspec.startswith('+'):
+                    plus_feats.append(subspec[1:])
+                elif subspec.startswith('-'):
+                    minus_feats.append(subspec[1:])
+                elif subspec.startswith('%'):
+                    key, val = subspec[1:].split('=')
+                    props[key] = val
+
+            have_plus_feats = all(
```
teojgo, line 292: This is just a question, in case on no `plus_feats` do we consider `have_plus_feats` to be True?

reframe/core/pipeline.py
```diff
@@ -82,6 +82,19 @@
 DEPEND_FULLY = 3
 
 
+# Valid systems/environments mini-language
+_N = r'(\w[-.\w]*)'         # name
```
teojgo, line 86: Do we want to allow `.` in the names?

teojgo (changes requested):

reframe/core/pipeline.py
vkarak: Yes, because as a future extension I had in mind to allow something like `%proc.arch=broadwell` which it would query the processor object of the partition.

reframe/core/runtime.py
vkarak: Good question, but yes we need to consider it as true due to the check below:

```python
            if have_plus_feats and not have_minus_feats and have_props:
                return True
```

One of the three will have at least an item to check, so if we didn't assume `True` for empty `plus_feats`, this test would fail. Similarly the default `False` for the `any()` of the `minus_feats` serves the same purpose.

unittests/test_fixtures.py
vkarak: In the context of these unit tests we need the test just for its `valid_systems`  and `valid_prog_environs`. See it as a wrapper in this case and a normal test. The registry's `add()` method takes the fixture to be registered as a separate argument and not from the `parent_test` argument.

teojgo (approved):

PR #2485, opened 2022-03-29
[bugfix] Fix test filtering by name when multiple exact matches are requested

### LINKED ISSUE #2483: Filtering tests by name does not work for multiple values
```
./bin/reframe --system=daint -C config/cscs.py -c cscs-checks/ -n  LAMMPSCPUCheck@1 -l  -n AlltoallvTest@0
```
The above will result in 0 tests found, while passing each name separately, works fine.

teojgo:
Fixes #2483


ekouts (changes requested):
I think there is something wrong with the PR. No matter what I try all tests are listed when I pass `-n Foo@1` (even if there is no test Foo).

reframe/frontend/cli.py
```diff
@@ -920,9 +920,19 @@ def print_infoline(param, value):
...
+
+                return _fn
+
+            # Differentiate between exact names and regexes since
+            # exact names cannot be combined in a single regex
+            exact_names = [n for n in options.names if '@' in n]
+            regex_names = [n for n in options.names if '@' not in n]
+
+            names = exact_names + ['|'.join(regex_names)]
+            testcases = filter(_filter_multiple_names(names), testcases)
```
vkarak, line 935: I think that we should rather introduce a separate filter just for matching the unique names. So you should split the names into `exact` and `regex` as you do now (in a single loop preferably), join the `regex_names` with `|` and use the current `have_name()` (but remove from it the logic about matching exact names) and then call another filter for matching any of the exact names.

unittests/test_cli.py
```diff
@@ -578,11 +578,11 @@ def test_filtering_multiple_criteria(run_reframe):
     returncode, stdout, stderr = run_reframe(
         checkpath=['unittests/resources/checks'],
         action='list',
-        more_options=['-t', 'foo', '-n', 'HelloTest']
+        more_options=['-t', 'foo', '-n', 'HelloTest', '-n', 'HelloMakeTest']
```
vkarak, line 581: I don't think that this change tests what we want to test. We should test a combination of `-n Test` and `-n TestX@0` options.

vkarak (changes requested):

teojgo:
> I think there is something wrong with the PR. No matter what I try all tests are listed when I pass `-n Foo@1` (even if there is no test Foo).

Indeed, thanks for checking, the problem is that even if `regex_names` is empty the `join` results in the empty string which matches everything after

vkarak (changes requested):
I have created a new filter function that basically combines the two filtering functions that you have. I think that we still need to remove the `have_name` and replace `have_not_name` now.

vkarak:
With my suggestion we could also avoid the extra tests in the cli if we added unit tests for the new filtering function.

ekouts (approved):

vkarak (approved):

PR #2486, opened 2022-03-30
[test] Revisit LAMMPS CPU large performance

teojgo:


lucamar (approved):
Thanks @teojgo, unfortunately the performance has degraded a lot in the past week: I assume this is due to the multicore nodes, recently busier than usual

cscs-checks/apps/lammps/lammps_check.py
```diff
@@ -91,7 +91,7 @@ class LAMMPSCPUCheck(LAMMPSCheck):
             'pilatus:mc': {'perf': (5000, -0.10, None, 'timesteps/s')}
         },
         'large': {
-            'daint:mc': {'perf': (5360, -0.30, None, 'timesteps/s')},
+            'daint:mc': {'perf': (3477, -0.30, None, 'timesteps/s')},
```
vkarak, line 94: According to my latest records, this is the value I get... 

```suggestion
            'daint:mc': {'perf': (2875, -0.65, None, 'timesteps/s')},
```

lucamar: The performance might increase at the beginning of the new quarter as of tomorrow: today I could get `4326 timesteps/s` for the large check (by the way, using `--strict` did not raise a failure).

teojgo:
Closing since the performance is back to the expected one.

PR #2492, opened 2022-04-05
[doc] Improve documentation about setting performance references in tests

### LINKED ISSUE #2426: Improve documentation about setting performance reference values
From Slack discussion [here](https://reframetalk.slack.com/archives/CF8ENLUDB/p1644918073022479):

At https://reframe-hpc.readthedocs.io/en/stable/tutorial_basics.html#adding-reference-values - I always struggle to understand what the performance value tuple will result in and always resort to making a test where it fails to understand how the tuple works. I'd find it helpful to have a few examples containing the tuple and what this results as in lower and upper value. I was thinking something like:

```
Performance Tuple               Expected      Lowest accepted   Highest accepted
(25200, -0.05, 0.05, 'MB/s')    25200 MB/s    23940 MB/s        26460 MB/s
(40, None, 0.1, 'GB/s')         40 GB/s       0                 44 GB/s
(100, -0.1, None, 'Gflop/s')    100 Gflop/s   99 Gflop/s        inf
```

ekouts:
Closes #2426 .


reframe/core/pipeline.py
```diff
@@ -609,6 +609,28 @@ def pipeline_hooks(cls):
...
+    #: | (-100, -0.01, 0.02, 'unit')   | -100 unit    | -101 unit           | -98 unit             |  # noqa: E501
+    #: +-------------------------------+--------------+---------------------+----------------------+  # noqa: E501
+    #: | (-100, -0.01, None, 'unit')   | -100 unit    | -101 unit           | inf unit             |  # noqa: E501
+    #: +-------------------------------+--------------+---------------------+----------------------+  # noqa: E501
+    #: | (-100, None, 0.02, 'unit')    | -100 unit    | -inf unit           | -98 unit             |  # noqa: E501
+    #: +-------------------------------+--------------+---------------------+----------------------+  # noqa: E501
+    #:
+    #: More information in the `assert_reference function
+    #: <deferrable_functions_reference.html#reframe.utility.sanity.assert_reference>`__.  # noqa: E501
+    #:
```
vkarak, line 633: This doesn't render correctly for me:

Screenshot 2022-04-05 at 23 09 45

Also better try using simple tables instead of grid tables. It should be easier to write them.

vkarak (changes requested):

reframe/core/pipeline.py
ekouts: ah, you are right. It's the noqa comment. I will try to find another way to skip the pep8 error.

vkarak: Also, don't use `unit` as a unit but a real thing, such as `s`, `MB/s`, `ns/day` etc. And perhaps the simple tables could be more compact.

ekouts: I had `MB/s` in the beginning but I wasn't sure what unit would make sense for a `-100` value, temperature in `C`? With simple table it is still more than 90 characters.

reframe/core/pipeline.py
```diff
@@ -609,6 +609,22 @@ def pipeline_hooks(cls):
...
+    #:   (100, -0.01, 0.02, 'MB/s')   100 MB/s      99 MB/s     102 MB/s
+    #:   (100, -0.01, None, 'MB/s')   100 MB/s      99 MB/s     inf MB/s
+    #:   (100, None, 0.02, 'MB/s')    100 MB/s      -inf MB/s   102 MB/s
+    #:   (-100, -0.01, 0.02, 'C')     -100 C        -101 C      -98 C
+    #:   (-100, -0.01, None, 'C')     -100 C        -101 C      inf C
+    #:   (-100, None, 0.02, 'C')      -100 C        -inf C      -98 C
+    #:   ===========================  ============  ==========  ===========
+    #:
+    #: More information in the `assert_reference function
+    #: <deferrable_functions_reference.html#reframe.utility.sanity.assert_reference>`__.
```
ekouts, line 626: I don't know how to shorten this line, so that pep8 doesn't complaint. `#noqa` comment doesn't work, because it will appear in the docs. I could omit it completely. I don't know if `assert_reference` gives much more details.

vkarak, line 626:
```suggestion
    #: :func:`~reframe.utility.sanity.assert_reference`.
```

teojgo (approved):

vkarak (approved):

PR #2494, opened 2022-04-11
[feat] Add new build system backend for custom builds

### LINKED ISSUE #2473: Add a minimal `Custom` build system
This should allow users to build their code with a custom command. It should not do any interpretation of the current environ. This feature should be used with caution, since environment management, reproducibility and any potential side effects are all controlled by the user's custom build system.

vkarak:
Fixes #2473

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -897,6 +897,29 @@ class Spack(BuildSystem):
         return cmds
 
 
+class CustomBuild(BuildSystem):
+    '''Custom build system.
+
+    This build system backend allows users to use custom build scripts to
+    build the test code. It does not do any interpretation of the current test
+    environment and it simply runs the supplied :attr:`commands`.
+
+    Users should use this build system with caution, because environment
+    management, reproducibility and any potential side effects are all
+    controlled by the user's custom build system.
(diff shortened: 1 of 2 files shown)


ekouts (approved):
lgtm

reframe/core/buildsystems.py
```diff
@@ -897,6 +897,29 @@ def prepare_cmds(self):
         return cmds
 
 
+class CustomBuild(BuildSystem):
```
teojgo, line 900: I would rename this to `Custom` since the user is going to use:
```
build_system = 'Custom'
```

teojgo (changes requested):
Since you are touching the buildsystems, can you change the remaining string formatting which uses `%` to f=strings?

reframe/core/buildsystems.py
vkarak: Yeah, I though about this, but `build_system = 'CustomBuild'` is not bad either, is it? In the sense that you are doing "custom build" after all.

teojgo: I am fine with both.

teojgo (approved):

PR #2514, opened 2022-05-12
[feat] Add `--mode` option to GitLab CI pipeline command

rsarm:
Closes https://github.com/reframe-hpc/reframe/issues/2467

### DIFF
--- a/reframe/frontend/ci.py
+++ b/reframe/frontend/ci.py
@@ -78,9 +78,10 @@ def _emit_gitlab_pipeline(testcases, child_pipeline_opts):
     return json
 
 
-def emit_pipeline(fp, testcases, child_pipeline_opts=[], backend='gitlab'):
+def emit_pipeline(fp, testcases, child_pipeline_opts=None, backend='gitlab'):
     if backend != 'gitlab':
         raise errors.ReframeError(f'unknown CI backend {backend!r}')
 
+    child_pipeline_opts = child_pipeline_opts or []
     yaml.dump(_emit_gitlab_pipeline(testcases, child_pipeline_opts), stream=fp,
               indent=2, sort_keys=False, width=sys.maxsize)
--- a/reframe/frontend/ci.py
+++ b/reframe/frontend/ci.py
@@ -11,7 +11,7 @@ import reframe.core.exceptions as errors
 import reframe.core.runtime as runtime
 
 
-def _emit_gitlab_pipeline(testcases):
+def _emit_gitlab_pipeline(testcases, child_pipeline_opts):
     config = runtime.runtime().site_config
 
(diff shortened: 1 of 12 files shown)


vkarak (changes requested):
I wouldn't add a configuration parameter to select the `mode`. Given that we already define the `modes` having another parameter to control which `mode` to select can be confusing. I suggest removing that and instead passing the mode options to the `emit_pipeline()` function with an argument named `child_pipeline_opts`.  We will also need to expand the documentation of the `--ci-generate` option and feature and explain which options are passed through always and how a user can options to the reframe invocation in the child pipeline.

reframe/frontend/ci.py
```diff
@@ -11,7 +11,7 @@
 import reframe.core.runtime as runtime
 
 
-def _emit_gitlab_pipeline(testcases):
+def _emit_gitlab_pipeline(testcases, child_pipeline_opts):
```
vkarak, line 14:
```suggestion
def _emit_gitlab_pipeline(testcases, child_pipeline_opts=None):
```

And avoid having to change the unit test.

reframe/frontend/cli.py
```diff
@@ -1111,7 +1110,11 @@ def _case_failed(t):
             list_checks(testcases, printer)
             printer.info('[Generate CI]')
             with open(options.ci_generate, 'wt') as fp:
-                ci.emit_pipeline(fp, testcases)
+                child_pipeline_opts = []
+                if options.mode:
+                    child_pipeline_opts.append(f'--mode={options.mode}')
+
+                ci.emit_pipeline(fp, testcases, child_pipeline_opts)
```
vkarak, line 1117: I was thinking of passing directly `options.mode` here. Why did yo prefer to do it this way?

rsarm: Just to check that `--mode` was passed and avoid passing `--mode=None` to the pipeline.

vkarak: Eventually, I think it makes sense, because otherwise you would have to reparse the mode options.

reframe/frontend/ci.py
```diff
@@ -78,7 +78,7 @@ def rfm_command(testcase):
     return json
 
 
-def emit_pipeline(fp, testcases, child_pipeline_opts, backend='gitlab'):
+def emit_pipeline(fp, testcases, child_pipeline_opts=[], backend='gitlab'):
```
vkarak, line 81: It's never safe to have a mutable object as a default argument. I can fix that.

```suggestion
def emit_pipeline(fp, testcases, child_pipeline_opts=None, backend='gitlab'):
```

vkarak (approved):

PR #2516, opened 2022-05-13
[test] Align the job report test to the new `gpu_burn_check` library test

teojgo:


cscs-checks/system/jobreport/gpu_report.py
```diff
@@ -67,32 +64,33 @@ def set_sanity_patterns(self):
...
             sn.all(
-                map(lambda x, y: self.burn_time/x <= y, time_reported, usage)
+                map(lambda x, y: self.duration/x <= y/100, time_reported, usage)
             ),
-            sn.assert_ge(sn.min(time_reported), self.burn_time)
+            sn.assert_ge(sn.min(time_reported), self.duration)
         ])
 
+    @deferrable
+    def count_successful_burns(self):
```
vkarak, line 88: This does not only count them, but asserts them.

cscs-checks/system/jobreport/gpu_report.py
```diff
@@ -23,14 +23,11 @@ class gpu_usage_report_check(GpuBurn):
     '''
 
     valid_systems = ['daint:gpu', 'dom:gpu']
-    valid_prog_environs = ['PrgEnv-gnu']
+    valid_prog_environs = ['PrgEnv-nvidia']
```
vkarak, line 26: There is no need to redefine this, since the  `gpu_burn_check` is now using `+cuda` or `+hip`. Scoping the systems makes sense though, since job report is available only on those two.

vkarak (changes requested):
I have two additional general comments for the test. I don't think it's a good idea to set `num_tasks=0` and also have the `production` tag. I would rather set explicitly `num_tasks` to a value. Second, I don't think that any sort of performance function is needed for the job report. The test is to verify the correctness of the reporting.

vkarak (approved):

PR #2520, opened 2022-05-20
[bugfix] Improve detection of ReFrame test files

### LINKED ISSUE #2502: ReFrame requires an `import reframe` statement to consider a Python file as reframe test
While this works for standalone tests, it may be problematic for tests deriving from library tests. Such tests may not need to import from `reframe`, but just from the library.

vkarak:
Now not only files that import `reframe` are consider, but also files that use the `@simple_test` decorator.

Fixes #2502.

### DIFF
--- a/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
+++ b/cscs-checks/microbenchmarks/gpu/gpu_burn/gpu_burn_test.py
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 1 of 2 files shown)


ekouts (approved):

victorusu (approved):
lgtm. this is a great idea! Does it make sense to have this as a helper function that we can reuse in other tests?

victorusu:
> lgtm. this is a great idea! Does it make sense to have this as a helper function that we can reuse in other tests?

Forget it! It is already there! 😄

PR #2533, opened 2022-06-08
[bugfix] Improve test selection by their variant id

### LINKED ISSUE #2513: When selecting tests by their id, their id is taken literally as a string
In a parameterised test with more than 10 variants, you can't select a test with a single digit variant, e.g., `testA@0` will match no test, whereas `testA@00` will. The variant ID should be interpreted as an integer when matching tests by variant.

ekouts:
Fixes #2513 .

### DIFF
--- a/reframe/frontend/filters.py
+++ b/reframe/frontend/filters.py
@@ -49,11 +49,11 @@ def have_any_name(names):
     regex_matches = []
     for n in names:
         if has_compact_names and '@' in n:
-            test_name, sep, id = n.rpartition('@')
-            if id.isdigit():
-                exact_matches.append((test_name, int(id)))
-            else:
-                exact_matches.append((test_name, id))
+            test, _, variant = n.rpartition('@')
+            try:
+                exact_matches.append((test, int(variant)))
+            except ValueError:
+                exact_matches.append((test, variant))
 
         else:
             regex_matches.append(n)
--- a/reframe/frontend/filters.py
+++ b/reframe/frontend/filters.py
@@ -50,10 +50,8 @@ def have_any_name(names):
     for n in names:
         if has_compact_names and '@' in n:
             test, _, variant = n.rpartition('@')
-            try:
+            if variant.isdigit():
                 exact_matches.append((test, int(variant)))
-            except ValueError:
-                exact_matches.append((test, variant))
 
         else:
             regex_matches.append(n)
@@ -67,12 +65,7 @@ def have_any_name(names):
         # Check if we have an exact match
         for m in exact_matches:
             cls_name = type(case.check).__name__
-            if case.check.is_fixture():
-                fixt_id = case.check._rfm_fixt_data.mashup()
-                if (cls_name, fixt_id) == m:
-                    return True
-
-            elif (cls_name, case.check.variant_num) == m:
+            if (cls_name, case.check.variant_num) == m:
                 return True
 
         display_name = case.check.display_name.replace(' ', '')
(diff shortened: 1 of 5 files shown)


victorusu (approved):
`lgtm`

reframe/frontend/filters.py
```diff
@@ -49,7 +49,12 @@ def have_any_name(names):
     regex_matches = []
     for n in names:
         if has_compact_names and '@' in n:
-            exact_matches.append(n.replace('@', '_'))
+            test_name, sep, id = n.rpartition('@')
```
vkarak, line 52:
```suggestion
            test_name, sep, variant = n.rpartition('@')
```

reframe/frontend/filters.py
```diff
@@ -49,7 +49,12 @@ def have_any_name(names):
...
+            if id.isdigit():
+                exact_matches.append((test_name, int(id)))
+            else:
+                exact_matches.append((test_name, id))
```
vkarak, line 56: Why not in the pythonic way?

```suggestion
            try:
                exact_matches.append((test, int(id)))
             except ValueError:
                exact_matches.append((test, id))
```

reframe/frontend/filters.py
```diff
@@ -61,7 +66,13 @@ def have_any_name(names):
     def _fn(case):
         # Check if we have an exact match
         for m in exact_matches:
-            if m == case.check.unique_name:
+            cls_name = type(case.check).__name__
+            if case.check.is_fixture():
+                fixt_id = case.check._rfm_fixt_data.mashup()
+                if (cls_name, fixt_id) == m:
+                    return True
```
vkarak, line 73: Why should we deal with fixtures here? If you select a test, its fixtures are also selected by default. And we should not allow selecting fixtures only.

vkarak: And if we need what you have here, we should not be accessing private fields of the test...

vkarak (changes requested):

reframe/frontend/filters.py
ekouts: > And we should not allow selecting fixtures only.

Before it was working so I was trying to maintain the same behavior. I think it make sense to run just the fixture when you investigate errors. If we decide to not allow it we should remove the `Rerun with ` in the failure report.

ekouts: > And if we need what you have here, we should not be accessing private fields of the test...

Yes you are right... I was trying to avoid adding another property in the pipeline, we already have `variant_num`, `param_variant` and `fixture_variant` that are different. Should add another property about data or something that only returns the `hash_id`?

vkarak: > Before it was working so I was trying to maintain the same behavior. I think it make sense to run just the fixture when you investigate errors. If we decide to not allow it we should remove the Rerun with in the failure report.

I disagree with this. Fixtures have a meaning *only* in relation to the test that are attached to, so they should not run stand alone. The fact this was allowed before it was just an artifact, never on purpose :-)

reframe/frontend/filters.py
ekouts: The only reason to keep non-digit ids was for the fixtures so I removed the else

reframe/schemas/runreport.json
```diff
@@ -40,6 +40,7 @@
                 "fail_reason": {"type": ["string", "null"]},
                 "fail_severe": {"type": "boolean"},
                 "filename": {"type": "string"},
+                "is_fixture": {"type": "boolean"},
```
vkarak, line 43:
```suggestion
                "fixture": {"type": "boolean"},
```

vkarak (commented):
Just a minor comment still, which I can fix.

PR #2535, opened 2022-06-09
[bugfix] Skip test file when an import error is encountered

### LINKED ISSUE #2517: Run session stops when an import error is encountered in one of the tests
The correct behaviour should be to report the error and continue with the rest of the tests.

rsarm:
Closes #2517


victorusu (approved):
`lgtm`

vkarak (approved):

PR #2537, opened 2022-06-17
[feat] Do not require to set `container_platform` for running containerised tests

### LINKED ISSUE #2396: Make tests container-runtime agnostic
Currently, the test has to define the container platform and the configuration simply states how that platform is made available. Ideally, we want the test to be able to run without changes on a container platform that is entirely configured at the system level.

vkarak:
This PR improves how tests handle the `container_platform` attribute. Users are no more required to set the `container_platform` in their tests. This will be taken from the default `container_platform` defined for the current partition in the configuration. Also, this PR brings back the `workdir` attribute of the container platform and each backend emits the right option.

The documentation and the tutorial are also adapted and a new tutorial example is added which show cases a tests of GROMACS that runs both natively and from different containers.

## Implementation details

In order to allow setting the `container_platform` attributes in the class body, we have employed the following trick. We assign a bogus `ContainerPlatform` that implements its abstract methods simply by raising a `NotImplementedError`. The purpose of this container platform is to hold any attributes set by the user. During the test's setup, when we know the current system partition, we replace the `container_platform` by creating a concrete one and copying any set attributes from the bogus. This whole mechanism is completely transparent to the users, which they can also set explicitly the `container_platform` if they wish, as in the past.

Fixes #2396.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -472,6 +472,16 @@ ReFrame can launch containerized applications, but you need to configure properl
    - ``Singularity``: The `Singularity <https://sylabs.io/>`__ container runtime.
 
 
+.. js:attribute:: .systems[].partitions[].container_platforms[].default
+
+   :required: No
+
+   If set to ``true``, this is the default container platform of this partition.
+   If not specified, the default container platform is assumed to be the first in the list of :js:attr:`container_platforms`.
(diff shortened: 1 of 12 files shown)


victorusu (approved):
`lgtm`

PR #2539, opened 2022-06-21
[feat] Add new command line option to repeat selected tests

### LINKED ISSUE #2523: Add a CLI option to run selected tests repeatedly
This is now very straightforward to implement using the machinery for the `--distribute` option. Also we should be able to combine that new option with the `--distribute` one and practically have a test run repeatedly on the same node.

vkarak:
The new option is called `--repeat` and it can be combined with the `--distribute`. Behind the scenes it uses the same mechanism as the `--distribute` option to generate the new tests.

Fixes #2523.


victorusu (approved):
`lgtm`

PR #2557, opened 2022-07-04
[feat] Build configuration incrementally by allowing to combine multiple configuration files

### LINKED ISSUE #1725: Support configuration scopes
Quoting from @akesandgren:

> It would be somewhat more user friendly if it augmented the settings.py from the user with the core/settings.py. Although I would prefer if it did read and augment config files in the order -C, $HOME/..., the other ones from "Locating the Configuration File", core/settings.py.
> Or perhaps one should do it the other way around and override with whatever values comes from the read file, i.e. read core/settings.py, XDG_DATA/reframe.d/settings.py, $HOME/... (also using the XDG_USER style), and finally any -C config files.

I'm not entirely sure about chaining the `-C` option, but I'm totally in favor of supporting configuration scopes, so that you only redefine the relevant parts of the documentation.

ekouts:
This PR introduces some changes for the configuration:
- ReFrame will always start from a basic minimal configuration with a `generic` system (local scheduler and launcher) and the `builtin` environment. It also includes the basic logging handler and handlers_perflog
- The user can pass multiple configuration files and they will be added one after the other. If systems/env etc have the same name the last one will be kept.

EDIT @vkarak: This PR also adds and changes the following:

- We warn on system, partition and environment redefinitions

- All sections with unnamed objects are merged and resolved based on the
  selected system. As a result, this sections can be combined arbitrarily through
  as the chain of configuration files is loaded and the resulting internal config
  will always has a single such object, so that it is guaranteed that calls, such
  as `get('general/0/<option>')` will always yield the correct value.

- Add a special section named `handlers$` that contains the basic stream handler
  that handles the framework's output. This allows users to override the
  standard `handlers` without having to override the basic output handler.

- Only a single stdout and stderr stream handler is considered and if a second
  one is defined this is silently ignored. This was done primarily for
  backward compatibility with the current user configurations, so that their
  framework stdout handler in `handlers` will be simply ignored in favor of the
  one in the special section `handlers$`.

- Unit tests were updated and cleaned up.

Will close #1725 .

### Todos

- [x] Update the docs with the latest updates related to this PR.


vkarak (changes requested):
We also need a unit test to explicitly test the `update_config()` (your `add_config()` currently) as well as to update the docs. Also regarding the order that the various sections are combined we should be careful, since in several places in the code we assume the place 0 as the "default", since we don't expect to have another entry (see for example `general/0/option-x`). I've seen that you had tweak the unit tests to pass specifically for this and this can be dangerous. Perhaps we need place 0 to be the latest, but we must have a look at it more thoroughly. Playing with the `--show-config` option should help with this.

vkarak (commented):
I think we need to focus a bit more on how we intend to use this feature.

vkarak:
Regarding the usage of this feature, I was thinking the following:

1. We should not change`-C` to accept multiple configuration files.
2. Have a `RFM_CONFIG_PATH` type of variable that reframe will look for a `settings.py` or `settings.json` file in each path and load it in order.
3. The `-C config_file` will replace completely the configuration chain as of now unless it is written as `-C +config_file` in which case it will be applied incrementally on top of the configurations from the `RFM_CONFIG_PATH`.
4. We could also cover the case of different configuration files per system by interpreting the `--system` option and then looking for a specific configuration file (we lose the auto-selection in this case, though).

@ekouts @victorusu What do you think?

ekouts:
> Regarding the usage of this feature, I was thinking the following:
> 
>     1. We should not change`-C` to accept multiple configuration files.
> 
>     2. Have a `RFM_CONFIG_PATH` type of variable that reframe will look for a `settings.py` or `settings.json` file in each path and load it in order.
> 
>     3. The `-C config_file` will replace completely the configuration chain as of now unless it is written as `-C +config_file` in which case it will be applied incrementally on top of the configurations from the `RFM_CONFIG_PATH`.
> 
>     4. We could also cover the case of different configuration files per system by interpreting the `--system` option and then looking for a specific configuration file (we lose the auto-selection in this case, though).
> 
> 
> @ekouts @victorusu What do you think?

1. I am fine with that, the original suggestion from the issue was `-C file1:file2:file3` so this is why I started with multiple files.
2. I guess we also accept full path of configurations? Or do we want only files named `settings.py`?
3. 👍
4. I don't understand what you mean here. How we would look for the right config file?
5. What happens now if the user doesn't give a config at all? We used to look by default in `${HOME}/.reframe/settings.{py,json}`, `${RFM_INSTALL_PREFIX}/settings.{py,json}`, `/etc/reframe.d/settings.{py,json}`. I guess we keep that the same?

victorusu:
> Regarding the usage of this feature, I was thinking the following:
> 
> 1. We should not change`-C` to accept multiple configuration files.
> 2. Have a `RFM_CONFIG_PATH` type of variable that reframe will look for a `settings.py` or `settings.json` file in each path and load it in order.
> 3. The `-C config_file` will replace completely the configuration chain as of now unless it is written as `-C +config_file` in which case it will be applied incrementally on top of the configurations from the `RFM_CONFIG_PATH`.
> 4. We could also cover the case of different configuration files per system by interpreting the `--system` option and then looking for a specific configuration file (we lose the auto-selection in this case, though).
> 
> @ekouts @victorusu What do you think?

1. My issue with multiple files is the consistency... We have a plethora of different ways to pass various options. For instance,  the `-J` option can be passed multiple times to aggregate different job options. If we use the `-C` with the proposed syntax `-C file1:file2:file3`, then it becomes inconsistent. I prefer to avoid accepting multiple files in a single option. We need to review this anyway for all the other options to make everything consistent. Probably, we should move to action-based commands, as the "new" tools use in the wild, as proposed by @teojgo. 
2. If I understand correctly, the `RFM_CONFIG_PATH` environment will accept the syntax `path1:path2:path3`, as is the normality for BASH. Then the `settings.py` and `settings.json` can be provided in those paths. I vote for that!
3. I am not so sure if we should do the `+` syntax here... I would vote for consistency and use the same construction that we use for `-J`, based on the arguments I gave on 1. 
4. I agree with this. I am not sure how this will be used in practice, but this is in principle the right way of doing it.

ekouts:
@victorusu about (1) I think @vkarak suggests to keep `-C` as it was before, it will accept only one file and the user cannot pass multiple config file from the command line, so no inconsistency there. I only mentioned the user's suggestion since the intention was to give multiple config files from the cli. I think the env var would be enough.
About (2) I assume it would make sense to have config files with the machine name instead of different directories, like `config/dom.py`, `config/daint.py` etc instead of `dom/settings.py`, `daint/settings.py`, etc.. But if you both think it's better I am fine with it.

vkarak:
> I guess we also accept full path of configurations? Or do we want only files named settings.py?

I would say no and just use directories with `settings.py` inside them, so as to be consistent with other PATH variables in Linux, as @victorusu explained.

> I don't understand what you mean here. How we would look for the right config file?

That's easy; we simply look for a specific config file pattern that includes the system name, e.g., `settings-daint.py` instead of `settings.py`. But as @victorusu pointed out that it's not clear how we use that in practice, we can wait for this behaviour until we have a concrete example.

> What happens now if the user doesn't give a config at all? We used to look by default in ${HOME}/.reframe/settings.{py,json}, ${RFM_INSTALL_PREFIX}/settings.{py,json}, /etc/reframe.d/settings.{py,json}. I guess we keep that the same? But then the user still needs to write all parts of his configurations, they cannot avoid the logging unless they know the path to the reframe/core/settings and use RFM_CONFIG_PATH.

For your first part of the question, the answer is yes. We will put those in the path. For the second, the user only needs to do `-C +myconfig`. But, frankly, it might be better to *always* include the `reframe/core/settings.py` regardless of whether the path is followed or not.

> I am not so sure if we should do the + syntax here... I would vote for consistency and use the same construction that we use for -J, based on the arguments I gave on 1.

I cannot understand this very well.

vkarak:
I am a bit puzzled of what is the best for the default behaviour of the `-C` option, because eventually, I believe that we should always build incrementally on top of `reframe/core/settings.py`. So the most consistent behaviour would be for `-C` to add to the configuration to those found in the path, including the builtin. We could have a syntax such as `-C :myconfig.py` (or pick one) that would allow users to tell reframe to use their config file exclusively. What do you think?

victorusu:
> I am a bit puzzled of what is the best for the default behaviour of the `-C` option, because eventually, I believe that we should always build incrementally on top of `reframe/core/settings.py`. So the most consistent behaviour would be for `-C` to add to the configuration to those found in the path, including the builtin. We could have a syntax such as `-C :myconfig.py` (or pick one) that would allow users to tell reframe to use their config file exclusively. What do you think?

I agree that `-C` should always build incrementally. Than we should allow to pass multiple `-C`s... Which means that this cmdline would be acceptable `-C config1.py -C config2.json -C config3.py` 

Having said that, I am not sure if the `-C :myconfig.py` is the right syntax. Probably yes, but I am not so sure...
When passing multiple options one would do the following... 
```
reframe -C config1.py -C config2.json -C :config3.py
```
Is it ok?

victorusu:
> > I am a bit puzzled of what is the best for the default behaviour of the `-C` option, because eventually, I believe that we should always build incrementally on top of `reframe/core/settings.py`. So the most consistent behaviour would be for `-C` to add to the configuration to those found in the path, including the builtin. We could have a syntax such as `-C :myconfig.py` (or pick one) that would allow users to tell reframe to use their config file exclusively. What do you think?
> 
> I agree that `-C` should always build incrementally. Than we should allow to pass multiple `-C`s... Which means that this cmdline would be acceptable `-C config1.py -C config2.json -C config3.py`
> 
> Having said that, I am not sure if the `-C :myconfig.py` is the right syntax. Probably yes, but I am not so sure... When passing multiple options one would do the following...
> 
> ```
> reframe -C config1.py -C config2.json -C :config3.py
> ```
> 
> Is it ok?

The person who wrote this (👆) doesn't know how to write in English... Sorry for that... The correct spelling is the following... 

> I am a bit puzzled of what is the best for the default behaviour of the `-C` option, because eventually, I believe that we should always build incrementally on top of `reframe/core/settings.py`. So the most consistent behaviour would be for `-C` to add to the configuration to those found in the path, including the builtin. We could have a syntax such as `-C :myconfig.py` (or pick one) that would allow users to tell reframe to use their config file exclusively. What do you think?

 I agree that `-C` should always build incrementally. Then we should allow passing multiple `-C`s... This means that this command line would be acceptable `-C config1.py -C config2.json -C config3.py`

Having said that, I am not sure if the `-C :myconfig.py` is the right syntax. Probably yes, but I am not so sure... When passing multiple options one would do the following...

```
reframe -C config1.py -C config2.json -C :config3.py
```

Is it ok? What do you think?

vkarak:
If we say that the `-C` option can be specified multiple times, then we don't need the `RFM_CONFIG_PATH`. Regarding the `-C :myconfig.py` option, we need a way to break the chain, that's why I propose it. If a `:config` entry appears, then this will supersede all configs previously loaded. Let me think a bit about it, cos this is another breaking change.

unittests/test_config.py
```diff
@@ -12,10 +12,34 @@
...
     config.load_config('reframe/core/settings.py')
 
 
+def test_load_multiple_configs():
+    site1 = config.load_config('unittests/resources/config/settings.py')
+    site2 = config.load_config(
+        'unittests/resources/config/settings-part1.py',
+        'unittests/resources/config/settings-part2.py',
+        'unittests/resources/config/settings-part3.py'
+    )
```
vkarak, line 39: I would probably generate those on the fly through a pytest fixture if we also need to reuse them. It's too much hassle to keep four (!) almost the same configuration files. You load as raw json the `resources/settings.py`, change some stuff and then dump it again with a separate name.

unittests/test_config.py
```diff
@@ -12,10 +12,34 @@
...
+    assert site1._site_config == site2._site_config
```
vkarak, line 40: Unit tests should assert API behaviour. The `_site_config` is an implementation detail and should not be read like that... Better add a couple for assertions for the things that you expect to have changed.

vkarak (changes requested):

unittests/test_config.py
```diff
@@ -7,37 +7,60 @@
 import pytest
 
 import reframe.core.config as config
+import reframe.utility as util
 from reframe.core.exceptions import ConfigError
 from reframe.core.systems import System
 
 
+@pytest.fixture
+def part_config_files(tmp_path):
```
vkarak, line 16: What about?

```suggestion
def generate_partial_configs(tmp_path):
```

unittests/test_config.py
```diff
@@ -7,37 +7,60 @@
...
+    }
+    config_3 = {
+        'systems': full_config['systems'][:-3],
+        'logging': full_config['logging'],
+        'general': full_config['general'][3:],
+    }
+    part1.write_text(f'site_configuration = {config_1!r}')
+    part2.write_text(f'site_configuration = {config_2!r}')
+    part3.write_text(f'site_configuration = {config_3!r}')
+
```
vkarak, line 42: Perhaps remove this line.

reframe/frontend/cli.py
```diff
@@ -709,28 +709,10 @@ def restrict_logging():
...
-                if f == '':
-                    conf_file = []
-                elif f.startswith(':'):
-                    conf_file = [f[1:]]
-                else:
-                    conf_file.append(f)
-
-        site_config = config.load_config(*conf_file)
+        conf_files = config.find_config_files(
+            options.config_path, options.config_file
```
vkarak, line 713: I think that you should get the `config_path` from the runtime, don't you? Since it can also be set from the `RFM_CONFIG_PATH` environment variable.

reframe/core/config.py
```diff
@@ -440,6 +439,31 @@ def select_subconfig(self, system_fullname=None,
         self._subconfigs[system_fullname] = local_config
 
 
+def find_config_files(config_path=None, config_file=None):
+    res = []
+    if config_path:
+        for p in config_path:
+            if os.path.exists(p + '/settings.py'):
```
vkarak, line 446:
```suggestion
            if os.path.exists(os.path.join(p, '/settings.py')):
```

reframe/core/config.py
```diff
@@ -440,6 +439,31 @@ def select_subconfig(self, system_fullname=None,
...
+                res.append(p + '/settings.py')
```
vkarak, line 447: Actually, why don't you construct the path once at the beginning of the for loop?

reframe/core/config.py
```diff
@@ -440,6 +439,31 @@ def select_subconfig(self, system_fullname=None,
...
+            elif os.path.exists(p + '/settings.json'):
+                res.append(p + '/settings.json')
+            else:
+                getlogger().debug('Loading the generic configuration')
```
vkarak, line 451:
```suggestion
        for p in config_path:
            py_config = os.path.join(p, 'settings.py')
            json_config = os.path.join(p, 'settings.json')
            if os.path.exists(py_config):
                res.append(py_config)
            elif os.path.exists(json_config):
                res.append(json_config)
```

I think this debug message makes more sense to be emitted from the actual loader, that will eventually fall back to the builtin config.

reframe/frontend/cli.py
ekouts: Currently the user can only set it from the variable (and `options.config_path` will have the env var). Do you think we should add also a cli option, something like `--config-path`?

reframe/frontend/cli.py
```diff
@@ -945,7 +948,7 @@ def print_infoline(param, value):
         f"{session_info['user'] or '<unknown>'}@{session_info['hostname']}"
     )
     print_infoline('working directory', repr(session_info['workdir']))
-    print_infoline('settings file', f"{session_info['config_file']!r}")
+    print_infoline('settings files', f"{session_info['config_files']!r}")
```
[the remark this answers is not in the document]
ekouts: What do you think of
```bash
settings files:    '<builtin>', 'config/test_settings.py'
```
? I simply removed the brackets but let me know if you have a better idea.

reframe/core/config.py
ekouts: This is actually an unfortunate copy-paste 😛 I wanted to print a message that says that this path will be omitted. I changed it, let me know if you think we should skip it

ekouts:
@jenkins-cscs retry all

reframe/core/config.py
vkarak: I think we could omit the log message here altogether as there are different places where we can see which configuration files were picked.

reframe/frontend/cli.py
```diff
@@ -921,7 +921,10 @@ def print_infoline(param, value):
         f"{session_info['user'] or '<unknown>'}@{session_info['hostname']}"
     )
     print_infoline('working directory', repr(session_info['workdir']))
-    print_infoline('settings files', f"{session_info['config_files']!r}")
+    print_infoline(
+        'settings files',
+        f"{', '.join(map(lambda x: repr(x), session_info['config_files']))}"
```
vkarak, line 926:
```suggestion
        ', '.join(repr(x) for x in session_info['config_files'])
```

reframe/frontend/cli.py
vkarak: You are right, I was confused!

reframe/core/config.py
ekouts: I am worried that users may not notice this the directory needs to have a `settings.py` or json file in the directory. As it is now with -vv I am getting this output:
```bash
Loading user configuration
No 'settings.py' or 'settings.json' found in 'wrong_path', path will be ignored
Loading the generic configuration
Detecting system using method: 'hostname'
...
```
Without the message `No 'settings.py' or 'settings.json' found in 'wrong_path', path will be ignored` I think it can be hard to understand what you are doing wrong.

vkarak: Ok, I agree, it  helps.

vkarak (commented):
Lgtm overall. I could groom the docs a bit and it'll be ready to go.

vkarak (changes requested):
There are still a couple of things that we need to address (especially for the docs):

1. I suggest renaming the `-C` option to `--config-files` and add a `--config-file` option as an alias with a deprecation warning. This will affect those users that are using the long option, but not those using the short one. The same we can do with the `RFM_CONFIG_FILE` envvar.
2. We need to adapt the documentation of the option and the various environment variables in the man page. It is there that we should detail their syntax and its meaning.
3. I think it'd be better if we also adapted the tutorial to reflect this new feature, as it will certainly make things easier. And perhaps we need to rethink whether the `configure.rst` is relevant as it is written now, because it assumes that we have a single config file. I updated already its introduction to reflect the changes of this PR, but I haven't read through it all again to see if it's still consistent and makes sense.

docs/configure.rst
```diff
@@ -56,9 +51,12 @@ For the complete listing and description of all configuration options, you shoul
    :start-after: # rfmdocstart: site-configuration
    :end-before: # rfmdocend: site-configuration
 
-There are three required sections that each configuration file must provide: ``systems``, ``environments`` and ``logging``.
+There are three required sections that each configuration must provide: ``systems``, ``environments`` and ``logging``.
```
ekouts, line 54: The reason I had removed this completely is because they are not required anymore, from the user's perspective. He can skip any of them and ReFrame won't complain.

vkarak: Yes, I understood this, but essentially they are still required to be present in the final configuration, therefore defined somewhere. That's why I kept it and rephrased it.

reframe/frontend/cli.py
```diff
@@ -502,10 +502,16 @@ def main():
...
     # Miscellaneous options
     misc_options.add_argument(
-        '-C', '--config-file', action='append', metavar='FILE',
+        '--config-file', action='append', metavar='FILE',
+        dest='config_file_alias',
+        help="Deprecated, alias of `--config-files'",
+        envvar='RFM_CONFIG_FILE :'
+    )
+    misc_options.add_argument(
+        '-C', '--config-files', action='append', metavar='FILE',
```
vkarak, line 511: Now I just realised that this option is `append`, so it can't be named `--config-files`. I thought we were passing all the configs with a single argument. So we can keep the `--config-file` name. Sorry for the back and forth! But using `RFM_CONFIG_FILES` instead of `RFM_CONFIG_FILE` is correct.

docs/manpage.rst
```diff
@@ -775,11 +775,18 @@ It does so by leveraging the selected system's environment modules system.
...
+   The user can pass multiple configuration files that will be added on top of the ``${RFM_INSTALL_PREFIX}/reframe/core/settings.py``.
+   To ignore previous configuration files you need to pass ``-C :new_config.py``
 
-   This option can also be set using the :envvar:`RFM_CONFIG_FILE` environment variable.
+   This option can also be set using the :envvar:`RFM_CONFIG_FILES` environment variable.
+
+   ReFrame first loads the builtin config unconditionally and then starts to look for configs in the :envvar:`RFM_CONFIG_PATH` and starts chaining them.
+   :envvar:`RFM_CONFIG_PATH` containe directories where a file named ``setting.py`` or ``setting.json`` is.
+   If both ``settings.py`` and ``settings.json`` are found, the Python file is preferred.
+   Then accumulates the configs of the command line option, potentially replacing completely those from the :envvar:`RFM_CONFIG_PATH`.
```
vkarak, line 789: We should add a `versionchanged` annotation here.

vkarak (changes requested):
I was thinking of a bit more drastic changes to the tutorial. For example, we should no more require the users to copy the builtin config and start modifying it. And instead of setting the `RFM_CONFIG_FILS`, we could show the use of `RFM_CONFIG_PATH`. What do you think?

docs/manpage.rst
```diff
@@ -778,9 +778,18 @@ Miscellaneous options
 .. option:: -C --config-file=FILE
 
    Use ``FILE`` as configuration file for ReFrame.
+   The user can pass multiple configuration files that will be added on top of the ``${RFM_INSTALL_PREFIX}/reframe/core/settings.py``.
+   To ignore previous configuration files you need to pass ``-C :new_config.py``
 
    This option can also be set using the :envvar:`RFM_CONFIG_FILE` environment variable.
 
+   ReFrame first loads the builtin config unconditionally and then starts to look for configs in the :envvar:`RFM_CONFIG_PATH` and starts chaining them.
+   :envvar:`RFM_CONFIG_PATH` containe directories where a file named ``setting.py`` or ``setting.json`` is.
```
jgphpc, line 787: contain`s`

docs/tutorial_basics.rst
```diff
@@ -652,7 +652,7 @@ We will only do so with the final versions of the tests from the previous sectio
 
 .. code-block:: console
 
-   export RFM_CONFIG_FILES=$(pwd)/tutorials/config/settings.py
+   export RFM_CONFIG_FILE=$(pwd)/tutorials/config/settings.py
```
vkarak, line 655:
```suggestion
   export RFM_CONFIG_PATH=$(pwd)/tutorials/config
```

docs/tutorial_advanced.rst
```diff
@@ -9,7 +9,7 @@ Finally, to avoid specifying the tutorial configuration file each time, make sur
 
 .. code:: bash
 
-   export RFM_CONFIG_FILES=$(pwd)/tutorials/config/settings.py
+   export RFM_CONFIG_PATH=$(pwd)/tutorials/config
```
vkarak, line 12: We will need also to change perhaps the explanation in the test.

reframe/frontend/cli.py
```diff
@@ -502,16 +502,10 @@ def main():
...
-        help="Deprecated, alias of `--config-files'",
-        envvar='RFM_CONFIG_FILE :'
-    )
-    misc_options.add_argument(
-        '-C', '--config-files', action='append', metavar='FILE',
+        '-C', '--config-file', action='append', metavar='FILE',
         dest='config_file',
         help='Set configuration file',
-        envvar='RFM_CONFIG_FILES :'
+        envvar='RFM_CONFIG_FILE :'
```
vkarak, line 508: This should remain `RFM_CONFIG_FILES`...

unittests/test_argparser.py
```diff
@@ -153,7 +153,7 @@ def test_option_precedence(default_exec_ctx, extended_parser):
         assert options.recursive is None
         assert options.timestamp == '%FT%T'
         assert options.non_default_craype is True
-        assert options.config_files is None
+        assert options.config_file is None
```
vkarak, line 156: And this should remain `config_files`.

vkarak (changes requested):
Perhaps I was not so clear before, sorry... We should revert _only_ the command line option, all the rest of the references to `files` are correct, e.g., the environment variable should be in plural and the same for the internal variable. I can try fixing that.

vkarak:
@ekouts I will also update the tutorial to avoid the step of copying the builtin config. It's good if this PR is complete.

docs/tutorial_basics.rst
vkarak: I fixed that.

docs/tutorial_advanced.rst
vkarak: I fixed that and fine tuned further the docs.

vkarak:
All open issues that we have discussed are done, but giving it a last try, I realised that although configs are nicely combined for systems and environments, the same is not the same for logging. The moment you decide to change a single thing in logging, e.g., add another logger that logs debug2 info to a file, you are obliged to rewrite the whole logging section, even the `perflog_handlers`. The same would happen in the most common situation, where somebody wants to change only the `perflog_handlers`; they would have to rewrite the whole logging section, which is not very nice...

vkarak:
@ekouts could you rerun the doc listings on Daint? I have changed substantially the way the configuration is presented and we'd need to rerun.

tutorials/basics/hello/hello2.py
```diff
@@ -15,13 +14,10 @@ class HelloMultiLangTest(rfm.RegressionTest):
...
     @run_before('compile')
     def set_sourcepath(self):
         self.sourcepath = f'hello.{self.lang}'
-    # rfmdocend: set_sourcepath
 
     @sanity_function
     def assert_hello(self):
-        return sn.assert_found(r'Hello, World\!', self.stdout)
-# rfmdocend: hellomultilang
+        return sn.assert_found(r'Hello, World\!', sn.print(self.stdout))
```
ekouts, line 23: This is changed by mistake, right?

vkarak: Most likely!

vkarak (approved):

tutorials/config/daint.py
```diff
@@ -27,7 +27,23 @@
...
+                    'container_platforms': [
+                        {
+                            'type': 'Sarus',
+                            'modules': ['sarus']
+                        },
+                        {
+                            'type': 'Singularity',
+                            'modules': ['singularity']
+                        }
+                    ]
```
vkarak, line 46: I have another config file that defines the containers: `daint-containers.py`. You should probably use that, because this will destroy the listings in the tutorial. I didn't like before that we were listing a whole big config that most of its parts were irrelevant to the context.

vkarak (changes requested):

tutorials/config/daint.py
ekouts: Where is this configuration? I didn't find it

ekouts:
@jenkins-cscs retry all

vkarak (approved):

ekouts:
@vkarak looks good to me. If the CI passes let's merge :)

PR #2565, opened 2022-07-14
[refactor] Using filter operator instead of for loops

maldil:
Using Python "filter" operators rather than Python `for` loops is more pythonic and succinct. What do you think about the change?

### DIFF
--- a/reframe/core/fixtures.py
+++ b/reframe/core/fixtures.py
@@ -336,10 +336,7 @@ class FixtureRegistry:
     def _filter_valid_environs(self, part, candidate_environs):
         ret = []
         environs = self._env_by_part[part]
-        for e in candidate_environs:
-            if e in environs:
-                ret.append(e)
-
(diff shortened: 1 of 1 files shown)


reframe/core/fixtures.py
```diff
@@ -336,10 +336,7 @@ def _filter_valid_partitions(self, candidate_parts):
     def _filter_valid_environs(self, part, candidate_environs):
         ret = []
         environs = self._env_by_part[part]
-        for e in candidate_environs:
-            if e in environs:
-                ret.append(e)
-
+        ret = list(filter(lambda e:e in environs, candidate_environs))
```
vkarak, line 339: I would probably go for a list comprehension which is even more Pythonic ;-)

vkarak (changes requested):

vkarak:
Replaced by #2573.

PR #2575, opened 2022-08-05
[feat] Add command line option to set the execution order of independent tests

### LINKED ISSUE #2548: RFE: randomize test execution order
I don't think this is possible right now. It would be useful to be able to add a CLI argument to randomize the execution order of all the tests. 

Ideally it would randomize the tests across all files passed to Reframe, e.g. with `-c test1.py -c test2.py -c test3.py` it would randomize and interleave tests between all 3 files, not just within each file. To replay particular sequence of execution, perhaps the argument could accept an optional seed for the RNG.

The rationale is that some test A might change the state of the system like making the system very hot and throttling clocks, that could cause a failure in a later test C only if you run A -> C -> B instead of always executing A -> B -> C, for example if test B is long but does not stress the system (giving the system the time to cool down).
Similarly, if the same Reframe test starts in parallel on 10 machines, they might all reach a network test at the same time, creating heavy congestion (which is useful in itself, but might not be the target of the test).

If possible, it would also be great if the performance report printed would always stay in the same order to make it easier to compare runs despite the randomized execution order.

vkarak:
This PR introduces a new command line option, `--exec-order`, that will impose an order of execution on the independent tests. This is particularly useful when you want to randomise the order of tests that have been generated with either the `--repeat` or the `--distribute` option.

Although, currently, the order of execution is implementation defined and we could simply shuffle the tests, I decided that it's better introducing a new option for a couple of reasons:

1. Shuffling the tests as an implementation detail could very easily change in the future and break the "expected" behaviour.
2. As the various actions like `--list` and `--run` happen after the test processing and after the test DAG is built, shuffling the tests, implies that the listing would also be shuffled, which is not very helpful for two reasons: (a) it's not easy to find a test in a listing (b) break the current expectation that the listing order follows the (serial) execution order.
3. Having an option to control the order makes things clearer both from the implementation's point of view but also from the user perspective.

## Implementation details

Currently, all the DAG manipulation routines and the topological sorting respect the order of the test cases that are passed as input to them, which means that simply ordering or shuffling the test cases beforehand is enough. However, there is the subtle case we need to exclude some test cases due to unresolved dependencies. We were doing that by subtracting two sets, which randomised the test cases in a new way. By replacing these sets with ordered sets, the problem is solved.

Fixes #2548.


teojgo:
@vkarak do you think instead of the `random` option it would make sense to allow passing an integer value which would then be treated as the `seed` for the shuffling?

vkarak:
@teojgo I was thinking to keep it simple for the moment. If we see the need for a seed, then we extend the arguments that are accepted.

teojgo (approved):

victorusu (approved):
`lgtm`

ekouts (approved):

PR #2578, opened 2022-08-19
[feat] Associate a unique hash with each test

### LINKED ISSUE #2577: Parameters values for filelog perflogs
Feel free to tell me that you don't want to support this use case if it's not useful for your sites.

We have a unit test that executes on each IB NIC/HCA on a node, this is done through a parameterized test:
```python
hca = parameter(["mlx5_0", "mlx5_1", "mlx5_2", "mlx5_3", "mlx5_6", "mlx5_7", "mlx5_8", "mlx5_9"])
```
With the naming scheme in 3.12.0, we could simply access the historical performance for one HCA given that the path was predictable:
```
$ cat ./logs/node0042/perflogs/ib_write_bw_loopback_mlx5_6.log
2022-08-07T09:54:13|reframe 3.12.0|ib_write_bw_loopback %hca=mlx5_6 ....
2022-08-08T10:28:11|reframe 3.12.0|ib_write_bw_loopback %hca=mlx5_6 ....
```
With 833c6ea3582bdcf4cceb6dc7a1aa667d0ca029e2, it's now a little more complex as the output file name is not predictable anymore, we would have to grep across all logs for this test to find the right HCA.

I understand the limitations of the old naming scheme mentioned in https://github.com/reframe-hpc/reframe/blob/833c6ea3582bdcf4cceb6dc7a1aa667d0ca029e2/docs/manpage.rst#test-naming-scheme, but perhaps we could have a way to modify the name of the `filelog` output file (doesn't seem possible today: https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#the-filelog-log-handler)? And then a new log record attribute that would allow to get parameters names in the file name.

vkarak:
Each test is now associated with a hash code, which is generated by the test name and the key value pairs of its parameters. This hash code is printed along with the display name with the `-l` option and tests can be selected also by their hash code with the `-n` option, by passing `-n /<hash>`. Tests do still have their unique name that is comprised by the class name and the variant number. Test-specific files associated with the test execution, such as the stage and output directories and the performance log file of the `filelog` handler use the test's hash code in their filename. Conversely to the variant number, the test's hash code does not change if new values are added to a parameter, or values are deleted or the list of values are shuffled. However, you might get filesystem conflicts if you run a parameterised test that has a parameter that takes the same value many times. Such tests must run serially to avoid this. Note that if a parameter is added to or deleted from the test or the name of parameter changes, the hash will change.

Fixes #2577.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1007,9 +1007,9 @@ The additional properties for the ``filelog`` handler are the following:
      {basedir}/
         system1/
             partition1/
-                test_name.log
+                test_short_name.log
             partition2/
-                test_name.log
+                test_short_name.log
             ...
         system2/
         ...
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -125,6 +125,8 @@ This happens recursively so that if test ``T1`` depends on ``T2`` and ``T2`` dep
 
    If the special notation ``<test_name>@<variant_num>`` is passed as the ``NAME`` argument, then an exact match will be performed selecting the variant ``variant_num`` of the test ``test_name``.
 
+   You may also select a test by its hash code using the notation ``/<test-hash>`` for the ``NAME`` argument.
+
    .. note::
 
       Fixtures cannot be selected.
@@ -132,6 +134,11 @@ This happens recursively so that if test ``T1`` depends on ``T2`` and ``T2`` dep
    .. versionchanged:: 3.10.0
 
       The option's behaviour was adapted and extended in order to work with the updated test naming scheme.
+
+   .. versionchanged:: 4.0.0
+
+      Support selecting tests by their hash code.
+
 
 .. option:: -p, --prgenv=NAME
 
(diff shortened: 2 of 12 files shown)


vkarak:
@flx42 Could you test this PR with your scenario? Now filelog performance files should not change across reframe invocations.

flx42:
Yes, I confirm it solve a the problem of multiple tests ending in the same
perflog file, thanks!
But it doesn't solve the part about being able to customize the perflog
output file to get a behavior similar to 3.x if you want to, not sure what
is your opinion on this.

vkarak:
> But it doesn't solve the part about being able to customize the perflog
output file to get a behavior similar to 3.x if you want to, not sure what
is your opinion on this.

Actually, 4.x does not change how the `filelog` handler generates the log filenames. In the `filelog` handler you could only (and still can) configure the prefix directory of the log file, but not the name of the filename. The name of the filename was composed by the test's `unique_name`, thus the ambiguity with the variant names, which is now solved with hashes. In the old naming scheme, the `unique_name` contained a stringification of the test parameters and their values, which is essentially the same as of now, except that the string representation is now different.

I would see this request as a separate feature request, especially if the current behaviour breaks your workflow or just annoys you 😂. Technically, it is possible, but it might need to add another configuration parameter to the `filelog` handler to control the log file name per se.

flx42:
Sounds good! Then I will file a separate feature request at a later time
when I'm back in the office as I'm currently travelling.

Thanks!

PR #2584, opened 2022-09-01
[testlib] memory_bandwidth.cu: Handle `size` argument larger than int

akesandgren:
The default of 1GB might be too small for newer systems to get a consistent measurement of memory bandwidth.
Need to be able to five larger values so fuse std:.stol instead of std:.stoi


vkarak (approved):

vkarak:
ok to test

PR #2592, opened 2022-09-09
[doc] Fix documentation of modules test attribute

ekouts:
We support passing dictionaries in the modules attribute, but the documentation is not up-to-date.


ekouts:
@jenkins-cscs retry all

vkarak (approved):

PR #2598, opened 2022-09-16
[feat] Add support for the Flux scheduler

vsoch:
This will add a basic scheduler that can interact with Flux: https://github.com/flux-framework/flux-core

Today is the first day I've ever used (and developed for) reframe, so I will need to ask for guidance on the best way to run tests, as either we will need a container base with flux or to mock something. The way I was able to develop and test is by adding a tutorial that I ran in a container! I also suspect I'll need to add proper documentation, if you could point me to the best places to do that. Thank you!

 - [x] Discuss testing strategy
 - [x] Documentation updates?

Signed-off-by: vsoch <vsoch@users.noreply.github.com>

### DIFF
--- a/.github/workflows/test-flux.yaml
+++ b/.github/workflows/test-flux.yaml
@@ -1,0 +1,46 @@
+name: Test Flux Scheduler
+on:
+  pull_request: []
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      packages: read
+    strategy:
+      fail-fast: false
+      matrix:
+        container: ['fluxrm/flux-sched:focal']
+
+    container:
+      image: ${{ matrix.container }}
+      options: "--platform=linux/amd64 --user root -it"
+
+    name: ${{ matrix.container }}
+    steps:
+      - name: Make Space
+        run: |        
+          rm -rf /usr/share/dotnet
+          rm -rf /opt/ghc
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Reframe
+        run: |
+          wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh           
+          /bin/bash ~/miniconda.sh -b -p /opt/conda
+          export PATH=/opt/conda:/bin:$PATH
+          /bin/bash ./bootstrap.sh
+          pip install pytest
+          export PATH=$PWD/bin:$PATH
+          which reframe
+
+      # Any additional examples added here will be tested
+      - name: Start Flux and Run Test
+        run: |
+          export PATH=$PWD/bin:$PATH
+          which reframe           
+          flux start reframe -c tutorials/flux -C tutorials/flux/settings.py -l
+          flux start reframe -c tutorials/flux -C tutorials/flux/settings.py --run
+          flux start python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv unittests/test_schedulers.py
--- a/.github/workflows/test-flux.yaml
+++ b/.github/workflows/test-flux.yaml
@@ -40,4 +40,11 @@ jobs:
           which reframe   
           flux start reframe -c tutorials/flux -C tutorials/flux/settings.py -l
           flux start reframe -c tutorials/flux -C tutorials/flux/settings.py --run
-          flux start --test-size=1 python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv
+          # For unit tests, start flux in the background
+          flux start sleep inf & FLUX_URI=$(flux uri pid:$!)
+          echo "Flux is running at URI: ${FLUX_URI}"
+          export FLUX_URI
+          flux resource info
+          python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv
+          # force stop flux instance
+          flux shutdown
--- a/.github/workflows/test-flux.yaml
+++ b/.github/workflows/test-flux.yaml
@@ -28,11 +28,8 @@ jobs:
 
       - name: Install Reframe
         run: |
-          wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh           
-          /bin/bash ~/miniconda.sh -b -p /opt/conda
-          export PATH=/opt/conda:/bin:$PATH
           /bin/bash ./bootstrap.sh
-          pip install pytest
+          python3 -m pip install pytest
           export PATH=$PWD/bin:$PATH
           which reframe
 
@@ -40,7 +37,8 @@ jobs:
       - name: Start Flux and Run Test
         run: |
           export PATH=$PWD/bin:$PATH
-          which reframe           
+          which reframe   
           flux start reframe -c tutorials/flux -C tutorials/flux/settings.py -l
           flux start reframe -c tutorials/flux -C tutorials/flux/settings.py --run
-          flux start python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv unittests/test_schedulers.py
+          flux mini run python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv unittests/test_schedulers.py::test_cancel_term_ignore unittests/test_schedulers.py::test_cancel_with_grace
+          flux mini run --env SKIP_FLAKY_CANCEL=true python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv
--- a/.github/workflows/test-flux.yaml
+++ b/.github/workflows/test-flux.yaml
@@ -40,5 +40,4 @@ jobs:
           which reframe   
           flux start reframe -c tutorials/flux -C tutorials/flux/settings.py -l
           flux start reframe -c tutorials/flux -C tutorials/flux/settings.py --run
-          flux mini run python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv unittests/test_schedulers.py::test_cancel_term_ignore unittests/test_schedulers.py::test_cancel_with_grace
-          flux mini run --env SKIP_FLAKY_CANCEL=true python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv
+          flux start --test-size=1 python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -223,6 +223,7 @@ System Partition Configuration
    Supported schedulers are the following:
 
    - ``local``: Jobs will be launched locally without using any job scheduler.
+   - ``flux``: Jobs will be launched using the `Flux Framework <https://flux-framework.org/>`_ scheduler.
    - ``oar``: Jobs will be launched using the `OAR <https://oar.imag.fr/>`__ scheduler.
    - ``pbs``: Jobs will be launched using the `PBS Pro <https://en.wikipedia.org/wiki/Portable_Batch_System>`__ scheduler.
    - ``sge``: Jobs will be launched using the `Sun Grid Engine <https://arc.liv.ac.uk/SGE/htmlman/manuals.html>`__ scheduler.
--- a/docs/tutorial_flux.rst
+++ b/docs/tutorial_flux.rst
@@ -32,7 +32,7 @@ And then reframe will again be in the local ``bin`` directory:
    # which reframe
    /code/bin/reframe
 
-Then we can run reframe with the custom config `config.py <config.py>`__
+Then we can run ReFrame with the custom config `config.py <config.py>`__
 for flux.
 
 .. code:: bash
--- a/docs/tutorial_flux.rst
+++ b/docs/tutorial_flux.rst
@@ -1,0 +1,108 @@
+========================================
+Tutorial 7: The Flux Framework Scheduler
+========================================
+
+This is a tutorial that will show how to use refame with `Flux
+Framework <https://github.com/flux-framework/>`__. First, build the
+container here from the root of reframe.
+
+.. code:: bash
+
+   $ docker build -f tutorials/flux/Dockerfile -t flux-reframe .
+
+Then shell inside, optionally binding the present working directory if
+you want to develop.
+
+.. code:: bash
+
+   $ docker run -it -v $PWD:/code flux-reframe
+   $ docker run -it flux-reframe
+
+Note that if you build the local repository, you’ll need to bootstrap
+and install again, as we have over-written the bin!
+
+.. code:: bash
+
+   ./bootstrap.sh
+
+And then reframe will again be in the local ``bin`` directory:
+
+.. code:: bash
+
+   # which reframe
+   /code/bin/reframe
+
+Then we can run reframe with the custom config `config.py <config.py>`__
+for flux.
+
+.. code:: bash
+
+   # What tests are under tutorials/flux?
+   $ cd tutorials/flux
+   $ reframe -c . -C settings.py -l
+
+.. code:: console
+
+   [ReFrame Setup]
+     version:           4.0.0-dev.1
+     command:           '/code/bin/reframe -c tutorials/flux -C tutorials/flux/settings.py -l'
+     launched by:       root@b1f6650222bc
+     working directory: '/code'
+     settings file:     'tutorials/flux/settings.py'
+     check search path: '/code/tutorials/flux'
+     stage directory:   '/code/stage'
+     output directory:  '/code/output'
+
+   [List of matched checks]
+   - EchoRandTest /66b93401
+   Found 1 check(s)
+
+   Log file(s) saved in '/tmp/rfm-ilqg7fqg.log'
+
+This also works
+
+.. code:: bash
+
+   $ reframe -c tutorials/flux -C tutorials/flux/settings.py -l
+
+And then to run tests, just replace ``-l`` (for list) with ``-r`` or
+``--run`` (for run):
+
+.. code:: bash
+
+   $ reframe -c tutorials/flux -C tutorials/flux/settings.py --run
+
+.. code:: console
+
+   root@b1f6650222bc:/code# reframe -c tutorials/flux -C tutorials/flux/settings.py --run
+   [ReFrame Setup]
+     version:           4.0.0-dev.1
+     command:           '/code/bin/reframe -c tutorials/flux -C tutorials/flux/settings.py --run'
+     launched by:       root@b1f6650222bc
+     working directory: '/code'
+     settings file:     'tutorials/flux/settings.py'
+     check search path: '/code/tutorials/flux'
+     stage directory:   '/code/stage'
+     output directory:  '/code/output'
+
+   [==========] Running 1 check(s)
+   [==========] Started on Fri Sep 16 20:47:15 2022 
+
+   [----------] start processing checks
+   [ RUN      ] EchoRandTest /66b93401 @generic:default+builtin
+   [       OK ] (1/1) EchoRandTest /66b93401 @generic:default+builtin
+   [----------] all spawned checks have finished
+
+   [  PASSED  ] Ran 1/1 test case(s) from 1 check(s) (0 failure(s), 0 skipped)
+   [==========] Finished on Fri Sep 16 20:47:15 2022 
+   Run report saved in '/root/.reframe/reports/run-report.json'
+   Log file(s) saved in '/tmp/rfm-0avso9nb.log'
+   
+For advanced users or developers, here is how to run tests within the container:
+
+Testing
+-------
+
+.. code:: console
+
+    ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py unittests/test_schedulers.py -xs
--- a/reframe/core/schedulers/flux.py
+++ b/reframe/core/schedulers/flux.py
@@ -6,17 +6,18 @@
 #
 # Flux-Framework backend
 #
-# - Initial version submitted by Vanessa Sochat, Lawrence Livermore National Lab
+# - Initial version submitted by Vanessa Sochat,
+#   Lawrence Livermore National Lab
 #
 
+import itertools
 import os
 import time
 
 import reframe.core.runtime as rt
 from reframe.core.backends import register_scheduler
-from reframe.core.exceptions import JobSchedulerError, JobError
-from reframe.core.schedulers.pbs import PbsJobScheduler
-
+from reframe.core.exceptions import JobError
+from reframe.core.schedulers import JobScheduler, Job
 
 # Just import flux once
 try:
@@ -24,18 +25,58 @@ try:
     import flux.job
     from flux.job import JobspecV1
 except ImportError:
-    flux = None
+    error = "no flux Python bindings found"
+else:
+    error = None
 
 waiting_states = ["QUEUED", "HELD", "WAITING", "PENDING"]
 
 
-@register_scheduler("flux")
-class FluxJobScheduler(PbsJobScheduler):
+class _FluxJob(Job):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._cancelled = False
+
+        # This is set by the scheduler when both the job's state is
+        # 'COMPLETED' and the job's stdout and stderr are written back
+        self._completed = False
+        self.create_job(**kwargs)
+
+    def create_job(self, **kwargs):
+        """
+        Create the flux job (and future) to watch.
+        """
+        # Generate the flux job
+        self.fluxjob = JobspecV1.from_command(
+            command=["/bin/bash", self.script_filename],
+            num_tasks=self.num_tasks_per_core or 1,
+            cores_per_task=self.num_cpus_per_task or 1,
+        )
+
+        # We must use absolute paths for Flux
+        out = os.path.join(os.path.abspath(self.workdir), self.stdout)
+        err = os.path.join(os.path.abspath(self.workdir), self.stderr)
+
+        # A duration of zero (the default) means unlimited
+        self.fluxjob.duration = self.time_limit or 0
+        self.fluxjob.stdout = out
+        self.fluxjob.stderr = err
+        self.fluxjob.cwd = os.path.abspath(self.workdir)
+        self.fluxjob.environment = dict(os.environ)
+
+    @property
+    def cancelled(self):
+        return self._cancelled
+
+    @property
+    def completed(self):
+        return self._completed
+
+
+@register_scheduler("flux", error=error)
+class FluxJobScheduler(JobScheduler):
     def __init__(self):
-        if not flux:
-            raise JobSchedulerError(
-                "Cannot import flux. Is a cluster available to you with Python bindings?"
-            )
         self._fexecutor = flux.job.FluxExecutor()
         self._submit_timeout = rt.runtime().get_option(
             f"schedulers/@{self.registered_name}/job_submit_timeout"
@@ -45,45 +86,41 @@ class FluxJobScheduler(PbsJobScheduler):
         # We don't need to submit with a file, so we don't need a preamble.
         return []
 
+    def make_job(self, *args, **kwargs):
+        return _FluxJob(*args, **kwargs)
+
     def submit(self, job):
         """
         Submit a job to the flux executor.
         """
-        # Output and error files
-        script_prefix = job.script_filename.split(".")[0]
-        output = os.path.join(job.workdir, f"{script_prefix}.out")
-        error = os.path.join(job.workdir, f"{script_prefix}.err")
-
-        # Generate the flux job
-        # Assume the filename includes a hashbang
-        # flux does not support mem_mb, disk_mb
-        fluxjob = JobspecV1.from_command(
-            command=["/bin/bash", job.script_filename],
-            num_tasks=job.num_tasks_per_core or 1,
-            cores_per_task=job.num_cpus_per_task or 1,
-        )
-
-        # A duration of zero (the default) means unlimited
-        fluxjob.duration = job.time_limit or 0
-        fluxjob.stdout = output
-        fluxjob.stderr = error
-
-        # This doesn't seem to be used?
-        fluxjob.cwd = job.workdir
-        fluxjob.environment = dict(os.environ)
-        flux_future = self._fexecutor.submit(fluxjob)
+        flux_future = self._fexecutor.submit(job.fluxjob)
         job._jobid = str(flux_future.jobid())
         job._submit_time = time.time()
         job._flux_future = flux_future
 
     def cancel(self, job):
-        # Job cannot cancel once running or completed
+        """
+        Cancel a running Flux job.
+        """
+        # Job future cannot cancel once running or completed
         if not job._flux_future.cancel():
             # This will raise JobException with event=cancel (on poll)
             flux.job.cancel(flux.Flux(), job._flux_future.jobid())
+
+        # In testing, cancel is too "soft" - or often the pid is still remaining
+        # after. We thus do a cancel in addition to a kill.
+        try:
+            flux.job.kill(flux.Flux(), job._flux_future.jobid())
+
+        # Job is inactive
+        except EnvironmentError:
+            pass
         job._is_cancelling = True
 
     def poll(self, *jobs):
+        """
+        Poll running Flux jobs for updated states.
+        """
         if jobs:
             # filter out non-jobs
             jobs = [job for job in jobs if job is not None]
@@ -93,18 +130,15 @@ class FluxJobScheduler(PbsJobScheduler):
 
         # Loop through active jobs and act on status
         for job in jobs:
-
             if job._flux_future.done():
                 # The exit code can help us determine if the job was successful
                 try:
                     exit_code = job._flux_future.result(0)
-
-                # Currently the only state we see is cancelled here
                 except flux.job.JobException:
+                    # Currently the only state we see is cancelled here
                     self.log(f"Job {job.jobid} was likely cancelled.")
                     job._state = "CANCELLED"
                     job._cancelled = True
-
                 except RuntimeError:
                     # Assume some runtime issue (suspended)
                     self.log(f"Job {job.jobid} was likely suspended.")
@@ -113,21 +147,41 @@ class FluxJobScheduler(PbsJobScheduler):
                     # the job finished (but possibly with nonzero exit code)
                     if exit_code != 0:
                         self.log(f"Job {job.jobid} did not finish successfully")
-                    job._state = "COMPLETED"
-                job._completed = True
 
+                    job._state = "COMPLETED"
+
+                job._completed = True
             elif job.state in waiting_states and job.max_pending_time:
                 if time.time() - job.submit_time >= job.max_pending_time:
                     self.cancel(job)
                     job._exception = JobError(
                         "maximum pending time exceeded", job.jobid
                     )
-
-            # Otherwise, we are still running
             else:
+                # Otherwise, we are still running
                 job._state = "RUNNING"
+
+    def allnodes(self):
+        """
+        Return a set of all free nodes.
+        """
+        rpc = flux.resource.list.resource_list(flux.Flux())
+        listing = rpc.get()
+        return {str(node) for node in listing.free.nodelist}
+
+    def filternodes(self, job, nodes):
+        raise NotImplementedError("flux backend does not support node filtering")
+
+    def wait(self, job):
+        """
+        Wait until a job is finished.
+        """
+        intervals = itertools.cycle([1, 2, 3])
+        while not self.finished(job):
+            self.poll(job)
+            time.sleep(next(intervals))
 
     def finished(self, job):
         if job.exception:
             raise job.exception
-        return job.state in ["COMPLETED", "CANCELLED", "SUSPENDED"]
+        return job.completed
--- a/reframe/core/schedulers/flux.py
+++ b/reframe/core/schedulers/flux.py
@@ -41,9 +41,9 @@ class _FluxJob(Job):
         # This is set by the scheduler when both the job's state is
         # 'COMPLETED' and the job's stdout and stderr are written back
         self._completed = False
-        self.create_job(**kwargs)
+        self.create_job()
 
-    def create_job(self, **kwargs):
+    def create_job(self):
         """
         Create the flux job (and future) to watch.
         """
@@ -115,7 +115,6 @@ class FluxJobScheduler(JobScheduler):
         # Job is inactive
         except EnvironmentError:
             pass
-        job._is_cancelling = True
 
     def poll(self, *jobs):
         """
@@ -162,12 +161,7 @@ class FluxJobScheduler(JobScheduler):
                 job._state = "RUNNING"
 
     def allnodes(self):
-        """
-        Return a set of all free nodes.
-        """
-        rpc = flux.resource.list.resource_list(flux.Flux())
-        listing = rpc.get()
-        return {str(node) for node in listing.free.nodelist}
+        raise NotImplementedError("flux backend does not support node listing")
 
     def filternodes(self, job, nodes):
         raise NotImplementedError("flux backend does not support node filtering")
--- a/reframe/core/schedulers/flux.py
+++ b/reframe/core/schedulers/flux.py
@@ -25,31 +25,22 @@ try:
     import flux.job
     from flux.job import JobspecV1
 except ImportError:
-    error = "no flux Python bindings found"
+    error = 'no flux Python bindings found'
 else:
     error = None
 
-waiting_states = ["QUEUED", "HELD", "WAITING", "PENDING"]
+WAITING_STATES = ('QUEUED', 'HELD', 'WAITING', 'PENDING')
 
 
 class _FluxJob(Job):
     def __init__(self, *args, **kwargs):
+        '''Create the flux job (and future) to watch.'''
+
         super().__init__(*args, **kwargs)
 
-        self._cancelled = False
-
-        # This is set by the scheduler when both the job's state is
-        # 'COMPLETED' and the job's stdout and stderr are written back
-        self._completed = False
-        self.create_job()
-
-    def create_job(self):
-        """
-        Create the flux job (and future) to watch.
-        """
         # Generate the flux job
         self.fluxjob = JobspecV1.from_command(
-            command=["/bin/bash", self.script_filename],
+            command=['/bin/bash', self.script_filename],
             num_tasks=self.num_tasks_per_core or 1,
             cores_per_task=self.num_cpus_per_task or 1,
         )
@@ -66,20 +57,16 @@ class _FluxJob(Job):
         self.fluxjob.environment = dict(os.environ)
 
     @property
-    def cancelled(self):
-        return self._cancelled
-
-    @property
     def completed(self):
-        return self._completed
+        return not self.state in WAITING_STATES
 
 
-@register_scheduler("flux", error=error)
+@register_scheduler('flux', error=error)
 class FluxJobScheduler(JobScheduler):
     def __init__(self):
         self._fexecutor = flux.job.FluxExecutor()
         self._submit_timeout = rt.runtime().get_option(
-            f"schedulers/@{self.registered_name}/job_submit_timeout"
+            f'schedulers/@{self.registered_name}/job_submit_timeout'
         )
 
     def emit_preamble(self, job):
@@ -90,36 +77,24 @@ class FluxJobScheduler(JobScheduler):
         return _FluxJob(*args, **kwargs)
 
     def submit(self, job):
-        """
-        Submit a job to the flux executor.
-        """
+        '''Submit a job to the flux executor.'''
+
         flux_future = self._fexecutor.submit(job.fluxjob)
         job._jobid = str(flux_future.jobid())
         job._submit_time = time.time()
         job._flux_future = flux_future
 
     def cancel(self, job):
-        """
-        Cancel a running Flux job.
-        """
+        '''Cancel a running Flux job.'''
+
         # Job future cannot cancel once running or completed
         if not job._flux_future.cancel():
             # This will raise JobException with event=cancel (on poll)
             flux.job.cancel(flux.Flux(), job._flux_future.jobid())
 
-        # In testing, cancel is too "soft" - or often the pid is still remaining
-        # after. We thus do a cancel in addition to a kill.
-        try:
-            flux.job.kill(flux.Flux(), job._flux_future.jobid())
-
-        # Job is inactive
-        except EnvironmentError:
-            pass
-
     def poll(self, *jobs):
-        """
-        Poll running Flux jobs for updated states.
-        """
+        '''Poll running Flux jobs for updated states.'''
+
         if jobs:
             # filter out non-jobs
             jobs = [job for job in jobs if job is not None]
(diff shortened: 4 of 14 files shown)


vsoch:
Maybe @jenkins-cscs ... maybe! I just need an education for how to best go about it.

vkarak:
Hi @vsoch and thanks for you contribution! And also for taking the time to provide a full tutorial as well! Regarding testing, I think that testing it through a container triggered by a Github action is totally fine. We do the same for the various modules systems backends. As for the scheduler backends, we don't use Github actions yet, because simply we didn't have a containerised workflow for them and, therefore, we're relying on the @jenkins-cscs buddy. I know that @teojgo has been experimenting with testing a Slurm installation for CI/CD with containers, so ideally, we want to move that to GH actions, too.

As for the docs, they are in RST format (sorry for that, I see your tutorial –of course, logically– is in Markdown). We will need a simple entry in the [list](https://github.com/reframe-hpc/reframe/blob/0219a7a94f79e0f49db75b6dbd9c57033c67d7bc/docs/config_reference.rst?plain=1#L2250) of available scheduler backends with the link to the Flux docs.

As for your tutorial, if you could possibly convert it to RST, that'd be great. Then you can add an entry [here](https://github.com/reframe-hpc/reframe/blob/0219a7a94f79e0f49db75b6dbd9c57033c67d7bc/docs/tutorials.rst?plain=1#L6) and it will be picked up, built and listed in the tutorials page.

vsoch:
Will get these fixes in soon! And rst is totally good - I use it for a lot of my docs as well.

vsoch:
ok! This took a hot minute, but the new workflow should be ready to test here!

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,152 @@
...
+                self.job_attrs.update((field,))
+            else:
+                self.job_attrs.update(fields2attrs[field])
+
+
+    def emit_preamble(self, job):
+        """
+        We don't need to submit with a file, so we don't need a preamble.
+        """
+        return ["echo $PWD"]
```
vkarak, line 82: I think return an empty list should be fine, unless you need the workind dir for debug purposes.

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,152 @@
...
+    def submit(self, job):
+        """
+        Submit a job to the flux executor.
+        """
+        self._prepare_job_formatter()
+
+        # Output and error files
+        script_prefix = job.script_filename.split('.')[0]
+        output = os.path.join(job.workdir, "%s.out" % script_prefix)
+        error = os.path.join(job.workdir, "%s.err" % script_prefix)
```
vkarak, line 93: We generally prefer the f-strings in new code:

```suggestion
        output = os.path.join(job.workdir, f'{script_prefix}.out')
        error = os.path.join(job.workdir, f'{script_prefix}.err')
```

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,152 @@
...
+        self._submit_timeout = rt.runtime().get_option(
+            f'schedulers/@{self.registered_name}/job_submit_timeout'
+        )
+
+    def _prepare_job_formatter(self):
+        """
+        A job doesn't have an easy status command - instead we do a job
+        listing with a particular format. See src/cmd/flux-jobs.py#L44
+        in flux-framework/flux-core for more attributes.
+        """
```
vkarak, line 47: I would probably make that an internal comment starting with `#`.

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,152 @@
...
+            if field not in fields2attrs:
+                self.job_attrs.update((field,))
+            else:
+                self.job_attrs.update(fields2attrs[field])
+
+
+    def emit_preamble(self, job):
+        """
+        We don't need to submit with a file, so we don't need a preamble.
+        """
```
vkarak, line 81: Just a standard inline comment should be better off here.

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,152 @@
...
+            "exception.type": ("exception_type",),
+            "exception.note": ("exception_note",),
+            "runtime": ("t_run", "t_cleanup"),
+            "status": ("state", "result"),
+            "status_abbrev": ("state", "result"),
+            "t_remaining": ("expiration", "state", "result"),
+        }
+
+        # Set job attributes we will use later to get job statuses
+        self.job_attrs = set()
```
vkarak, line 70: This doesn't seem to be used elsewhere, is it?

vkarak (changes requested):
Thanks again for the PR! Apart from the various minor comments, there is one additional thing that I just realised about testing. You could expand the unit tests in `test_schedules.py` and run the unit tests inside the container with your settings file using the `--rfm-user-config` and possibly the `--rfm-system` options of `./test_reframe.py`. I could also give it a try this week.

.github/workflows/test-flux.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Reframe
+        run: |
+          wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh           
+          /bin/bash ~/miniconda.sh -b -p /opt/conda
+          export PATH=/opt/conda:/bin:$PATH
+          /bin/bash ./bootstrap.sh
```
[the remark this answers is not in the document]
vsoch: yes! Generally I find it more reproducible (e.g., if someone wants to make one on their HPC and try to emulate what I'm doing) than using system Python. I usually only go for system Python if I don't expect someone is going to look at it and try to copy it (perhaps on a different OS) and/or I want to try to make the container smaller.

docs/config_reference.rst
```diff
@@ -223,6 +223,7 @@ System Partition Configuration
    Supported schedulers are the following:
 
    - ``local``: Jobs will be launched locally without using any job scheduler.
+   - ``flux``: Jobs will be launched using the `Flux Framework <https://flux-framework.org/>`_ scheduler.
```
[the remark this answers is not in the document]
vsoch: Yes when you install flux the Python bindings are installed - there was a setting that you could build without them but actually my PR to remove this entirely (because Flux needs Python) was merged this morning! So if they have flux available to use, they will have the Python bindings, generally speaking. The admin that sets up flux would have to go out of their way to make them not available, which (in my developer experience using flux) I haven't seen yet. So I think we don't need to put that comment, at least for the time being until it's a proven issue.

reframe/core/schedulers/flux.py
vsoch: This was mostly for my debugging to see where it was running - will remove it asap!

vsoch:
Okay changes are in!

> You could expand the unit tests in test_schedules.py and run the unit tests inside the container with your settings file using the --rfm-user-config and possibly the --rfm-system options of ./test_reframe.py. I could also give it a try this week.

And yes I'd be happy to do that - are there a standard set that you like to run for reframe, or should I ping the Flux core devs to see what they would use?

vkarak:
> And yes I'd be happy to do that - are there a standard set that you like to run for reframe, or should I ping the Flux core devs to see what they would use?

No, these are unit tests for reframe to test the various scheduler backends, practically, that the API is implemented correctly. They are in the `unittests/test_schedulers.py`. Apart from the Slurm-specific ones (check the `slurm_only` fixture), there are some that are meant to be valid for all schedulers. These are the `test_prepare*` unit tests which exercise the `emit_preamble()` for the various schedulers (that should be trivial for flux to implement), the `test_submit*`, the `test_cancel*` and a couple more that check some corner cases. For the `test_prepare*` family of tests you essentially need to provide an implementation of the directives that you expect in the form of `_expected_flux_directives(job)` and `_expected_flux_directives_minimal(job)`. These exercise that the backends emit the right preamble if they get a minimally defined job or one with more details. For the rest, I would simply add `flux` in the parameterized `scheduler` [fixture](https://github.com/reframe-hpc/reframe/blob/5218831c94b7dff9004078bbed96a2034996cc7d/unittests/test_schedulers.py#L29-L30) and wait to see the failures and try to fix them :-) The tests that are meant to run only for a specific backend will be skipped, so you don't need to worry about them. But you're not done yet! If you try to run the unit tests right away, only those that can be executed locally will run (the `test_prepare*` practically). You need to pass a custom config that defines a partition with a `flux` scheduler and then run the unit tests with `./test_reframe.py --rfm-user-config=flux-config.py`. Inside you container, you could also run reframe's unit tests this way before your tutorial example. I hope it's less complicated than it sounds, but l am happy to help!

vsoch:
Ok, I can take a shot hopefully soon! I'll ping you if I have any questions.

vsoch:
@vkarak I took a first shot! I got the tests to run locally so (fingers crossed) the container in the CI environment works too! 

https://github.com/reframe-hpc/reframe/pull/2598/commits/5109b8ec6ca732e8d293d64865e43269187f0664

I'm glad we did this - I was totally missing functionality to check for the max wait time and to cancel.

reframe/core/schedulers/flux.py
```diff
@@ -74,6 +76,12 @@ def submit(self, job):
         job._submit_time = time.time()
         job._flux_future = flux_future
 
+    def cancel(self, job):
+        # Job cannot cancel once running or completed
+        if not job._flux_future.cancel():
+            # This will raise JobException with event=cancel
+            flux.job.cancel(flux.Flux(), job._flux_future.jobid())
```
vkarak, line 83: If this is gonna raise, it's better if you wrap it and reraise it as a `JobError`. Reframe's runtime can cancel jobs if a timeout is reached or if the user aborts the execution. If the raised error is a `JobError` and generally any other soft error, reframe will just print its message without a stack trace, otherwise the stack trace will be printed. Generally, you can enable stack traces also for soft errors, such as `JobError` if you increase verbosity, but users wouldn't want to see a full stack trace by default if the job was not cancelled.

```python
try:
    flux.job.cancel(flux.Flux(), job._flux_future.jobid())
except flux.job.JobException as err:
    raise JobError('some message') from err
```

unittests/test_schedulers.py
```diff
@@ -148,6 +148,12 @@ def _expected_lsf_directives(job):
         f'#DW stage_in source=/foo',
     ])
 
+def _expected_flux_directives(job):
+    return set([])
```
vkarak, line 152: Just `return set()` should be fine here.

unittests/test_schedulers.py
```diff
@@ -445,11 +453,14 @@ def test_cancel(make_job, exec_ctx):
...
     assert minimal_job.finished()
-    assert t_job < 30
 
     # Additional scheduler-specific checks
     sched_name = minimal_job.scheduler.registered_name
-    if sched_name in ('slurm', 'squeue'):
+
+    # This is a little over 30
+    assert t_job < 30
+
```
vkarak, line 462: I would move that back before the `sched_name` so that the `sched_name` is used close to its assignment. Also I don't understand vey much the comment. This generally should be less than 30s, unless the submit is super slow (which can happen occasionally when testing Slurm, but that generally should not be case).

reframe/core/schedulers/flux.py
vsoch: The second command (the one you targeted) won't raise directly - as I understand it, it's a future so will only raise when we retrieve the result (in the poll function). So I don't think this is necessary here - we can bring in the flux maintainers after a first round of review to double check this.

unittests/test_schedulers.py
vsoch: Oh that comment is wrong now (and I should have removed it!) I was testing before I had this actually implemented in flux and sort of taking notes. :laughing:

vsoch:
And it looks like we need to disable running flux tests for the normal CI - how would you like to try doing that?
Also I forgot to run the tests with flux start in the container - doh!

reframe/core/schedulers/flux.py
vkarak: If that's the case, then I think that's the desired behaviour. I was confused a bit the comment and thought that this would raise.

vsoch: Sorry bad comment, I’ll tweak it for next round of pushes to make it clear that happens later!

vkarak:
> And it looks like we need to disable running flux tests for the normal CI - how would you like to try doing that? Just a figure that is based on being able to import flux?

I've seen that and I'm trying to figure out what's the way to do that. It's the first scheduler backend that uses Python bindings of some sort. One option is to skip the test if you get a `JobSchedulerError` in the `make_job` fixture. Another could be to avoid registering the backend if it's not there.

vsoch:
OK just made a tweak that (fingers crossed) should result in green! `python` isn't in the container (but `python3` is).

vsoch:
Ah so close!! Linting failed!! I added `assert flux` so the variable is used. :crossed_fingers:

vsoch:
I might need some help thinking through this one: https://github.com/reframe-hpc/reframe/actions/runs/3104434148/jobs/5037509247#step:6:449 as it works for me locally (so I cannot reproduce to debug!)

vkarak:
> I might need some help thinking through this one: https://github.com/reframe-hpc/reframe/actions/runs/3104434148/jobs/5037509247#step:6:449 as it works for me locally (so I cannot reproduce to debug!)

You were faster than me! I was about to open a PR to your branch, but you just rebased your branch and invalidated mine :-P I will rebase mine and open it now. It improves how scheduler (and generally) backends are registered, so that you get nicer messages when you try to use flux and the bindings are not there:

```
reframe: failed to initialize runtime: failed to initialize partition 'tresa:default': could not register scheduler backend: no flux Python bindings found
```

And in the unit tests:

```
unittests/test_schedulers.py::test_prepare[flux] SKIPPED (could not register scheduler backend: no flux Python bindings found)                                                                           [  0%]
```

vsoch:
Oops sorry about that! I'll wait for your PR before I take another shot!

vkarak:
My PR is done now!

vsoch:
OK - my hypothesis is that the bug was a result of adding another miniconda environment, and installing pytest to it, so then even though we use flux start, the pytest we use is from that environment and somehow can't see flux. We will find out soon!

vsoch:
okay I'm going to look closer at the `test_cancel_with_grace` test- there must be something different on my host than in the CI.

unittests/test_schedulers.py
```diff
@@ -655,7 +668,7 @@ def test_cancel_with_grace(minimal_job, scheduler, local_only):
 
     # Verify that the spawned sleep is killed, too, but back off a bit in
     # order to allow the sleep process to wake up and get the signal
-    time.sleep(0.1)
+    time.sleep(5)
```
vsoch, line 671: I still can't reproduce this locally - the tests always pass. I'm assuming there is some delay in the CI container and testing adding a longer sleep here.

vsoch: Nope that didn't work! @vkarak do you have any ideas? I can't reproduce this error locally.

.github/workflows/test-flux.yaml
```diff
@@ -0,0 +1,43 @@
...
+          which reframe
+
+      # Any additional examples added here will be tested
+      - name: Start Flux and Run Test
+        run: |
+          export PATH=$PWD/bin:$PATH
+          which reframe   
+          flux start reframe -c tutorials/flux -C tutorials/flux/settings.py -l
+          flux start reframe -c tutorials/flux -C tutorials/flux/settings.py --run
+          flux start python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv unittests/test_schedulers.py
```
vkarak, line 43: Hmm, I see you're running only the scheduler unit tests here. I'm trying to run the full suite and I'm getting another error earlier on:

```
____________________________________________________________________________________________ test_remote_autodetect ____________________________________________________________________________________________

remote_exec_ctx = <unittests.conftest._ExecutionContext object at 0x7fc809c96070>

    def test_remote_autodetect(remote_exec_ctx):
        # All we can do with this test is to trigger the remote auto-detection
        # path; since we don't know what the remote user system is, we cannot test
        # if the topology is right.
        partition = test_util.partition_by_scheduler()
        if not partition:
            pytest.skip('job submission not supported')

>       autodetect.detect_topology()

unittests/test_autodetect.py:116:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
reframe/frontend/autodetect.py:246: in detect_topology
    part._processor = ProcessorInfo(_remote_detect(part))
reframe/frontend/autodetect.py:154: in _remote_detect
    _log_contents(job.stdout)
reframe/frontend/autodetect.py:36: in _log_contents
    getlogger().debug(f'--- {filename} ---\n'
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

filename = '/code/rfm.6mxhbs0h/rfm-detect-job.out'

    def _contents(filename):
        '''Return the contents of a file.'''

>       with open(filename) as fp:
E       FileNotFoundError: [Errno 2] No such file or directory: '/code/rfm.6mxhbs0h/rfm-detect-job.out'

reframe/frontend/autodetect.py:30: FileNotFoundError
```

Have you tried the full suite? Did it work for you?

vsoch: I have not - let me give that a spin!

vkarak: Ok, I found this. The problem is that you haven't implemented the `wait()` in `FluxJobScheduler` and the fact that Python didn't complain, as it had to since the method is supposed to abstract, is because you inherit your implementation from the `PbsScheduler`, thus here reframe calls `PbsScheduler`'s `wait()`. Since Flux is a scheduler of its own, I would suggest to simply inherit from the abstract `JobScheduler` class. It could be that somehow this has messed up with your other unit test failure, perhaps not. But we should better fix that first.

unittests/test_schedulers.py
vkarak: Check my comment in your Dockerfile. I found another bug, which I don't know if it's related to this failure, but let's first fix that.

.github/workflows/test-flux.yaml
vsoch: OK reproduced! And working on it! :gear:

vsoch: okay so I took a look at wait - and actually I had checked the version in the PbsScheduler and I don't see any issue. The reason is that the wait function is just sleeping for different intervals and then running another poll, and checking for finished:

```python
    def wait(self, job):
        intervals = itertools.cycle([1, 2, 3])
        while not self.finished(job):
            self.poll(job)
            time.sleep(next(intervals))
```
And then finished it just checking for an exception and returning if the job is completed (and actually we implement this in flux to check for finished based on the job state

```python
    def finished(self, job):
        if job.exception:
            raise job.exception

        return job.state in ['COMPLETED', 'CANCELLED', 'SUSPENDED']```
```
I'm trying to dig into the failure you showed above - it looks like this raises the error:

```python
autodetect.detect_topology()
```
Will report back in a few minutes (hopefully) with what is going on - I think I just found the part that is erroring out on the missing file! (I love debugging!) <3

vsoch: okay confirmed the erroneous "part" is indeed flux!
```
{
  "name": "default",
  "descr": "",
  "scheduler": "flux",
  "launcher": "local",
  "access": [],
  "container_platforms": [],
  "modules": [],
  "variables": [],
  "environs": [
    "builtin"
  ],
  "max_jobs": 8,
  "resources": []
}
```
We found device information (but not processor information):

```python
part.devices
Out[8]: [<reframe.core.systems.DeviceInfo at 0x7f7bf56b03d0>]
```
And then most of the if logic is skipped, and we wind up entering the `elif detect_remote_systems` and this is the command that fails:

```
                with runtime.temp_environment(modules=modules, variables=vars):
                    part._processor = ProcessorInfo(_remote_detect(part))
```
So I think what's happening here is that I haven't providing enough info about processor information, so it erroneously is looking for remote information, and I'm guessing those files (that aren't found) are what might hold it? So I need to figure out how to get this populated I think?

```
part.processor.info
Out[26]: {}
```

vsoch: I'm also seeing these directories scattered in the root:
```
# ls rfm.9xoaersf/
bin  bootstrap.sh  reframe  requirements.txt  rfm-detect-job.sh
```
and it could be flux doesn't know how to make one? Or run a requirement to work with one?

vsoch: @vkarak can you give me a sense of what this check is doing, and what might be missing for flux?

vkarak: Ok, I think that I am getting closer to understanding what is happening. Let me explain first how jobs are submitted and managed in ReFrame. The `compile` and `run` pipeline stages (see `reframe.core.pipeline`) of a `reframe.core.pipeline.RegressionTest` create `reframe.core.scheduler.Job` instances for the corresponding test jobs. These contain all the information related to the submitted job (e.g. the script filename, submission parameters, the output files, the status, the jobid etc.). The jobs are managed (submitted, collected, cancelled etc.) by the `reframe.core.schedulers.JobScheduler` backends, which implement concretely a set of abstract methods for those operations. All these method take the job as an argument and they update the job status (e.g., the job's `state` or any other information that needs to be updated). The framework uses the various `Job` instances to get *any* information related to the job, whereas the `JobScheduler` is merely used for managing the jobs (submitting, cancelling etc.).

Now this is what happens in your case. ReFrame has a mechanism for auto-detecting the processor topology not only for the local partitions but also for remote partitions, so that your tests can use that information to configure themselves (e.g., set the thread numbers or the CPU affinities, set compilation options etc.). To do so for remote partitions, reframe submits a job that executes itself with the `--detect-host-topology` option from a temporary directory and uses that output (after caching it for future) to configure the processor information for the target partition. That unit test executes only for remote partitions (i.e., any partition with a non-local scheduler) and does exactly that. Now what I see in your code is that you never set `Job.stdout` and `Job.stderr` (they have their [default](https://github.com/reframe-hpc/reframe/blob/9f0d16742166af0482c52ca53a01e4e5cf86ce29/reframe/core/schedulers/__init__.py#L295-L296) values), but instead you set them in your local `fluxjob` variable during submit. Although the values match the default ones and are in the correct directory (see later on my suggestion on how to better implement this), it seems that the stdout or stderr files are not created after your job finishes (supposedly, after the `wait()` is called). So I don't know if you need to do something specifically for Flux to make those appear or if Flux deletes them, or if there is delay before appearing? But then I'm wondering how the example works for your tutorial example? (Note: to debug this, I suggest removing the `os.rmtree()` call in the `_copy_reframe.__exit__()` method). This is where I reached so far.

Back to the design of `Job` vs. `JobScheduler`, I would suggest the following:

- Do not inherit from `PbsScheduler` but just `JobScheduler` (even conceptually is wrong; Flux has a completely different mentality and besides you would be getting errors mentioning PBS 😄 ).
- Define a `_FluxJob` extending `Job` (as the other schedulers do) and keep there *all* the information related to the Flux, see it as a wrapper to the result `flux.job.JobSpecV1.from_command`. Also, all this machinery can move to `_FluxJob.__init__()` and leave in the `JobScheduler.submit()` the actual job submission through the Flux executor.

That would make the design more compatible with the concepts of the framework and hopefully will avoid us surprises in the future.

vsoch: This was really helpful! So I didn't look closely at the wrappers to sched.Job (and this is how self.completed/self.cancelled make sense). As I was rewriting that part, I realized that the stdout and stderr that I set aren't from the job - they are custom paths that I wrote based on the script because I didn't realize the base job class had already specified one. So the bug was just that flux didn't know about the correct files to write to.

And then for the next bug for the autodetect, it looks like the files for output/error are relative, which works given the context when it's defined, but flux seemed to be writing to the original PWD. So the fix there is to take the abspath of the self.workdir and provide Flux with the full, absolute paths!

Just ran all the tests locally and green? Will push soon and fingers crossed it reproduces!

vsoch: Ok pushed! I restored the sleep I had extended to the original value, and tried to remove the empty newlines I had added in autodetect.py. Fingers crossed! :crossed_fingers:

vsoch:
ok - two test failures that I still can't reproduce locally.  Any things I can try?

vsoch:
I could try increasing the sleep again in case flux needs more time to cancel given the CI worker (maybe it's slower?)

vsoch:
okay I have an idea! Flux also has a kill functionality for a job - the cancel might be too soft here.

vsoch:
okay just pushed a change so the job is killed and rebased with master. If that doesn't work I think we need to look more closely at exactly what PID is expected to be killed (vs. what Flux is doing).

vkarak:
@vsoch The problem with the unit test failure is not Flux's problem. It's something wrong with the local scheduler implementation; I will have a look at it and try to fix it. This test is a bit flaky and I also cannot reproduce it locally with your container 😬

vsoch:
okay sounds good! I'll be on my computer most of the day so please ping me if I can be of help. Really excited to get this working!

vkarak:
Still no luck, but I just noticed that you run the unit tests and reframe etc. in the container using flux, like `flux start ...`. Could you try removing that and see if Github actions become happy again? (PS: I could not reproduce it locally even with `flux start`...)

Also the intention of that unit test is to check that the local scheduler kills any deeply spawned process. Practically, the local scheduler spawns its job as a new session, so that later when it kills it, all the processes in the session get killed too.

vsoch:
Without flux start we won't have a flux instance running - so I don't think the flux tests would pass. Using flux start (with the command) is the typical / suggested use case, and the other option would be interactive, e.g.,:

```bash
$ flux start --test-size=4
# the above is akin to starting an interactive shell you can exit from
# interact with flux here
```
if we start a flux instance before (via an interactive shell) that wouldn't work in the CI. If you just want to do a sanity check to remove flux start I can do it, but I don't think it's going to work.

vsoch:
okay - just pushed a skip condition that will skip the failing tests when running with flux, and run them after. This should serve as a sanity check to see if the local tests are failing because of flux start. If it's not that, my next guess is something about running in the container.

vkarak:
> if we start a flux instance before (via an interactive shell) that wouldn't work in the CI. If you just want to do a sanity check to remove flux start I can do it, but I don't think it's going to work.

Hmm, until I spotted this I was running the unit tests in the container as usual without `flux start` and they seemed to work:

```
unittests/test_schedulers.py::test_submit[flux] PASSED
unittests/test_schedulers.py::test_submit_max_pending_time[flux] PASSED
```

vsoch:
Oh that's weird - well I guess we can try it then!

vsoch:
haha, yeah that's about what I'd expect! I'll try adding back the skips for those two tests, and running the rest with flux start.

vkarak:
> haha, yeah that's about what I'd expect! I'll try adding back the skips for those two tests, and running the rest with flux start.

😱 Then why does it work on me without `flux start`? So strange!

vsoch:
oops I had a bug. :bug:

vsoch:
the envar didn't seem to take - let me look into that.

vsoch:
ok I'm trying the same idea with flux mini run.

vkarak:
Do these unit test fail also when you run your container locally or do they fail only on Github?

vsoch:
They work locally and fail on GitHub - that’s why debugging this is so hard.

vsoch:
The other thing I'm curious about is if it could be an issue with having multiple brokers for the instance. Let me see if flux start supports just using one.

vsoch:
hey @trws @grondo we need some help from the big leagues! So the flux start command is actually leading to a local test failure, specifically a PID that should not be running (a sleep command) is found to be running. We've been trying different ways to start flux (or not) but haven't found a solution - without flux start we can't run the flux tests, but with it this weird thing happens. Is there something else we can try? Ideally the test suite works under flux start.

vkarak:
@teojgo Could you also have a look at the CI failure? We can't reproduce it anywhere else except on Github...

grondo:
I'm just guessing (without knowing how all this fits together), but by default when canceling a job Flux first attempts to send `SIGTERM` then waits 5s by default, then sends `SIGKILL`. If the job ignores SIGTERM, then you'll have to wait for ~5s before SIGKILL is sent. This can be tuned by reloading the job-exec module
```
$ flux module reload job-exec kill-timeout=.25s
```
Or by configuration, which may be more convenient here:
```toml
[exec]
kill-timeout = "0.25s"
```
Try placing the above TOML into a file `conf/exec.toml`, then running the test like:
```
$ flux start -o,--config-path=conf ...
```
And see if that helps.

vsoch:
@grondo it's actually not the flux cancel that's failing - it's the local scheduler that is failing, and we (think?) because it's being run under flux start (but there is no flux being run for this test failure, aside from having it run in that environment!)

```console
unittests/test_schedulers.py:613: Failed
=========================== short test summary info ============================
FAILED unittests/test_schedulers.py::test_cancel_with_grace[local] - Failed: ...  # if this were flux it would say flux instead of local
FAILED unittests/test_schedulers.py::test_cancel_term_ignore[local] - Failed:...
======= 2 failed, 972 passed, 208 skipped, 6 rerun in 152.61s (0:02:32) ========
2022-09-25T22:43:21.960481Z broker.err[0]: rc2.0: python3 ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -vvvv Exited (rc=1) 153.3s
flux-start: 0 (pid 376) exited with rc=1
Error: Process completed with exit code 1.

```

Super weird, affirmative.

grondo:
Oh, sorry! For some reason I thought there was a failure due to a `sleep` job still running at the end of the test, and jumped to conclusions! Sorry, I was off in the weeds!

Well, in case it is helpful `flux start` starts one or more flux-brokers (by default just one), waits for the instance to be ready, then invokes its arguments as a child of the rank 0 broker (we call the command and arguments the "initial program"). Once the command exits, then the Flux instance will tear itself down and exit with the same exit code as the initial program.

> it's the local scheduler that is failing, and we (think?) because it's being run under flux start 

I don't know what the local scheduler is here, but another thing to try (if possible) to avoid running the tests as a child of the flux-broker would be to run an instance in the background, then connect to that instance by exporting its `FLUX_URI`, e.g. something like:

```console
$ flux start sleep inf & FLUX_URI=$(flux uri pid:$!)
$ echo $FLUX_URI
local:///tmp/flux-FWVTqp/local-0
$ flux resource info
1 Nodes, 4 Cores, 0 GPUs
$ run-my-reframe-tests
$ flux shutdown  # force-stop flux instance
2022-09-25T23:25:22.251326Z broker.err[0]: rc2.0: sleep inf Hangup (rc=129) 99.4s
broker.err[0]: rc2.0: sleep inf Hangup (rc=129) 99.4s
broker.info[0]: rc2-fail: run->cleanup 1.65669m
broker.info[0]: cleanup.0: flux queue stop --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup.1: flux job cancelall --user=all --quiet -f --states RUN Exited (rc=0) 0.0s
broker.info[0]: cleanup.2: flux queue idle --quiet Exited (rc=0) 0.0s
broker.info[0]: cleanup-success: cleanup->shutdown 0.111324s
broker.info[0]: children-none: shutdown->finalize 0.081045ms
broker.info[0]: rc3.0: /home/grondo/git/flux-core.git/etc/rc3 Exited (rc=0) 0.6s
broker.info[0]: rc3-success: finalize->goodbye 0.595873s
flux: flux_open: No such file or directory
flux: flux_open: No such file or directory
[1]+  Exit 129                flux start sleep inf
```

vsoch:
ooh lala! I'll try that out. GitHub actions might exit when it sees us trying to do that, but it's worth a shot.

vsoch:
I tried removing the line to ask for resource info, but not sure if that will help. At least we might get to running the tests and can pick up tomorrow (1:30am here).

teojgo:
I have done a little bit of investigation of what is happening. So, it seems that `flux start <command>` sets up the correct PYTHONPATH to execute flux commands which is valid for the `<command>`. If you run locally the with docker without specifying a command, `docker run -it <image_name>` then the command that gets executed is the default one (that comes from the image) : `flux start /bin/bash` then executing the commands one by one seems to work fine. In the github commands case, this does not happen (I think) since the actual command is overriden by the actual run commands given on the yaml of the github actions. 
@vsoch can you try adding this before you do the reframe test run?
`export PYTHONPATH=$(flux start bash -c 'echo $PYTHONPATH')`

vsoch:
> Ok, so now we have only the two failing tests. Out of curiosity, could you try increasing the sleep time we wait before checking that the pid is still valid and see what happens?

Sure! Just for perspective we were at this spot before, without the changes, just doing the flux start with the command. So I'm not sure much has changed. But I think you found something interesting - I'm wondering if I should try to customize the entrypoint of the container so it mirrors what happens locally.

teojgo:
> > Ok, so now we have only the two failing tests. Out of curiosity, could you try increasing the sleep time we wait before checking that the pid is still valid and see what happens?
> 
> Sure! Just for perspective we were at this spot before, without the changes, just doing the flux start with the command. So I'm not sure much has changed. But I think you found something interesting - I'm wondering if I should try to customize the entrypoint of the container so it mirrors what happens locally.

The entrypoint seems to execute a script that starts munge

vsoch:
okay just pushed a test with increased sleep time for both tests, 0.1 to 5 seconds.

teojgo:
It seems to be reduced to 1 test now. So It's not easy to reproduce. One final thought that I have is to use the `--init` option for the container

vsoch:
Just for kicks and giggles, I tried doubling it.

vsoch:
>  One final thought that I have is to use the --init option for the container

We could try that - could you expand upon that idea?

teojgo:
> > One final thought that I have is to use the --init option for the container
> 
> We could try that - could you expand upon that idea?

I remember that we have used it sometime in the past and then we found a workaround for reaping zombie processes.

vsoch:
oh the actual `--init` to the container!! Yes I understand! Just pushed.

vsoch:
Did that work!? 🥹

teojgo:
> Did that work!? 🥹

Yes, the other was a codecov error. I will force rerun

teojgo:
Ok to test

.github/workflows/test-flux.yaml
```diff
@@ -0,0 +1,50 @@
...
+          rm -rf /usr/share/dotnet
+          rm -rf /opt/ghc
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install Reframe
+        run: |
+          /bin/bash ./bootstrap.sh
+          python3 -m pip install pytest
```
teojgo, line 32: pytest is already installed by requirements so this is not needed.

teojgo (changes requested):

teojgo:
> It’s green!! @teojgo you nailed it - it was the zombies all along!
> 
> i am hugely not a morning person (I might be a zombie myself atm! zombie_woman) so I’ll be back later today to finish this up). But hooray! So happy it’s working! grinning

Can you also revert the sleep time to the normal levels?

vsoch:
I'm going to test restoring the sleep (and the original testing commands) to see if the `--init` flag actually solved the issue.

vsoch:
okay - the git history is a bit muddled (hopefully you allow squash and merge?) but I've done a final test - removing pytest and restoring the original sleep and command: https://github.com/reframe-hpc/reframe/pull/2598/commits/f9b5fc44afb9817ab15e5f4bbe940bf937c90405

vsoch:
Nice!! :partying_face: 

@teojgo GitHub actions usually aren't this tricky - this is the hardest one to debug I've encountered I think.

teojgo:
> Nice!! partying_face
> 
> @teojgo GitHub actions usually aren't this tricky - this is the hardest one to debug I've encountered I think.

Yes, in other tests that use containers, what we do is ensure that all the necessary steps that need to run, are embedded in the container image itself. Then we build the image by adding only the latest source code, as in https://github.com/reframe-hpc/reframe/blob/9f0d16742166af0482c52ca53a01e4e5cf86ce29/.github/workflows/main.yml#L60
An example dockerfile can be found here: https://github.com/reframe-hpc/reframe/blob/master/ci-scripts/dockerfiles/Lmod.dockerfile

Therefore it's just a matter of performing a `docker run` after.

vsoch:
That's an approach I've used as well, but we'd need to have that automated build somewhere (and we do not). We'd still run into this `--init` bug regardless.

I think this approach makes sense because it mirrors the small Flux tutorial we provide in the docs, and an interested user looking for a container is going to use this approach to start with a flux base and install reframe. With the current setup they have several examples of that to stumble upon.

vkarak:
Thanks @teojgo and @vsoch for getting it to work! You gave me very useful hints to understand of what is probably happening. If not started with `--init`, the first process in the container (the entrypoint or the command) gets the PID 1. Any orphaned process in Linux gets inherited by PID 1, which in a non containerised environment corresponds to the init or systemd process, whose role is also to `wait(2)` on orphaned children and reap zombies (any process who's dead but not waited for). If a random process inside a container becomes PID 1 and this process spawns more processes, which spawn other processes and so on, the initial process has to take on the role of system's init process and reap any zombies left by its successors. Obviously, you don't expect that from a random process, that's why the `--init` option starts a `docker-init` process with PID 1 inside the container to collect any zombies left behind. If you query the status of a zombie process with `kill(pid, 0)` the process will show as alive, because it has not yet been waited for. And this is what is happening now with the notorious `test_cancel_with_grace()` unit test. In that test we spawn this process tree: `reframe --- bash --- sleep`. When the grace period finishes, ReFame kills the whole session led by `bash` and therefore `sleep`. ReFrame waits for `bash`, so it reaps it, but it has no clue of what bash has spawned. It knows it has killed it, but it can't reap it. This gets inherited by PID 1, which should reap it. If PID 1 is meant to do that, i.e., being the `init` process or `bash` or `docker-init` etc. then all fine, we just wait [a bit](https://github.com/reframe-hpc/reframe/blob/9f0d16742166af0482c52ca53a01e4e5cf86ce29/unittests/test_schedulers.py#L658) (the comment is deceiving – I should fix it) so that the PID 1 process takes its chance to call wait on it. Running inside a container, it might also happen that reframe itself is PID 1! For this case, we have [this](https://github.com/reframe-hpc/reframe/blob/9f0d16742166af0482c52ca53a01e4e5cf86ce29/unittests/test_schedulers.py#L595-L598) code snippet in `assert_process_died`, where we reap the zombie, because nobody else could :-)

Running `docker run reframe-flux:focal flux start ps -j` you can see that `flux-broker` is PID 1:

```
  PID  PGID   SID TTY          TIME CMD
    1     1     1 ?        00:00:00 flux-broker-0
   79     1     1 ?        00:00:00 ps
```

Nonetheless, this still works if you run the unit tests:

```
docker run -t reframe-flux:focal flux start ./test_reframe.py --rfm-user-config=tutorials/flux/settings.py -v unittests/test_schedulers.py
```

So `flux-broker` is probably not the problem. One last thing to try out, out of pure curiosity only, is to add a `ps -j` command in Github's workflow `run` section to see the process hierarchy and who is PID 1 when you don't pass `--init` to docker.

vsoch:
Do we need to do that if everything is working? Is this a request / requirement for the PR?

vsoch:
actually I'll just open a separate PR to test, that's easy enough!

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,187 @@
...
+        super().__init__(*args, **kwargs)
+
+        self._cancelled = False
+
+        # This is set by the scheduler when both the job's state is
+        # 'COMPLETED' and the job's stdout and stderr are written back
+        self._completed = False
+        self.create_job(**kwargs)
+
+    def create_job(self, **kwargs):
```
vkarak, line 46: I would probably not use `kwargs` at all here, since you don't need them in this function. They're good and evil at the same time, so in principle I use them only when needed.

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # In testing, cancel is too "soft" - or often the pid is still remaining
+        # after. We thus do a cancel in addition to a kill.
+        try:
+            flux.job.kill(flux.Flux(), job._flux_future.jobid())
+
+        # Job is inactive
+        except EnvironmentError:
+            pass
+        job._is_cancelling = True
```
vkarak, line 118: That's probably a leftover from the previous implementation.

reframe/core/schedulers/flux.py
```diff
@@ -0,0 +1,187 @@
...
+                if time.time() - job.submit_time >= job.max_pending_time:
+                    self.cancel(job)
+                    job._exception = JobError(
+                        "maximum pending time exceeded", job.jobid
+                    )
+            else:
+                # Otherwise, we are still running
+                job._state = "RUNNING"
+
+    def allnodes(self):
```
vkarak, line 164: Nice that you implemented this one, but alone it is not very useful. It is used in conjunction with the `filternodes()` below to support flexible node allocation in tests as well as the `--distribute` option. Essentially, the `JobScheduler.guess_num_tasks` tries to guess the available nodes to run the test on, based on various constraints (e.g., the `access` partition attribute, the `-J` scheduler specific options etc.). You can leave the implementation of `filternodes()` for future work in another PR 😉

vkarak (changes requested):
Just a couple of minor comments still and then I could take care of the style conformance, as well as fix a comment in the `test_cancel_with_grace()` and we're good to go!

vkarak:
> Do we need to do that if everything is working? Is this a request / requirement for the PR?

No, it's not! Feel free to skip it!

vsoch:
Ok I’m going out to get flu shot but will work on these when I get back!

vkarak:
> One last thing to try out, out of pure curiosity only, is to add a ps -j command in Github's workflow run section to see the process hierarchy and who is PID 1 when you don't pass --init to docker.

So this explains it all!

```
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.1  0.0   7272   572 pts/0    Ss+  20:02   0:00 tail -f /dev/null
root         145  0.0  0.0   2616   596 ?        Ss   20:02   0:00 sh -e /__w/_temp/3cfab0e9-9359-4809-bb2c-59de1fbf5770.sh
root         152  0.0  0.0  10620  3296 ?        R    20:02   0:00 ps aux
```

And this:

```
  /usr/bin/docker create --name 908acf52a594482898b84c799eba7612_fluxrmfluxschedfocal_33c4b1 --label 817f70 --workdir /__w/reframe/reframe --network github_network_e9fe0fd3a5f84ae18e6f0f3aea6fe8fa --platform=linux/amd64 --user root -it -e "HOME=/github/home" -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work":"/__w" -v "/home/runner/runners/2.296.2/externals":"/__e":ro -v "/home/runner/work/_temp":"/__w/_temp" -v "/home/runner/work/_actions":"/__w/_actions" -v "/opt/hostedtoolcache":"/__t" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" --entrypoint "tail" fluxrm/flux-sched:focal "-f" "/dev/null"
```

The entrypoint for the Github action containers is `tail`, so the failure we were seeing was 100% deterministic! I will add a comment in the corresponding unit tests to make sure to pass `--init` if the container machinery of Github workflows is used.

vsoch:
okay I'm back! Updates coming soon!

reframe/core/schedulers/flux.py
vsoch: okay I'll raise NotImplementedError for both.

vsoch:
okay updated! The git history here is a bit wonky and I can no longer rebase - hopefully that's not an issue and we can squash and merge?

vsoch:
Green! :green_apple: :green_book: :green_heart: :green_salad:

docs/tutorial_flux.rst
```diff
@@ -0,0 +1,108 @@
...
+   ./bootstrap.sh
+
+And then reframe will again be in the local ``bin`` directory:
+
+.. code:: bash
+
+   # which reframe
+   /code/bin/reframe
+
+Then we can run reframe with the custom config `config.py <config.py>`__
```
teojgo, line 35:
```suggestion
Then we can run ReFrame with the custom config `config.py <config.py>`__
```

tutorials/flux/README.md
```diff
@@ -0,0 +1,4 @@
+# Flux Framework Tutorial
+
+This is a demo that will show how to use refame with [Flux Framework](https://github.com/flux-framework/).
```
teojgo, line 3:
```suggestion
This is a demo that will show how to use ReFrame with [Flux Framework](https://github.com/flux-framework/).
```

unittests/test_schedulers.py
```diff
@@ -381,11 +392,12 @@ def test_submit(make_job, exec_ctx):
 
     # Additional scheduler-specific checks
     sched_name = minimal_job.scheduler.registered_name
+
     if sched_name == 'local':
         assert [socket.gethostname()] == minimal_job.nodelist
         assert minimal_job.exitcode == 0
         assert minimal_job.state == 'SUCCESS'
-    elif sched_name == ('slurm', 'squeue', 'pbs', 'torque'):
+    elif sched_name == ('slurm', 'squeue', 'pbs', 'torque', 'flux'):
```
teojgo, line 400:
```suggestion
    elif sched_name in ('slurm', 'squeue', 'pbs', 'torque', 'flux'):
```

Could the `==` be wrong in the first place?

teojgo (changes requested):
Mostly cosmetic changes needed

unittests/test_schedulers.py
vkarak: Oh, indeed! Good catch!

vsoch: Fixed with https://github.com/reframe-hpc/reframe/pull/2598/commits/e2032ba5884c3380676c976aea90182798e1760c

unittests/test_schedulers.py
```diff
@@ -381,11 +392,12 @@ def test_submit(make_job, exec_ctx):
 
     # Additional scheduler-specific checks
     sched_name = minimal_job.scheduler.registered_name
+
     if sched_name == 'local':
         assert [socket.gethostname()] == minimal_job.nodelist
         assert minimal_job.exitcode == 0
         assert minimal_job.state == 'SUCCESS'
-    elif sched_name == ('slurm', 'squeue', 'pbs', 'torque'):
+    elif sched_name in ('slurm', 'squeue', 'pbs', 'torque', 'flux'):
```
teojgo, line 400: This might have revealed an actual bug. Is there a way to retrieve the nodes of a job using flux ?

vkarak:
@vsoch I'm looking now at the CI failure for Flux and I'm about to do a PR to your branch fixing it + with some cosmetic changes.

unittests/test_schedulers.py
vkarak: @teojgo I think that this has triggered a bug also in other backends, that's why the "CSCS CI" fails. Could you check it?

@vsoch Putting `flux` in that list makes sense only if there is a way to retrieve the nodes where a flux job has run.

vkarak:
@vsoch PR [done](https://github.com/researchapps/reframe/pull/3). After that on is merged, this would be good to go from my side! Only thing is for @teojgo to check the CSCS CI failure.

vkarak:
@vsoch I just realised that I have probably broken the `test_cancel()`... Let me try to fix it and I'll fire up a new PR.

vsoch:
Thank you - yes some of the logic I added was very specifically because it was needed!

vsoch:
I can also add it back with logic to list node (to test) if you prefer - let me know.

vkarak:
> I can also add it back with logic to list node (to test) if you prefer - let me know.

No problem! I have just submitted another PR that passes.

unittests/test_schedulers.py
```diff
@@ -381,11 +392,12 @@ def test_submit(make_job, exec_ctx):
 
     # Additional scheduler-specific checks
     sched_name = minimal_job.scheduler.registered_name
+
     if sched_name == 'local':
         assert [socket.gethostname()] == minimal_job.nodelist
         assert minimal_job.exitcode == 0
         assert minimal_job.state == 'SUCCESS'
-    elif sched_name == ('slurm', 'squeue', 'pbs', 'torque'):
+    elif sched_name in ('slurm', 'squeue', 'pbs', 'torque'):
```
teojgo, line 400: Remove the `squeue` because in that case, neither of the `nodelist`, `exitcode` are retrieved. This is why the CSCS CI fails.

teojgo (changes requested):

teojgo (approved):
It looks fine now. Update it to the latest main branch and it's ready to be merged.

vsoch:
Done!

vkarak (approved):
Lgtm! Thanks again for the contribution and the patience for debugging intricate failures!

vsoch:
woohoo!! :partying_face:

PR #2603, opened 2022-09-23
[feat] Stop execution with an error when a cli module fails to load

### LINKED ISSUE #2597: failed load of user module with command line option --module does not cause failed tests.
if a module is provided on the command line, for example with `--module doesnotexist`, a warning is issued `could not load module 'doesnotexit' correctly; skipping...` but the tests continue.

https://github.com/reframe-hpc/reframe/blob/337fdcca763cc4081d2c53c429a3f3b2402f0b5f/reframe/frontend/cli.py#L1245

the safer behavior would be that ~the tests fail~ ReFrame errors out because they're not doing what the user asked, so the results are misleading.
if that's not possible (e.g. to not break user's code that depends on this), can you add an option that makes sure the tests fail?

ekouts:
The output will look something like this:
```bash
$ reframe -m wrong_module -v ...
[ReFrame Setup]
...

./bin/reframe: could not load module 'wrong_module' correctly
Log file(s) saved in ...
```

With `-vv` then the output is:
```bash
$ reframe -m wrong_module -vv ...
[ReFrame Setup]
...

./bin/reframe: could not load module 'wrong_module' correctly
./bin/reframe: could not execute module operation: command 'modulecmd python show wrong_module' failed with exit code 0:
--- stdout ---
--- stdout ---
--- stderr ---
ModuleCmd_Display.c(157):ERROR:105: Unable to locate a modulefile for 'wrong_module'


--- stderr ---
Log file(s) saved in ...
```

Closes #2597 .

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -1241,9 +1241,10 @@ def main():
                 rt.modules_system.load_module(**m, force=True)
             except errors.EnvironError as e:
                 printer.error(
-                    f'could not load module {m["name"]!r} correctly'
+                    f'could not load module {m["name"]!r} correctly; rerun '
+                    f'with -vv for more information'
                 )
-                printer.error(str(e))
+                printer.debug(str(e))
                 sys.exit(1)
 
         options.flex_alloc_nodes = options.flex_alloc_nodes or 'idle'
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -1240,11 +1240,11 @@ def main():
             try:
                 rt.modules_system.load_module(**m, force=True)
             except errors.EnvironError as e:
-                printer.warning(
-                    f'could not load module {m["name"]!r} correctly; '
-                    f'skipping...'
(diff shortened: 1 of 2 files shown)


reframe/frontend/cli.py
```diff
@@ -1240,11 +1240,11 @@ def module_unuse(*paths):
...
                 rt.modules_system.load_module(**m, force=True)
             except errors.EnvironError as e:
-                printer.warning(
-                    f'could not load module {m["name"]!r} correctly; '
-                    f'skipping...'
+                printer.error(
+                    f'could not load module {m["name"]!r} correctly'
                 )
-                printer.debug(str(e))
+                printer.error(str(e))
```
vkarak, line 1246: I'd print that in verbose mode. It's too much info for printing all this right away. You can also extend the error message with a `rerun with increased verbosity for more information`.

vkarak (changes requested):

vkarak (approved):

PR #2605, opened 2022-09-26
[feat] Allow user defined launchers in the configuration

### LINKED ISSUE #2560: Allow defining custom MPI launchers for systems
There are some systems I have access to where it'd be beneficial to use custom MPI launchers, different from the standard ones.  Latest case I ran into is https://www.dur.ac.uk/icc/cosma/support/rockport/, where `mpirun` should be called as `mpirun $RP_OPENMPI_ARGS`, where `$RP_OPENMPI_ARGS` is an environment variable set by a module, which has to be loaded when running the tests (for example by setting [`systems.partitions.modules`](https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#systems-.partitions-.modules)).

I may have time to work on this, but I need some guidance, especially in terms of what the API should be.  In general, I want this to be system-specific (so to be customised in the system configuration), I don't want to entangle tests with system-specific details.

ekouts:
Based on @giordano 's suggestion I removed the schema restriction and added an example in the documentation.
Closes #2560 .

@jgphpc @teojgo this will also make it easier for us to test on hohgant

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -322,6 +322,39 @@ System Partition Configuration
    - ``upcrun``: Parallel programs will be launched using the `UPC <https://upc.lbl.gov/>`__ ``upcrun`` command.
    - ``upcxx-run``: Parallel programs will be launched using the `UPC++ <https://bitbucket.org/berkeleylab/upcxx/wiki/Home>`__ ``upcxx-run`` command.
 
+   .. tip::
+
+      ReFrame also allows you to register your own custom launchers simply by defining them in the configuration.
+      Here is an example:
+
+      .. code:: python
+
+         import reframe.core.launchers.mpi as mpi
+
+         @mpi.register_launcher('custom_launcher')
+         class MyLauncher(mpi.SrunLauncher):
+             def command(self, job):
+                 return super().command(job) + [...]
+
+         site_configuration = {
+             'systems': [
+                 {
+                     'name': 'my_system',
+                     'partitions': [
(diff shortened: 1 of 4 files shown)


docs/config_reference.rst
```diff
@@ -322,6 +322,39 @@ System Partition Configuration
    - ``upcrun``: Parallel programs will be launched using the `UPC <https://upc.lbl.gov/>`__ ``upcrun`` command.
    - ``upcxx-run``: Parallel programs will be launched using the `UPC++ <https://bitbucket.org/berkeleylab/upcxx/wiki/Home>`__ ``upcxx-run`` command.
 
+   .. tip::
+
+      ReFrame also allows you to register your own custom launchers simply by defining them in the configuration.
+      Here is an example:
+
```
vkarak, line 329: What about expanding this a bit with what users are expected to implement in `command` and move that to a small tutorial of its own? At some we should also have a how-to implement a new backend scheduler.

vkarak (changes requested):

docs/config_reference.rst
ekouts: I made some adjustments based on your comment, what do you think now? If we make a longer tutorial about the internals we could have a separate tutorial but I thought it fits nicely in the `tutorial_advanced.html`.

teojgo (approved):

vkarak (approved):
I did some fine tuning to the docs and lgtm.

PR #2606, opened 2022-09-26
Test/flux again

vsoch:
This is a requested test for #2598


vkarak:
It doesn't recognise `ps -j`. No worries! Let's close that and just focus on merging the #2598. We can do that today actually!

PR #2608, opened 2022-09-28
[doc] Adapt README for dark and light Github modes

### LINKED ISSUE #2204: ReFrame logo in README does not show up correctly in Github's dark mode
Screenshot:

image (1)

Thanks @branfosj for spotting this.

ekouts:
@rsarm found an example from DeepSpeed's README.md https://github.com/microsoft/DeepSpeed/blob/master/README.md

You can see directly how it looks in https://github.com/ekouts/reframe/tree/dark_logo

(Hadn't noticed that there was an issue and a slightly different solution from Teo actually 😅)
Closes #2204

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
-<div align="left">
- <img src="docs/_static/img/reframe_logo-width400p.png#gh-light-mode-only" width="400px">
- <img src="docs/_static/img/reframe-logo-dark-bg.png#gh-dark-mode-only" width="400px">
-</div>
+<picture>
+  <source media="(prefers-color-scheme: dark)" srcset="docs/_static/img/reframe-logo-dark-bg.png" width="400px">
+  <source media="(prefers-color-scheme: light)" srcset="docs/_static/img/reframe_logo-full.png" width="400px">
+  <img alt="ReFrame logo" src="https://raw.githubusercontent.com/reframe-hpc/reframe/master/docs/_static/img/reframe_logo-full.png">
+</picture>
 
 [![Build Status](https://github.com/reframe-hpc/reframe/workflows/ReFrame%20CI/badge.svg)](https://github.com/reframe-hpc/reframe/actions?query=workflow%3A%22ReFrame+CI%22)
 [![Documentation Status](https://readthedocs.org/projects/reframe-hpc/badge/?version=latest)](https://reframe-hpc.readthedocs.io/en/latest/?badge=latest)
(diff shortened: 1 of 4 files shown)


rsarm (approved):

ekouts:
I close it so that I don't spam you with my tries, the suggested way doesn't work for me for some reason.

teojgo:
> In the documentation of github they suggest the `<picture>` element but i couldn't make it work. I think for now `#gh-dark-mode-only` and `#gh-light-mode-only` are good enough.

The only problem that this has, is that because we extract the logo for the pypi installation via the README, it will show the logo twice.

ekouts:
> The only problem that this has, is that because we extract the logo for the pypi installation via the README, it will show the logo twice.

Aha, is that why we also had `https://raw.githubusercontent.com/reframe/...` paths instead of of relevant paths? Do you know if the picture element would solve the problem?

victorusu (approved):
`lgtm`

ekouts:
I added an second README.md for pypi. It has the old picture so it should appear normally. It is not a great solution, because we have to maintain two READMEs, almost identical other that the picture on top. Any better idea?

teojgo:
> I added an second README.md for pypi. It has the old picture so it should appear normally. It is not a great solution, because we have to maintain two READMEs, almost identical other that the picture on top. Any better idea?

For me it's fine. I don't know how else it could work.

vkarak:
@ekouts @teojgo Couldn't that second README be generated on the fly before deployment to PyPI?

teojgo:
> @ekouts @teojgo Couldn't that second README be generated on the fly before deployment to PyPI?

yes but this would need to happen from the github action before publishing, but I would like to avoid forgetting to update the corresponding github action if the README changes

vkarak:
> but I would like to avoid forgetting to update the corresponding github action if the README changes

Wouldn't you have the same problem with maintaining a copy of the README in parallel? How can we make sure that the copy is always updated. That's why I proposed to generate the PyPI README just before publishing if possible?

ekouts:
In my opinion the pypi readme is not that important and we don't even need to have all the information from github. What if we have a more minimal readme for pypi that has the `nutshell` section + links to github and docs + the contact section about mailing list + slack etc. Like this we won't have to generate and we avoid mentioning something that is not up-to-date.

vkarak:
Perhaps rename it to `README_minimal.md` instead of `README_pypi.md`.

vkarak (approved):

PR #2611, opened 2022-09-29
[bugfix] Submit jobs using the correct module path when testing scheduler backends

### LINKED ISSUE #2609: Unittest module path is not used when submitting the test jobs
This happends in the `test_schedulers.py` where the job submission should be wrapped in a:
```
with rt.module_use(test_util.TEST_MODULES):
```
This can be easily reproduced by executing the `test_submit[slurm]` test and looking at the stderr where the `testmod_foo` cannot be found.

teojgo:
Fixes #2609


teojgo:
@jenkins-cscs retry all

vkarak (approved):

PR #2613, opened 2022-09-30
[feat] Add support for the Apptainer container platform

### LINKED ISSUE #2602: Support apptainer container platform
As far as I understand from @teojgo , it should be very easy to add as we already support Singularity.

teojgo:
Fixes #2602

### DIFF
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -268,7 +268,11 @@ class Singularity(ContainerPlatform):
 
 class Apptainer(Singularity):
     '''Container platform backend for running containers with `Apptainer
-    <https://apptainer.org/>`__.'''
+    <https://apptainer.org/>`__.
+
+    .. versionadded:: 4.0.0
+
+    '''
 
     def __init__(self):
         super().__init__()
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -243,6 +243,7 @@ class Singularity(ContainerPlatform):
     def __init__(self):
         super().__init__()
         self.with_cuda = False
+        self._launch_command = 'singularity'
 
     def emit_prepare_commands(self, stagedir):
         return []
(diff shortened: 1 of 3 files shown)


rsarm (approved):

reframe/core/containers.py
```diff
@@ -259,10 +260,19 @@ def launch_command(self, stagedir):
...
+            return (f'{self._launch_command} exec {" ".join(run_opts)} '
                     f'{self.image} {self.command}')
 
-        return f'singularity run {" ".join(run_opts)} {self.image}'
+        return f'{self._launch_command} run {" ".join(run_opts)} {self.image}'
+
+
+class Apptainer(Singularity):
+    '''Container platform backend for running containers with `Apptainer
+    <https://apptainer.org/>`__.'''
```
vkarak, line 271: Add an annotation here that this was added in version 4.0.0.

unittests/test_containers.py
```diff
@@ -101,7 +102,7 @@ def expected_cmd_mount_points(container_variant):
                 '--mount=type=bind,source="/path/two",destination="/two" '
                 '--mount=type=bind,source="/foo",destination="/rfm_workdir" '
                 'load/library/image:tag cmd')
-    elif container_variant in {'Singularity', 'Singularity+nopull'}:
+    elif container_variant in {'Singularity'}:
```
vkarak, line 105:
```suggestion
    elif container_variant == 'Singularity':
```

unittests/test_containers.py
```diff
@@ -189,6 +203,15 @@ def expected_cmd_run_opts(container_variant):
     elif container_variant == 'Singularity+nocommand':
         return ('singularity run -B"/path/one:/one" -B"/foo:/rfm_workdir" '
                 '--foo --bar image:tag')
+    elif container_variant in {'Apptainer'}:
```
vkarak, line 206:
```suggestion
    elif container_variant == 'Apptainer':
```

vkarak (changes requested):

reframe/core/containers.py
vkarak: You need to list it also here: https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#systems-.partitions-.container_platforms-.type

unittests/test_containers.py
```diff
@@ -13,7 +13,8 @@
     'Sarus', 'Sarus+nocommand', 'Sarus+nopull', 'Sarus+mpi', 'Sarus+load',
     'Shifter', 'Shifter+nocommand', 'Shifter+mpi', 'Shifter+nopull',
     'Shifter+load',
-    'Singularity', 'Singularity+nocommand', 'Singularity+cuda'
+    'Singularity', 'Singularity+nocommand', 'Singularity+cuda',
+    'Apptainer', 'Apptainer+nocommand', 'Apptainer+cuda'
```
ekouts, line 17: It looks like these params are in alphabetical order, but it's fine to leave like that

ekouts (approved):

unittests/test_containers.py
teojgo: I was thinking to change them but in the end I left it like that to be side by side with the Singularity one

vkarak (approved):

PR #2618, opened 2022-10-01
[feat] Improve framework messages and the performance report

vkarak:
This PR introduces several small improvements that imo enhance the daily experience.

1. The log files are printed before the run start. Therefore it is easy to inspect the logs live for long runs, especially with the default log configuration, which creates a temporary log file.
2. The performance data of a  performance test is printed immediately after the test finishes (either successfully or not). This is especially useful for long performance exploration experiments, where you don't want to wait until the end of the run to see the performance in the performance report.
3. The performance report is modernised and improved. More information is listed but in a more compact and concise way.
4. All error and warning messages are prefixed with `ERROR` and `WARNING` always. This makes it easier spot errors and warnings, even when no colors are used.
5. Some debug messages where moved to debug2.
6. A warning when building the docs was fixed.
7. The docs and local run listings are updated.

TODOS

- [x] Update the remote run listings (@ekouts Could you do that on Daint?)

### DIFF
--- a/docs/listings/deps_rerun_t6.txt
+++ b/docs/listings/deps_rerun_t6.txt
@@ -1,22 +1,25 @@
 [ReFrame Setup]
-  version:           3.10.0-dev.3+149af549
+  version:           4.0.0-dev.1+36ffa085
   command:           './bin/reframe --restore-session --keep-stage-files -n T6 -r'
-  launched by:       user@host
+  launched by:       user@tresa.local
   working directory: '/home/user/Repositories/reframe'
-  settings file:     '<builtin>'
+  settings file:     '/home/user/Repositories/reframe/tutorials/config/settings.py'
   check search path: '/home/user/Repositories/reframe/unittests/resources/checks_unlisted/deps_complex.py'
   stage directory:   '/home/user/Repositories/reframe/stage'
   output directory:  '/home/user/Repositories/reframe/output'
+  log files:         '/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/rfm-xwildgxp.log'
 
 [==========] Running 1 check(s)
-[==========] Started on Sat Jan 22 23:44:25 2022
+[==========] Started on Sat Oct  1 20:06:40 2022 
 
 [----------] start processing checks
-[ RUN      ] T6 @generic:default+builtin
-[       OK ] (1/1) T6 @generic:default+builtin
+[ RUN      ] T6 /6dbdaf93 @catalina:default+gnu
+[ RUN      ] T6 /6dbdaf93 @catalina:default+clang
+[       OK ] (1/2) T6 /6dbdaf93 @catalina:default+gnu
+[       OK ] (2/2) T6 /6dbdaf93 @catalina:default+clang
 [----------] all spawned checks have finished
 
-[  PASSED  ] Ran 1/1 test case(s) from 1 check(s) (0 failure(s), 0 skipped)
-[==========] Finished on Sat Jan 22 23:44:25 2022
-Run report saved in '/home/user/.reframe/reports/run-report.json'
-Log file(s) saved in '/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/rfm-mug0a4cb.log'
+[  PASSED  ] Ran 2/2 test case(s) from 1 check(s) (0 failure(s), 0 skipped)
+[==========] Finished on Sat Oct  1 20:06:41 2022 
+Run report saved in '/home/user/.reframe/reports/run-report-111.json'
+Log file(s) saved in '/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/rfm-xwildgxp.log'
--- a/docs/listings/deps_rerun_t6.txt
+++ b/docs/listings/deps_rerun_t6.txt
@@ -1,25 +1,23 @@
 [ReFrame Setup]
-  version:           4.0.0-dev.1+36ffa085
+  version:           4.0.0-dev.1+3b06b112
   command:           './bin/reframe --restore-session --keep-stage-files -n T6 -r'
-  launched by:       user@tresa.local
+  launched by:       user@host
   working directory: '/home/user/Repositories/reframe'
-  settings file:     '/home/user/Repositories/reframe/tutorials/config/settings.py'
+  settings file:     '<builtin>'
   check search path: '/home/user/Repositories/reframe/unittests/resources/checks_unlisted/deps_complex.py'
   stage directory:   '/home/user/Repositories/reframe/stage'
   output directory:  '/home/user/Repositories/reframe/output'
-  log files:         '/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/rfm-xwildgxp.log'
+  log files:         '/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/rfm-r1gopvcr.log'
 
 [==========] Running 1 check(s)
-[==========] Started on Sat Oct  1 20:06:40 2022 
+[==========] Started on Fri Oct  7 23:06:23 2022 
 
 [----------] start processing checks
-[ RUN      ] T6 /6dbdaf93 @catalina:default+gnu
-[ RUN      ] T6 /6dbdaf93 @catalina:default+clang
-[       OK ] (1/2) T6 /6dbdaf93 @catalina:default+gnu
-[       OK ] (2/2) T6 /6dbdaf93 @catalina:default+clang
+[ RUN      ] T6 /6dbdaf93 @generic:default+builtin
+[       OK ] (1/1) T6 /6dbdaf93 @generic:default+builtin
 [----------] all spawned checks have finished
 
-[  PASSED  ] Ran 2/2 test case(s) from 1 check(s) (0 failure(s), 0 skipped)
-[==========] Finished on Sat Oct  1 20:06:41 2022 
-Run report saved in '/home/user/.reframe/reports/run-report-111.json'
-Log file(s) saved in '/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/rfm-xwildgxp.log'
+[  PASSED  ] Ran 1/1 test case(s) from 1 check(s) (0 failure(s), 0 skipped)
+[==========] Finished on Fri Oct  7 23:06:24 2022 
+Run report saved in '/home/user/.reframe/reports/run-report-125.json'
+Log file(s) saved in '/var/folders/h7/k7cgrdl13r996m4dmsvjq7v80000gp/T/rfm-r1gopvcr.log'
--- a/docs/listings/deps_run_t6.txt
+++ b/docs/listings/deps_run_t6.txt
@@ -1,41 +1,31 @@
 [ReFrame Setup]
-  version:           4.0.0-dev.1+36ffa085
+  version:           4.0.0-dev.1+3b06b112
   command:           './bin/reframe -c unittests/resources/checks_unlisted/deps_complex.py -n T6 -r'
-  launched by:       user@tresa.local
+  launched by:       user@host
   working directory: '/home/user/Repositories/reframe'
-  settings file:     '/home/user/Repositories/reframe/tutorials/config/settings.py'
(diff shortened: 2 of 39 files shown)


docs/listings/hello2_list_verbose.txt
```diff
@@ -1,25 +1,27 @@
...
 Initializing runtime
-Selecting subconfig for 'generic:default'
 Initializing system partition 'default'
-Selecting subconfig for 'generic'
-Initializing system 'generic'
+Initializing system 'catalina'
 Initializing modules system 'nomod'
-detecting topology info for generic:default
-> found topology file '/home/user/.reframe/topology/generic-default/processor.json'; loading...
+detecting topology info for catalina:default
```
ekouts, line 12: This listing is not correct. I looked up where we load it and it is in tutorial_tips_tricks.rst. There we say:
"You can see all the different phases ReFrame’s frontend goes through when loading a test. The first “strange” thing to notice in this log is that ReFrame picked the generic system configuration. This happened because it couldn’t find a system entry with a matching hostname pattern. However, it did not impact the test loading, because these tests are valid for any system, ..."

docs/listings/hellomp2.txt
```diff
@@ -1,60 +1,45 @@
...
-[ RUN      ] HelloThreadedExtendedTest @catalina:default+clang
-[     FAIL ] (1/2) HelloThreadedExtendedTest @catalina:default+clang
-==> test failed during 'sanity': test staged in '/path/to/reframe/stage/catalina/default/clang/HelloThreadedExtendedTest'
-[     FAIL ] (2/2) HelloThreadedExtendedTest @catalina:default+gnu
-==> test failed during 'sanity': test staged in '/path/to/reframe/stage/catalina/default/gnu/HelloThreadedExtendedTest'
+[ RUN      ] HelloThreadedExtendedTest /4733a67d @catalina:default+gnu
+[ RUN      ] HelloThreadedExtendedTest /4733a67d @catalina:default+clang
+[     FAIL ] (1/2) HelloThreadedExtendedTest /4733a67d @catalina:default+gnu
+==> test failed during 'sanity': test staged in '/home/user/Repositories/reframe/stage/catalina/default/gnu/HelloThreadedExtendedTest'
+[       OK ] (2/2) HelloThreadedExtendedTest /4733a67d @catalina:default+clang
```
ekouts, line 20: This test must be passing by accident (?), I think it should also be failing.

reframe/frontend/cli.py
```diff
@@ -1394,7 +1396,6 @@ def module_unuse(*paths):
             log_files = logging.log_files()
             if site_config.get('general/0/save_log_files'):
                 log_files = logging.save_log_files(rt.output_prefix)
-
```
ekouts, line 1397: I think we still want this empty line, right?

reframe/frontend/statistics.py
```diff
@@ -106,11 +107,12 @@ def json(self, force=False):
                     ],
                     'description': check.descr,
                     'display_name': check.display_name,
-                    'filename': inspect.getfile(type(check)),
                     'environment': None,
+                    'filename': inspect.getfile(type(check)),
```
ekouts, line 111: I have no problem with the order, i am just making sure. Did you move it in order for them be in alphabetical order? If yes, it should be moved a bit lower, after `fail_reason`.

ekouts (changes requested):

ekouts:
I was trying to run the listings on daint and realized that the performance report is not reporting all the progenvs. This is the new performance report from `alltests_daint`:
```bash
[ReFrame Setup]
  version:           4.0.0-dev.1+6c51b548
...
[ RUN      ] HelloThreadedExtended2Test /57223829 @daint:mc+cray
[ RUN      ] StreamWithRefTest /f925207b @daint:login+gnu
[ RUN      ] StreamWithRefTest /f925207b @daint:gpu+gnu
[ RUN      ] StreamWithRefTest /f925207b @daint:mc+gnu
...
[       OK ] (16/42) StreamWithRefTest /f925207b @daint:login+gnu
P: Copy: 69777.8 MB/s (r:0, l:None, u:None)
P: Scale: 42953.6 MB/s (r:0, l:None, u:None)
P: Add: 46951.0 MB/s (r:0, l:None, u:None)
P: Triad: 47320.6 MB/s (r:0, l:None, u:None)
[       OK ] (17/42) HelloMultiLangTest %lang=cpp /71bf65a3 @daint:mc+gnu
...
[       OK ] (25/42) StreamWithRefTest /f925207b @daint:mc+gnu
P: Copy: 49051.3 MB/s (r:0, l:None, u:None)
P: Scale: 31809.1 MB/s (r:0, l:None, u:None)
P: Add: 33709.2 MB/s (r:0, l:None, u:None)
P: Triad: 33932.1 MB/s (r:0, l:None, u:None)
[       OK ] (26/42) HelloMultiLangTest %lang=c /7cfa870e @daint:gpu+cray
...
[       OK ] (41/42) StreamWithRefTest /f925207b @daint:gpu+gnu
P: Copy: 50739.1 MB/s (r:0, l:None, u:None)
P: Scale: 34674.5 MB/s (r:0, l:None, u:None)
P: Add: 38353.3 MB/s (r:0, l:None, u:None)
P: Triad: 38683.2 MB/s (r:0, l:None, u:None)
[       OK ] (42/42) HelloThreadedExtended2Test /57223829 @daint:gpu+nvidia

[  PASSED  ] Ran 42/42 test case(s) from 4 check(s) (0 failure(s), 0 skipped)
[==========] Finished on Wed Oct  5 14:22:30 2022 

================================================================================
PERFORMANCE REPORT
--------------------------------------------------------------------------------
[StreamWithRefTest /f925207b @daint:mc:gnu]
  num_gpus_per_node: 0
  num_tasks: 1
  performance:
    - Copy: 49051.3 MB/s (r: 0 MB/s l: -inf% u: +inf%)
    - Scale: 31809.1 MB/s (r: 0 MB/s l: -inf% u: +inf%)
    - Add: 33709.2 MB/s (r: 0 MB/s l: -inf% u: +inf%)
    - Triad: 33932.1 MB/s (r: 0 MB/s l: -inf% u: +inf%)
--------------------------------------------------------------------------------
Run report saved in '/home/user/.reframe/reports/run-report-59.json'
Log file(s) saved in '/tmp/rfm-vju06n8f.log'
```

ekouts:
I have updated the remote listings, but we need to re-run `alltests_daint.txt` and `stream4_daint.txt` after the problem with the performance report is fixed. Please let me know if I missing any other listing.

I also needed to change some tests because the `nvidia` environment of daint needed a fix. I didn't push the changes because I think they may complicate the tutorial for no reason. What do you think? I needed to add this in a couple of tests.
```python
    @run_before('compile')
    def prgenv_nvidia_workaround(self):
        ce = self.current_environ.name
        if ce == 'nvidia':
            self.build_system.cppflags += [
                '-D__GCC_ATOMIC_TEST_AND_SET_TRUEVAL'
            ]
```

docs/listings/hello2_list_verbose.txt
vkarak: Good point. I think that this sentence should be removed, because it doesn't make sense. When we wrote this, almost 2 years ago, I was probably seeing that failure because I was running from VPN and reframe could not retrieve match the local hostname. Later on, we introduced the auto-detection from the fqdn and sometimes my mac's fqdn was rather weird, so reframe could not still pick the pattern. Both behaviours were not expected actually and that was an explanation for that listing. I will remove this now.

docs/listings/hellomp2.txt
vkarak: Indeed, that's a random success! :-) I could rerun it to get two failures.

reframe/frontend/statistics.py
vkarak: Exactly! I will sort them better ;-)

reframe/frontend/cli.py
vkarak: No, because in the line afterwards, we continue one level up.

vkarak:
> I was trying to run the listings on daint and realized that the performance report is not reporting all the progenvs. This is the new performance report from alltests_daint:

Hmm, that smells like a bug. I will have a look at it.

vkarak:
Btw, what do you think about the `P:` lines print outs? Should they be in the verbose mode?

ekouts:
> Btw, what do you think about the P: lines print outs? Should they be in the verbose mode?

I honestly don't know what I prefer. Initially I thought about asking to put it in verbose, but I kinda like it and find it useful. I think most people don't run in verbose mode and wouldn't see it probably. But again, I don't have a strong opinion on this.

> Good point, it will complicate it without a reason, but perhaps this deserves a nice small tip-n-trick section, where we can show the functionality of disabling the workaround hook. ;-)

Hm how about I put it under `# rfmdocend:` and then we show it only in the tips-and-tricks section?

vkarak:
> I honestly don't know what I prefer. Initially I thought about asking to put it in verbose, but I kinda like it and find it useful. I think most people don't run in verbose mode and wouldn't see it probably. But again, I don't have a strong opinion on this.

I had the same feelings as well, that's why I asked. I'm glad that we agree! I kinda prefer it in the non-verbose mode as well.

> Hm how about I put it under # rfmdocend: and then we show it only in the tips-and-tricks section?

Yes, if that works, that'd be good!

victorusu (approved):
`lgtm`

teojgo (approved):

reframe/frontend/statistics.py
```diff
@@ -315,42 +313,63 @@ def print_failure_stats(self, printer):
...
-        return ''
+        width = shutil.get_terminal_size()[0]
+        lines = ['', width*'=', 'PERFORMANCE REPORT', width*'-']
+
+        # Collect all the records from performance tests
+        perf_records = {}
+        for run in self.json():
+            for tc in run['testcases']:
+                if tc['perfvars']:
+                    key = tc['unique_name']
```
ekouts, line 324: I think this is why the performance report contains only one env-system per test
```suggestion
                    key = (tc['unique_name'], tc['system'], tc['environment'])
```

ekouts (changes requested):

reframe/frontend/statistics.py
ekouts: oh, sorry, my bad. Didn't realize you had already fixed the bug.

docs/listings/deps_rerun_t6.txt
```diff
@@ -1,22 +1,25 @@
 [ReFrame Setup]
-  version:           3.10.0-dev.3+149af549
+  version:           4.0.0-dev.1+36ffa085
   command:           './bin/reframe --restore-session --keep-stage-files -n T6 -r'
-  launched by:       user@host
+  launched by:       user@tresa.local
   working directory: '/home/user/Repositories/reframe'
-  settings file:     '<builtin>'
+  settings file:     '/home/user/Repositories/reframe/tutorials/config/settings.py'
```
ekouts, line 6: If I am not mistaken this should run without configuration file.

docs/listings/deps_run_t6.txt
```diff
@@ -1,30 +1,41 @@
 [ReFrame Setup]
-  version:           3.10.0-dev.3+149af549
+  version:           4.0.0-dev.1+36ffa085
   command:           './bin/reframe -c unittests/resources/checks_unlisted/deps_complex.py -n T6 -r'
-  launched by:       user@host
+  launched by:       user@tresa.local
   working directory: '/home/user/Repositories/reframe'
-  settings file:     '<builtin>'
+  settings file:     '/home/user/Repositories/reframe/tutorials/config/settings.py'
```
ekouts, line 6: This should have been run without config based on gendoclistings. Maybe you had the env var before running the script?

ekouts (changes requested):
Some very minor comments. `deps_rerun_t6` and `deps_rerun_t6` should be run without config file according to gendoclistings. We should update the listingsa or the `gendoclistings.py`

docs/listings/deps_rerun_t6.txt
vkarak: Strange, I haven't change the `gendoclistings.py`. I will fix it.

vkarak:
@ekouts The tutorial looks good to me; I just did a little bit of refinement.

ekouts (approved):

PR #2620, opened 2022-10-03
[feat] Add `perflog_ignore` attribute in httpjson handler

ekouts:
Since Kibana rejects a record that includes a list of values of different types (for example `check_perfvalues`) I added the `perflog_ignore` in the configuration which will exclude these attributes from the json that we eventually send.
I don't think we should exclude the key  from all logs. What do you think?

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1179,9 +1179,9 @@ The additional properties for the ``httpjson`` handler are the following:
    A set of optional key/value pairs to be passed with each log record to the server.
    These may depend on the server configuration.
 
-.. js:attribute:: .logging[].handlers[].perflog_ignore
+.. js:attribute:: .logging[].handlers[].ignore_keys
 
-.. object:: .logging[].handlers_perflog[].perflog_ignore
+.. object:: .logging[].handlers_perflog[].ignore_keys
 
    :required: No
    :default: ``[]``
@@ -1203,7 +1203,7 @@ An example configuration of this handler for performance logging is shown here:
            'facility': 'reframe',
            'data-version': '1.0'
        },
-       'perflog_ignore': ['check_perfvalues']
+       'ignore_keys': ['check_perfvalues']
    }
 
 
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1179,6 +1179,15 @@ The additional properties for the ``httpjson`` handler are the following:
    A set of optional key/value pairs to be passed with each log record to the server.
    These may depend on the server configuration.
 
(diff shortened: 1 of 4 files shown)


ekouts:
@jenkins-cscs retry dom

docs/config_reference.rst
```diff
@@ -1179,6 +1179,15 @@ The additional properties for the ``httpjson`` handler are the following:
    A set of optional key/value pairs to be passed with each log record to the server.
    These may depend on the server configuration.
 
+.. js:attribute:: .logging[].handlers[].perflog_ignore
```
vkarak, line 1182: Since it's already a perflog-only property, I would simply name it as `ignore_keys`.

reframe/core/logging.py
```diff
@@ -386,20 +387,30 @@ class HTTPJSONHandler(logging.Handler):
...
         self._extras = extras
+        self._perflog_ignore = perflog_ignore
 
     def _record_to_json(self, record):
+        def _log_record(key):
+            if key.startswith('_') or key in HTTPJSONHandler.LOG_ATTRS:
+                return False
+
+            if self._perflog_ignore and key in self._perflog_ignore:
+                return False
```
vkarak, line 402: Why don't you combine this in a single if?

reframe/core/logging.py
```diff
@@ -386,20 +387,30 @@ class HTTPJSONHandler(logging.Handler):
...
+
+            return True
+
         json_record = {
             k: v for k, v in record.__dict__.items()
-            if not (k.startswith('_') or k in HTTPJSONHandler.LOG_ATTRS)
+            if _log_record(k)
```
vkarak, line 408: Perhaps all this can fit in a single line now; same below.

reframe/core/logging.py
```diff
@@ -386,20 +387,30 @@ class HTTPJSONHandler(logging.Handler):
...
 
-    def __init__(self, url, extras=None):
+    def __init__(self, url, extras=None, perflog_ignore=None):
         super().__init__()
         self._url = url
         self._extras = extras
+        self._perflog_ignore = perflog_ignore
 
     def _record_to_json(self, record):
+        def _log_record(key):
```
vkarak, line 397: What about this?

```suggestion
        def _can_send(key):
```

vkarak (changes requested):

reframe/core/logging.py
ekouts: I thought it might be more readable but you are right, I changed it.

vkarak (approved):

teojgo (approved):

PR #2625, opened 2022-10-13
[bugfix] Get performance info from within the changedir

teojgo:
This allows performing more complicated performance queries such as:
```
def reference_meminfo(self):
        regex = 'memory from sysconf: total: \S+ \S+ avail: (?P<mem>\S+) GB'
        return sn.extractsingle(regex, self.stdout, 'mem', int)
```
If the above is executed outside the stagedir then `self.stdout` is not found.

### DIFF
--- a/reframe/frontend/executors/policies.py
+++ b/reframe/frontend/executors/policies.py
@@ -43,14 +43,15 @@ def _cleanup_all(tasks, *args, **kwargs):
 def _print_perf(task):
     '''Get performance info of the current task.'''
     
-    with change_dir(task.testcase.check.stagedir or '.'):
-        perfvars = task.testcase.check.perfvalues
-        for key, info in perfvars.items():
-            name = key.split(':')[-1]
-            getlogger().info(
-                f'P: {name}: {info[0]} {info[4]} '
-                f'(r:{info[1]}, l:{info[2]}, u:{info[3]})'
-            )
+    perfvars = task.testcase.check.perfvalues 
+    if perfvars:
+        with change_dir(task.testcase.check.stagedir):
+            for key, info in perfvars.items():
+                name = key.split(':')[-1]
+                getlogger().info(
+                    f'P: {name}: {info[0]} {info[4]} '
+                    f'(r:{info[1]}, l:{info[2]}, u:{info[3]})'
+                )
 
 
 class _PollController:
--- a/reframe/frontend/executors/policies.py
+++ b/reframe/frontend/executors/policies.py
@@ -43,7 +43,7 @@ def _cleanup_all(tasks, *args, **kwargs):
 def _print_perf(task):
     '''Get performance info of the current task.'''
     
-    with change_dir(task.testcase.check.stagedir):
+    with change_dir(task.testcase.check.stagedir or '.'):
         perfvars = task.testcase.check.perfvalues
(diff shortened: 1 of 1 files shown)


teojgo:
@jgphpc can you check this?

reframe/frontend/executors/policies.py
```diff
@@ -41,14 +42,15 @@ def _cleanup_all(tasks, *args, **kwargs):
...
-
-    perfvars = task.testcase.check.perfvalues
-    for key, info in perfvars.items():
-        name = key.split(':')[-1]
-        getlogger().info(
-            f'P: {name}: {info[0]} {info[4]} '
-            f'(r:{info[1]}, l:{info[2]}, u:{info[3]})'
-        )
+    
+    with change_dir(task.testcase.check.stagedir or '.'):
```
ekouts, line 46: Which is the case that we might not have stagedir?

ekouts:
@jenkins-cscs retry daint

reframe/frontend/executors/policies.py
teojgo: It seems that it was the case for some unittests. Maybe because it was a temporary directory.

ekouts: Hm, I have a look and I think it's because we are calling `_print_perf` in every failure (even if the test fails early, before stagedir is set). I think it's better we only change directory if there are perfvars:
```python
    perfvars = task.testcase.check.perfvalues
    if perfvars:
        with change_dir(task.testcase.check.stagedir):
```
 If we have perfvars then the test should have run and have a stagedir, right?

ekouts (changes requested):

reframe/frontend/executors/policies.py
teojgo: thanks @ekouts

vkarak: This `_print_perf()` is therefore a bug that I introduced with #2618 😬 What was your scenario that caused it to fail. It'd be good to add a unit test to check this.

vkarak (changes requested):
It looks like that this fix is a workaround for something different. At this point the test has finished and `perfvalues` is a "static" value; it does not compute anything on-the-fly, so I was wondering how this could fail?

teojgo:
> It looks like that this fix is a workaround for something different. At this point the test has finished and `perfvalues` is a "static" value; it does not compute anything on-the-fly, so I was wondering how this could fail?

The problem was coming when trying to run the test given here: https://github.com/eth-cscs/cscs-reframe-tests/pull/30

It was because the performance info was querying `self.stdout` which is just the name of the output file and cannot be retrieved as is if outside the stage dir.

vkarak:
The problem in this case is not how we call the `_print_perf()` in that part of the code. The problem is two-fold here. The test is ill-formed and ReFrame accepts it nonetheless. More specifically, the problem is that `reference` contains a deferred expression (the call to `reference_meminfo()`), which shouldn't be accepted. Then when we try to print the reference from the `perfvalues` after the test has finished, we (re)trigger it's evaluation and of course we fail at a strange location. The fix atm should be at the test's side, because it's ill formed and as far as the framework is concerned, I'll open issue to restrict the types accepted as references.

PR #2627, opened 2022-10-13
[feat] Allow tests to require references

vkarak:
This PR adds a new boolean test variable, named `require_reference`, that if set to true, it will cause the test to fail if there is no reference defined for the current system.

I believe that this is probable better than introducing a global configuration parameter, since variables give you both the ability to set them globally, but also per test.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -676,8 +676,11 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
     #: Require that a reference is defined for each system that this test is
     #: run on.
     #:
+    #: If this is set and a reference is not found for the current system, the
+    #: test will fail.
+    #:
     #: :type: boolean
-    #: :default: :obj:`False`
+    #: :default: :const:`False`
     #:
     #: .. versionadded:: 4.0.0
     require_reference = variable(typ.Bool, value=False)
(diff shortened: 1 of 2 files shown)


victorusu (approved):
`lgtm`

reframe/core/pipeline.py
```diff
@@ -673,6 +673,15 @@ def pipeline_hooks(cls):
     # FIXME: There is not way currently to express tuples of `float`s or
     # `None`s, so we just use the very generic `object`
 
+    #: Require that a reference is defined for each system that this test is
+    #: run on.
+    #:
+    #: :type: boolean
+    #: :default: :obj:`False`
```
ekouts, line 680:
```suggestion
    #: :default: :class:`False`
```
I think this is typo, but I am not sure what the `class` keyword means in the other examples.

ekouts (changes requested):
lgtm, I have a small comment for the configuration

reframe/core/pipeline.py
vkarak: Practically, they are rendered the same and `obj` is more correct as `False` is an object. In the past, I have started using `class` because that was it giving me the rendering that I wanted and didn't know that I could use `obj` too. However, now that I'm looking at the docs [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-obj), it should probably be `:const:` instead and perhaps we get also a link to the Python docs.

ekouts: Hm okay, I see. I tried the `:const:` and I got the same rendering and no link in the local docs. I guess const is a more accurate option but I am not sure if it will make a difference.

vkarak: No, it won't, but I can change it.

ekouts (approved):

PR #2634, opened 2022-10-18
[bugfix] Keep updating the node list during the job

### LINKED ISSUE #2567: Incomplete node list
On lumi the command `sacct -S <start_tim> -P -j <jobid> -o jobid,state,exitcode,end,nodelist` doesn't always give the complete node list since the beginning of the job. It would print for instance
```
JobID|State|ExitCode|End|NodeList
1305969|PENDING|0:0|Unknown|nid005828
1305969.batch|RUNNING|0:0|Unknown|nid005828
```
and a bit later it starts giving the whole information
```
JobID|State|ExitCode|End|NodeList
1305969|RUNNING|0:0|Unknown|nid[005828-005831]
1305969.batch|RUNNING|0:0|Unknown|nid005828
1305969.0|RUNNING|0:0|Unknown|nid[005829-005831]
```
This is a problem for reframe because it updates the node list only once https://github.com/reframe-hpc/reframe/blob/a5b66c7c41d7cc884893642fd4d9331b146a3c16/reframe/core/schedulers/slurm.py#L383-L392

rsarm:
Closes #2567

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -380,18 +380,12 @@ class SlurmJobScheduler(sched.JobScheduler):
         node_descriptions = completed.stdout.splitlines()
         return _create_nodes(node_descriptions)
 
-    def _expand_nodelist(self, nodespec):
-        completed = osext.run_command(f'scontrol show hostname {nodespec}')
-        return completed.stdout.splitlines()
-
     def _update_nodelist(self, job, nodespec):
-        if nodespec and nodespec != 'None assigned':
-            current_nodelist = self._expand_nodelist(nodespec)
+        if job.nodelist is not None:
+            return
 
-            if job.nodelist == current_nodelist:
-                return
-            else:
-                job._nodelist = current_nodelist
+        if nodespec and nodespec != 'None assigned':
+            job._nodelist = [n.name for n in self._get_nodes_by_name(nodespec)]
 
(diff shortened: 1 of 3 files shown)


rsarm:
Here the idea is to use the command `scontrol show hostname {nodespec}`, which I don't think it looks at all to the Slurm database. I think it's just an utility to expand a node list string. For instance
```bash
> scontrol show hostname xxx0[00-05]
xxx000
xxx001
xxx002
xxx003
xxx004
xxx005
```

I was considering checking if `len(job.nodelist)` is equal to number of nodes and in that case, stop updating  the node list. I'm a bit confused though with `job.num_tasks_per_node` in the case of flexible allocations. If `job.num_tasks_per_node` is `None` it means that there's only one node? Or that's not valid for flexible allocations?

vkarak (changes requested):
@rsarm I think the best approach is to simply store the `nodespec` as we get it back from the `sacct` or `squeue` and then convert it to a nodelist using `scontrol` once at the end after the job has finished. This is the easiest implementation, imho, until we implement the inverse of `nodelist_abbrev`.

vkarak (approved):
I've updated the PR according to my last suggestion. We save the `nodespec` with every `sacct` and we expand it into the `nodelist` the first time this is requested. I have also added a utility function to expand a nodespec into a nodelist.

PR #2638, opened 2022-10-19
[feat] Change default value of `num_gpus_per_node` to `None`

vkarak:
This is to match the rest of non-required resource-related variables that default to `None`. This will also make `num_gpus_per_node` disappear from the new performance report, if it is not set, as it happens with the rest of the variables.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -556,7 +556,7 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
     #: For more information on test resources, have a look at the
     #: :attr:`extra_resources` attribute.
     #:
-    #: :type: integral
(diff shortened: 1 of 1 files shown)


reframe/core/pipeline.py
```diff
@@ -557,8 +557,11 @@ def pipeline_hooks(cls):
     #: :attr:`extra_resources` attribute.
     #:
     #: :type: integral
```
ekouts, line 559:
```suggestion
    #: :type: integral or :class:`None`
```

ekouts (changes requested):

ekouts (approved):

ekouts:
@jenkins-cscs retry daint

PR #2643, opened 2022-10-20
[doc] Remove mailing list from README

vkarak:


ekouts (changes requested):
We should also remove it from the `README_minimal.md`.

ekouts (approved):

PR #2644, opened 2022-10-21
[feat] Extend `cray_cdt_version()` utility function to recognise CPE versions too

rsarm:
For this I guess with should change the name of the function since it has `cdt`. Maybe `cray_pe_version` and deprecate `cray_cdt_version `?

### DIFF
--- a/reframe/utility/osext.py
+++ b/reframe/utility/osext.py
@@ -573,7 +573,7 @@ def unique_abs_paths(paths, prune_children=True):
 
 
 def cray_cdt_version():
-    '''Return eithe the Cray Development Toolkit (CDT) version, the Cray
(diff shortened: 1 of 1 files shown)


ekouts (commented):
Looks fine, besides a small typo. I am also not sure if we should rename the function. If we do we will also have to update all the reframe tests that use it.

vkarak (approved):
The PR lgtm, but generally, I think that the use of this function, in general should be deprecated in favor of the new syntax of `valid_prog_environs` that can match environments based on features and extras. That function and, probably similar ones in the future should be called from a feature autodetection mechanism from the framework. I think also that CSCS' config and tests should move towards that direction abandoning the use of explicit system names and environments in the `valid_systems` and `valid_prog_environs`.

PR #2645, opened 2022-10-21
[test] Add optional validation `-c` option to the OSU hpclib tests

mszpindler:
I would like to propose optional `validation` parameter for the checks class that enables built-in OSU validation.

### DIFF
--- a/hpctestlib/microbenchmarks/mpi/osu.py
+++ b/hpctestlib/microbenchmarks/mpi/osu.py
@@ -113,6 +113,14 @@ class osu_benchmark(rfm.RunOnlyRegressionTest):
     num_tasks = required
     num_tasks_per_node = 1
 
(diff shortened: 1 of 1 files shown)


ekouts:
Ok to test

vkarak (commented):
That's a good proposal and I would propose to replace our current validation completely. It's always better to use the benchmark's validation. @rsarm What do you think?

rsarm:
I agree. I think it's a good idea. @mszpindler had told me about it offline.

vkarak (changes requested):

vkarak:
@mszpindler I have made a PR to your branch with my suggestion to use always the `-c` option. Could you please merge it, so that we can merge this one as well?

vkarak:
Replaced by #2679.

PR #2647, opened 2022-10-27
[feat] Introduce variable aliases

vkarak:
This practically extends on the field aliasing that we used to have and brings the same concept to variables. Users can now define a variable as an alias of another one. This is quite useful for transitional periods and deprecations when a variable is renamed. An aliased can be defined as follows:

```python
x = variable(int, value=3)
y = variable(alias=x)
```

The variable `y` may not have a default value, but it can be independently loggable. Any update of variable `x` will be visible to `y` and vice versa. Aliased variables can refer also to inherited variables.

This feature is useful for transitional periods when a variable is renamed and a deprecation is introduced and we want to keep both names for a while. The new variable can also be wrapped with `deprecate()` to mark a deprecation. This allows us to introduces variable aliases and deprecations effortlessly.

This commit also introduces implementation improvements on how variables are inherited. In particular, inherited variables are no more set redundantly as they are being inherited by classes lower in the hierarchy.

### DIFF
--- a/reframe/core/variables.py
+++ b/reframe/core/variables.py
@@ -158,6 +158,14 @@ class TestVar:
     from :class:`EchoBaseTest` to throw an error indicating that the variable
     ``what`` has not been set.
 
+    Finally, variables may alias each other. If a variable is an alias of
+    another one it behaves in the exact same way as its target. If a change is
+    made to the target variable, this is reflected to the alias and vice
+    versa. However, alias variables are independently loggable: an alias may
+    be logged but not its target and vice versa. Aliased variables are useful
+    when you want to rename a variable and you want to keep the old one for
+    compatibility reasons.
+
     :param `types`: the supported types for the variable.
     :param value: the default value assigned to the variable. If no value is
         provided, the variable is set as ``required``.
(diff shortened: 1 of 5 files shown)


reframe/core/variables.py
```diff
@@ -172,16 +172,40 @@ class MyRequiredTest(HelloTest):
...
+                 '_loggable', '_name', '_target')
+
+    __mutable_props = ('_default_value',)
 
     def __init__(self, *args, **kwargs):
         field_type = kwargs.pop('field', fields.TypedField)
-        self._default_value = kwargs.pop('value', Undefined)
+        alias = kwargs.pop('alias', None)
+
+        if alias and not isinstance(alias, TestVar):
```
victorusu, line 194: Does it make sense to move these tests to a function?

victorusu (approved):
`lgtm`

reframe/core/variables.py
vkarak: Why? They are only used here.

victorusu:
@teojgo and @ekouts, could you please take a look at this PR?

teojgo (approved):
lgtm, I only have a question. Is the `null` string the default one when the alias is not loggable? I couldn't find where it comes from

vkarak:
> lgtm, I only have a question. Is the null string the default one when the alias is not loggable? I couldn't find where it comes from

The alias get their default value from their target. See the `_default_value` property.

ekouts:
@jenkins-cscs retry dom

teojgo:
> > lgtm, I only have a question. Is the null string the default one when the alias is not loggable? I couldn't find where it comes from
> 
> The alias get their default value from their target. See the `_default_value` property.

Ok, yeah it was difficult to find where the `null` comes from because in the end it comes from the `json.dumps`.

PR #2648, opened 2022-10-29
[feat] Deprecate `variables` in favour of the new `env_vars`

### LINKED ISSUE #1774: Replace variables with env_vars

vkarak:
The use of `variables` is deprecated in both the configuration and the tests. Users should use `env_vars` instead.

This PR builds on top of #2647.

Closes #1774.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,6 +38,16 @@ sys.path = [prefix, external] + sys.path
 import reframe                           # noqa: F401, F403
 import reframe.utility.osext as osext    # noqa: F401, F403
 
+# Sphinx erroneously uses `repr()` to display values and not the
+# human-readable `str()`. For this reason we define the following variable in
+# the `reframe` module so that objects can redirect to their `str()` method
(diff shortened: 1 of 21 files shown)


teojgo (approved):

giordano:
Out of curiosity, what changed since https://github.com/reframe-hpc/reframe/issues/1774#issuecomment-1252841310

> As discussed, we will not deprecate `variables` because this would cause a major disruption.

?

vkarak:
> Out of curiosity, what changed since https://github.com/reframe-hpc/reframe/issues/1774#issuecomment-1252841310

We were quite ambivalent with this change to be honest, but since 4.0 is a major version bump, we eventually thought that it's probably the right moment to introduce such a change. After all, all it requires is a search-and-replace. Deprecating `perf_patterns` in favour of `@performance_function`s on the other hand would be more tedious for the users, so atm we choose not to do so and just update the docs.

FYI, though, there is a bug in the master regarding the use of `variables` (see #2681). We will fix this asap.

PR #2649, opened 2022-10-30
[feat] Relax the value syntax of `env_vars`

### LINKED ISSUE #56: Some fields of `RegressionTest` are too restrictive
For example, the `variables` field. We now require it to be of type `dict[str, str]`, but in fact any type of values could be used, since the `str()` is called internally. This would allow us to use the following syntax:

```python
self.variables = {
    'OMP_NUM_THREADS': self.num_cpus_per_task
}
```

instead of the current one:

```python
self.variables = {
    'OMP_NUM_THREADS': str(self.num_cpus_per_task)
}
```

Internal issue: https://madra.cscs.ch/scs/reframe/issues/562

vkarak:
Now the value of environment variables can accept any Python type. ReFrame will invoke `str(value)` automatically to emit it in a script.

This PR depends on #2648.

Fixes #56.

### DIFF
--- a/reframe/core/environments.py
+++ b/reframe/core/environments.py
@@ -45,7 +45,15 @@ class Environment(jsonext.JSONSerializable):
         self._name = name
         self._modules = normalize_module_list(modules)
         self._module_names = [m['name'] for m in self._modules]
-        self._env_vars = collections.OrderedDict(env_vars)
+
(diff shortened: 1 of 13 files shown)


ekouts:
@jenkins-cscs retry daint

teojgo (approved):

PR #2656, opened 2022-11-15
[bugfix] Pass explicitly the `--cpus-per-task` option to `srun` for Slurm >= 22.05

### LINKED ISSUE #2655: Change of behaviour in Slurm version 22.05
From the release notes: https://slurm.schedmd.com/news.html
```
RELEASE NOTES FOR SLURM VERSION 22.05
...
-- srun will no longer read in SLURM_CPUS_PER_TASK. This means you will
    implicitly have to specify --cpus-per-task on your srun calls, or set the
    new SRUN_CPUS_PER_TASK env var to accomplish the same thing.
```

To achieve the same behavior as before we need either to pass explicitly `--cpus-per-task=$SLURM_CPUS_PER_TASK` or set the new `SRUN_CPUS_PER_TASK` env var.
```
export SRUN_CPUS_PER_TASK=$SLURM_CPUS_PER_TASK
```

As far as I understand we can pass the option for older version of slurm as well so there is no need to check for the version.

ekouts:
For Slurm version >= 22.05 we pass explicitly `--cpus-per-task` to `srun` if `num_cpus_per_task` is set. For older versions we simply emit the `srun` command. If we cannot get the version of Slurm we also pass the option since it shouldn't hurt in theory for older versions.

Fixes #2655 .

### DIFF
--- a/reframe/core/launchers/mpi.py
+++ b/reframe/core/launchers/mpi.py
@@ -17,7 +17,7 @@ from reframe.utility import seconds_to_hms
 class SrunLauncher(JobLauncher):
     def __init__(self):
         self.options = []
-        self.explicit_cpus_per_task = True
+        self.use_cpus_per_task = True
         try:
             out = osext.run_command('srun --version')
             match = re.search('slurm (\d+).(\d+).(\d+)', out.stdout)
@@ -31,14 +31,12 @@ class SrunLauncher(JobLauncher):
                     )
                 )
                 if slurm_version < semver.VersionInfo(22, 5, 0):
-                    self.explicit_cpus_per_task = False
-
+                    self.use_cpus_per_task = False
             else:
                 getlogger().warning(
                     'could not get version of Slurm, --cpus-per-task will be '
                     'set according to the num_cpus_per_task attribute'
                 )
-
         except Exception:
             getlogger().warning(
                 'could not get version of Slurm, --cpus-per-task will be set '
@@ -47,7 +45,7 @@ class SrunLauncher(JobLauncher):
 
     def command(self, job):
         ret = ['srun']
-        if self.explicit_cpus_per_task and job.num_cpus_per_task:
+        if self.use_cpus_per_task and job.num_cpus_per_task:
             ret.append(f'--cpus-per-task={job.num_cpus_per_task}')
 
         return ret
--- a/reframe/core/launchers/mpi.py
+++ b/reframe/core/launchers/mpi.py
@@ -20,7 +20,7 @@ class SrunLauncher(JobLauncher):
         self.explicit_cpus_per_task = True
         try:
             out = osext.run_command('srun --version')
-            match = re.search('slurm (\d+).0*(\d+).0*(\d+)', out.stdout)
+            match = re.search('slurm (\d+).(\d+).(\d+)', out.stdout)
             if match:
                 # We cannot pass to semver strings like 22.05.1 directly
                 # because it is not a valid version string for semver. We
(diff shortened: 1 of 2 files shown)


vkarak (changes requested):
You have the `job` so you can set the `--cpus-per-task` based on the `job`'s value, from `job.num_cpus_per_task`. If it's `None`, you simply don't add the option. Also, I would also suggest to check the `srun --version` during the initialization of the launcher and emit this option only if it is greater than 22.05.

ekouts:
> You have the job so you can set the --cpus-per-task based on the job's value, from job.num_cpus_per_task. If it's None, you simply don't add the option.

I thought about this but I think it wouldn't work if the user is setting the option from the cli with `-Jcpus-per-task`, right? Or if it's part of the partition options in the configuration. I will try it to make sure

reframe/core/launchers/mpi.py
```diff
@@ -3,15 +3,54 @@
...
 
 
 @register_launcher('srun')
 class SrunLauncher(JobLauncher):
+    def __init__(self):
+        self.options = []
+        self.explicit_cpus_per_task = True
+        try:
+            out = osext.run_command('srun --version')
+            match = re.search('slurm (\d+).0*(\d+).0*(\d+)', out.stdout)
```
vkarak, line 23: There is no need to match the 0s. You can then convert them to an int and pass them to `VersionInfo`. Also you must match explicitly the dots.

```suggestion
            match = re.search('slurm (\d+)\.(\d+)\.(\d+)', out.stdout)
```

reframe/core/launchers/mpi.py
```diff
@@ -3,15 +3,54 @@
...
+            if match:
+                # We cannot pass to semver strings like 22.05.1 directly
+                # because it is not a valid version string for semver. We
+                # need to remove all the leading zeros.
+                slurm_version = (
+                    semver.VersionInfo(
+                        match.group(1), match.group(2), match.group(3)
```
vkarak, line 30:
```suggestion
                        int(match.group(1)), int(match.group(2)), int(match.group(3))
```

reframe/core/launchers/mpi.py
```diff
@@ -3,15 +3,54 @@
...
+                    )
+                )
+                if slurm_version < semver.VersionInfo(22, 5, 0):
+                    self.explicit_cpus_per_task = False
+
```
vkarak, line 35: Remove this blank line.

reframe/core/launchers/mpi.py
```diff
@@ -3,15 +3,54 @@
...
 from reframe.core.launchers import JobLauncher
+from reframe.core.logging import getlogger
 from reframe.utility import seconds_to_hms
 
 
 @register_launcher('srun')
 class SrunLauncher(JobLauncher):
+    def __init__(self):
+        self.options = []
+        self.explicit_cpus_per_task = True
```
vkarak, line 20:
```suggestion
        self.use_cpus_per_task = True
```

reframe/core/launchers/mpi.py
```diff
@@ -3,15 +3,54 @@
...
+                )
+                if slurm_version < semver.VersionInfo(22, 5, 0):
+                    self.explicit_cpus_per_task = False
+
+            else:
+                getlogger().warning(
+                    'could not get version of Slurm, --cpus-per-task will be '
+                    'set according to the num_cpus_per_task attribute'
+                )
+
```
vkarak, line 41: Remove this blank line.

vkarak (changes requested):

reframe/core/launchers/mpi.py
ekouts: I didn't need to call `int` because when I pass it to `semver.VersionInfo`, it doesn't complain with the string and does the conversion itself

vkarak (approved):

Issue #2093, opened 2021-07-23
#SBATCH --ntasks value


yellowhat:
Hi,
first of all thank you for this magnificent piece of work.

I would like to run tests using `spack` as build system and `slurm` as launcher.
This is my `settings.py` file:
```python
site_configuration = {
    "systems": [
        {
            "name": "cluster",
            "descr": "cluster",
            "hostnames": [".*"],
            "partitions": [
                {
                    "name": "hpc",
                    "scheduler": "slurm",
                    "launcher": "srun",
                    "access": ["--exclusive", "--partition hpc"],
                    "max_jobs": 2,
                    "environs": ["gnu"],
                },
            ],
        },
    ],
    "environments": [
          {
            "name": "gnu",
            "cc": "gcc-9",
            "cxx": "g++-9",
            "ftn": "gfortran-9"
        },
    ],
    "logging": [
        {
            "level": "debug",
            "handlers": [
                {
                    "type": "stream",
                    "name": "stdout",
                    "level": "info",
                    "format": "%(message)s"
                },
                {
                    "type": "file",
                    "level": "debug",
                    "format": "[%(asctime)s] %(levelname)s: %(check_info)s: %(message)s",  
                    "append": False
                }
            ],
            "handlers_perflog": [
                {
                    "type": "filelog",
                    "prefix": "%(check_system)s/%(check_partition)s",
                    "level": "info",
                    "format": (
                        "%(check_job_completion_time)s|reframe %(version)s|"
                        "%(check_info)s|jobid=%(check_jobid)s|"
                        "%(check_perf_var)s=%(check_perf_value)s|"
                        "ref=%(check_perf_ref)s "
                        "(l=%(check_perf_lower_thres)s, "
                        "u=%(check_perf_upper_thres)s)|"
                        "%(check_perf_unit)s"
                    ),
                    "append": True
                }
            ]
        }
    ],
}
```

This is the test that I am trying to run:
```python
import reframe as rfm
import reframe.utility.sanity as sn

@rfm.simple_test
class Fio(rfm.RegressionTest):
    descr = "fio benchmark"
    valid_systems = ["cluster:hpc"]
    valid_prog_environs = ["*"]
    vars_dirs = parameter(["/mnt/resource"])
    vars_rw = parameter(["read", "write"])
    vars_bs_size = parameter([("4K", "128M"), ("4M", "2G")])
    prerun_cmds = ["set -euo pipefail"]
    executable = "fio"
    executable_opts = [
[...]

Issue #2669, opened 2022-11-17
Remove the `schedulers` section from the configuration


vkarak:
And move it under the `partitions`. These are options tied to the scheduler of each partition.

ireed:
Is there a way to change or set the scheduler inside a test? For example, i can do 

> @run_before('run')
>  def prepare_run(self):
>      self.job.launcher = getlauncher('srun')()

 to set the launcher, or change the change it from whatever is in the config file. But using something like 

> @run_before('run')
> def prepare_run(self):
>    self.job.scheduler= getscheduler('slurm')()

will either be ignored or give a runtime error such as -- _Reason: attribute error: can't set attribute_ -- if i already have a config file.

I see no code, tests, or examples in the repo that demonstrates changing the scheduler from what is in the config file.

vkarak:
Hi @ireed, no you can't change the scheduler cos there is one instance per partition, as the scheduler needs to know of all the jobs on the partition in order to optimise its queries, e.g., check the status of all jobs at once. You could potentially hack it in, but wouldn't defining a new partition help?

PR #2693, opened 2022-11-26
[feat] Add ability to specify number of nodes for slurm scheduler

ethanjjjjjjj:
I believe that being able to explicitly ask for a number of nodes would be a useful feature, especially where the same set of tests are reused between clusters and partitions. This should give the ability to request full nodes from the Slurm scheduler without having to be explicit about how many tasks to run with, or how many cores the node has.

Helps solve some of the issues mentioned in #2093

### DIFF
--- a/docs/regression_test_api.rst
+++ b/docs/regression_test_api.rst
@@ -335,6 +335,7 @@ Mapping of Test Attributes to Job Scheduler Backends
    Test attribute               Slurm option                  Torque option                                                                              PBS option
    ============================ ============================= ========================================================================================== ==================
    :attr:`num_tasks`            :obj:`--ntasks`:sup:`1`       :obj:`-l nodes={num_tasks//num_tasks_per_node}:ppn={num_tasks_per_node*num_cpus_per_task}` :obj:`-l select={num_tasks//num_tasks_per_node}:mpiprocs={num_tasks_per_node}:ncpus={num_tasks_per_node*num_cpus_per_task}`
+   :attr:`num_nodes`            :obj:`--nodes`                n/a                                                                                        n/a
    :attr:`num_tasks_per_node`   :obj:`--ntasks-per-node`      see :attr:`num_tasks`                                                                      see :attr:`num_tasks`
    :attr:`num_tasks_per_core`   :obj:`--ntasks-per-core`      n/a                                                                                        n/a
    :attr:`num_tasks_per_socket` :obj:`--ntasks-per-socket`    n/a                                                                                        n/a
--- a/reframe/core/launchers/mpi.py
+++ b/reframe/core/launchers/mpi.py
@@ -134,6 +134,9 @@ class SrunAllocationLauncher(JobLauncher):
 
         if job.num_tasks:
             ret += ['--ntasks=%s' % str(job.num_tasks)]
+        
+        if job.num_nodes:
+            ret += ['--nodes=%s' % str(job.num_nodes)]
 
         if job.num_tasks_per_node:
             ret += ['--ntasks-per-node=%s' % str(job.num_tasks_per_node)]
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -541,7 +541,20 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
     #:
     #: .. |--flex-alloc-nodes| replace:: :attr:`--flex-alloc-nodes`
     #: .. _--flex-alloc-nodes: manpage.html#cmdoption-flex-alloc-nodes
-    num_tasks = variable(int, value=1, loggable=True)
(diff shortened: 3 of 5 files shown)


ethanjjjjjjj:
> Can I test this patch?

Go ahead

ethanjjjjjjj:
I'd appreciate some guidance on how to cleanly introduce this feature into ReFrame, as it stands, my needs are met by this small change, but I'm happy to keep working on it until many other people are satisfied.

vkarak:
Wouldn't setting `num_tasks = N` and `num_tasks_per_node = 1` have the same effect? Also, you can instruct reframe to emit the `-N` option always by setting the `use_nodes_option` configuration [option](https://reframe-hpc.readthedocs.io/en/stable/config_reference.html?highlight=use_nodes_option#schedulers-.use_nodes_option).

If you use that option, note that in 4.0 all the scheduler-specific options will be moved inside the partition definition (see also issue #2669).

ethanjjjjjjj:
> Wouldn't setting num_tasks = N and num_tasks_per_node = 1 have the same effect? Also, you can instruct reframe to emit the -N option always by setting the use_nodes_option configuration [option](https://reframe-hpc.readthedocs.io/en/stable/config_reference.html?highlight=use_nodes_option#schedulers-.use_nodes_option).


This is very close to what I want, it adds the --nodes line to the job script, but then only allows me to allocate a fixed number of mpi tasks per node.

 I'd like that to be automatically filled in by the scheduler, which slurm will do as long as you leave --ntasks out of your script. Allowing Slurm to come up with the number of tasks, means that I can run the test on a new cluster without having to specify how many tasks will be required to fill the node (assuming 1 cpu per task).

vkarak:
You are right, I wasn't aware of this capability of the `--cpus-per-task`:

> If -c is specified without -n, as many tasks will be allocated per node as possible while satisfying the -c restriction.

My concern regarding the implementation is that Introducing simply a `num_nodes` in parallel with the `num_tasks` will require changes in many places and might have strange side-effects if both are specified. It's not only the Slurm backend that should be updated; all the other backends should treat that option and should do something if it is specified along with the `num_tasks` and `num_tasks_per_node`. Slurm gives much more context in all of these options compared to other schedulers, and I would rather avoid implementing Slurm's behaviour in all other backends. That was also the rationale of keeping minimal and well defined what is specified in a test by default. Adding a `num_nodes` would require us to introduce logic and possibly reframe-specific interpretations on what to do if a job spec is over-specified. There is also the flexible node allocation and the `--distribute` option that should take care of a test that specifies `num_nodes`, `num_tasks` and `num_tasks_per_node`. On the other hand, what you propose here is a valid request that reframe should have a way to support this behaviour for the Slurm backend. I'm thinking what could be an alternative implementation that would be rather Slurm-specific and would not have to be exposed to the test API, if possible.

ethanjjjjjjj:
It seems to me that Slurm is already given special treatment with many options as shown by the table [here](https://reframe-hpc.readthedocs.io/en/latest/regression_test_api.html#mapping-of-test-attributes-to-job-scheduler-backends). While adding another is definitely not the cleanest way to deal with this looking to the future it certainly wouldn't be out of place with the current state of things. I agree and I'm on board with finding a more long term solution to all those other examples too though.

In this case, my requirements depends on the omission of `--ntasks` from the job script, therefore I believe there should still be a way to set `num_tasks=None` and this requires at least some small modification to the test API.

I think having the option of being explicit within the frontend about exactly how the job is run by the  scheduler behind the scenes would be desirable to many people. They might prefer to set `num_nodes` and `num_tasks` explicitly as you would in a job script, which is currently how this PR deals with assigning both of these options. Perhaps a dictionary of options within something like `RegressionTest.scheduler_opts.slurm` could satisfy this?

I think `--disribute` still launches one test per node that meets the filter which can still be multi node jobs, so in this case, I think specifying `num_nodes` explicitly doesn't change the behaviour here too much.

vkarak:
What about the following? If `num_tasks` could be set to `None`, which I agree is unavoidable for achieving this scenario, then you could write your test as follows:

```python
class my_test(...):
    num_tasks = None

    @run_before('run')    # or anywhere after setup
    def setup_job(self):
        self.job.options = [f'--nodes={N}']
```

Setting `num_tasks` to `None`, which is the only scheduler-specific variable that is not allowed to be `None` now, would be somehow the equivalent of telling the framework "I know what scheduler I'm using and I know what I'm doing with my job options." That's fine, I think, and then you can pass any additional options as I've shown above. Regarding the implementation, we should treat the case for other backends, but I believe this should also be straightforward: we don't emit anything derived from `num_tasks` and we simply process the user specified job options (that's already done).

ethanjjjjjjj:
Yep, that makes sense to me. Would definitely give me the control that I need over the job script.

vkarak:
@ethanjjjjjjj I will close and reopen this against the `develop` branch.

PR #2694, opened 2022-11-28
[feat] Remove the `schedulers` section from the configuration

### LINKED ISSUE #2669: Remove the `schedulers` section from the configuration

vkarak:
The scheduler options are now moved under in the `sched_options` inside each partition definition.

Closes #2669.


victorusu (approved):
`lgtm`.
I have to confess that I took some time to put my mind around this... 
Now that I have used it for a few days and have created several configurations, I think that this is actually an improvement in life! 
The config feels more consistent...

PR #2695, opened 2022-11-29
[feat] Control verbosity level of the immediate performance info

### LINKED ISSUE #2666: The immediate performance printout is too verbose
Some users don't like that the output is "spoiled" always with the `P:` during the run and would like that to be requested in a verbose mode. 

Generally, we should reconsider what is being printed in verbose and debug modes, as if this printout were to go now to the current verbose mode, then those that like the `P:` would start complaining of lots of irrelevant information being printed, while they just wanted the `P:` lines. Perhaps we need another verbosity level, or move some output to `debug`.

vkarak:
This PR introduces the `general/perf_info_level` configuration parameter and the `RFM_PERF_INFO_LEVEL` environment variable that control the level at which the immediate performance info is printed (i.e., the `P:` lines after a performance test finishes). The default level is `info`, which causes the info to be printed always, but it can be set to `verbose` if users want this output to be printed only with the `-v` option.

Closes #2666.


victorusu (commented):
I loved this feature! Thanks for this!
I am wondering here whether having a dedicated flag just for performance debugging would be better because the performance debugging increases also for everything else. This is a bit annoying when you are "just interested" in debugging the performance lines, which is in general my case... 
Having said that, I agree that having a single cmd option for debugging is consistent and this is just another debugging compared to any other one... The main difference is that in this case, we have dedicated control...
What do you think?

victorusu:
I have to confess that my tests here are not that bad! I am so used to the debugging info that they actually do not fog my ability to focus on the performance one... 😄

vkarak:
Can you post an example of what is your case? This PR is just about which level the immediate performance info will be printed. If you set it to `verbose`, then this info is going to show up only when the `-v` option is passed. As described in the issue, it's useful to control whether you want that information or not: if you run interactively, then you need that information printed (without all the verbose messages), so in that case you set its level info to `info` (the default). If you are not running interactive and you want a cleaner output, you select to print this information only in `verbose` level. Both scenarios are valid.

ekouts (changes requested):
Overall it looks good to me and I think it's on the right direction. I noticed that the error can be very misleading in case the log level has a typo:
```bash
$ export RFM_PERF_INFO_LEVEL=infus

$ ./bin/reframe ...

[ RUN      ] Foo /1cbec737 @generic:default+builtin
[       OK ] (1/1) Foo /1cbec737 @generic:default+builtin
[==========] Finished on Mon Dec 12 09:07:03 2022 
Run report saved in '/Users/ekoutsaniti/.reframe/reports/run-report-218.json'
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/frontend/cli.py", line 1275, in main
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/core/logging.py", line 887, in _fn
    return fn(*args, **kwargs)
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/frontend/executors/__init__.py", line 516, in runall
    self._runall(testcases)
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/frontend/executors/__init__.py", line 577, in _runall
    self._policy.exit()
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/frontend/executors/policies.py", line 340, in exit
  File "/Users/ekoutsaniti/repos/reframe-fork/reframe/frontend/executors/policies.py", line 411, in _advance_all

Log file(s) saved in '/var/folders/fk/vjppffg16jd1yv63wfbgsjd40000gp/T/rfm-_znxnhb0.log'
```
I think we should check that the string can be translated to a log level before running the tests and issue at least a warning. What do you think?

vkarak:
> I think we should check that the string can be translated to a log level before running the tests and issue at least a warning. What do you think?

In the config, I'm using an enum, but I'm not checking the value when it comes from the environment variable. I'll fix that.

ekouts (approved):

PR #2702, opened 2022-12-01
[bugfix] Restrict reference tuples acceptable types

### LINKED ISSUE #2630: Restrict the types accepted in reference tuples
Currently the tuples can accept anything, but this is problematic and can lead to subtle issues with ill-formed tests as the one described in #2625.

vkarak:
Reference tuples cannot accept a deferrable as their elements.

This is achieved nicely by extending the type checking system to support composite types and by allowing to insert in the type check subsystem any custom type. For example, one can do the following:

```python
Deferrable = typ.make_meta_type('Deferrable', _DeferredExpression)

# x can be either an integer or a deferrable or a list of integers or floats
x = variable(typ.Integer|Deferrable|typ.List[typ.Integer|typ.Float])

# y can be anything except Deferrable
y = variable(~typ.Deferrable)
```

This mechanism allows us to remove the `other_type` argument of `TypedField`, but it will probably break existing `variable` definitions of users.

Fixes #2630.

### DIFF
--- a/reframe/utility/typecheck.py
+++ b/reframe/utility/typecheck.py
@@ -93,8 +93,9 @@ of :class:`List[int]`.
 
 '''
 
-import abc
+import pdb
 import re
+import abc
 
 
 class ConvertibleType(abc.ABCMeta):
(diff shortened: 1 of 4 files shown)


unittests/test_typecheck.py
```diff
@@ -251,3 +251,42 @@ def __init__(self, x, y):
 
     assert Y('1').y == 1
     assert Z(5, 3).z  == 8
+
+
+def test_composition_of_types():
```
victorusu, line 256: Is this list of tests complete or exhaustive? Should it be?

vkarak: What do you think it's missing? Perhaps I could add a composition for Dicts and Tuples. Is it something else?

victorusu: Dicts were in my mind, but I didn't want to be explicit because I was unsure if it was the only missing type to add... 
Tuples, as you have said, are also missing... Do we have any other type that we support?
If so, we should add it here... What do you think?

vkarak: I've added more unit tests and it actually caught an error in the implementation.

reframe/utility/typecheck.py
```diff
@@ -93,8 +93,9 @@
 
 '''
 
-import abc
+import pdb
 import re
+import abc
```
ekouts, line 98: I guess `pdb` is left by mistake. Also `abc` should go where it was (before `re`).

ekouts (changes requested):
lgtm, just a small comment about imports

ekouts (approved):

victorusu (approved):
`lgtm`

PR #2703, opened 2022-12-05
[ci] Add support for Python 3.11 and update CI workflow for Python 3.6

vkarak:
We need to use an older container to test python 3.6 now, as the Github action with `ubuntu-latest` does not include it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,17 +34,22 @@ jobs:
         fail_ci_if_error: true
 
   unittest-py36:
-    runs-on: ubuntu-latest
-    permissions:
-      packages: read
-    container:
-      image: "reframehpc/rfm-ci-ubuntu-py36:latest"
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -33,6 +33,19 @@ jobs:
       with:
         fail_ci_if_error: true
 
+  unittest-py36:
```
teojgo, line 36: You can copy paste the the above and just use ubuntu 18.04 without relying on a container.

teojgo (changes requested):

.github/workflows/main.yml
vkarak: No, you can't do that anymore. Github complains. See the description of this PR. The only solution for supporting 3.6 is to test inside an older container.

vkarak: That was much simpler! Thanks @teojgo!

teojgo (approved):

PR #2707, opened 2022-12-08
[feat] Update `temp_environment` to use `env_vars`  as an argument

vkarak:
This is a leftover from the change of `variables` to `env_vars`.


victorusu (approved):
`lgtm`

PR #2716, opened 2022-12-20
[bugfix] Fix cleanup crash for non-performance tests when the `perflog_compat` option is on

### LINKED ISSUE #2715: Problem with multiline logging
I am not sure why exactly this happens but when I set the `perflog_compat` to `True` I am getting a weird error:
```bash
  File "/users/eirinik/repos/reframe-test/reframe/frontend/cli.py", line 1267, in main
  File "/users/eirinik/repos/reframe-test/reframe/core/logging.py", line 892, in _fn
    return fn(*args, **kwargs)
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/__init__.py", line 518, in runall
    self._runall(testcases)
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/__init__.py", line 579, in _runall
    self._policy.exit()
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/policies.py", line 340, in exit
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/policies.py", line 411, in _advance_all
```
I think the problem is coming from https://github.com/reframe-hpc/reframe/blob/master/reframe/core/logging.py#L737
Somehow the `self.extra['__rfm_check__']` is `None` and then an exception should have been raised in `check.perfvalues.items`. No exception is actually raised there but the test fails subsequently with the error above.

This is one test that reproduces the failure but I don't see anything obviously wrong with it: https://github.com/eth-cscs/cscs-reframe-tests/blob/main/checks/prgenv/affinity_check.py

ekouts:
This is first attempt to fix the bug, but maybe we should investigate more:
1. how to handle errors in `log_performance`. It's not a good idea to completely suppress them, because the error that ReFrame gives in this case can be very confusing
2. if it's correct to have `check == None` for non-performance tests

Fixes #2715 .

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -735,15 +735,16 @@ class LoggerAdapter(logging.LoggerAdapter):
         if multiline:
             # Log one record for each performance variable
             check = self.extra['__rfm_check__']
-            for var, info in check.perfvalues.items():
-                val, ref, lower, upper, unit = info
-                self.extra['check_perf_var'] = var.split(':')[-1]
-                self.extra['check_perf_value'] = val
-                self.extra['check_perf_ref'] = ref
-                self.extra['check_perf_lower_thres'] = lower
-                self.extra['check_perf_upper_thres'] = upper
-                self.extra['check_perf_unit'] = unit
-                self.log(level, msg)
+            if check:
+                for var, info in check.perfvalues.items():
+                    val, ref, lower, upper, unit = info
+                    self.extra['check_perf_var'] = var.split(':')[-1]
+                    self.extra['check_perf_value'] = val
+                    self.extra['check_perf_ref'] = ref
+                    self.extra['check_perf_lower_thres'] = lower
+                    self.extra['check_perf_upper_thres'] = upper
+                    self.extra['check_perf_unit'] = unit
+                    self.log(level, msg)
         else:
             self.log(level, msg)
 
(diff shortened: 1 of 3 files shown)


teojgo (changes requested):
I think that what you need to do is to check:
```
self.check.is_performance_check()
```

And perform the performance logging in https://github.com/reframe-hpc/reframe/blob/1724a35cdaff557eba3f3e3a049a43760ae7f64b/reframe/frontend/executors/__init__.py#L389

only when it is actually a performance check. If it's not a performance test, then the perflogger will not have been set properly with the check name here: https://github.com/reframe-hpc/reframe/blob/1724a35cdaff557eba3f3e3a049a43760ae7f64b/reframe/frontend/executors/__init__.py#L372

ekouts:
> I think that what you need to do is to check:
> 
> ```
> self.check.is_performance_check()
> ```
> 
> And perform the performance logging in
> 
> https://github.com/reframe-hpc/reframe/blob/1724a35cdaff557eba3f3e3a049a43760ae7f64b/reframe/frontend/executors/__init__.py#L389
> 
> only when it is actually a performance check. If it's not a performance test, then the perflogger will not have been set properly with the check name here:
> 
> https://github.com/reframe-hpc/reframe/blob/1724a35cdaff557eba3f3e3a049a43760ae7f64b/reframe/frontend/executors/__init__.py#L372

Yes, I think your solution is nicer. I fixed it

teojgo (approved):

vkarak (approved):

PR #2735, opened 2023-01-14
[feat] Relax the regex matching for test names and tags

### LINKED ISSUE #2658: Use `search` instead of `match` when matching regexes for the various command line options
The `match` matches at the beginning of the string, which requires to add a `.*` at the pattern if we want to match something in the middle of the name. This is not very convenient for the end user.

vkarak:
We now use `re.search()` instead of `re.match()` for matching the name or tag patterns in the related cli options. So there is no need any more to prepend  `.*` to a name pattern so as to match something in the middle of the test name.

Closes #2658


vkarak:
Will reopen against `develop`.

PR #2746, opened 2023-01-18
[feat] Introduce a `--dry-run` option

### LINKED ISSUE #2242: Support a dry-run mode in ReFrame
This mode can simply emit job scripts without executing them.

ekouts:
Adding a "dry-run" attribute instead of a cli option gives to the user more flexibility, in case they want to run a fixture/dependency for some reason. The user can always run with `-Sdry_run_mode=1` so that they don't actually change the test in most cases.

I also changed the color of the "OK" and "RUN" messages to yellow for tests that have `dry_run_mode=True` but we can easily change it back if you don't like it. I just thought that this feature can easily lead to mistakes if someone is setting the test in `dry_run_mode` without noticing.

Closes #2242.

### DIFF
--- a/docs/tutorial_basics.rst
+++ b/docs/tutorial_basics.rst
@@ -440,12 +440,13 @@ For running the benchmark, we need to set the OpenMP number of threads and pin t
 You can set environment variables in a ReFrame test through the :attr:`~reframe.core.pipeline.RegressionTest.env_vars` dictionary.
 
 What makes a ReFrame test a performance test is the definition of at least one :ref:`performance function<deferrable-performance-functions>`.
-Similarly to a test's :func:`@sanity_function<reframe.core.pipeline.RegressionMixin.sanity_function>`, a performance function is a member function decorated with the :attr:`@performance_function<reframe.core.pipeline.RegressionMixin.performance_function>` decorator, which binds the decorated function to a given unit.
-These functions can be used by the regression test to extract, measure or compute a given quantity of interest; where in this context, the values returned by a performance function are referred to as performance variables.
-Alternatively, performance functions can also be thought as `tools` available to the regression test for extracting performance variables.
-By default, ReFrame will attempt to execute all the available performance functions during the test's ``performance`` stage, producing a single performance variable out of each of the available performance functions.
-These default-generated performance variables are defined in the regression test's attribute :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` during class instantiation, and their default name matches the name of their associated performance function.
-However, one could customize the default-generated performance variable's name by passing the ``perf-key`` argument to the :attr:`@performance_function<reframe.core.pipeline.RegressionMixin.performance_function>` decorator of the associated performance function.
+Similarly to a test's :func:`@sanity_function<reframe.core.builtins.sanity_function>`, a performance function is a member function decorated with the :func:`@performance_function<reframe.core.builtins.performance_function>` decorator that merely extracts or computes a performance metric from the test's output and associates it with a unit.
+By default, every performance function defined in the test is assigned to a *performance variable* with the function's name.
+A performance variable is a named quantity representing a performance metric that ReFrame will report on, log and can also check against a reference value.
+The performance variables of a test are stored in the :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` dictionary.
+The keys are the names of the metrics, whereas the values are :ref:`performance functions <deferrable-performance-functions>`.
+The :func:`@performance_function<reframe.core.builtins.performance_function>` decorator apart from turning an ordinary method into a "performance function", it also creates an entry in the :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` dictionary.
+The optional ``perf_key`` argument can be used to assign a different name to the newly created performance variable.
 
 In this example, we extract four performance variables, namely the memory bandwidth values for each of the "Copy", "Scale", "Add" and "Triad" sub-benchmarks of STREAM, where each of the performance functions use the :func:`~reframe.utility.sanity.extractsingle` utility function.
 For each of the sub-benchmarks we extract the "Best Rate MB/s" column of the output (see below) and we convert that to a float.
@@ -479,9 +480,11 @@ This is especially useful if you run long suites of performance exploration test
 Setting explicitly the test's performance variables
 ---------------------------------------------------
 
-In the above STREAM example, all four performance functions were almost identical except for a small part of the regex pattern, which led to some code repetition.
-Even though the performance functions were rather simple and the code repetition was not much in that case, this is still not a good practice and it is certainly an approach that would not scale when using more complex performance functions.
-Hence, in this example, we show how to collapse all these four performance functions into a single function and how to reuse this single performance function to create multiple performance variables.
+Users are allowed to manipulate the test's :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` dictionary directly.
+This is useful to avoid code repetition or in cases that relying on decorated methods to populate the :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` is impractical, e.g., creating multiple performance variables in a loop.
+
+You might have noticed that in our STREAM example above, all four performance functions are almost identical except for a small part of the regex pattern.
+In the following example, we define a single performance function, :func:`extract_bw`, that can extract any of the requested bandwidth metrics, and we populate the :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` ourselves in a pre-performance hook:
 
 .. code-block:: console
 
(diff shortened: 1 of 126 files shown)


docs/tutorial_basics.rst
```diff
@@ -119,6 +119,17 @@ Let's inspect what files ReFrame produced for this test:
...
+   .. tip::
+
+      .. versionadded:: 4.1.0
+
+        Until you get familiar with all the test attributes and options of ReFrame you can run a test in dry-run mode, with the :attr:`~reframe.core.pipeline.RegressionTest.dry_run_mode` attribute.
+        When this attribute is set to `True`, ReFrame will create the build and run script for the test but will not actually submit to the scheduler.
+        You can then find the created job script in the output directory of the test.
+
+        It is recommended to set this attribute directly in the cli with the :option:`--setvar` option instead of marking a test in the test file.
+        Just try adding `-Sdry_run_mode=1` to your command.
```
vkarak, line 131: I suggest providing an actual `--dry-run` option that would set a special `__rfm_dry_run__` variable inside the test.

reframe/frontend/executors/policies.py
```diff
@@ -98,9 +98,13 @@ def __init__(self):
 
     def runcase(self, case):
         super().runcase(case)
-        check, partition, environ = case
+        check, partition, _ = case
         task = RegressionTask(case, self.task_listeners)
-        self.printer.status('RUN', task.info())
+        if check.dry_run_mode:
+            self.printer.status('DRUN', task.info())
```
vkarak, line 104:
```suggestion
            self.printer.status('DRY', task.info())
```

reframe/frontend/executors/policies.py
```diff
@@ -205,7 +209,11 @@ def on_task_failure(self, task):
 
     def on_task_success(self, task):
         msg = f'{task.info()}'
-        self.printer.status('OK', msg, just='right')
+        if task.check.dry_run_mode:
+            self.printer.status('DOK', msg, just='right')
+        else:
+            self.printer.status('OK', msg, just='right')
+
```
vkarak, line 216: I'd rather keep the original behaviour and colors everywhere, except for changing `RUN` to `DRY`.

unittests/test_cli.py
```diff
@@ -298,6 +298,27 @@ def test_check_sanity_failure(run_reframe, tmp_path):
...
+    assert 'Traceback' not in stderr
+    assert returncode == 0
+    assert os.path.exists(
+        tmp_path / 'output' / 'generic' / 'default' /
+        'builtin' / 'SanityFailureCheck' / 'rfm_job.sh'
+    )
+    assert not os.path.exists(
+        tmp_path / 'output' / 'generic' / 'default' /
+        'builtin' / 'SanityFailureCheck' / 'rfm_job.out'
+    )
```
vkarak, line 319: I think that the creation of the output directory should skipped completely in the dry-run mode. And here we should check that the job script is generated in the stage directory.

vkarak (changes requested):
Besides, my individual comments in the code, I think the proposed changes are very intrusive. I don't think we should touch the scheduler backends. The dry-run mode mechanics should stop in the `RegressionTest`. If the test is in dry-run mode, we should simply skip the job submission. Similarly, the `compile|run_wait()` should return immediately and the `compile|run_complete()` should always return `True`. As for the sanity and performance phases, I suggest that we enter normally in those stages and exit just before we start the evaluation of the sanity or performance functions. That would allow to reveal test problems even in dry-run mode. Finally, we should probably omit completely the creation of the output directory as nothing will be eventually copied there and the "cleanup" stage should just return immediately. The idea of the dry-run mode is primarily to be able to inspect the generated scripts, which will be left in the stage dir. I would also claim that the copy from `sourcesdir` to the `stagedir` should still happen, as it will also reveal potential problems without actually running the test.

reframe/frontend/executors/policies.py
ekouts: I thought the color were useful when we suggested people to run with `-Sdry_run_mode=1`, especially for tests with fixtures.
Even if you see a lot of `DRY` messages it can be confusing in the end if the `OK` belongs to a dry-run or a normal run. But if I understand correctly the user will have to run everything in dry-run, we don't allow them to choose if they want to run something normally, right?

vkarak: Exactly, I would vote for everything in dry-run. Technically, depending on where the `--dry-run` option sets the corresponding variable, we could allow exceptionally to reset the dry run mode for a specific test. But I would start simple. And then `DOK` instead of `OK` is a bit weird as a name, isn't it?

ekouts: > And then DOK instead of OK is a bit weird as a name, isn't it?

Yes, I was removing the `D` from the message when it was printed, I just used it to differentiate from normal `OK` and set a different color. But if we are running all the tests in dry-run, `OK` is enough, I will remove it.

reframe/frontend/cli.py
```diff
@@ -380,6 +380,11 @@ def main():
         help=('Distribute the selected single-node jobs on every node that'
               'is in STATE (default: "idle"')
     )
+    run_options.add_argument(
+        '--dry-run', action='store_true',
```
vkarak, line 384: I think that's an action option not a run option.

unittests/test_cli.py
```diff
@@ -301,22 +301,37 @@ def test_check_sanity_failure(run_reframe, tmp_path):
...
+        tmp_path / 'stage' / 'generic' / 'default' /
         'builtin' / 'SanityFailureCheck' / 'rfm_job.out'
     )
+    assert not os.path.exists(
+        tmp_path / 'output' / 'generic' / 'default' /
+        'builtin' / 'SanityFailureCheck' / 'rfm_job.sh'
+    )
+
+
+def test_check_sanity_failure_distribute_dry_run(run_reframe, tmp_path):
```
vkarak, line 324: Instead of copy-pasting the tests, I suggest to have parameterised fixture to effectively duplicate them. Something like this:

```python
@pytest.fixture(params=['run', 'dry_run'])
def action(request):
    return request.param

def test_check_sanity_failure_distribute(run_reframe, tmp_path, action):
    ...
```

We will also need a test to check the `--dry-run` and  `--repeat` combination.

vkarak (changes requested):

reframe/frontend/executors/policies.py
```diff
@@ -138,7 +142,9 @@ def runcase(self, case):
 
             self._pollctl.reset_snooze_time()
             while True:
-                sched.poll(task.check.job)
+                if not task.check.is_dry_run():
+                    sched.poll(task.check.job)
```
vkarak, line 146: All the special treatment in the policies is not need, as the `RegressionTest` takes care of dealing with the dry run mode. Only the printout out should be changed here.

reframe/core/pipeline.py
```diff
@@ -2275,6 +2271,9 @@ def cleanup(self, remove_files=False):
               more details.
 
         '''
+        if self.is_dry_run():
+            return
```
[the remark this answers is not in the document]
vkarak: I think it's correct where it is now, since all the dry-run adaptation happen in `RegressionTest`.

vkarak (approved):

vkarak:
@ekouts The PR is now fine from my side.

reframe/frontend/printer.py
```diff
@@ -48,15 +48,15 @@ def status(self, status, message='', just=None, level=logging.INFO):
...
+            if status_stripped in ('DRY', 'SKIP'):
                 status = color.colorize(status, color.YELLOW)
-            elif status_stripped in ['FAIL', 'FAILED', 'ERROR']:
+            elif status_stripped in ('FAIL', 'FAILED', 'ERROR'):
                 status = color.colorize(status, color.RED)
             else:
                 status = color.colorize(status, color.GREEN)
 
         final_msg = f'[ {status} ] '
-        if status_stripped in ['OK', 'SKIP', 'FAIL']:
```
teojgo, line 59: is `'DRY'` needed here as well?

ekouts: No, the counter is increased to have an output like:
```bash
...
[ DRY      ] jupyterhub_submit_test /45d367fb @daint:jupyter_mc+builtin
[ DRY      ] NCO_MergeNC4CTest /6636720f @daint:gpu+builtin
[ DRY      ] NCO_MergeNC4CTest /6636720f @daint:mc+builtin
[       OK ] (  5/890) nvidia_smi_check %gpu_mode=ecc /78390e2f @daint:gpu+builtin
[       OK ] (  6/890) nvidia_smi_check %gpu_mode=compute /a936ce18 @daint:gpu+builtin
...
```
Only when the test is completed we want to increase the counter with the completed tests

teojgo (approved):

reframe/frontend/printer.py
vkarak: Nope, that's just the outcome of the test. We are not printing the `DRY` there :-)

PR #2761, opened 2023-01-25
[feat] Print the first lines of stderr and stdout in case of sanity failures

### LINKED ISSUE #2668: Print always the first lines of stderr (and stdout) in case of sanity failures
This will help to immediately spot easy-to-fix problems, such as wrong commands/options etc.

ekouts:
We do something similar in case of `BuildError`, where we add the `stderr` in the message of the exception: https://github.com/reframe-hpc/reframe/blob/develop/reframe/core/exceptions.py#L164
I print stderr and stdout in statistics because we have checks at CSCS that raise their own `SanityError` ([like this example](https://github.com/eth-cscs/cscs-reframe-tests/blob/main/checks/prgenv/affinity_check.py#L335)) and I thought other people may be using it like this.
What do you think? Should I change the PR and go with the approach of `BuildError`?

Closes #2668 .

### DIFF
--- a/reframe/frontend/statistics.py
+++ b/reframe/frontend/statistics.py
@@ -243,7 +243,6 @@ class TestStats:
                             lines.append(line[:-1])
 
             lines += [f'--- {filename} ---']
-
             return lines
 
         line_width = shutil.get_terminal_size()[0]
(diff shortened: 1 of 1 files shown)


ekouts:
I started the branch from master, I will close and redo the PR based on develop.

PR #2762, opened 2023-01-25
[feat] Print the first lines of stderr and stdout in case of sanity failures

### LINKED ISSUE #2668: Print always the first lines of stderr (and stdout) in case of sanity failures

ekouts:
We do something similar in case of `BuildError`, where we add the `stderr` in the message of the exception: https://github.com/reframe-hpc/reframe/blob/develop/reframe/core/exceptions.py#L164
I print stderr and stdout in statistics because we have checks at CSCS that raise their own `SanityError` ([like this example](https://github.com/eth-cscs/cscs-reframe-tests/blob/main/checks/prgenv/affinity_check.py#L335)) and I thought other people may be using it like this.
What do you think? Should I change the PR and go with the approach of `BuildError`?

Closes #2668 .

### DIFF
--- a/reframe/frontend/statistics.py
+++ b/reframe/frontend/statistics.py
@@ -231,18 +231,25 @@ class TestStats:
         return self._run_data
 
     def print_failure_report(self, printer, rerun_info=True):
-        def first_n_lines_of_file(filename, prefix, num_lines=10):
-            lines = [
-                f'--- {filename} (first {num_lines} lines) ---'
-            ]
-            with contextlib.suppress(OSError):
+        def _head_n(filename, prefix, num_lines=10):
+            # filename and prefix are `None` before setup
+            if filename is None or prefix is None:
+                return []
+
+            try:
                 with open(os.path.join(prefix, filename)) as fp:
+                    lines = [
+                        f'--- {filename} (first {num_lines} lines) ---'
+                    ]
                     for i, line in enumerate(fp):
                         if i < num_lines:
                             # Remove trailing '\n'
-                            lines.append(line[:-1])
+                            lines.append(line.rstrip())
 
-            lines += [f'--- {filename} ---']
+                lines += [f'--- {filename} ---']
+            except OSError as e:
+                lines = [f'--- {filename} ({e}) ---']
+
             return lines
 
         line_width = shutil.get_terminal_size()[0]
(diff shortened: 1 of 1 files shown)


reframe/frontend/statistics.py
```diff
@@ -229,6 +231,20 @@ def json(self, force=False):
         return self._run_data
 
     def print_failure_report(self, printer, rerun_info=True):
+        def first_n_lines_of_file(filename, prefix, num_lines=10):
+            lines = [
+                f'--- {filename} (first {num_lines} lines) ---'
+            ]
+            with contextlib.suppress(OSError):
```
vkarak, line 238: I would not suppress the OSError here. Instead, I would change the message to

```python
f'--- {filename} (<the-os-error-message>) ---'
```

reframe/frontend/statistics.py
```diff
@@ -229,6 +231,20 @@ def json(self, force=False):
         return self._run_data
 
     def print_failure_report(self, printer, rerun_info=True):
+        def first_n_lines_of_file(filename, prefix, num_lines=10):
```
vkarak, line 234: Let's be less verbose :-)

```suggestion
        def _head_n(filename, prefix, num_lines=10):
```

reframe/frontend/statistics.py
```diff
@@ -229,6 +231,20 @@ def json(self, force=False):
...
+            lines = [
+                f'--- {filename} (first {num_lines} lines) ---'
+            ]
+            with contextlib.suppress(OSError):
+                with open(os.path.join(prefix, filename)) as fp:
+                    for i, line in enumerate(fp):
+                        if i < num_lines:
+                            # Remove trailing '\n'
+                            lines.append(line[:-1])
```
vkarak, line 243: What if there is no trailing `\n`? I would rather `rstrip()` the line.

reframe/frontend/statistics.py
```diff
@@ -264,7 +280,14 @@ def print_failure_report(self, printer, rerun_info=True):
                              f" -p {r['environment']} --system "
                              f"{r['system']} -r'")
 
-            printer.info(f"  * Reason: {r['fail_reason']}")
+            mssg = r['fail_reason']
```
vkarak, line 283:
```suggestion
            msg = r['fail_reason']
```

vkarak: Also I would rather check against the `r['fail_info']['exc_type']` or the `exc_value`, which is more precise.

reframe/frontend/statistics.py
```diff
@@ -264,7 +280,14 @@ def print_failure_report(self, printer, rerun_info=True):
...
+            if mssg.startswith('sanity error'):
```
vkarak, line 284:
```suggestion
            if isinstance(r['fail_info']['exc_value'], SanityError):
```

vkarak (changes requested):
I agree that it's better to add in the failure report instead of bloating the `SanityError`.

vkarak (approved):
lgtm

PR #2766, opened 2023-01-25
[bugfix] Pass long job options with `=`

### LINKED ISSUE #2745: Invalid sbatch file generated with `--job-option nice=-1000`
Using `'scheduler': 'slurm'`, an invalid sbatch file is generated with `--job-option nice=-1000`:
```bash
#!/bin/bash
[...]
#SBATCH --nice -1000
```

ReFrame split the option in two arguments by removing the `=`, but that's not valid for Slurm as the second argument will begin with a `-` and is interpreted as another option:
```console
$ srun --nice -1000 true
srun: invalid option -- '1'
Try "srun --help" for more information
```
One could also argue it's a command-line parsing bug in Slurm, but also I don't think the split on `=` is documented right now.

vkarak:
Fixes #2745.

@flx42 Could you test this, as I don't have access to a Slurm cluster as a [privileged](https://slurm.schedmd.com/srun.html#OPT_nice) user?

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -935,7 +935,7 @@ def main():
             elif len(optstr) == 1:
                 parsed_job_options.append(f'-{optstr} {valstr}')
             else:
-                parsed_job_options.append(f'--{optstr} {valstr}')
+                parsed_job_options.append(f'--{optstr}={valstr}')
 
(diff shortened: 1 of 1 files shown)


flx42:
It works, thanks!

ekouts (approved):

PR #2769, opened 2023-01-26
[bugfix] Fix confusing stack trace dump when a logging error occurs

### LINKED ISSUE #2747: When performance logging fails, ReFrame crashes with a confusing dump
That doesn't even help you to debug:

```
  File "/users/eirinik/repos/reframe-test/reframe/frontend/cli.py", line 1267, in main
  File "/users/eirinik/repos/reframe-test/reframe/core/logging.py", line 892, in _fn
    return fn(*args, **kwargs)
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/__init__.py", line 518, in runall
    self._runall(testcases)
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/__init__.py", line 579, in _runall
    self._policy.exit()
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/policies.py", line 340, in exit
  File "/users/eirinik/repos/reframe-test/reframe/frontend/executors/policies.py", line 411, in _advance_all
```

The backtrace is the same as in #2715 and its root cause is that errors raised inside the `on_task_{success,failure}` execution policy callbacks are not handled correctly and execution proceeds with the `_current_tasks` not in a consistent state; so then it  tries to advance a test case that has already finished (RETIRED state) and reframe crashes at a later time producing the above error, which although it was caused by an earlier failure, there is no path that allows you to track the root cause.

This crash has appeared repeatedly when something was wrong with the performance logging, but it's not limited to it.

vkarak:
This is more general than a logging error, but this is the case that exposes the bug to the users. The problem was that we were unwittingly suppressing all exceptions (except `TaskExit`, which effectively denotes test errors) inside the `_advance_completing()` function. This was also posing a problem during development as it was masquerading programming errors that caused exceptions to be thrown from within this function. So much intricate was the bug to spot as so much trivial is the fix!

Fixes #2747.

### DIFF
--- a/reframe/frontend/executors/policies.py
+++ b/reframe/frontend/executors/policies.py
@@ -513,7 +513,6 @@ class AsynchronousExecutionPolicy(ExecutionPolicy, TaskEventListener):
                 return 1
             else:
                 return 0
-
         except TaskExit:
             self._partition_tasks[partname].remove(task)
             self._current_tasks.remove(task)
@@ -530,9 +529,9 @@ class AsynchronousExecutionPolicy(ExecutionPolicy, TaskEventListener):
             task.finalize()
             self._retired_tasks.append(task)
             self._current_tasks.remove(task)
+            return 1
         except TaskExit:
             self._current_tasks.remove(task)
-        finally:
             return 1
 
     def deps_failed(self, task):


vkarak:
@jenkins-cscs retry all

ekouts (approved):

PR #2770, opened 2023-01-31
[bugfix] Register performance functions even if `perf_variables` are defined in the class body and improve the performance function docs

### LINKED ISSUE #2704: The documentation of mixing `@performance_function`s and `perf_variables` is confusing
The problem is that when no `perf_variables` are set by the test explicitly, ReFrame will use the `performance_function`s to generate  the `perf_variables`. However, if the test sets the `perf_variables` explicitly, `performance_function`s have no (visible) effect. They are not used as performance variables at all. However, they are used as performance value extractors that can be assigned as values to the `perf_variables` dictionary. This is the equivalent of wrapping any function with the `make_performance_function()` and assigning to a performance variable. This behaviour is exactly displayed by `stream4.py` tutorial:

```python
    @performance_function('MB/s')
    def extract_bw(self, kind='Copy'):
        if kind not in {'Copy', 'Scale', 'Add', 'Triad'}:
            raise ValueError(f'illegal value in argument kind ({kind!r})')

        return sn.extractsingle(rf'{kind}:\s+(\S+)\s+.*',
                                self.stdout, 1, float)

    @run_before('performance')
    def set_perf_variables(self):
        self.perf_variables = {
            'Copy': self.extract_bw(),
            'Scale': self.extract_bw('Scale'),
            'Add': self.extract_bw('Add'),
            'Triad': self.extract_bw('Triad'),
        }
```

The `extract_bw()` is not a performance variable but rather a data extractor. At the absence of the `set_perf_variables`, though, it would be a performance variable. That's confusing! And it can lead to strange results in case of inheritance. For example, imagine that a base test defines  a set of `performance_function`s to be treated as performance variables and a subclass wants to add some more performance variables by setting the `perf_variables` directly, because that's more convenient and would avoid code repetition, then immediately it loses all the performance functions defined in the base class!

In my opinion, `performance_function` and `perf_variables` must be allowed at the same time and the dictionary would be simply built incrementally by all of them. With this solution, you could still override any of the performance functions of the base class by simply overriding the function. The following shows an example on how you can revert the performance function of the `StreamMultiSysTest` to that of the basic `StreamTest` of the tutorial:

```python
@rfm.simple_test
class my_test(StreamMultiSysTest):
    def extract_bw(self):
        pass

    def set_perf_variables(self):
        pass

    @performance_function('MB/s', perf_key='Copy')
    def extract_copy_perf(self):
        return sn.extractsingle(r'Copy:\s+(\S+)\s+.*', self.stdout, 1, float)
```

vkarak:
Although this PR is primarily about the updating and improving the docs, I use the bug fix as a title, so that it shows up in the release notes.

Closes #2704.

### DIFF
--- a/docs/tutorial_basics.rst
+++ b/docs/tutorial_basics.rst
@@ -429,12 +429,13 @@ For running the benchmark, we need to set the OpenMP number of threads and pin t
 You can set environment variables in a ReFrame test through the :attr:`~reframe.core.pipeline.RegressionTest.env_vars` dictionary.
 
 What makes a ReFrame test a performance test is the definition of at least one :ref:`performance function<deferrable-performance-functions>`.
-Similarly to a test's :func:`@sanity_function<reframe.core.pipeline.RegressionMixin.sanity_function>`, a performance function is a member function decorated with the :attr:`@performance_function<reframe.core.pipeline.RegressionMixin.performance_function>` decorator, which binds the decorated function to a given unit.
-These functions can be used by the regression test to extract, measure or compute a given quantity of interest; where in this context, the values returned by a performance function are referred to as performance variables.
-Alternatively, performance functions can also be thought as `tools` available to the regression test for extracting performance variables.
-By default, ReFrame will attempt to execute all the available performance functions during the test's ``performance`` stage, producing a single performance variable out of each of the available performance functions.
-These default-generated performance variables are defined in the regression test's attribute :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` during class instantiation, and their default name matches the name of their associated performance function.
-However, one could customize the default-generated performance variable's name by passing the ``perf-key`` argument to the :attr:`@performance_function<reframe.core.pipeline.RegressionMixin.performance_function>` decorator of the associated performance function.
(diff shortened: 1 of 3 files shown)


docs/tutorial_basics.rst
```diff
@@ -429,12 +429,13 @@ For running the benchmark, we need to set the OpenMP number of threads and pin t
...
-These default-generated performance variables are defined in the regression test's attribute :attr:`~reframe.core.pipeline.RegressionTest.perf_variables` during class instantiation, and their default name matches the name of their associated performance function.
-However, one could customize the default-generated performance variable's name by passing the ``perf-key`` argument to the :attr:`@performance_function<reframe.core.pipeline.RegressionMixin.performance_function>` decorator of the associated performance function.
+Similarly to a test's :func:`@sanity_function<reframe.core.builtins.sanity_function>`, a performance function is a member function decorated with the :func:`@performance_function<reframe.core.builtins.performance_function>` decorator that merely extracts or computes a performance metric from the test's output and associates it with a unit.
+By default, every performance function defined in the test is assigned to a *performance variable* with the function's name.
+A performance variable is a named quantity representing a performance metric that ReFrame will report on it, log it and can also check it against a reference value.
```
ekouts, line 434:
```suggestion
A performance variable is a named quantity representing a performance metric that ReFrame will report on, log and can also check against a reference value.
```
It sounds better to me like this but not that important.

ekouts (approved):

PR #2780, opened 2023-02-07
[bugfix] Do not issue a name conflict error for the same fixture imported from different test files

### LINKED ISSUE #2759: Reframe skips tests when a library test is used in multiple files
Tests that are coming from the same library have to come from the same file in order to not be skipped.
This will work:
```python
import reframe as rfm
from hpctestlib.microbenchmarks.gpu.gpu_burn import gpu_burn_check

@rfm.simple_test
class gpu_burn_check_benchmark(gpu_burn_check):
    use_dp = True
    duration = 40
    exclusive_access = True
    valid_systems = ['*']
    valid_prog_environs = ['*']
    reference = {
        'dom:gpu': {
            'gpu_perf_min': (4115, -0.10, None, 'Gflop/s'),
        },
    }

    tags = {'benchmark'}

@rfm.simple_test
class gpu_burn_check_diagnostic(gpu_burn_check):
    use_dp = True
    duration = 5
    exclusive_access = True
    valid_systems = ['*']
    valid_prog_environs = ['*']
    reference = {
        'dom:gpu': {
            'gpu_perf_min': (4115, -0.10, None, 'Gflop/s'),
        },
    }

    tags = {'diagnostic'}
```
We get two tests:
```bash
[List of matched checks]
- gpu_burn_check_diagnostic /6edf14bd
    ^gpu_burn_build ~generic:default+builtin /9cd86ed5
- gpu_burn_check_benchmark /77499166
    ^gpu_burn_build ~generic:default+builtin /9cd86ed5
Found 2 check(s)
```

But splitting it in two files we get this error:
```bash
WARNING: skipping test file '/Users/ekoutsaniti/repos/reframe/checks/gpu_burn_conflict_2.py': name conflict error: test 'gpu_burn_build_6698ebea' from '/Users/ekoutsaniti/repos/reframe/checks/gpu_burn_conflict_2.py' is already defined in '/Users/ekoutsaniti/repos/reframe/checks/gpu_burn_conflict_1.py' (rerun with '-v' for more information)
[List of matched checks]
- gpu_burn_check_diagnostic /6edf14bd
    ^gpu_burn_build ~generic:default+builtin /9cd86ed5
Found 1 check(s)
```

teojgo:
Fixes #2759 

- [x] Create unittest


ekouts (commented):
Would it be easy to make the PR to master, since it is a bugfix? Otherwise, it looks good to me

vkarak (changes requested):
I think you started off the branch from the `develop` and now trying to merge with `master`. You should branch off master and apply your changes there.

ekouts (approved):

vkarak (changes requested):
I think that this fix introduces another bug. :-) The reason that the framework refuses to run tests with conflicting names is because they cannot execute safely in parallel, as they will step on each other in the stage directory. This fix basically allows conflicting tests defined in different files to run together, which is wrong. However, the problem in #2759 is that _fixtures_ with the same name are disallowed. This is a bug, because fixtures with the same name (i.e., with the same scope and same variables defined) are guaranteed to run only once, so the framework should not complain about them being defined multiple times. Therefore the fix should be to ignore name conflicts in case of fixtures as fixtures are guaranteed to get a unique name and each one of those will executed once per scope.

unittests/resources/testlib/simple.py
```diff
@@ -0,0 +1,39 @@
...
+    valid_prog_environs = ['builtin']
+    executable = 'echo'
+    message = variable(str, value='World') 
+    hello_output = fixture(simple_echo, scope='environment')
+
+    @run_before('run')
+    def add_exec_prefix(self):
+        fixture_output = os.path.join(self.hello_output.stagedir,
+                                      str(self.hello_output.stdout))
+        self.executable_opts = [f'$(cat {fixture_output})', self.message]
```
vkarak, line 35: This doesn't have a purpose I think. I would rather completely omit this.

unittests/resources/testlib/simple.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class simple_echo(rfm.RunOnlyRegressionTest, pin_prefix=True):
+    descr = 'Simple Echo build fixture'
+    executable = 'echo'
+    executable_opts = ['Hello']
+
+    @sanity_function
+    def assert_success(self):
+        return sn.assert_found(r'Hello', self.stdout)
```
vkarak, line 19: This can also be made very dummy and compact as it's sole purpose is to create dependency:

```suggestion
class dummy_fixture(rfm.RunOnlyRegressionTest):
    executable = 'echo'
    sanity_patterns = sn.assert_true(1)
```

unittests/resources/testlib/simple.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@rfm.simple_test
+class simple_echo_check(rfm.RunOnlyRegressionTest):
+    descr = 'Simple Echo Test'
+    valid_systems = ['*']
+    valid_prog_environs = ['builtin']
+    executable = 'echo'
+    message = variable(str, value='World') 
+    hello_output = fixture(simple_echo, scope='environment')
```
vkarak, line 29:
```suggestion
    dummy = fixture(dummy_fixture, scope='environment')
```

vkarak (changes requested):

vkarak (approved):

PR #2782, opened 2023-02-08
[ci] Improve the release notes generation

vkarak:
Now the script recognises also the `[enhancement]` prefix and a new section about enhancement is added. Also, empty sections are skipped.

### DIFF
--- a/ci-scripts/genrelnotes.py
+++ b/ci-scripts/genrelnotes.py
@@ -15,8 +15,8 @@ def usage():
     sys.stderr.write('Usage: %s PREV_RELEASE CURR_RELEASE\n' % sys.argv[0])
 
 
-def extract_release_notes(git_output, tag):
-    return re.findall(r'pull request (#\d+).*\s*\[%s\] (.*)' % tag, git_output)
+def extract_release_notes(git_output):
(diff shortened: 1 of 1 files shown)


ci-scripts/genrelnotes.py
```diff
@@ -40,12 +40,14 @@ def extract_release_notes(git_output, tag):
...
-        'feat': '## New features and enhancements',
+        'feat': '## New features',
+        'enhancement': '## Enhancements',
+        'doc': '## Enhancements',
         'bugfix': '## Bug fixes',
         'testlib': '## Test library'
     }
     sections = collections.OrderedDict()
-    for tag in ['feat', 'bugfix', 'testlib', 'ci', 'doc']:
+    for tag in ['feat', 'enhancement', 'doc', 'bugfix', 'testlib', 'ci']:
```
teojgo, line 50:
```suggestion
    for tag in titles.keys():
```

teojgo (changes requested):

ci-scripts/genrelnotes.py
vkarak: The reason I did that was that the `ci` tag didn't have a dedicated section, removing it from here would make it not to appear in the release notes. I changed this loop completely to make the intent clearer.

vkarak:
@jenkins-cscs retry all

PR #2787, opened 2023-02-12
[enhancement] Properly combine `append`-type command-line options defined in execution modes

### LINKED ISSUE #2785: The `-S` option resets any test variable set in an execution mode using the `-S` option
## How to reproduce

Create the following `mode_config.py`:

```python
site_configuration = {
    'modes': [
        {
            'name': 'foo',
            'options': ['-c unittests/resources/checks/hellocheck.py', '-n ^HelloTest', '-S descr=foo_test']
        }
    ]
}
```

and then run the following:

```
reframe -C mode_config.py --mode=foo --describe | jq .[].descr
```

The test description is as expected:

```json
"foo_test"
```

However if we set an irrelevant variable, the variable set in the mode will be lost:

```
reframe -C mode_config.py --mode=foo -S num_tasks=1 --describe | jq .[].descr
```

Then the test description is reset:

```json
"C Hello World test"
```

vkarak:
The documentation for this case is a bit grey area. In documentation of the `modes` configuration parameter it is stated:

> The options of an execution mode will be passed to ReFrame as if they were specified in the command line.

And the docs for the `--mode` read as:

> If an option is specified both in an execution mode and in the command-line, then command-line takes precedence.

These are consistent for the options that can be specified only once: every time such an option is re-passed, it overwrites the previous one. But it's unclear what happens with options that are of an `append` type. The first excerpt implies that the options will be simply unpacked and combined with the same options in the command line (desired behaviour that this PR provides), but the second excerpt allows the interpretation that if such an option is specified in the command line will still overwrite completely that in the execution mode (current behaviour and what was validated in the unit tests).

I believe that options that can be specified multiple options must be combined and not overwritten, because this would allow users to set test variables in an execution mode and then set additional variables or reset existing ones from the command line without the need to repass all options.

Since the current behaviour was validated in the existing unit tests, I am not classifying this as a bug fix, but as an enhancement, so it targets the `develop` branch.

Fixes #2785.


vkarak:
Messed up the branches. Will reopen soon :-D

PR #2791, opened 2023-02-16
[ci] Use release/v1 version for the publish gh action

teojgo:


ekouts (approved):

vkarak:
@teojgo Can you explain the difference?

teojgo:
> @teojgo Can you explain the difference?

Yes, you can see for example in https://github.com/reframe-hpc/reframe/actions/runs/4087047455 there is a warning and if you visit the corresponding repo for the action: https://github.com/pypa/gh-action-pypi-publish#-master-branch-sunset- they say that the master should not be used anymore

PR #2793, opened 2023-02-17
[bugfix] Properly define `sched_options` in the configuration schema

### LINKED ISSUE #2792: Configuration validation error when `sched_options` is set inside a partition
Although this is the only way to define scheduler options in ReFrame 4.0, this does not work. 

## How to reproduce

Add the following in one of your partitions:

```python
                    'sched_options': {
                        'use_nodes_options': True
                    }
```

Then loading this configuration file, you get the following error:

```console
ERROR: failed to load configuration: could not validate configuration files: '['<builtin>', 'config/machines.py']': Additional properties are not allowed ('sched_options' was unexpected)
```

vkarak:
The unit tests are also adapted in order to reproduce the initial error.

Fixes #2792.

### DIFF
--- a/reframe/schemas/config.json
+++ b/reframe/schemas/config.json
@@ -102,18 +102,6 @@
                 }
             ]
         },
-        "sched_options": {
-            "type": "object",
-            "properties": {
-                "ignore_reqnodenotavail": {"type": "boolean"},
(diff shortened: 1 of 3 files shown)


victorusu (approved):
`lgtm`

PR #2796, opened 2023-02-18
[bugfix] Fix use of `sched_options` by the backend schedulers

### LINKED ISSUE #2794: `use_nodes_option` scheduler option does not work in ReFrame 4.0

### LINKED ISSUE #2795: `sched_options` not allowed in `systems` configuration section
Conversely to what the docs state.

vkarak:
This PR fixes three bugs related to `sched_options`. Two were introduced in 4.0 and one was present since the introduction of the `use_nodes_option` parameter.

1. Scheduler backends were not accessing the correct configuration section due in order to get the partition's `sched_options` due to a typo! This is fixed, unit tests are added to verify it and the configuration schema for accessing the `sched_options` has become stricter in order to avoid typos going undetected in the future.
2. Fix the configuration schema and allow `sched_options` to be defined inside the `systems` section, as it is also stated in the docs.
3. Fix crash of the `use_nodes_option` when then `num_tasks_per_node` was not set. This bug was present since the introduction of the `use_nodes_option`. If `num_tasks_per_node` is not set in this case, it is considered as `1`.

Fixes #2794.
Fixes #2795.

Fyi @smoors

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -359,7 +359,6 @@ System Partition Configuration
    :default: ``false``
 
    Always emit the ``--nodes`` Slurm option in the preamble of the job script.
-   This option is relevant to Slurm backends only.
 
    This option is relevant for the Slurm backends only.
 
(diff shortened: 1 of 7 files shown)


ekouts (approved):

vkarak:
I'm not sure that this would help, as the docs for 4.0 (not the `latest` or `stable` to) would still claim that `sched_options` work.

giordano:
I was reading `stable` docs and assumed I could use `sched_options` with v4.0.0 and [discovered the hard way](https://github.com/ukri-excalibur/excalibur-tests/pull/111#issuecomment-1460436145) that's not true.

PR #2798, opened 2023-02-22
[bugfix] Fix listing of non-fixture-based dependencies

teojgo:


vkarak:
@teojgo Can you explain a bit what problem this PR solves?

vkarak (changes requested):
Perhaps we should add a small cli test that you list test with dependencies, e.g., `./bin/reframe -c unittests/resources/checks_unlisted/deps_complex.py -l`

teojgo:
> @teojgo Can you explain a bit what problem this PR solves?

It fixes the problem where during listing, ReFrame tries to access fixture data even for non fixture tests.

vkarak (approved):

PR #2819, opened 2023-03-06
[feat] Support arbitrary Spack environment configuration

jack-morrison:
`reframe.core.buildsystems.Spack` supports setting `install_tree`, but there are many other ways to configure a Spack environment. This proposal introduces support for setting arbitrary Spack environment configuration options. The implementation is nearly identical to how `install_tree` works.

As an example with the changes proposed here, one could set the following:
`self.build_system.config_opts = ['concretizer:unify:false']`

If the configuration option is not already present in the default `spack.yaml`, it is added. If the configuration option is present in the default `spack.yaml`, Spack overwrites them with the values given via `spack config add`.

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -873,6 +873,8 @@ class Spack(BuildSystem):
     #:
     #: :type: :class:`List[str]`
     #: :default: ``[]``
+    #:
+    #: .. versionadded:: 4.2
     config_opts = variable(typ.List[str], value=[])
 
     def __init__(self):
@@ -889,7 +891,7 @@ class Spack(BuildSystem):
 
         config_opts += self.config_opts
         for opt in config_opts:
-       		ret.append(f'spack -e {self.environment} config add "{opt}"')
+            ret.append(f'spack -e {self.environment} config add "{opt}"')
 
         if self.specs:
             specs_str = ' '.join(self.specs)
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -869,6 +869,12 @@ class Spack(BuildSystem):
     #: :default: ``[]``
     install_opts = variable(typ.List[str], value=[])
 
+    #: A list of Spack configurations in flattened YAML.
+    #:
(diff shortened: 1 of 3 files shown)


jack-morrison:
Assuming this is moving forwards and there're no name changes, etc, I'm happy to add docs changes for this PR as well.

reframe/core/buildsystems.py
```diff
@@ -881,6 +887,10 @@ def emit_build_commands(self, environ):
             ret.append(f'spack -e {self.environment} config add '
                        f'"config:install_tree:root:{install_tree}"')
```
vkarak, line 888: Since you are doing exactly the same later with the `config_opts`, I would instead add an entry in a local `config_opts` that I would then iterate on. Something like this:

```python
config_opts = []
if self._auto_env:
    config_opts.append(...)

config_opts += self.config_opts
for opt in config_opts:
    ... 
```

vkarak (changes requested):
Hi @jack-morrison, thanks for the PR, I think it makes sense.

I would also add a unit test in `test_buildsystems.py` that simply checks whether the `spack config` commands are emitted correctly (there are already some examples that you could use as a template for your unit test).

Regarding the docs, I think the documentation of the `config_opts` is sufficient. Perhaps, you can also add a comment about this in the Spack build system tutorial.

reframe/core/buildsystems.py
jack-morrison: Good call. Is [9721b13](https://github.com/reframe-hpc/reframe/pull/2819/commits/9721b131a0473f1326f27850bfc690572e090189) what you're suggesting?

jack-morrison:
> Perhaps, you can also add a comment about this in the Spack build system tutorial.

Done via [f9bc8e6](https://github.com/reframe-hpc/reframe/pull/2819/commits/f9bc8e62db33804fcf8b6923fd04052a4edf578d)

vkarak (approved):
Lgtm!

vkarak:
ok to test

PR #2821, opened 2023-03-10
[testlib] support overriding executable_opts on the cmd line for gromacs test

smoors:
currently, the gromacs option `-nb` is always set in the hpctestlib, which is not always what you want.

this PR modifies the behavior if the user sets executable_opts on the cmd line via `--setvar executable_opts=...`.
in that case, assume that the user knows what they are doing and don't set any options (except for the input file).

### DIFF
--- a/hpctestlib/sciapps/gromacs/benchmarks.py
+++ b/hpctestlib/sciapps/gromacs/benchmarks.py
@@ -65,8 +65,8 @@ class gromacs_check(rfm.RunOnlyRegressionTest):
         self.prerun_cmds = [
             f'curl -LJO https://github.com/victorusu/GROMACS_Benchmark_Suite/raw/{self.benchmark_version}/{self.__bench}/benchmark.tpr'  # noqa: E501
         ]
-        if not self.executable_opts:
-            self.executable_opts = ['-nb', self.nb_impl]
+        if '-nb' not in ' '.join(self.executable_opts).split()
+            self.executable_opts += ['-nb', self.nb_impl]
         self.executable_opts += ['-s benchmark.tpr']
 
     @loggable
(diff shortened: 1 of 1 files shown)


hpctestlib/sciapps/gromacs/benchmarks.py
```diff
@@ -65,7 +65,9 @@ def prepare_test(self):
         self.prerun_cmds = [
             f'curl -LJO https://github.com/victorusu/GROMACS_Benchmark_Suite/raw/{self.benchmark_version}/{self.__bench}/benchmark.tpr'  # noqa: E501
         ]
-        self.executable_opts += ['-nb', self.nb_impl, '-s benchmark.tpr']
+        if not self.executable_opts:
+            self.executable_opts = ['-nb', self.nb_impl]
+        self.executable_opts += ['-s benchmark.tpr']
```
vkarak, line 70: `nb_impl` is a test parameter, so ignoring it completely here violates the whole purpose of having a parameter. Perhaps, defining another value, e.g., `auto` that would skip adding the `-nb` would make more sense to me? @victorusu what do think?

vkarak (changes requested):

hpctestlib/sciapps/gromacs/benchmarks.py
smoors: i agree that completely overriding `executable_opts` is a bit too big of a hammer.
however, adding another parameter value will multiply the number of tests for n good reason, so that's not optimal either.

i've modified this PR so that `-nb` will be overridden *only* if `-nb` is actually provided on the cmd line.
in all other cases, the original behavior remains the same, i.e. `['-nb', self.nb_impl, '-s benchmark.tpr']` is appended to the cmd line options. see https://github.com/reframe-hpc/reframe/pull/2821/commits/024c70a877a73b180344d3416d27e7947c243d72

what do you think of this solution?

smoors: on further thought, this is not a good solution either:
the test name always includes the `self.nb_impl` value, but with this change the real `-nb` value might be different, so this is misleading.

it's probably better to not add the `-nb` option on the cmd line and filter out the tests that have the `nb_impl` that you don't want.

i'll close this PR.

PR #2822, opened 2023-03-11
[feat] Add support for user-defined `spack env create` options

jack-morrison:
ReFrame, unless targeting a user-specified existing Spack environment, automatically passes the `-d rfm_spack_env` option to create an ephemeral environment. This is a small addition that allows users to define extra `spack env create` options ([link to Spack docs for list of options](https://spack.readthedocs.io/en/latest/command_index.html#spack-env-create)). 

With the changes proposed here, a user could define the following:
```python
self.build_system.env_create_opts = ["--without-view"]
```

which produces following in the build job script:
```sh
spack env create -d rfm_spack_env --without-view
```

While technically this allows for any `spack env create` options, my motivation for this feature is use of `--without-view`. ReFrame's use of temporary Spack environments leaves little use for a filesystem view, and disabling it can reduce the possibility of `spack load`-time spec collisions.

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -863,7 +863,7 @@ class Spack(BuildSystem):
     #: :default: :obj:`True`
     emit_load_cmds = variable(bool, value=True)
 
-    #: Options to pass to ``spack install``
+    #: Options to pass to ``spack install``.
     #:
     #: :type: :class:`List[str]`
     #: :default: ``[]``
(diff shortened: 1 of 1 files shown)


jack-morrison:
Failing unittests, so marking as a draft until I touch those up. Same for docs.

reframe/core/buildsystems.py
```diff
@@ -908,7 +916,7 @@ def _create_env_cmds(self):
 
         self.environment = 'rfm_spack_env'
         self._auto_env = True
-        return [f'spack env create -d {self.environment}']
+        return [f'spack env create -d {self.environment} {" ".join(self.env_create_opts)}']
```
vkarak, line 919: This adds an additional trailing whitespace if `env_create_opts` is empty and that's why unit tests are failing.

PR #2823, opened 2023-03-11
[bugfix] Allow proper assignment of `time_limit` from command line

### LINKED ISSUE #2815: Cannot set `time_limit` from cli using an integer
Trying to set with `-S time_limit=60` fails with the following error:

```console
WARNING: skipping test 'my_test': value error: invalid format for timer field (rerun with '-v' for more information)
```

However, the `time_limit` can be set as an integer according to the [docs](https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.pipeline.RegressionTest.time_limit).

vkarak:
This PR fixes the same problem for all other `TimerField` variables. 

To allow setting these fields properly from the command line, we get rid of the `TimerField` and replace it with the augmented type `Duration`, which is similar in concept with the `Bool` type. This type allows conversion from floats, int and strings and applies essentially the logic of the former `TimerField`. Since for this field we want to allow conversions also when it's set from inside the test, e.g., `time_limit = '10s'`, we add a new argument `allow_implicit` to the `TypedField` constructor (also exposed through the `variable` builtin), which make the field try implicit conversions before assigning the value to the test attribute. Therefore a variable declared as `x = variable(int, allow_implicit=True)` can be set with either `x = 1` or `x = '1'`.

Closes #2815.

### DIFF
--- a/reframe/utility/typecheck.py
+++ b/reframe/utility/typecheck.py
@@ -94,6 +94,7 @@ of :class:`List[int]`.
 '''
 
 import abc
+import datetime
 import re
 
 
@@ -152,19 +153,6 @@ class ConvertibleType(abc.ABCMeta):
 # container types
 
 
-class _CompositeType(abc.ABCMeta):
-    def __instancecheck__(cls, inst):
-        assert hasattr(cls, '_types') and len(cls._types) == 2
-        return (issubclass(type(inst), cls._types[0]) or
-                issubclass(type(inst), cls._types[1]))
-
-
-class _InvertedType(abc.ABCMeta):
-    def __instancecheck__(cls, inst):
-        assert hasattr(cls, '_xtype')
-        return not issubclass(type(inst), cls._xtype)
-
-
 class _BuiltinType(ConvertibleType):
     def __init__(cls, name, bases, namespace):
         # Make sure that the class defines `_type`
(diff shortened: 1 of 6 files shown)


ekouts (approved):

reframe/utility/typecheck.py
```diff
@@ -424,3 +411,50 @@ def make_meta_type(name, cls, metacls=_BuiltinType):
...
+    def __rfm_cast_float__(cls, val):
+        return Duration._assert_non_negative(val)
+
+    @classmethod
+    def __rfm_cast_str__(cls, val):
+        # First try to convert to a float
+        try:
+            val = float(val)
+        except ValueError:
+            time_match = re.match(r'^((?P<days>\d+)d)?'
```
teojgo, line 448: @vkarak Do you think we should allow floating point durations as well? If so the regex `\d*\.*\d+` should be used. I am asking since timedelta accepts floating point values.

vkarak: You can still set it as a float, either from the command line or from within the test. This checks only the case when you assign a string to it. So there we don't allow `3.5d`. I haven't actually changed that from the `TimerField`.

teojgo (approved):

PR #2825, opened 2023-03-13
[enhancement] Add log files in the run report

### LINKED ISSUE #2811: Store the log files in the run report
Similarly, we should also store the full list of config files, instead of a single one.

ekouts:
Closes #2811 .


vkarak:
I suggest making this PR against master, as it's a small enhancement rather than a new feature.

ekouts:
Closing since #2828 is the same but against master.

PR #2832, opened 2023-03-14
[feat] Support custom system auto-detection methods

### LINKED ISSUE #2744: Expand reframe core code for VM and cloud integration
Reframe in its current form cannot be used for cloud clusters without additional edits to the core API files, due to fundamental differences between VM vs baremetal behavior. This causes issues with both creating and maintaining cloud-based tests using Reframe.

The issue from a SW dev perspective is VMs needing to 'ask' the cloud provider for its specs. For example, because not all cores and HW on a physical machine are allocated to a cloud node, certain commands such as lscpu or checking locality/affinity can show infothat is misleading or incorrect. Getting this info in Azure VMs require a [server ping](https://github.com/JonShelley/reframe/blob/52e4ff0473703ab0b6f7f6002b6b003bf5df6050/reframe/core/config.py#L283) that returns the HW/VM info.

The issue from a SW management perspective is cloud users needing to maintain a special branch of Reframe that has [cloud-specific logic](https://github.com/JonShelley/reframe/blob/52e4ff0473703ab0b6f7f6002b6b003bf5df6050/reframe/core/config.py#L279-L298) - as @vkarak pointed out in #2444. This means not being able to access the latest Reframe code without doing regular merges, which is not good for automated or regression testing.

There are a few ways I can think of to handle both issues. None are ideal.

- A slight variation of [#2440](https://github.com/reframe-hpc/reframe/issues/2440). Introduce new cloud-specific functions that are equivalent to regular reframe functions (ex: self.current_system_**VM**.node_data) with the same behavior as the baremetal counterparts. But, different clouds likely have different behaviors - and, clouds themselves are not necessarily static, unlike baremetal OS - making this is option a stability, validation, and compatibility nightmare.
- Add conditionals inside existing reframe functions, similar to how reframe has different logic for various schedulers. But, that may require adding support for only one VM/cloud at a time, with each provider needing to wait their turn.
- Put VM function calls in various parts of the configuration that cloud users can fill in a separate, stand-alone python file. For example, with the above referenced [_detect_system()](https://github.com/JonShelley/reframe/blob/52e4ff0473703ab0b6f7f6002b6b003bf5df6050/reframe/core/config.py#L262) in config.py you can include a sub-function _detect_vm() that runs before any _detect_system() code and is empty by default. Cloud users fill in _detect_vm() in a separate file - cloud_config.py - with desired behavior. This would prevent the need for Reframe to account for different VM/cloud usage scenarios. And, automated cloud tests can pull the main reframe code, then pull their additional file(s) when building/initializing the environment.  The downside is outsourcing reframe development to cloud devs (basically, having mini-versions of reframe for every cloud provider), causing reframe devs to [rightfully] decline customer support from any cloud user.

Thoughts?

### LINKED ISSUE #2797: generalize autodetect setting
On some NERSC machines we are using `/etc/clustername` instead of `/etc/xthostname`.

Could the auto detection method be generalized to support run time specification of which path to check for cluster identity?

cc: @bcfriesen

ekouts:
~~There is a new section in the configuration that defines how to detect the system:~~
```python
# NOT VALID anymore
site_configuration = {
    'autodetect': [
        # The old methods can be defined like this
        {
            "method": "hostname",
        },
        {
            "method": "fqdn",
        },
        {
            "method": "xthostname",
        },
        # These are new methods
        {
            "method": "custom_cmd",
            "cmd": "echo sys",
            # we could also add a field to choose stdout or stderr
        },
        {
            "method": "custom_file",
            "file": "myhostnameo.txt"
        },
    ],
    'systems': [
...
```
~~Reframe will check the list and pick the system based on the first successful method. The default configuration has the standard `hostname` method. Reframe picks the first method that will be successful so as a result additional autodetect sections will be put on top of the list, like we do for systems.~~

EDIT from @vkarak 

I updated this PR so that the syntax of autodection methods is simplified but also generalised. More specifically, we introduce a new top-level section in the configuration, `autodetect_methods`. This is a list of strings which describe a method of autodetection. A normal string is a command to execute in order to retrieve a hostname. If this string is prefixed with `py::`, then the remaining string is interpreted as a Python method, which will be called to retrieve the hostname, e.g. `py::socket.gethostname`, will invoke `socket.gethostname()`. If the python callable name is not prefixed by a module name, then it will be searched in the configuration file.

Example config:

```python
site_configuration = {
    'autodetect_methods': ['cat /etc/xthostname', 'py::detect_azure_vm', 'py::socket.gethostname'],
    'systems': [...],
    'environments': [...]
}
```

A new `RFM_AUTODETECT_METHODS` environment variable is introduced, which is a comma-separated list of methods as described above. The former `RFM_AUTODETECT_METHOD`, `RFM_AUTODETECT_FQDN` and `RFM_AUTODETECT_XTHOSTNAME` are all deprecated as they can be replaced by the newly introduced env. variable.

We are still missing:
- [x] Documentation
- [x] Unit tests

Closes #2797.
Closes #2744.

@brandongc You should now be able to set `autodetect_methods` in your config to `['cat /etc/clustername', 'hostname']` and it should work.

@ireed This PR addresses also #2744 in the sense that you can now have reframe calling a custom Python method or a custom external script to retrieve a "hostname" that would match against the system's hostname patterns. Let me know if that would work in your scenario.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -68,15 +68,15 @@ It consists of the following properties, which we also call conventionally *conf
 
    A list of system auto-detection methods for identifying the current system.
 
-   The list can contain two type of methods:
+   The list can contain two types of methods:
 
    1. Python methods: These are prefixed with ``py::`` and should point to a Python callable taking zero arguments and returning a string.
-      If the specified Python callable is not prefixed with a module, it will be looked up in the loaded configuration file chain starting from the last file in the chain.
+      If the specified Python callable is not prefixed with a module, it will be looked up in the loaded configuration files starting from the last file.
       If the requested symbol cannot be found, a warning will be issued and the method will be ignored.
    2. Shell commands: Any string not prefixed with ``py::`` will be treated as a shell command and will be executed *during auto-detection* to retrieve the hostname.
       The standard output of the command will be used.
 
-   If the :option:`--system` option is not passed, ReFrame will try to autodetect the current system trying the methods in this list one after the other, until one of them succeeds.
+   If the :option:`--system` option is not passed, ReFrame will try to autodetect the current system trying the methods in this list successively, until one of them succeeds.
    The resulting name will be matched against the :attr:`~config.systems.hostnames` patterns of each system and the system that matches first will be used as the current one.
 
    The auto-detection methods can also be controlled through the :envvar:`RFM_AUTODETECT_METHODS` environment variable.
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -61,6 +61,29 @@ It consists of the following properties, which we also call conventionally *conf
    A list of `general configuration objects <#general-configuration>`__.
 
 
+.. py:data:: autodetect_methods
+
+   :required: No
+   :default: ``["py::socket.gethostname"]``
+
+   A list of system auto-detection methods for identifying the current system.
+
+   The list can contain two type of methods:
+
+   1. Python methods: These are prefixed with ``py::`` and should point to a Python callable taking zero arguments and returning a string.
+      If the specified Python callable is not prefixed with a module, it will be looked up in the loaded configuration file chain starting from the last file in the chain.
+      If the requested symbol cannot be found, a warning will be issued and the method will be ignored.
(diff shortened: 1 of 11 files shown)


vkarak (approved):
Lgtm now.

PR #2834, opened 2023-03-17
[feat] Add a symlink to the latest run report

vkarak:
This PR adds a `latest.json` symlink to the latest run report. This is useful when there are many reports and you would like to see all the details of the last run. Another extension that we could consider is to add an option to "tag" a report for quick future reference. This would create a symlink to the report using the tag name. What do you think?


ekouts (approved):
Lgtm. I am not sure how useful the tagging of the report would be, as the user could simply give a name for the json report file, right?

PR #2839, opened 2023-03-30
[feat] Add support for repeated testing through two new options: `--reruns` and `--duration`

### LINKED ISSUE #619: Keeping a test system busy, new execution policy?
I would like to extend reframe for the use case of generating a constant workload on a test/ warm spare system. 

```
./reframe -C my_config.py --tag workload1 -r  --exec-policy=continuous --queue-depth=10
```

Where the main loop would be something like:

```python
while queue_depth < arg.queue_depth:
    start(random.choice(tests))
```

with the desired behavior to always maintain a queue of queue_depth pending jobs for the scheduler to evaluate.

I am willing to do the implementation myself, but would like some feedback on the reception of this idea and possibly some suggestions on how to implement it. e.g. Should it be a new execution policy? If so what should it be called?

vkarak:
This PR introduces the following two options that are meant to allow users to use reframe for stress testing a target system:

- The `--reruns=N` option will rerun the same test suite N times (N+1 total runs).
- The `--duration=TIMEOUT` option will run the same test suite repeatedly until `TIMEOUT` expires.

Compared to the similar `--max-retries` option that reruns the failed tests, these options will report statistics as well as failures from _all_ runs.

Closes #619.

@brandongc FYI

### Todos

- [x] Update the docs

### DIFF
--- a/reframe/frontend/executors/__init__.py
+++ b/reframe/frontend/executors/__init__.py
@@ -11,14 +11,17 @@ import sys
 import time
 import weakref
 
+import reframe.core.fields as fields
 import reframe.core.logging as logging
 import reframe.core.runtime as runtime
 import reframe.frontend.dependencies as dependencies
 import reframe.utility.jsonext as jsonext
+import reframe.utility.typecheck as typ
 from reframe.core.exceptions import (AbortTaskError,
                                      JobNotStartedError,
                                      FailureLimitError,
                                      ForceExitError,
+                                     RunSessionTimeout,
                                      SkipTestError,
                                      TaskExit)
 from reframe.core.schedulers.local import LocalJobScheduler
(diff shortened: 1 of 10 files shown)


vkarak:
ok to test

teojgo:
Ok to test

reframe/frontend/executors/__init__.py
```diff
@@ -461,7 +467,11 @@ def on_task_skip(self, task):
 
     @abc.abstractmethod
     def on_task_failure(self, task):
-        '''Called when a regression test has failed.'''
+        '''Called when a RegressionTask has failed.'''
+
+    @abc.abstractmethod
+    def on_task_abort(self, task):
+        '''Called when a RegressionTask has failed.'''
```
ekouts, line 474:
```suggestion
        '''Called when a RegressionTask has aborted.'''
```
Maybe you meant to write aborted here?

ekouts (changes requested):
Overall it looks good, if you fix the very small comment I made I will approve.

ekouts (approved):

teojgo (approved):

PR #2864, opened 2023-04-18
[feat] From performance logger to result logger

### LINKED ISSUE #2853: Logging not performance oriented tests in a rich fashion
The logging mechanism of 4.x introduced new mechanics for performance tests where a test log is produced at the completion of each test, with large improvement within the  handlers .

A similar behaviour should be achieved by not performance oriented tests while using, for instance, the `httpjson` and `filelog` handlers.

In order to understand the actual outcome, one should also think about taking into account when a test fails the sanity stage. In this circumstance, the handlers should provide a detailed log, independently of the nature of the test. 

This addition would allow a more concise log management and greater integration with remote end data storage services.

Lewih:
Closes #2853.

As discussed also on Slack, performance loggers should log a test record for any kind of test. 
This PR has the purpose to discuss and develop this feature.

**Checklist:**
- [x] Extend the output of perflog_handlers (e.g. `httpjson` and `filelog`) to tests not performing the performance stage.
- [ ] Check that the log record content is correct for every kind of test.
- [ ] Rename "performance loggers" to "result loggers" (or alternative name).  
- [ ] Update Config Schema
- [ ] Update the tests.
- [ ] Update the documentation.

The first commit should tackle the first point.


vkarak:
With addition or results storage in 4.7 we now store and can retrieve all the test information. Although currently focused on performance tests, we can easily extend it to sanity-only checks.

PR #2876, opened 2023-05-02
[feat] Allow setting nested mapping types using the `-S` option

### LINKED ISSUE #2868: Feature request: set multi-level dictionaries from the command line
I'd like to set `extra_resources` from the command line, but it's currently not possible because it's a dictionary of dictionaries, and there's currently no syntax for that.

I should have some time to work on this, but I'd need some directions, in terms of where the code is located (I haven't tried to search it yet) and discuss what the syntax for this would look like.

giordano:
The advantage is that we can automatically convert nested mappings, like for example the `extra_resources` attribute, which can thus be set from the command line.

This is the kind of change I'd like to propose to fix #2868.  If this is deemed acceptable, I'll add some tests.

### DIFF
--- a/reframe/utility/typecheck.py
+++ b/reframe/utility/typecheck.py
@@ -95,6 +95,7 @@ of :class:`List[int]`.
 
 import abc
 import datetime
+import json
 import re
 
 
@@ -322,19 +323,25 @@ class _MappingType(_BuiltinType):
         mappping_type = cls._type
         key_type = cls._key_type
         value_type = cls._value_type
-        seq = []
-        for key_datum in s.split(','):
-            try:
-                k, v = key_datum.split(':')
-            except ValueError:
-                # Re-raise as TypeError
-                raise TypeError(
-                    f'cannot convert string {s!r} to {cls.__name__!r}'
-                ) from None
 
-            seq.append((key_type(k), value_type(v)))
+        try:
+            d = json.loads(s)
+        except json.JSONDecodeError:
+            seq = []
+            for key_datum in s.split(','):
+                try:
+                    k, v = key_datum.split(':')
+                except ValueError:
+                    # Re-raise as TypeError
+                    raise TypeError(
+                        f'cannot convert string {s!r} to {cls.__name__!r}'
+                    ) from None
 
-        return mappping_type(seq)
+                seq.append((key_type(k), value_type(v)))
+
+            return mappping_type(seq)
+        else:
+            return mappping_type(d)
 
 
 class _StrType(_SequenceType):
--- a/reframe/utility/typecheck.py
+++ b/reframe/utility/typecheck.py
@@ -325,9 +325,9 @@ class _MappingType(_BuiltinType):
         value_type = cls._value_type
 
         try:
-            d = json.loads(s)
+            items = json.loads(s)
         except json.JSONDecodeError:
-            seq = []
+            items = []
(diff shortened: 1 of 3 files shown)


reframe/utility/typecheck.py
```diff
@@ -322,19 +323,15 @@ def __rfm_cast_str__(cls, s):
...
-
-        return mappping_type(seq)
+        # Convert string to JSON format: wrap it with braces and add double
+        # quotes where necessary.
+        s = f'{{{s}}}'
+        s = re.sub('([:,{] *)([^":,{}]+)', r'\1"\2"', s)
+        # Parse the JSON string.
+        d = json.loads(s)
+        # Enforce the desired output types.
+        d = {key_type(key): value_type(value) for key, value in d.items()}
```
giordano, line 333: I added this to force conversion to the desired types and make existing tests pass, but this now breaks parsing of `extra_resources`, I get again `type error: failed to set field 'extra_resources': could not convert to any of the supported types: ['Dict[str,Dict[str,object]]']` :smile:

giordano: Uhm, something like
```suggestion
        if not issubclass(value_type, object):
            d = {key_type(key): value_type(value) for key, value in d.items()}
```
should do the trick.  Does it make sense?

vkarak:
What would be the syntax for setting a nested dict from the command line? Could you give an example?

giordano:
Something like
```
-S extra_resources=a:{b:c, d:{e:f}},g:h
```

reframe/utility/typecheck.py
vkarak: I don't think we should do any manipulation of the syntax here. What if we first tried to convert from json and if that failed, fall back to the current conversion? So, for a normal dict, those two syntaxes would be equivalent:

1. `-S env_vars=OMP_NUM_THREADS:4`
2. `-S env_vars='{"OMP_NUM_THREADS": "4"}'`

For the nested dicts one should write always valid json: `-S extra_resources='{"gpu":{"num_gpus_per_node":8}}'`. 

We should do the same for the other aggregate types, too.

unittests/test_typecheck.py
```diff
@@ -208,7 +208,7 @@ def test_mapping_type():
     # Test conversions
     assert typ.Dict[str, int]('a:1,b:2') == {'a': 1, 'b': 2}
 
-    with pytest.raises(TypeError):
+    with pytest.raises(json.decoder.JSONDecodeError):
```
vkarak, line 211: These conversion should always raise a `TypeError`.

reframe/utility/typecheck.py
giordano: I was just trying to support current brace- and quotes-less syntax, but if you prefer valid JSON syntax and fallback on current syntax, that'd be good for me.

giordano: Ok, I changed this PR to first try to parse with json and then fall back to the old parsing strategy unchanged, if `json.loads` fails.  I also added a test.

giordano:
Bump :slightly_smiling_face:

vkarak (changes requested):
Lgtm; I've done a couple of style fixes. We'd need also to update the docs of the `-S` option.

vkarak (approved):
I updated the docs; it's fine now.

PR #2878, opened 2023-05-08
[doc] List the `httpjson` and `filelog` handlers as perflog handlers only

ekouts:


vkarak (commented):
Could you also check the same for the `filelog` handler? I don't think that it will work with the normal `handlers`.

vkarak (changes requested):

vkarak (approved):

vkarak:
@jenkins-cscs retry all

PR #2882, opened 2023-05-22
[feat] Combine perflogs for parameterised tests

### LINKED ISSUE #2591: RFE: new parameter for perflog file name
This is a follow-up of https://github.com/reframe-hpc/reframe/issues/2577 and the comment https://github.com/reframe-hpc/reframe/pull/2578#issuecomment-1231358999

The request is to have a way to control the filename of the `filelog` output file, in order to have a human-readable filename like `ib_write_bw_loopback_mlx5_6.log` (where `mlx5_6` is a test parameter) when listing files, instead of having `ib_write_bw_loopback_<hash>.log`

In addition to having a way to control the filename, I suppose we also need a way in the ReFrame settings file to stringify the test parameters in order to get a filename similar to what we had in ReFrame 3.x.

Thanks!

vkarak:
This PR provides the following:

1. The performance logs for paramterised tests using the filelog handler are combined in a single file that is named after the test's base name (i.e. the class name). This is very useful as a parameterised test is essentially the same test with differing parameters.
2. A new special log format specifier `%(check_#ALL)s` is added. If this is found in the log format, it will be replaced by all the loggable `%(check_<attr>)s`.
3. A new `ignore_keys` configuration parameter is added to the filelog log handler that allows users to ignore certain test attributes during logging. The builtin configuration already puts several attributes in that list in order to save space.
4. The default delimiter for the filelog output is reverted to `|` as the `,` may conflict with the logged `env_vars` and `extra_resources`. Users using the `%(check_#ALL)s` specifier should pay attention to the delimiter used in their format.

## Todos

- [x] Update docs
- [x] Update the configuration schema

Closes #2591.


reframe/core/logging.py
```diff
@@ -147,9 +170,21 @@ def __init__(self, prefix, mode='a', encoding=None, fmt=None, perffmt=None):
         self.__fmt = fmt
         self.__perffmt = perffmt
         self.__attr_patt = re.compile(r'\%\((.*?)\)s(.)?')
+        self.__ignore_keys = set(ignore_keys) if ignore_keys else set()
 
     def __generate_header(self, record):
         # Generate the header from the record and fmt
+
+        # Expand the for the special check_#ALL specifier
```
ekouts, line 178:
```suggestion
        # Expand the special check_#ALL specifier
```

ekouts (changes requested):
Lgtm, just some small comments.

ekouts (approved):

PR #2887, opened 2023-05-25
[feat] Support `prepare_cmds` at the environment level

### LINKED ISSUE #2884: Add option for `prepare_cmds` at the programming environment level
The commands would be emitted after the `prepare_cmds` that correspond to the system partition

teojgo:
Closes #2884


docs/config_reference.rst
```diff
@@ -783,6 +783,16 @@ They are associated with `system partitions <#system-partition-configuration>`__
    .. versionadded:: 3.9.1
 
 
+.. py:attribute:: environments.prepare_cmds
+
+   :required: No
+   :default: ``[]``
+
+   List of shell commands to be emitted after the :attr:`~config.systems.partitions.prepare_cmds` but before any module loading commands(if applicable) and environment variables are set.
```
vkarak, line 791: I would not correlate this with the partition semantically. Whatever is defined in the environment should be scoped to it. So better rephrase it as:

```suggestion
   List of shell commands to be emitted before any commands that load the environment.
```

unittests/test_environments.py
```diff
@@ -248,6 +259,33 @@ def test_emit_loadenv_commands(base_environ, user_runtime,
...
+        *env0.prepare_cmds,
+        ms.emit_load_commands('testmod_foo')[0],
+        'export _var0=val1',
+        'export _var2=$_var0',
+        'export _var3=${_var1}',
+    ]
+    assert expected_commands == rt.emit_loadenv_commands(env0)
+
+
+def test_emit_loadenv_nomod_commands(base_environ, user_runtime, env0):
```
vkarak, line 271: Do we need this test? I think it's the same as the `test_emit_loadenv_commands`, which btw is defined twice. Probably a copy-paste error?

unittests/test_schedulers.py
vkarak, line 1: Adding the `prepare_cmds` in the environments in the `test_schedulers.py` here does not offer anything. The scheduler are not involved. Environments are stand alone.

vkarak (changes requested):

unittests/test_environments.py
teojgo: I needed it so that I test when no modules system is defined. I added it to be able to test locally.

unittests/test_environments.py
```diff
@@ -256,6 +268,19 @@ def test_emit_loadenv_commands(base_environ, user_runtime,
     assert expected_commands == rt.emit_loadenv_commands(env0)
 
 
+def test_emit_loadenv_nomod_commands(base_environ, user_runtime, env0):
+    if test_util.has_sane_modules_system():
+        pytest.skip('test scenario not valid for a modules system')
```
vkarak, line 273: I don't think you need to skip the test as it's still valid when there is no modules system.

vkarak (changes requested):

vkarak (approved):

PR #2894, opened 2023-05-31
[feat] Support for `pdsh` and `clush` parallel launchers

brandonbiggs:
[pdsh](https://linux.die.net/man/1/pdsh) and [clush](https://clustershell.readthedocs.io/en/latest/tools/clush.html) are parallel ssh tools. I found the regular ssh launcher to be a bit limiting when trying to run tests on multiple nodes. These launchers add support for additional parallel tools to simplify basic ssh tests on many nodes.


docs/config_reference.rst
```diff
@@ -398,6 +398,7 @@ System Partition Configuration
    Available values are the following:
 
    - ``alps``: Parallel programs will be launched using the `Cray ALPS <https://pubs.cray.com/content/S-2393/CLE%205.2.UP03/cle-xc-system-administration-guide-s-2393-5203-xc/the-aprun-client>`__ ``aprun`` command.
+   - ``clush``: Parallel programs will be launched using the `clush` command. This launcher uses the partition's :attr:`~config.systems.partitions.access` property in order to determine the options to be passed to `clush`.
```
vkarak, line 401: Perhaps we should also add a link to the `clush` main page.

vkarak (changes requested):
We should also expand the unit tests in `test_launchers.py` by extending the fixture [here](https://github.com/reframe-hpc/reframe/blob/f49b0ecae2b475bac441ccdd9b275d13b7ad5816/unittests/test_launchers.py#L13-L16) and adapt the remaining unit tests.

I could have a look into that.

docs/config_reference.rst
brandonbiggs: I'll add that. Any other thoughts on things you'd like added?

vkarak:
I've added unit tests and moved the definitions of the new launchers in the same file with the `ssh` launcher, which I renamed to `rsh.py` for "remote shell launchers". With the minor doc fix above, I think it's ready to go.

vkarak:
Lgtm!

PR #2913, opened 2023-06-06
[feat] Parameterise test from command line

### LINKED ISSUE #2665: Allow the parameterisation of a test on an existing variable from the command line
It's a common pattern if you have a test with a variable that can take multiple values to be able to parametrise the test on that variable. Currently, this cannot be done from the command line and we have to write some boilerplate code. For example,

```python
class MyTest(...):
    x = variable(int, value=0)

class MyTestWithX(MyTest):
    xx = parameter(range(10))

    @run_after('init')
    def set_x(self):
        self.x = self.xx
```

Ideally, users should be able to do `--parameterise=x:1,2,3,4`

vkarak:
This PR introduces parameterisation from the command line on existing variables. Passing `-P num_tasks=1,2,4,8` will create a parameterised version of a test with different `num_tasks` values. This feature cannot parameterise fixtures.

Closes #2665.

### Todos

- [x] Update the docs

### DIFF
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -539,6 +539,18 @@ Options controlling ReFrame execution
    .. versionchanged:: 4.1
       Options that can be specified multiple times are now combined between execution modes and the command line.
 
+.. option:: -P, --parameterize=[TEST.]VAR=VAL0,VAL1,...
+
+   Parameterize a test on an existing variable.
+
+   This option will create a new test with a parameter named ``$VAR`` with the values given in the comma-separated list ``VAL0,VAL1,...``.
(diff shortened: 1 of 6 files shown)


ekouts (approved):
Looks good overall. Should we mention somewhere in the docs that this parameterisation doesn't work for fixtures?

vkarak:
> Looks good overall. Should we mention somewhere in the docs that this parameterisation doesn't work for fixtures?

It's not exactly that it does not work with fixtures; it parameterises the leaf test, which itself can have fixtures. It's that you can't parameterise a nested fixture, such as `-P TEST.FIXT.var=1,2,3,4`. I will add a small note in the docs.

PR #2919, opened 2023-06-10
[feat] Additional Spack build system options

jack-morrison:
## 1. Add support for user-defined `spack env create` options

ReFrame, unless targeting a user-specified existing Spack environment, automatically passes the `-d rfm_spack_env` option to create an ephemeral environment. This is a small addition that allows users to define extra `spack env create` options ([link to Spack docs for list of options](https://spack.readthedocs.io/en/latest/command_index.html#spack-env-create)). 

With the changes proposed here, a user could define the following:
```python
self.build_system.env_create_opts = ["--without-view"]
```

which produces following in the build job script:
```sh
spack env create -d rfm_spack_env --without-view
```

While technically this allows for any `spack env create` options, my motivation for this feature is use of `--without-view`. ReFrame's use of temporary Spack environments leaves little use for a filesystem view, and disabling it can reduce the possibility of `spack load`-time spec collisions.

## 2. Add support for inserting commands between Spack environment creation and environment installation

Under certain circumstances it may be necessary to run complex custom commands to prepare a Spack environment. This needs to happen post-creation, but pre-installation. This capability is added to the Spack build system via `build_system.preinstall_cmds`.

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -863,7 +863,7 @@ class Spack(BuildSystem):
     #: :default: :obj:`True`
     emit_load_cmds = variable(bool, value=True)
 
-    #: Options to pass to ``spack install``
+    #: Options to pass to ``spack install``.
     #:
     #: :type: :class:`List[str]`
     #: :default: ``[]``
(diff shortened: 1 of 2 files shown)


victorusu (approved):
`lgtm`

vkarak (approved):

PR #2923, opened 2023-06-13
[feat] Added the ability to set a container launcher for Singularity/Apptainer

brandonbiggs:
Per a discussion on Slack, I was looking for a way to run `singularity exec` or `singularity run` with `mpirun`. It did not seem like this capability existed, so this PR adds the ability to run singularity/apptainer commands with a container launcher.

This can be defined by a simple list:
```
@run_before("run")
def set_container_launcher(self):
  self.container_platform.container_launcher = ['mpirun']
```

I'm not sure if there are additional places where documentation needs to be updated. Please let me know if there are and I'd be happy to update those as well. I can also add this for other container runtimes if that would be beneficial.

### DIFF
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -78,6 +78,15 @@ class ContainerPlatform(abc.ABC):
     #:    This attribute is no more deprecated.
     workdir = fields.TypedField(str, type(None))
 
+    #: Use a separate launcher for your container
+    #:
+    #: This allows for a container to be lauched with a separate tool
+    #: If no launcher is provided, then the default container command is ran.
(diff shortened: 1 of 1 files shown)


teojgo:
@brandonbiggs I don't understand why the normal launcher of the system is not working in this case.

brandonbiggs:
This functionality already exists by either setting the launcher in the partition in the config file or by using `self.job.launcher = getlauncher('mpirun')()`
https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.schedulers.Job.launcher

PR #2925, opened 2023-06-13
[packaging] Fail on errors in `bootstrap.sh` and add a new `--ignore-errors` option

### LINKED ISSUE #2912: Make `./bootstrap.sh` script fail on errors
Currently, the script keeps going even if `pip` fails, which leads to an incomplete reframe installation. We should be default exit on errors. For special cases, we may also introduce a `--ignore-errors` option to support the current behaviour.

vkarak:
Closes #2912.


vkarak:
> Tested with ubuntu works fine (ubuntu does not have ensurepip for the system python)

Yes, I noticed that when the Tmod4 tests failed, so I had to fix my initial implementation :-D

PR #2927, opened 2023-06-14
[bugfix] Properly cleanup resources of dependencies when their dependents are filtered

### LINKED ISSUE #2916: Cleanup is not called for fixtures that are referenced also in test that are not run
This can be reproduced with the `fixtures_simple.py` example. The fixture setup is the following there:

```bash
./bin/reframe -c unittests/resources/checks_unlisted/fixtures_simple.py -l
```

```console
- TestB /cc291487
    ^HelloTest ~generic 'ff /98821229
    ^HelloFixture ~generic 'f /93cd42f0
- TestA /f41565ab
    ^HelloFixture ~generic 'f /93cd42f0
- HelloTest /2b3e4546
```

The `HelloFixture` is used by `TestA` and `TestB`. If we execute all the tests, the `HelloFixture`'s stage directory is cleaned up as well:

```bash
ls stage/generic/default/builtin/
# empty
```

However, if we only select `TestA` to run, then the `HelloFixture`'s cleanup is not called and the stage directory is not deleted:

```bash
./bin/reframe -c unittests/resources/checks_unlisted/fixtures_simple.py -n ^TestA -r
ls stage/generic/default/builtin/
# HelloFixture_93cd42f0/
```

The reason for that is that the reference count for the fixtures or, generally, the dependencies, is not updated after we filter the tests. As a result, when `TestA` finishes the reference count is 1, so it is not cleaned up. But since we don't execute `TestB`, the resources of `HelloFixture` are leaked.

vkarak:
The fix was straightforward; we need to re-calculate the the in-degree of the nodes of the pruned graph, which is the testcase graph after filtering.

Closes #2916.

### DIFF
--- a/reframe/frontend/dependencies.py
+++ b/reframe/frontend/dependencies.py
@@ -205,6 +205,14 @@ def prune_deps(graph, testcases, max_depth=None):
                 if adj not in pruned_graph:
                     unvisited.append(adj)
 
+    # Re-calculate the in-degree of the pruned graph nodes
+    for u in pruned_graph:
+        u.in_degree = 0
+
+    for u, adjacent in pruned_graph.items():
+        for v in adjacent:
+            v.in_degree += 1
+
     return pruned_graph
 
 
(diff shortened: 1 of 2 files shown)


ekouts (approved):

victorusu (approved):
`lgtm`

PR #2931, opened 2023-06-19
[feat] Update default `--timestamp` format

### LINKED ISSUE #2867: Update the default `--timestamp` value so as not to use `:`
The paths containing colons are problematic for CMake and Bash's autocompletion does not like them either. It seems that we can still have an [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) compliant default timestamp without dashes and colons: `20230422T151055` instead of the current `2023-04-22T15:10:55`

vkarak:
The new format is `%y%m%dT%H%M%S%z` which translates to `230620T005733+0200`.

Closes #2867.

### DIFF
--- a/reframe/core/runtime.py
+++ b/reframe/core/runtime.py
@@ -9,7 +9,7 @@
 
 import os
 import functools
-from datetime import datetime
+import time
 
 import reframe.core.config as config
 import reframe.utility.osext as osext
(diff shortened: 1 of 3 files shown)


teojgo (changes requested):
Probably the change should be made also in `test_timestamp_option_default` which is the one that failing in the ci.

teojgo (approved):

PR #2945, opened 2023-07-03
[bugfix] Update OAR scheduler backend to accept the `Job_Id:` syntax

### LINKED ISSUE #2944: When using OAR, depending on the configuration of the `oarstat` output format, ReFrame might mark tests as finished too early
After submission on an OAR system, ReFrame automatically marked all running jobs as "complete" and therefore continued with the pipeline before the jobs were actually finished, failing `sanity` and `performance` in the process.

The issue was tracked down to `reframe/core/schedulers/oar.py::141`  where the regex returned no match. The analysis of the output showed that `oarstat` returned the job id as `id:XXXXXX` instead of `Job_Id:XXXXXXX`.

Digging further, there appears to be a configuration variable in OAR, `$OAR_OARSTAT_OUTPUT_FORMAT` which defines which type of output format should be used.
If unset, oarstat defaults to a value of 1 which results in the output with `id:XXXXXX`. The same happens with a value of 3.
If set to 2 however, oarstat returns an output containing `Job_Id:XXXXX` instead, which is what is expected by ReFrame.

After discussions on Slack (https://reframetalk.slack.com/archives/CF7GS1SAU/p1688392891640309), we decided to adapt `oar.py` to allow for both output formats. 

A PR will follow.

UeliDeSchwert:
The RegEx now supports `id: XXX` as well as `Job_Id: XXX`. 

Additionally, the link to the oarstat source is updated.

Fixes #2944

### DIFF
--- a/reframe/core/schedulers/oar.py
+++ b/reframe/core/schedulers/oar.py
@@ -135,8 +135,12 @@ class OarJobScheduler(PbsJobScheduler):
             jobinfo = {}
 
             # Typical oarstat -fj <job_id> output:
+            #
             # https://github.com/oar-team/oar/blob/0fccc4fc3bb86ee935ce58effc5aec514a3e155d/sources/core/qfunctions/oarstat#L310
-            # Update 2023-07: oarstat now supports multiple types of output, once containing `id: XXX` and once containing `Job_Id: XXX`
+            #
+            # Update 2023-07: oarstat now supports multiple types of output,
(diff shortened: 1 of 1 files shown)


vkarak (approved):
Lgtm

vkarak:
Since this is a bugfix, it's better to re-open this PR against `master` so that it makes it to the next patch level release, otherwise it will not be released until 4.4.

UeliDeSchwert:
Superseded by #2946 (PR to master instead of develop)

PR #2956, opened 2023-07-21
[testlib] Add ssh_audit test to hpctestlist

victorusu:
The test downloads the ssh-audit tool from GitHub and checks if there is any configuration that is considered as a failure.

The test is compatible with all the job schedulers. Because it only checks the localhost, where the test script runs. So, it becomes the responsibility of ReFrame to spawn the test script in the desired systems.

### DIFF
--- a/hpctestlib/system/ssh/ssh_audit.py
+++ b/hpctestlib/system/ssh/ssh_audit.py
@@ -1,0 +1,33 @@
+# Copyright 2016-2023 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import reframe as rfm
+import reframe.utility as util
+import reframe.utility.sanity as sn
+
+
+@rfm.simple_test
+class ssh_audit_check(rfm.RunOnlyRegressionTest):
+    '''ssh audit config test.
+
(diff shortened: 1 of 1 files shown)


hpctestlib/system/ssh/ssh_audit.py
```diff
@@ -0,0 +1,32 @@
...
+    `ssh-audit is a tool for ssh server & client configuration auditing.
+
+    The check consist on performing the basic ssh server config auditing
+    using the master version of https://github.com/jtesta/ssh-audit.
+    '''
+
+    executable = './ssh-audit.py'
+    executable_opts = ['-n', '-l', 'fail', 'localhost']
+    sourcesdir = 'https://github.com/jtesta/ssh-audit'
+    tags = {'system', 'ssh'}
```
vkarak, line 23: Should we add a security-related tag? Maybe just `security` or something more specific.

vkarak (approved):
Lgtm, just a minor comment.

vkarak:
Also, have you checked the docs? Do these tests' docs appear correctly in the lib docs?

hpctestlib/system/ssh/ssh_audit.py
victorusu: I don't know. I don't want to give the impression that this is a real security check. what do you think?

vkarak (approved):

PR #2957, opened 2023-07-21
[testlib] New test for testing SSH host keys

victorusu:
This test checks the ssh host keys age. The list of host keys and the maximum age allowed are parameters and variables of the test.

If the key file is not found, the test is skipped

### DIFF
--- a/hpctestlib/system/ssh/host_keys.py
+++ b/hpctestlib/system/ssh/host_keys.py
@@ -5,8 +5,6 @@
 
 import time
 
-from datetime import timedelta
-
 import reframe as rfm
 import reframe.utility.sanity as sn
 import reframe.utility.typecheck as typ
--- a/hpctestlib/system/ssh/host_keys.py
+++ b/hpctestlib/system/ssh/host_keys.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2022 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# Copyright 2016-2023 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
 # ReFrame Project Developers. See the top-level LICENSE file for details.
 #
 # SPDX-License-Identifier: BSD-3-Clause
@@ -11,23 +11,23 @@ import reframe.utility.typecheck as typ
 
 
 @rfm.simple_test
-class host_keys_check(rfm.RunOnlyRegressionTest):
-    '''host keys age check
+class ssh_host_keys_check(rfm.RunOnlyRegressionTest):
+    '''SSH host keys age check
 
-    The host keys should be renewed regularly.
+    The ssh host keys should be renewed regularly.
     In this case, we are checking against the
     max_key_age variable
     '''
 
     #: Parameter list with all host keys to check
     #:
-    #: The test skips if a key is not found
+    #: The test is skipped if a key is not found
     #:
     #: :type: :class:`str`
     #: :values: ``['/etc/ssh/ssh_host_rsa_key',
     #:              '/etc/ssh/ssh_host_ecdsa_key',
     #:              '/etc/ssh/ssh_host_ed25519_key']``
-    host_keys = parameter([
+    ssh_host_keys = parameter([
         '/etc/ssh/ssh_host_rsa_key',
         '/etc/ssh/ssh_host_ecdsa_key',
         '/etc/ssh/ssh_host_ed25519_key',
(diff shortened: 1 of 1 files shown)


hpctestlib/system/ssh/host_keys.py
```diff
@@ -0,0 +1,64 @@
+# Copyright 2016-2022 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
```
vkarak, line 1:
```suggestion
# Copyright 2016-2023 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
```

hpctestlib/system/ssh/host_keys.py
```diff
@@ -0,0 +1,64 @@
...
+
+import time
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+
+
+@rfm.simple_test
+class host_keys_check(rfm.RunOnlyRegressionTest):
```
vkarak, line 14:
```suggestion
class ssh_host_keys_check(rfm.RunOnlyRegressionTest):
```

hpctestlib/system/ssh/host_keys.py
```diff
@@ -0,0 +1,64 @@
...
+    '''host keys age check
```
vkarak, line 15:
```suggestion
    '''SSH host keys age check
```

hpctestlib/system/ssh/host_keys.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @sanity_function
+    def assert_file_age(self):
+        current_time = time.time()
+
+        skip_me = sn.extractall('No such file or directory', self.stderr)
+        self.skip_if(skip_me, msg=f'Skipping test because {self.host_keys}'
+                                  f' was not found')
+
+        return sn.all([
```
vkarak, line 58: I don't think you need the `sn.all` here.

hpctestlib/system/ssh/host_keys.py
```diff
@@ -0,0 +1,64 @@
...
+    '''host keys age check
+
+    The host keys should be renewed regularly.
+    In this case, we are checking against the
+    max_key_age variable
+    '''
+
+    #: Parameter list with all host keys to check
+    #:
+    #: The test skips if a key is not found
```
vkarak, line 24:
```suggestion
    #: The test is skipped if a key is not found
```

hpctestlib/system/ssh/host_keys.py
```diff
@@ -0,0 +1,64 @@
...
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+
+
+@rfm.simple_test
+class host_keys_check(rfm.RunOnlyRegressionTest):
+    '''host keys age check
+
+    The host keys should be renewed regularly.
```
vkarak, line 17: I would skip this sentence, as this is a policy. I would simply state what the test does: checks if the list of SSH keys has been updated recently.

hpctestlib/system/ssh/host_keys.py
```diff
@@ -0,0 +1,64 @@
...
+
+    #: Parameter list with all host keys to check
+    #:
+    #: The test skips if a key is not found
+    #:
+    #: :type: :class:`str`
+    #: :values: ``['/etc/ssh/ssh_host_rsa_key',
+    #:              '/etc/ssh/ssh_host_ecdsa_key',
+    #:              '/etc/ssh/ssh_host_ed25519_key']``
+    host_keys = parameter([
```
vkarak, line 30: What's the reason you prefer a parameter here instead of a variable with a default value for the keys? Using a variable would make it easier atm to change the list of keys to check from the command line.

PS: I'm not against the parameters, but trying to understand as both solutions have their merits.

vkarak (changes requested):

hpctestlib/system/ssh/host_keys.py
victorusu: The `sn.all` that you commented on above was a leftover of the test when I used a variable instead of the parameter. I only used a parameter here because the test became a tiny simpler compared to the variable version.
I can have it as a variable too. I don't mind.

vkarak (approved):
Lgtm

vkarak:
You should also add the corresponding docs, but this is also mentioned in #2956, which can include the rest of the docs. I will merge this.

PR #2958, opened 2023-07-23
[testlib] Add fs mount options check to hpctstlib

victorusu:
The filesystem_option_check check if the file systems were properly configured based on their type

### DIFF
--- a/hpctestlib/system/fs/mnt_opts.py
+++ b/hpctestlib/system/fs/mnt_opts.py
@@ -65,6 +65,7 @@ class filesystem_options_check(rfm.RunOnlyRegressionTest):
         for opt in opts_str.split(','):
             if opt == '':
                 continue
+
             opt_parts = opt.split('=', maxsplit=2)
             keystr = opt_parts[0]
             valstr = opt_parts[1] if len(opt_parts) > 1 else ''
--- a/hpctestlib/system/fs/mnt_opts.py
+++ b/hpctestlib/system/fs/mnt_opts.py
@@ -29,10 +29,10 @@ class filesystem_options_check(rfm.RunOnlyRegressionTest):
     #: Parameter pack with listing the mounted file systems, mount points
     #: type and mount options
     #:
-    #: :type: `Tupe[str, str, str]`. The keys should be
-    #:   'mount_poinnt' with the filesystem mount point as a str value
+    #: :type: `Tupe[str, str, str]`. The tuple fields represent
+    #:   'mount_point' with the filesystem mount point as a str value
     #:   'type' with the filesystem type as a str value
-    #:   'options' with the filesystem options as a List[str] value
+    #:   'options' with the filesystem options as a str value
     #:   E.g., '/home', 'xfs', 'rw,nosuid,logbsize=32k'
     #:
     #: :values: Values are generated using the `/proc/mounts` file
--- a/hpctestlib/system/fs/mnt_opts.py
+++ b/hpctestlib/system/fs/mnt_opts.py
@@ -1,0 +1,127 @@
+# Copyright 2016-2022 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import reframe as rfm
+import reframe.utility.sanity as sn
+import reframe.utility.typecheck as typ
+
+
+def get_mounted_fs():
+    result = []
+    with open('/proc/mounts','r') as f:
+        for line in f.readlines():
+            mnt_fs = dict()
+            _, mnt_point, type, options, _, _ = line.split()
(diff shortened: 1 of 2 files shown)


vkarak (changes requested):

hpctestlib/system/fs/mnt_opts.py
```diff
@@ -0,0 +1,127 @@
...
+                   self.ref_opts),
+            sn.map(lambda x: sn.assert_eq(x[1], self.curr_opts[x[0]],
+                                          msg=f'Value {x[1]} not equal to '
+                                              f'{self.curr_opts[x[0]]} in '
+                                              f'the variable {x[0]} for the '
+                                              f'mount point '
+                                              f'{self.fs_mnt_point}, '
+                                              f'of type {self.fs_type}'),
+                   self.ref_opts.items()),
+        ))
```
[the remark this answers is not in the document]
victorusu: I liked the `map` version much more than the for loop I created in the first implementation. So, I kept the `map` because it straight forward. In practice, it is just a single-line implementation. The message is the one that is taking most of the space there. If the message didn't depend on the lambda values, I would have defined it outside of the sanity function. And it would be a single liner...
 
What do you think?

hpctestlib/system/fs/mnt_opts.py
```diff
@@ -0,0 +1,127 @@
...
+                            fmt=lambda x: x[0],
+                            loggable=True)
+
+    #: Reference mount options
+    #:
+    #: :type: `Dict[str, str]`. The key should be the file system type.
+    #:   and the value should be a string with options.
+    #:   E.g., {'xfs: 'nosuid,logbsize=32k'}
+    #: :default: ``{}``
+    fs_ref_opts = variable(typ.Dict, value={}, loggable=True)
```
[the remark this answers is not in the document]
victorusu: Yes! I forgot about it! Thanks!

hpctestlib/system/fs/mnt_opts.py
```diff
@@ -26,102 +20,91 @@ class filesystem_options_check(rfm.RunOnlyRegressionTest):
...
-
     #: Reference mount options
     #:
     #: :type: `Dict[str, str]`. The key should be the file system type.
-    #:   and the value should be a string with options.
+    #:   and the value should be a string with mount options.
     #:   E.g., {'xfs: 'nosuid,logbsize=32k'}
     #: :default: ``{}``
     fs_ref_opts = variable(typ.Dict, value={}, loggable=True)
+    fs_ref_opts = required
```
vkarak, line 30: You don't need this: simply not set the default value above.

hpctestlib/system/fs/mnt_opts.py
```diff
@@ -0,0 +1,110 @@
...
+        with open(stdout, 'r') as fp:
+            for line in fp:
+                _, mnt_point, type, options, _, _ = line.split()
+                self.filesystem.append((mnt_point, type, options))
+
+    @run_before('sanity')
+    def print_test_variables_to_output(self):
+        '''
+        Write the reference mount point options used by the test
+        at the time of execution.
```
vkarak, line 56:
```suggestion
        '''Write the reference mount point options used by the test
        at the time of execution.
```

hpctestlib/system/fs/mnt_opts.py
```diff
@@ -0,0 +1,110 @@
...
+        '''
+        stdout = os.path.join(self.stagedir, sn.evaluate(self.stdout))
+        with open(stdout, 'a') as fp:
+            fp.write('\n---- Reference mount options ----\n')
+            for mnt_type, options in self.fs_ref_opts.items():
+                fp.write(f'{mnt_type} {options}\n')
+
+    def explode_opts_str(self, opts_str):
+        result=dict()
```
vkarak, line 65:
```suggestion
        result = {}
```

hpctestlib/system/fs/mnt_opts.py
```diff
@@ -0,0 +1,110 @@
...
+        for opt in opts_str.split(','):
+            if opt == '':
+                continue
+            opt_parts = opt.split('=', maxsplit=2)
```
vkarak, line 69:
```suggestion
            if opt == '':
                continue

            opt_parts = opt.split('=', maxsplit=2)
```

hpctestlib/system/fs/mnt_opts.py
```diff
@@ -0,0 +1,110 @@
...
+            keystr = opt_parts[0]
+            valstr = opt_parts[1] if len(opt_parts) > 1 else ''
+            result[keystr] = valstr
+        return result
```
vkarak, line 73:
```suggestion
            result[keystr] = valstr
        
        return result
```

vkarak (changes requested):
Some minor style suggestions.

vkarak (approved):

PR #2959, opened 2023-07-23
[testlib] Add trivy scan hpctestlib

victorusu:
The test scans the file systems using aquasecurity/triyv (https://github.com/aquasecurity/trivy)

The test uses the fs option and test against three different scanners vuln, config, and secret.

### DIFF
--- a/hpctestlib/system/fs/trivy_scan.py
+++ b/hpctestlib/system/fs/trivy_scan.py
@@ -14,8 +14,7 @@ from reframe.core.exceptions import ReframeError
 
 @rfm.simple_test
 class trivy_scan_check(rfm.RunOnlyRegressionTest):
-    '''
-    filesystem scanning check
+    '''Filesystem scanning check
 
     Scan the file systems using aquasecurity/triyv
     (https://github.com/aquasecurity/trivy)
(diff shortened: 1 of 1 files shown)


hpctestlib/system/fs/trivy_scan.py
```diff
@@ -0,0 +1,271 @@
...
+                             f'have a corrupt {self.trivy_output_file} file'),
+            sn.assert_eq(self.trivy_json_schema_version,
+                         trivy_output_json[field_name],
+                         msg=f'{field_name} value not supported. Found '
+                             f'{trivy_output_json[field_name]}, expected '
+                             f'{self.trivy_json_schema_version}')
+        )
+
+    @deferrable
+    def config_sanity_check(self, trivy_output_json):
```
vkarak, line 178: I would name all these function as `assert_XXX` and construct the name below accordingly.

hpctestlib/system/fs/trivy_scan.py
```diff
@@ -0,0 +1,271 @@
...
+            '--scanners', f'{self.__scanner_type}'
+        ]
+
+    def load_json_output(self, filename):
+        with open(filename) as fp:
+            try:
+                trivy_output_json = json.loads(fp.read())
+                return trivy_output_json
+            except json.JSONDecodeError as e:
+                return None
```
vkarak, line 161: You should fail the test here by raising a `ValueError`.

vkarak (changes requested):

Issue #64, opened 2018-02-15
Launch ReFrame on a remote machine


vkarak:
It would be very convenient if I could run a similar command on any machine:
```
$ ./bin/reframe -c reframe_tests/ --system=<remote system> -r
```
and then ReFrame would automagically copy the various code into the specified remote system and run the regression there.

Internal issue: https://madra.cscs.ch/scs/reframe/issues/229

vkarak:
Won't do.

PR #2975, opened 2023-08-28
[feat] Add an SSH-based scheduler backend

vkarak:
The new backend is called `ssh` and will spawn a test job on a remote host. More specifically, it will copy the test's stage directory to the remote host, execute the job and copy back the results.

Multiple hosts can be assigned to a partition with an `ssh` scheduler, in which case the scheduler will pick the next free host to spawn a test on.

This PR also introduced a new `run_command_async2` utility for executing asynchronous commands. Conversely to the `run_command_async`, this returns a "future" encapsulating the spawned process. Process futures can be chained with the `then()` method and the chained futures will be called upon termination of the predecessor in the chain.

The SSH scheduler backend uses the process futures and spawns a chain of push artefacts-execute job-pull artefacts commands in its `submit()` method, so that it does not block. The progress is ensured by the scheduler's `poll()` method which polls the spawned futures for completion.

This PR essentially solves the feature request in #64.

More details about the new scheduler can be found in its docs.

### DIFF
--- a/reframe/core/schedulers/ssh.py
+++ b/reframe/core/schedulers/ssh.py
@@ -42,8 +42,8 @@ class _SSHJob(Job):
 
 @register_scheduler('ssh')
 class SSHJobScheduler(JobScheduler):
-    def __init__(self):
-        self._free_hosts = set(self.get_option('ssh_hosts'))
+    def __init__(self, *, hosts=None):
+        self._free_hosts = set(hosts or self.get_option('ssh_hosts'))
         self._allocated_hosts = set()
         if not self._free_hosts:
             raise ConfigError(f'no hosts specified for the SSH scheduler: '
--- a/reframe/core/schedulers/ssh.py
+++ b/reframe/core/schedulers/ssh.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import os
-import functools
 import time
 
 import reframe.utility.osext as osext
--- a/reframe/core/schedulers/ssh.py
+++ b/reframe/core/schedulers/ssh.py
@@ -1,0 +1,219 @@
+# Copyright 2016-2023 Swiss National Supercomputing Centre (CSCS/ETH Zurich)
+# ReFrame Project Developers. See the top-level LICENSE file for details.
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
+import os
+import functools
+import time
+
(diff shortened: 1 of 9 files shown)


reframe/core/schedulers/ssh.py
```diff
@@ -0,0 +1,230 @@
...
+        job.steps['exec'] = osext.run_command_async2(
+            f'ssh -o BatchMode=yes {options} {job.host} '
+            f'"cd {job.remotedir} && bash -l {job.script_filename}"'
+        )
+
+    def submit(self, job):
+        assert isinstance(job, _SSHJob)
+
+        # Check if `#host` pseudo-option is specified and use this as a host,
+        # stripping it off the rest of the options
```
vkarak, line 131: Forgot to document this.

reframe/core/schedulers/ssh.py
```diff
@@ -0,0 +1,230 @@
...
+        host = None
+        stripped_opts = []
+        options = job.sched_access + job.options + job.cli_options
+        for opt in options:
+            if opt.startswith('#host='):
+                _, host = opt.split('=', maxsplit=1)
+            else:
+                stripped_opts.append(opt)
+
+        # Host is pinned externally (`--distribute` option)
```
vkarak, line 141: A note about using `--distribute` with this backend is worth in the docs, too.

reframe/schemas/config.json
```diff
@@ -109,6 +109,10 @@
         "sched_options": {
             "type": "object",
             "properties": {
+                "hosts": {
```
vkarak, line 112: Should I change this to `ssh_hosts`, wdyt?

reframe/core/schedulers/ssh.py
```diff
@@ -0,0 +1,230 @@
...
+    def host(self):
+        return self._host
+
+    @property
+    def ssh_options(self):
+        return self._ssh_options
+
+
+@register_scheduler('ssh')
+class SSHJobScheduler(JobScheduler):
```
vkarak, line 45: I would also need to add some basic unit tests for this scheduler, too.

victorusu (approved):
`lgtm`. I wonder how we will do the unit tests for this scheduler. Do we have any plans?

vkarak:
> lgtm. I wonder how we will do the unit tests for this scheduler. Do we have any plans?

I just added some basic unit tests.

PR #2992, opened 2023-09-10
[feat] Allow pinning a hook as the last of its stage

### LINKED ISSUE #2749: Allow defining pipeline hooks that run after any same-stage subclass hooks
Currently, hooks in the same stage follow the reverse MRO order, i.e., hooks from the parent classes execute first. This is problematic in cases that the base class hooks provide final functionality needed by the subclasses and the subclasses simply need to set some test variables; this is a typical scenario in library tests. Due to this limitation, tests inheriting from a a base class have to be aware of where the base class hooks are placed and define their hooks accordingly. This harms portability. I would like to be able to define a base class test as such:

```python
class base_func(...):
    some_var = variable(str)

    @run_before('run', when='last')
    def prepare_run(self):
        self.executable_opts = ['-f', self.some_var]


@rfm.simple_test
class my_test(base_func):
    @run_before('run')
    def i_cannot_set_this_before(self):
        self.some_var = 'foo'
```

This example currently fails as `some_var` is undefined when the `prepare_run` hook is executed, but I would like to be able specify that this hooks should be the last of all to run in that phase.

vkarak:
This PR introduces a new argument `always_last` to the `@run_before` and `@run_after` decorators that will pin the current hook as the last of its stage. Only one such hook can exist for each stage in a test hierarchy. This feature should make writing test libraries easier, as they can pin their hooks at the end of a stage overriding therefore therefore the default hook execution order.

Closes #2749.

### DIFF
--- a/docs/regression_test_api.rst
+++ b/docs/regression_test_api.rst
@@ -67,9 +67,9 @@ The use of this module is required only when creating new tests programmatically
 
 .. autodecorator:: reframe.core.builtins.require_deps
 
-.. autodecorator:: reframe.core.builtins.run_after(stage)
+.. autodecorator:: reframe.core.builtins.run_after(stage, *, always_last=False)
 
-.. autodecorator:: reframe.core.builtins.run_before(stage)
+.. autodecorator:: reframe.core.builtins.run_before(stage, *, always_last=False)
 
 .. autodecorator:: reframe.core.builtins.sanity_function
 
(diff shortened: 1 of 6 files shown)


victorusu (approved):
`lgtm`
I wonder if we can have dedicated documentation about it.
I don't think it should come in this PR, but a session explaining this feature would be nice.
What do you think?

vkarak:
> I wonder if we can have dedicated documentation about it.
> I don't think it should come in this PR, but a session explaining this feature would be nice.

Yes, it makes sense. I'll have a look in the tutorials if this could be showed in the examples we already have. Otherwise we will have to address this separately.

PR #2993, opened 2023-09-11
[bugfix] Properly retrieve exit status for PBS Pro jobs

### LINKED ISSUE #2936: When using `pbs` scheduler, failures during non-local build jobs aren't fatal
This means that even if the build job fails, reframe moves on to the test job, which is bound to fail.

giordano:
This is a start to address #2936.  It doesn't quite work as desired at the moment, because I don't know what to do exactly with the exit status, how do I use it to make ReFrame error if the job failed?

Closes #2936.

### DIFF
--- a/reframe/core/schedulers/pbs.py
+++ b/reframe/core/schedulers/pbs.py
@@ -178,9 +178,21 @@ class PbsJobScheduler(sched.JobScheduler):
         job._nodelist = [x.split('/')[0] for x in nodespec.split('+')]
         job._nodelist.sort()
 
-    # The second argument is to specialise some code paths to PBS Pro only, but
-    # not Torque.
-    def _poll(self, is_pbs_pro, *jobs):
+    def _query_exit_code(self, job):
+        '''Try to retrieve the exit code of a past job.'''
+
+        # With PBS Pro we can obtain the exit status of a past job
+        extended_info = osext.run_command(f'qstat -xf {job.jobid}')
+        exit_status_match = re.search(
+            r'^ *Exit_status *= *(?P<exit_status>\d+)', extended_info.stdout,
+            flags=re.MULTILINE,
+        )
+        if exit_status_match:
+            return int(exit_status_match.group('exit_status'))
+
+        return None
+
+    def poll(self, *jobs):
         def output_ready(job):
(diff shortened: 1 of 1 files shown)


vkarak:
I would suggest re-opening this as a bugfix against `master`.

giordano:
Changed target branch to `master`.

giordano:
Ok, I think this is ready for review: before it wasn't working because I made a mistake in the regexp pattern (didn't use the `re.MULTILINE` flag and so `^` wasn't matching the beginning of a line within the whole string), after fixing that the build job is failing as expected.  It looks like that setting `job._exitcode` is enough for ReFrame to handle this correctly.

giordano:
Uhm, I think now this fails also when jobs are successful, but I don't understand why :confused:

vkarak:
ok to test

reframe/core/schedulers/pbs.py
```diff
@@ -209,6 +211,19 @@ def output_ready(job):
...
+                    extended_info = osext.run_command(
+                        f'qstat -xf {job.jobid}'
+                    )
+                    exit_status_match = re.search(
+                        r'^ *Exit_status *= *(?P<exit_status>\d+)',
+                        extended_info.stdout,
+                        flags=re.MULTILINE,
+                    )
+                    if exit_status_match:
+                        job._exitcode = int(exit_status_match.group('exit_status'))
```
vkarak, line 226: I would not use the `if` here but rather let inheritance take care of it.

```python
job._exitcode = self._query_exit_code(job.jobid)
```

Where each of the PBS-based backends implements `_query_exit_code()` accordingly. The `torque` backend will simply return `None`.

vkarak (changes requested):

vkarak (approved):

vkarak:
@giordano I updated the PR based on the last comments. Can you give it a try and check if anything is broken so that we can merge it?

teojgo (approved):

giordano:
Sorry, I just managed to run the test with the latest version of `develop` (which includes the latest changes in this PR), and I can confirm that the exit status of the build job is handled correctly in PBS Pro.  Thanks!

PR #2997, opened 2023-09-15
[bugfix] Respect execution order of overriden hooks

### LINKED ISSUE #2800: Do not change the order of execution of a base class hook when overriding it
Consider the following test:

```python
import reframe as rfm


@rfm.simple_test
    valid_systems = ['*']
    valid_prog_environs = ['*']
    x = variable(int)

    @run_before('run')
    def setx(self):
        self.x = 10

    @run_before('run')
    def setup_executable(self):
        self.executable = f'echo {self.x}'

    @sanity_function
    def validate_test(self):
        return sn.assert_found(f'{self.x}', self.stdout)


@rfm.simple_test
class TestB(TestA):
    @run_before('run')
    def setx(self):
        self.x = 20
```

`TestA` runs fine but `TestB` fails with the following error:

```
  * Reason: attribute error: ../misc/rfm-support/slack/fontenaille_hooks/simple.py:19: required variable 'x' has not been set
        self.executable = f'echo {self.x}'
```

The order of execution for `TestA` is the expected one:

```
reframe -c /path/to/test.py -n ^TestA --describe | jq .[].pipeline_hooks
```

```json
{
  "pre_run": [
    "setx",
    "setup_executable"
  ]
}
```

For `TestB` however, the order is inverted:

```json
{
  "pre_run": [
    "setup_executable",
    "setx"
  ]
}
```

Therefore, when `setup_executable()` runs, `x` is undefined. The problem is that pipeline hooks follow the reverse MRO order when they are executed, but this applies also when a hook is overriden. As a result, `TestB.setx` is now part of the derived class and will therefore execute after every hook of the base class.

The desired behaviour would be to replace the base hook but not to move its position in the execution order.

vkarak:
Closes #2800.

### DIFF
--- a/reframe/core/hooks.py
+++ b/reframe/core/hooks.py
@@ -9,13 +9,21 @@ import inspect
 import reframe.utility as util
 
 
+def is_hook(func):
+    return hasattr(func, '_rfm_attach')
+
+
+def is_dep_hook(func):
+    return hasattr(func, '_rfm_resolve_deps')
+
+
 def attach_to(phase):
     '''Backend function to attach a hook to a given phase.
 
     :meta private:
     '''
     def deco(func):
-        if hasattr(func, '_rfm_attach'):
+        if is_hook(func):
             func._rfm_attach.append(phase)
         else:
             func._rfm_attach = [phase]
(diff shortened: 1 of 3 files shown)


victorusu (approved):
`lgtm`

PR #3010, opened 2023-10-02
[feat] Filter tests using arbitrary expressions

### LINKED ISSUE #2847: Filtering tests based on the number of nodes
Today we can filter individual tests based on their name using a regex, but the number of nodes is not always part of the test's name, for instance when `num_nodes` is not parameterized. It would be great to be able to specify a maximum (and possibly minimum) number of nodes to use and have ReFrame automatically ignore tests that do not match the constraint.

I discussed with @vkarak and he pointed out related features in https://github.com/reframe-hpc/reframe/issues/2665 and https://github.com/reframe-hpc/reframe/issues/2674, but those features would still require knowledge of which configurations are supported by each test, and thus might require setting different parameters for each test.

Whereas the request here would be less fine grained but would not require a priori knowledge about the existing tests, for instance it would work with tests using any of those statements:
```python
num_nodes = 1
num_nodes = 16
num_nodes = parameter(range(16))
num_nodes = parameter([10, 40, 80])
num_nodes = parameter([2, 4, 8, 16, 32])
num_nodes = int(os.getenv("TEST_NNODES", "10"))
```

vkarak:
This PR introduces a new option, `-E` or `--filter-expr`, that allows user to filter tests on arbitrary expressions on any test variable or parameter, e.g., `-E 'num_nodes>10'` will select all tests that define set the custom variable `num_nodes` greater than 10.

It also deprecates the `--cpu-only` and `--gpu-only` options as these can be expressed 1-1 with `-E`.

Closes #2847.

### DIFF
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -68,6 +68,21 @@ This happens recursively so that if test ``T1`` depends on ``T2`` and ``T2`` dep
    The value of this attribute is not required to be non-zero for GPU tests.
    Tests may or may not make use of it.
 
+   .. deprecated:: 4.4
+
+      Please use ``-E 'not num_gpus_per_node'`` instead.
+
+.. option:: -E, --filter-expr=EXPR
+
+   Select only tests that satisfy the given expression.
+
+   The expression ``EXPR`` can be any valid Python expression on the test variables or parameters.
+   For example, ``-E num_tasks > 10`` will select all tests, whose :attr:`~reframe.core.pipeline.RegressionTest.num_tasks` exceeds ``10``.
+   You may use any test variable in expression, even user-defined.
+   Multiple variables can also be included such as ``-E num_tasks >= my_param``, where ``my_param`` is user-defined parameter.
+
+   .. versionadded:: 4.4
+
 .. option:: --failed
 
    Select only the failed test cases for a previous run.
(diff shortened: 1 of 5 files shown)


docs/manpage.rst
```diff
@@ -68,6 +68,21 @@ This happens recursively so that if test ``T1`` depends on ``T2`` and ``T2`` dep
    The value of this attribute is not required to be non-zero for GPU tests.
    Tests may or may not make use of it.
 
+   .. deprecated:: 4.4
+
+      Please use ``-E 'not num_gpus_per_node'`` instead.
```
victorusu, line 73: This is beautiful!

unittests/test_filters.py
```diff
@@ -140,3 +141,30 @@ def test_invalid_regex(sample_cases):
 
     with pytest.raises(errors.ReframeError):
         count_checks(filters.have_tag('*foo'), sample_cases).evaluate()
+
+
+def test_validates_expr(sample_cases, sample_param_cases):
```
victorusu, line 146: Does it make sense to add a `not` test case too? What do you think?

vkarak: Not necessarily as we can't check all valid Python expressions and this is essentially indirectly tested through the `have_cpu` filter.

victorusu (approved):
`lgtm`

PR #3032, opened 2023-10-31
Propagate module path option to script creation

felilxtomski:
The `module-path` option is not considered when creating the shell script. This causes issues with schedulers such as SLURM.

### DIFF
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -367,6 +367,13 @@ class ModulesSystem:
         '''The version of this module system.'''
         return self._backend.version()
 
+    @property
+    def emit_extra_module_paths(self):
+        '''Emit extra module paths.'''
+        if hasattr(self._backend, '_extra_module_paths'):
+            return self._backend._emit_extra_module_paths()
+        return []
+
     def unload_all(self):
         '''Unload all loaded modules.'''
         return self._backend.unload_all()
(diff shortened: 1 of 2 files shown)


vkarak:
Hi @felilxtomski could you give more details on the actual problem?

felilxtomski:
I want to use the `--module-path` option to modify (in my case extend) the modulepath which works fine for local execution. However, when a test is executed with a non-local scheduler, the modified `MODULEPATH` environment variable is not propagated to the remote execution (at least in my environment). So if one of the used modules in the generated shell script for the test is not available in the default modulepath, the module cannot be loaded as the modulepath is not extended in the remote shell.

I know that I could also achieve a similar behavior by using prepare_cmds of the config, but I want to modify the modulepath dynamically when executing reframe.

vkarak:
I think passing the `--module-path` back to remote script violates the concept of this option. This, as well as other similar ones, like the `--module`, control reframe's environment and not the test's environment.

You can still enable a dynamic behaviour with the `prepare_cmds` by using custom environment variables. For example, you can define them as such in your config:

```python
'prepare_cmds': [f'export MODULEPATH={os.getenv("MY_MODULE_PATH")}'] if os.getenv("MY_MODULE_PATH") else []
```

PR #3048, opened 2023-11-20
Coverity scan badge

victorusu:
This introduces the Coverity Scan (https://scan.coverity.com) as a badge to ReFrame


vkarak (changes requested):
I think you branched off from `master`, but you should do so from `develop`.

vkarak:
Also the badge shows as "pending".

PR #3054, opened 2023-11-29
[enhancement] Skip modules system sanity checking when module conflict resolution is off

ekouts:
Not sure if this PR is more suitable for `master` or `develop`. I can change it in any case.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -149,6 +149,11 @@ System Configuration
    .. versionadded:: 3.4
       The ``spack`` backend is added.
 
+   .. versionchanged:: 4.5.0
+
+      When the module conflict resolution is off, ReFrame will not check if the module system is available and sane where it is running.
+      It will only emit the module commands in the build and run scripts.
+
 .. py:attribute:: systems.modules
 
    :required: No
@@ -1748,11 +1753,6 @@ General Configuration
 
 
    .. versionadded:: 3.6.0
-
-   .. versionchanged:: 4.5.0
-
-      When the module conflict resolution is off, ReFrame will not check if the module system is available and sane where it is running.
-      It will only emit the module commands in the build and run scripts.
 
 
 .. py:attribute:: general.save_log_files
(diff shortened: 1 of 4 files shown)


teojgo (approved):

ekouts:
I made he changes that you proposed and added a note about the change in the docs. About:

> Also what is the behaviour in the following scenario:
>  1. Define a system with a modules system that is not present

You mean with `RFM_RESOLVE_MODULE_CONFLICTS=0`? In theory Reframe should just emit the commands in the script, not crash. I will also look into the other scenarios that you mentioned and add a unittest.

vkarak:
For the docs I would not add this explanation in the environment variable or configuration option, but rather in the system's `modules_system`: https://reframe-hpc.readthedocs.io/en/stable/config_reference.html#config.systems.modules_system, where we should also say that ReFrame will try to verify that the requested modules system is installed unless the `RFM_RESOLVE_MODULE_CONFLICTS` is unset.

As for the unit tests, we'd need a unit test to verify that no exception is thrown whenever we try to initialize a backend with `validate=False`.

As for the failure scenario that I mentioned, here it is:

```bash
RFM_RESOLVE_MODULE_CONFLICTS=0 ./bin/reframe -C config/machines.py --module foo -c unittests/resources/checks/hellocheck.py -r
```

```console
[ReFrame Setup]
  version:           4.5.0-dev.1+b62838cc
  command:           './bin/reframe -C config/machines.py --module foo -c unittests/resources/checks/hellocheck.py -r'
  launched by:       karakasv@tresa.local
  working directory: '/Users/karakasv/Repositories/reframe'
  settings files:    '<builtin>', 'config/machines.py'
  check search path: '/Users/karakasv/Repositories/reframe/unittests/resources/checks/hellocheck.py'
  stage directory:   '/Users/karakasv/Repositories/reframe/stage'
  output directory:  '/Users/karakasv/Repositories/reframe/output'
  log files:         '/Users/karakasv/Repositories/reframe/reframe.log'

WARNING: skipping test 'SkipTest': unsupported
ERROR: run session stopped: file not found error: [Errno 2] No such file or directory: 'modulecmd'
Log file(s) saved in '/Users/karakasv/Repositories/reframe/reframe.log'
```

In the `config/machines.py` I have set `modules_system` to `tmod`. I think we should somehow treat this case and force the validation of the modules system before this point. Maybe, the `RFM_RESOLVE_MODULE_CONFLICTS=0` should delay the validation instead of cancelling it, because the message here is confusing.

vkarak (approved):

PR #3069, opened 2023-12-04
[enhancement] Remove use of deprecated `%Z` directive in `strftime()`

### LINKED ISSUE #2973: Replace use of `%Z` time format as it's now deprecated
Reference: https://docs.python.org/3.8/library/time.html#id2

We only use it here:

https://github.com/reframe-hpc/reframe/blob/e2f59b627611d575795be45d7595a439f59579c7/reframe/frontend/printer.py#L71

vkarak:
Closes #2973.

### DIFF
--- a/reframe/frontend/printer.py
+++ b/reframe/frontend/printer.py
@@ -3,8 +3,7 @@
 #
 # SPDX-License-Identifier: BSD-3-Clause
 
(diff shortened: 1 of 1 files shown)


victorusu (approved):
`lgtm`

PR #3070, opened 2023-12-04
[feat] Issue an error when invoking an invalid mode

### LINKED ISSUE #3019: Passing an invalid mode should issue an error not a warning
Currently, passing an inexistent mode with the `--mode` option, only a warning will be issued and the option will be ignored. This is error prone, as reframe continues execution as if the option wasn't passed.

vkarak:
Closes #3019.


victorusu (approved):
`lgtm`

PR #3078, opened 2023-12-12
[feat] Add new `sched_bind_options` environment configuration option for emitting environment-specific scheduler options

### LINKED ISSUE #2991: Support injection of scheduler options from an environment
The concept that was discussed is the support of the following:

```
environs = [{
	'name': 'gnu',
	'sched_bind_options': ['--uenv-file={env.extras["uenv_file"]}',
						   '--uenv-mount={env.extras["uenv_mount"]']
}]
```

Using the above syntax, the `env` refers to the actual environment and use its extras.

ekouts:
Closes #2991.

I am a bit confused by a comment from @vkarak in the issue:

> Just to be clear, the above syntax is proposed for the partition definition.

Why would these options be defined in the partition?


teojgo (approved):

reframe/core/pipeline.py
```diff
@@ -1691,7 +1691,10 @@ def _create_job(self, job_type, force_local=False, **job_opts):
                           name=f'rfm_{self.short_name}',
                           script_filename=script_name,
                           workdir=self._stagedir,
-                          sched_access=self._current_partition.access,
+                          sched_access=(
+                              list(self._current_partition.access) +
+                              list(self._current_environ.sched_bind_options)
+                          ),
```
vkarak, line 1697: I would rather not mix it with `sched_access`. Instead, I would prepend it to the `sched_options` passed in `job_opts`.

vkarak (changes requested):
We would also need a unit test.

vkarak (approved):

vkarak:
I was about to merge this, but read this and the original issue:

> Why would these options be defined in the partition?

@ekouts Yes, it was proposed to be in the `environs` because it's a partition's property how you should bind the environment and not the environment's. 

I'm removing it from 4.5 as we need to understand what's the best way to implement it.

PR #3081, opened 2023-12-13
[bugfix] Fix crash when `--distribute` option is applied to a run-only test

### LINKED ISSUE #3074: `--distribute` option crashes if `build_locally=0` on a run-only test
The test crashes with as follows:

```console
  File "/home/vkarakasis/.local/reframe/reframe/frontend/executors/__init__.py", line 317, in _safe_call
    return fn(*args, **kwargs)
  File "/home/vkarakasis/.local/reframe/reframe/core/hooks.py", line 107, in _fn
    getattr(obj, h.__name__)()
  File "/home/vkarakasis/.local/reframe/reframe/core/hooks.py", line 38, in _fn
    func(*args, **kwargs)
  File "/home/vkarakasis/.local/reframe/reframe/frontend/testgenerators.py", line 94, in _rfm_pin_build_nodes
    obj.build_job.pin_nodes = pin_nodes
AttributeError: 'NoneType' object has no attribute 'pin_nodes'
```

The problem is that we try to pin the build job based on `build_locally`'s value. We should make sure that either `build_job` is not `None` or that the test is not an instance of a `RunOnlyRegressionTest`.

https://github.com/reframe-hpc/reframe/blob/06f3e8a7c40a2933a453494dc1ab8466ecd062f7/reframe/frontend/testgenerators.py#L93

vkarak:
Closes #3074.

### DIFF
--- a/reframe/frontend/testgenerators.py
+++ b/reframe/frontend/testgenerators.py
@@ -90,7 +90,7 @@ def distribute_tests(testcases, node_map):
 
     def _rfm_pin_build_nodes(obj):
         pin_nodes = getattr(obj, '$nid')
-        if not obj.local and not obj.build_locally:
+        if obj.build_job and not obj.local and not obj.build_locally:
             obj.build_job.pin_nodes = pin_nodes
 
     def _make_dist_test(testcase):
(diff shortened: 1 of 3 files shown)


ekouts (approved):
lgtm

PR #3083, opened 2023-12-17
[enhancement] Allow `always_last` to be defined in multiple hooks and add a note about users' workaround code for enforcing hook execution order

### LINKED ISSUE #3012: Fixed pipeline hook execution order may break existing workaround code
The fix #2997 may break existing code used to workaround this limitation in previous versions. More specifically, the following example will fail to run with reframe >= 4.3.4

```python
class my_base_test(...):
    x = variable(int)

    @run_before('run')
    def use_x(self):
        y = self.x

@rfm.simple_test
class my_test(my_base_test):
    @run_before('run')
    def set_x(self):
        self.x = 1

    # Possible workaround for reframe <= 4.3.3
    @run_before('run')
    def use_x(self):
        super().use_x()
```
 

Affected versions: >= 4.3.4

### How to fix

If you are using reframe >= 4.4, you should pin the base class hook to run last it its stage:

```python
class my_base_test(...):
    x = variable(int)

    @run_before('run', always_last=True)
    def use_x(self):
        y = self.x

@rfm.simple_test
class my_test(my_base_test):
    @run_before('run')
    def set_x(self):
        self.x = 1
```

For reframe 4.3.4 specifically, where the `always_last` argument is not available, you should rewrite the initial workaround as

```python
class my_base_test(...):
    x = variable(int)

    @run_before('run')
    def use_x(self):
        y = self.x

@rfm.simple_test
class my_test(my_base_test):
    @run_before('run')
    def use_x(self):
        self.x = 1
        super().use_x()
```

We should add a note about this in the docs and also review the current docs/tutorials on their hook syntax.

### LINKED ISSUE #3057: Expand `always_last` semantics in hook definitions to allow attaching more hooks at the end of the stage
Currently, only a single hook in a stage can set `always_last=True`. This quite restrictive if you combine mixins and library tests. We should allow multiple hooks to define `always_last` in which case these hooks will be "accumulated" at the end of the stage in reverse MRO order. Here is an example:

```python
class A(...):
    @run_before('run', always_last=True)
    def hook_a(self): pass

    @run_before('run')
    def hook_b(self): pass

class B(A):
    @run_before('run', always_last=True)
    def hook_c(self): pass

    @run_before('run')
    def hook_d(self): pass
```

The execution order of the pre-run hooks in this case should be:

```console
hook_b, hook_d, hook_c, hook_a
```

Currently this is not allowed and if we remove the `always_last` from everywhere, the execution order will be:

```console
hook_a, hook_b, hook_c, hook_d
```

vkarak:
Closes #3057.
Closes #3012.

### DIFF
--- a/docs/regression_test_api.rst
+++ b/docs/regression_test_api.rst
@@ -112,7 +112,7 @@ the post-init hook will execute *right after* the test is initialized.
 The framework will then continue with other activities and it will execute the pre-setup hook *just before* it schedules the test for executing its setup stage.
 
 Pipeline hooks are normally executed in reverse MRO order, i.e., the hooks of the least specialized class will be executed first.
-In the following exampel, :func:`BaseTest.x` will execute before :func:`DerivedTest.y`:
+In the following example, :func:`BaseTest.x` will execute before :func:`DerivedTest.y`:
 
 .. code-block:: python
 
--- a/docs/regression_test_api.rst
+++ b/docs/regression_test_api.rst
@@ -111,10 +111,10 @@ So although a "post-init" and a "pre-setup" hook will both run *after* a test ha
 the post-init hook will execute *right after* the test is initialized.
 The framework will then continue with other activities and it will execute the pre-setup hook *just before* it schedules the test for executing its setup stage.
 
-Pipeline hooks are executed in reverse MRO order, i.e., the hooks of the least specialized class will be executed first.
+Pipeline hooks are normally executed in reverse MRO order, i.e., the hooks of the least specialized class will be executed first.
 In the following example, :func:`BaseTest.x` will execute before :func:`DerivedTest.y`:
 
-.. code:: python
+.. code-block:: python
 
    class BaseTest(rfm.RegressionTest):
        @run_after('setup')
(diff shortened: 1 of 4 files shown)


docs/regression_test_api.rst
```diff
@@ -111,10 +111,10 @@ So although a "post-init" and a "pre-setup" hook will both run *after* a test ha
 the post-init hook will execute *right after* the test is initialized.
 The framework will then continue with other activities and it will execute the pre-setup hook *just before* it schedules the test for executing its setup stage.
 
-Pipeline hooks are executed in reverse MRO order, i.e., the hooks of the least specialized class will be executed first.
-In the following example, :func:`BaseTest.x` will execute before :func:`DerivedTest.y`:
+Pipeline hooks are normally executed in reverse MRO order, i.e., the hooks of the least specialized class will be executed first.
+In the following exampel, :func:`BaseTest.x` will execute before :func:`DerivedTest.y`:
```
ekouts, line 115:
```suggestion
In the following example, :func:`BaseTest.x` will execute before :func:`DerivedTest.y`:
```

victorusu (approved):
`lgtm`

PR #3094, opened 2024-01-12
[bugfix] Do not ignore the `use_login_shell` configuration option when generation the auto-detection script

ekouts:
This is useful for the firecrest scheduler, but probably also for other schedulers.


teojgo (approved):

vkarak:
Could we add a unit test for this?

vkarak (approved):

PR #3096, opened 2024-01-17
[bugfix] Detect missing `curl` in `bootstrap.sh`

tmjbios:
Curl is not a default package on most Linux distributions, and is not listed as a requirement for reFrame.

When curl is missing, bootstrap.sh will blow right past the curl failure and continue to attempt to execute commands.

Add a check for curl, and print a meaningful message + exit if it is not found.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -106,7 +106,7 @@ else
 fi
 
 if ! type "curl" > /dev/null 2>&1; then
-    echo "curl is missing; install curl and try again"
+    echo -e "could not find \`curl': please install curl and try again"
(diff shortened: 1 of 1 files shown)


tmjbios:
Yes. 

To test locally, I used a default Ubuntu 22.04 server, cloned the repo and ran bootstrap, installed curl, the re-ran bootstrap.

vkarak:
@tmjbios Thanks for the fix. I updated your PR to be against master as it's a bug fix and we could release it in the next patch release.

vkarak (approved):

PR #3098, opened 2024-01-25
[feat] Add `nvcc` as an environment configuration parameter

OliverPerks:
This commit allows you to specify the path to nvidia's nvcc a part of a system environment. This allows for multiple environments with different paths to nvcc without the use of modules or the PATH environment. It allows for nvcc being unset in the environment and will default to 'nvcc'.

### DIFF
--- a/reframe/core/environments.py
+++ b/reframe/core/environments.py
@@ -321,7 +321,7 @@ class ProgEnvironment(Environment):
 
     @property
     def nvcc(self):
-        '''The NVIDIA nvcc compiler of this programming environment.
+        '''The NVIDIA CUDA compiler of this programming environment.
 
         :type: :class:`str`
         '''
--- a/reframe/core/environments.py
+++ b/reframe/core/environments.py
@@ -219,6 +219,7 @@ class ProgEnvironment(Environment):
     _cc = fields.TypedField(str)
     _cxx = fields.TypedField(str)
     _ftn = fields.TypedField(str)
+    _nvcc = fields.TypedField(str)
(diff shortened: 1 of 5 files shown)


reframe/core/environments.py
```diff
@@ -320,4 +321,8 @@ def ldflags(self):
 
     @property
     def nvcc(self):
+        '''The NVIDIA nvcc compiler of this programming environment.
```
vkarak, line 324:
```suggestion
        '''The NVIDIA CUDA compiler of this programming environment.
```

vkarak (approved):
Thanks for the PR! Lgtm. Just a minor comment.

reframe/core/environments.py
vkarak: We could also need to update the configuration reference docs.

reframe/core/systems.py
```diff
@@ -538,6 +538,7 @@ def create(cls, site_config):
                     cc=site_config.get(f'environments/@{e}/cc'),
                     cxx=site_config.get(f'environments/@{e}/cxx'),
                     ftn=site_config.get(f'environments/@{e}/ftn'),
+                    nvcc=site_config.get(f'environments/@{e}/nvcc', 'nvcc'),
```
vkarak, line 541: You should not add a default here, but it instead should come from the schema.

vkarak (changes requested):

vkarak (approved):
I updated the PR with the suggested changes, lgtm now.

PR #3101, opened 2024-01-30
[feat] Adding support for CMake + Ninja

OliverPerks:
Previously the CMake config would just run 'make'. 
This commit runs 'cmake --build .' which enables you to specify different build systems within cmake. Specifically this is useful for Ninja, using the 'cmake -GNinja' flag. If no build system is set it will fall back to make.

### DIFF
--- a/reframe/core/buildsystems.py
+++ b/reframe/core/buildsystems.py
@@ -584,7 +584,7 @@ class CMake(ConfigureBasedBuildSystem):
         else:
             cmake_cmd += ['.']
 
-        make_cmd = ['make -j']
+        make_cmd = ['cmake --build . -j']
         if self.max_concurrency is not None:
             make_cmd += [str(self.max_concurrency)]
 


vkarak:
I think the unit tests should be updated (I expect them to fail). Another question does this change requires a specific/recent version of CMake that would otherwise make reframe's backend break for users with older version? We could add a build system variable to control whether to emit `make` or `cmake --build` in this case.

teojgo:
This needs a little bit more thought. In particular, since we now assume that the build system is Make, all the options are passed to make verbatim. So if you pass the 'install' as an option the command will be `make install`. With the above syntax it would have to be changed to `cmake --build . -t install -j <n>` or `cmake --build . -j -- install`.

victorusu:
What is the status of this? Is anyone working to make the `cmake` options compatible with this new syntax?

PR #3107, opened 2024-02-04
[feat] Add CPU vendor and model name to topology files

jack-morrison:
By having the CPU vendor identifier available, test authors can use "GeniuneIntel", "AuthenticAMD", etc. as a logic condition where high-level characterization is acceptable. There are cases where this may actually be more useful than the particular CPU model name.

Fixes #https://github.com/reframe-hpc/reframe/issues/2995

### DIFF
--- a/reframe/utility/cpuinfo.py
+++ b/reframe/utility/cpuinfo.py
@@ -283,8 +283,8 @@ def cpuinfo():
     ret = {
         'arch': archspec.cpu.host().name,
         'vendor': archspec.cpu.host().vendor,
-        'model':
-            archspec.cpu.detect.raw_info_dictionary().get("model name", "N/A")
+        'model': archspec.cpu.detect.raw_info_dictionary().get('model name',
+                                                               'N/A')
     }
 
     # Try first to get information from the filesystem
--- a/reframe/utility/cpuinfo.py
+++ b/reframe/utility/cpuinfo.py
@@ -281,7 +281,10 @@ def _sysctl_topo():
 
 def cpuinfo():
     ret = {
-        'arch': archspec.cpu.host().name
+        'arch': archspec.cpu.host().name,
+        'vendor': archspec.cpu.host().vendor,
+        'model':
+            archspec.cpu.detect.raw_info_dictionary().get("model name", "N/A")
     }
 
     # Try first to get information from the filesystem
--- a/reframe/schemas/config.json
+++ b/reframe/schemas/config.json
@@ -218,6 +218,8 @@
             "type": "object",
             "properties": {
                 "arch": {"type": "string"},
+                "vendor": {"type": "string"},
+                "model": {"type": "string"},
                 "num_cpus": {"type": "number"},
                 "num_cpus_per_core": {"type": "number"},
                 "num_cpus_per_socket": {"type": "number"},


jack-morrison:
Got some tests to fix up here, it seems.

teojgo:
@jack-morrison since you are mentioning the model name of the CPU, do you think it could be added too?

jack-morrison:
> @jack-morrison since you are mentioning the model name of the CPU, do you think it could be added too?

 I think that's already available in the topology files as `arch` ([relevant archspec docs](https://archspec.readthedocs.io/en/latest/apidocs.html#archspec.cpu.Microarchitecture)), on the line immediately above this change.

vkarak:
@jack-morrison The CI fails because you need to update the configuration schema as well:

https://github.com/reframe-hpc/reframe/blob/f2a23261e7ae2927bc145cdc8f3d6a30f566f8b6/reframe/schemas/config.json#L219-L226

teojgo:
> > @jack-morrison since you are mentioning the model name of the CPU, do you think it could be added too?
> 
> I think that's already available in the topology files as `arch` ([relevant archspec docs](https://archspec.readthedocs.io/en/latest/apidocs.html#archspec.cpu.Microarchitecture)), on the line immediately above this change.
> 
> Edit: I guess it's the microarchitecture name that's already presented as `arch`. This what I was intending to refer to. I'm happy to add any (or all) other attributes from `archspec.cpu` though if that's desirable.

Yeah, it's indeed the microarchitecture, the model name I can retrieve using: `archspec.cpu.detect.raw_info_dictionary()['model name']`.

reframe/utility/cpuinfo.py
```diff
@@ -281,7 +281,9 @@ def _sysctl_topo():
 
 def cpuinfo():
     ret = {
-        'arch': archspec.cpu.host().name
+        'arch': archspec.cpu.host().name,
+        'vendor': archspec.cpu.host().vendor,
+        'model': archspec.cpu.detect.raw_info_dictionary()['model name']
```
teojgo, line 286: Maybe retrive the model name with `.get` and add a default value of Unknown for example here? What do you think @vkarak

vkarak: Yes, because otherwise it will crash if for some reason the `model name` key is not there.

jack-morrison: Makes sense, will do!

jack-morrison: addressed via [0c07e8c](https://github.com/reframe-hpc/reframe/pull/3107/commits/0c07e8c31b7447a2bfe190466b25bbb22f0a6c6d)

vkarak (approved):

reframe/utility/cpuinfo.py
jack-morrison: Good call @teojgo 😄 --
https://github.com/archspec/archspec/pull/149 looks like it'll probably break `archspec.cpu.detect.raw_info_dictionary()`. 

@vkarak / @teojgo, how can I best help update this line when RFM bumps the pinned version of archspec?

vkarak: The unit tests run the cpu auto-detection code, so I hope that this will be caught from the CI when the version is bumped. I could tag you in that case, if the fix is not obvious.

PR #3113, opened 2024-02-09
[feat] Enable multiple inheritance of test variables

vkarak:
Currently, multiple inheritance of test or mixin variables is prohibited leads to a test load error. This is restrictive, especially in case of mixins, which may be inherited multiple times in a complex setup. For example, the following would cause an error when reframe is processing the `Z` class:

```python
    class Mixin(rfm.RegressionMixin):
        x = variable(typ.List[int], value=[])

    class X(Mixin):
        x = [3, 4]

    class Y(Mixin):
        x = [10, 1]

    class Z(X, Y):
        pass
```

This PR introduces a new variable argument, `combine`, that defines how the default values of the parent variables can be combined in case of multiple inheritance. Since `combine=None` by default, the default behaviour of variables does not change. It just that we now offer the extra flexibility to users to define the variable behaviour in case of multiple inheritance.

In the example above, if we define `x` as follows:

```python
    class Mixin(rfm.RegressionMixin):
        x = variable(typ.List[int], value=[],
                     combine=lambda x, y: list(map(max, zip(x, y))))

    class X(Mixin):
        x = [3, 4]

    class Y(Mixin):
        x = [10, 1]

    class Z(X, Y):
        pass
```

its value in `Z` class will be `[10, 4]`.

### Todos
- [ ] Write docs

### DIFF
--- a/reframe/core/variables.py
+++ b/reframe/core/variables.py
@@ -202,7 +202,7 @@ class TestVar:
     # to denote the "private" fields.
 
     __slots__ = ('_p_default_value', '_p_field',
-                 '_loggable', '_name', '_target', '_refs')
+                 '_loggable', '_name', '_target', '_refs', '_combine_fn')
 
     __mutable_props = ('_default_value',)
 
@@ -224,6 +224,7 @@ class TestVar:
         else:
             self._p_default_value = kwargs.pop('value', Undefined)
 
+        self._combine_fn = kwargs.pop('combine', None)
         self._loggable = kwargs.pop('loggable', True)
         if not issubclass(field_type, fields.Field):
             raise TypeError(
@@ -248,6 +249,7 @@ class TestVar:
                                               kind, from_version)
         ret._p_default_value = var._default_value
         ret._loggable = var._loggable
+        ret._combine_fn = var._combine_fn
         ret._target = var._target
         ret._refs = var._refs
         if var.is_alias():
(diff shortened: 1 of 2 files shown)


reframe/core/variables.py
```diff
@@ -38,13 +38,16 @@ class TestVar:
...
-    might be caused by accidentally overriding a class attribute. See the
-    example below.
+    declared through the :func:`variable` built-in has a number of advantages:
+
+    1. Variables are type checked, so attempts to assign a value of a wrong
+       type will cause the test to fail.
+    2. You can set the values of variables from the command line using the
+       :option:`-S` option.
+    3. You can avoid variable redefinitions.
+    4. You can control whether a varaible can be inherited multiple times.
```
teojgo, line 48:
```suggestion
    4. You can control whether a variable can be inherited multiple times.
```

reframe/core/variables.py
```diff
@@ -180,7 +183,22 @@ class MyRequiredTest(HelloTest):
         variable will become a log record attribute under the name
         ``check_NAME``, where ``NAME`` is the name of the variable (default
         :obj:`True`).
-    :param `kwargs`: keyword arguments to be forwarded to the constructor of
+    :param merge_func: Enable multiple inheritance for this variable by
+        defining a merge strategy of their default values
```
teojgo, line 187:
```suggestion
        defining a merge strategy of their default values.
```

teojgo (changes requested):

victorusu (approved):
`lgtm`

teojgo (approved):

PR #3115, opened 2024-02-12
[feat] Enable default value type checking and implicit conversions at the variable declaration level

vkarak:
Currently, the default values of variables or the values assigned to them in the test class body are not type checked and no implicit conversions are applied to the value being assigned. This is deferred at the class instantiation time where the default values of variables are injected to the newly created test object and type checked through the associated descriptor.

This PR moves the type checking and implicit conversions, i.e., the functionality of the descriptors at the class level. We do so, with two tricks:

1. We extend the Python descriptor model by having `__set__()` return the value that the descriptor would write into the object if the object passed to `__set__()` is `None`.
2. We anticipate the call to the descriptor's `__set_name__()` at the same time that this function is called for the variable. And it is at this point that we also call the descriptor's `__set__()` method to force the type checking and any type conversion. For type conversions we do respect the `allow_implicit` argument of variables.

The advantages of this method are

- Variables before and after the test's instantiation behave the same way.
- Type errors during variable assignments are now more precise as the source file information is included in the error message.

Todos

- [x] Remove unnecessary conversions and type checking (type checking in `inject()` should not be needed)
- [x] Double-check the `setvar()` impl. for redundant operations.

### DIFF
--- a/docs/regression_test_api.rst
+++ b/docs/regression_test_api.rst
@@ -41,6 +41,13 @@ These are called *builtins* because they are directly available for use inside t
 However, almost all of these builtins are also available from the :obj:`reframe.core.builtins` module.
 The use of this module is required only when creating new tests programmatically using the :func:`~reframe.core.meta.make_test` function.
 
+.. versionchanged:: 3.7.0
+   Expose :func:`@deferrable <reframe.core.builtins.deferrable>` as a builtin.
+
+.. versionchanged:: 3.11.0
+   Builtins are now available also through the :obj:`reframe.core.builtins` module.
+
+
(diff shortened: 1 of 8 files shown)


vkarak:
The test instantiation time increases by about 5% and the overall time of the following benchmark increases by 3.5%:

```python
reframe -c unittests/resources/checks/hellocheck.py -n ^HelloTest --repeat=1000 -l
```

I believe the difference comes from the fact that we now validate every assignment of a variable, whereas before we were validating once during test instantiation.

victorusu (approved):
`lgtm`

PR #3150, opened 2024-04-02
[feat] Make processor model available through `ProcessorInfo`

vkarak:
This is a follow up of #3107 and defines the `model` attribute in `ProcessorInfo`.


ekouts (changes requested):
I would also add a short entry in `docs/config_reference.rst` under "Processor Info"

vkarak:
> I would also add a short entry in `docs/config_reference.rst` under "Processor Info"

I've added also corresponding entry for the devices which was also missing.

ekouts (approved):
lgtm

PR #3151, opened 2024-04-03
[feat] Include hardware platform info in `ProcessorInfo`

### LINKED ISSUE #3125: Collect some of the `platform` info in the auto-detected processor info
We are particularly interested in `platform.machine()` and maybe also `uname()`.

vkarak:
Closes #3125

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1923,6 +1923,15 @@ A *processor info object* in ReFrame's configuration is used to hold information
 
    .. versionadded:: 4.6
 
+.. attribute:: systems.partitions.processor.platform
+
+   :required: No
+   :default: ``None``
+
+   The OS platform for this processor (e.g., ``x86_64``, ``arm64`` etc.)
+
+   .. versionadded:: 4.6
+
 .. attribute:: systems.partitions.processor.num_cpus
 
    :required: No
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1928,7 +1928,7 @@ A *processor info object* in ReFrame's configuration is used to hold information
    :required: No
    :default: ``None``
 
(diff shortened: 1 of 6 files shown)


docs/config_reference.rst
```diff
@@ -1923,6 +1923,15 @@ A *processor info object* in ReFrame's configuration is used to hold information
 
    .. versionadded:: 4.6
 
+.. attribute:: systems.partitions.processor.platform
+
+   :required: No
+   :default: ``None``
+
+   The OS platform for this processor (e.g., ``x86_64``, ``arm64`` etc.)
```
jack-morrison, line 1931: Not sure this is really OS related. The Python docs just call this the ["machine type."](https://docs.python.org/3/library/platform.html#platform.machine)

vkarak: I could update the docs, but do you think we should also change the name of the actual option?

jack-morrison: > do you think we should also change the name of the actual option?

Now that you mention it, what do you think about `instruction_set`?

I know I'm nit picking here... Because there's a `platform.platform()` and a `platform.machine()`, I don't like the idea of crossing them. Maybe the name `machine` is better? It would at least map "machine" to `platform.machine()`.

vkarak: I would prefer to go with something that is used more generally for this string. For example, Docker calls this "platform", Linux's `uname` calls it "machine" but also in the docs (for MacOS) describes as "hardware platform". That's why I have used the term "platform" in first place, but I'm open also for "machine" or "machine_type".

jack-morrison: OK good points 😄 , I'm good with "platform", but would still recommend dropping OS from the docs.

ekouts (approved):
lgtm

teojgo (approved):

PR #3156, opened 2024-04-11
[feat] Allow ReFrame to pass the access options in command line instead of the script for Slurm

### LINKED ISSUE #2970: Allow users to pass extra options to the job submission command in the partition config
This is requested by https://github.com/EESSI/test-suite/issues/68 as well as #2968. #2968 introduces a solution that is LSF-specific (in terms of naming), but the idea is to generalise it to the rest of job schedulers.

@boegel @casparvl Would that work in your case?

ekouts:
~Right now I am only changing this for Slurm. I leave these options in the slurm script, mostly to remind whoever might be checking the script that the options are not ignored. Do you think I should remove them from the script?~

Closes #2970 .


reframe/schemas/config.json
```diff
@@ -625,6 +626,7 @@
         "systems/partitions/time_limit": null,
         "systems/partitions/devices": [],
         "systems/partitions/extras": {},
+        "systems*/sched_options/access_on_submission_command": false,
```
vkarak, line 629: I would suggest a shorter name:

```suggestion
        "systems*/sched_options/sched_access_in_submit": false,
```

vkarak (changes requested):
I would suggest to (a) extend this to all the schedulers and (b) do not emit the options in the job script.

Also I don't know how would be able to test this fully, but at least we need unit tests that set this option and verify that the `sched_access` options are not passed in the script if this configuration option is set.

reframe/core/schedulers/lsf.py
```diff
@@ -76,7 +80,11 @@ def emit_preamble(self, job):
 
     def submit(self, job):
         with open(job.script_filename, 'r') as fp:
-            completed = _run_strict('bsub', stdin=fp)
+            cmd_opts = (
+                ' '.join(job.sched_access) if self._sched_access_in_submit
+                else ''
+            )
+            completed = _run_strict('bsub {cmd_opts}', stdin=fp)
```
vkarak, line 87:
```suggestion
            cmd_parts = ['bsub']
            if self._sched_access_in_submit:
                cmd_parts += job.sched_access
            
            completed = _run_strict(' '.join(cmd_parts), stdin=fp)
```

reframe/core/schedulers/oar.py
```diff
@@ -101,9 +107,12 @@ def emit_preamble(self, job):
...
         # OAR batch submission mode needs full path to the job script
         job_script_fullpath = os.path.join(job.workdir, job.script_filename)
-
+        cmd_opts = (
+            ' '.join(job.sched_access) if self._sched_access_in_submit
+            else ''
+        )
         # OAR needs -S to submit job in batch mode
-        cmd = f'oarsub -S {job_script_fullpath}'
+        cmd = f'oarsub {cmd_opts} -S {job_script_fullpath}'
```
vkarak, line 115: I'd do the same thing as in the LSF.

reframe/core/schedulers/pbs.py
```diff
@@ -139,9 +147,14 @@ def filternodes(self, job, nodes):
...
     def submit(self, job):
+        cmd_opts = (
+            ' '.join(job.sched_access) if self._sched_access_in_submit
+            else ''
+        )
         # `-o` and `-e` options are only recognized in command line by the PBS
         # Slurm wrappers.
-        cmd = f'qsub -o {job.stdout} -e {job.stderr} {job.script_filename}'
+        cmd = (f'qsub {cmd_opts} -o {job.stdout} -e {job.stderr} '
+               f'{job.script_filename}')
```
vkarak, line 157: Same pattern with the `cmd_parts`.

reframe/core/schedulers/slurm.py
```diff
@@ -209,21 +212,22 @@ def emit_preamble(self, job):
...
+                    preamble.append('%s %s' % (self._prefix, opt))
+
+            # To avoid overriding a constraint that's passed into
+            # `sched_access`, we AND it with the `--constraint` option
+            # passed either in `options` or in `cli_options`
+            parsed_options, _ = constraint_parser.parse_known_args(
+                job.sched_access
+            )
+            if parsed_options.constraint:
+                constraints.append(parsed_options.constraint.strip())
```
vkarak, line 230: I think that we should do the same thing with constraints in both cases.

reframe/core/schedulers/slurm.py
```diff
@@ -259,7 +263,11 @@ def emit_preamble(self, job):
         return list(filter(None, preamble))
 
     def submit(self, job):
-        cmd = f'sbatch {job.script_filename}'
+        cmd_opts = (
+            ' '.join(job.sched_access) if self._sched_access_in_submit
+            else ''
+        )
+        cmd = f'sbatch {cmd_opts} {job.script_filename}'
```
vkarak, line 270: I'd use the `cmd_parts` pattern again.

vkarak (changes requested):

reframe/core/schedulers/slurm.py
ekouts: But in the case the `-C c1` is passed in with `sbatch`and then they are joined in the script `-C c1&c2` the option in the script will probably be ignored completely, right?
So are you suggesting that we join them in the submission command as well? Then the behaviour would be that we only add the constraint options from the test in the job, while passing the rest in the script?

reframe/core/schedulers/oar.py
vkarak: I could do it myself and try to push it for the release today.

reframe/core/schedulers/slurm.py
vkarak: Currently, `sched_access` is combined with the last constraint passed as an option, either from the script or from the command-line. The rationale behind that was that a constraint passed either in the test or in the command line would completely override the ReFrame partition definition if we didn't do that. But put aside this rationale, passing the `sched_access` from the command line should be completely equivalent to passing them through the script, so this logic must be preserved.

vkarak: I think additional constraints should be combined with `sched_access` and passed from the command line in this case, because this is the equivalent of what we are doing now.

vkarak: The logic becomes a bit complicated to properly support this, so I suggest postponing this to 4.7 to properly test. Maybe we should change the logic and stop treating constraints specially and pass the responsibility of combining the constraints to the test.

docs/manpage.rst
```diff
@@ -1214,6 +1214,21 @@ Here is an alphabetical list of the environment variables recognized by ReFrame.
...
+
+   .. table::
+      :align: left
+
+      ================================== ==================
+      Associated command line option     N/A
+      Associated configuration parameter :attr::attr:`~config.systems.partitions.sched_options.sched_access_in_submit`
+      ================================== ==================
+
+.. versionadded:: 4.6.0
```
victorusu, line 1229: I think we need to update this number.

reframe/core/schedulers/slurm.py
vkarak: I compacted the constraint code in the Slurm backend and change its logic to fit also the case where `sched_access_in_submit=True`. What I'm practically doing is to "move out" the constraint options from either the `job.options` or the `job.cli_options` to the `sched_access` and then we can continue as with the rest of the schedulers.

vkarak (approved):

PR #3158, opened 2024-04-15
[bugfix] Treat OR Slurm constraints in flexible node allocation

vkarak:
This PR fixes also the way `sched_access` constraints are combined with those specified in the CLI or in the `job.options` by parenthesizing appropriately. It also updates the docs.


teojgo (approved):

vkarak:
I would still need to add a unit test to check the behaviour with an invalid expression.

PR #3173, opened 2024-04-26
[feat] Skip by default flexible tests when not enough nodes found

### LINKED ISSUE #3164: Allow for skipping a flexible test if not enough nodes are available instead of failing
I would probably be also in favour of changing the default behaviour to be skipping the test and add a new command-line option to enforce the strict behaviour.

vkarak:
Flexible tests are now by default skipped when not enough nodes are found. We introduce a new option, `--flex-alloc-strict` that will cause the test to fail instead (the current default).

Closes #3164.

### DIFF
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -752,6 +752,18 @@ If no node can be selected, the test will be marked as a failure with an appropr
 
       Slurm OR constraints and parenthesized expressions are supported in flexible node allocation.
 
+   .. versionchanged:: 4.7
+      The test is not failed if not enough nodes are available, but it is skipped instead.
+      To enforce a failure, use :option:`--flex-alloc-strict`
+
+.. option:: --flex-alloc-strict
+
+   Fail flexible tests if their minimum task requirement is not satisfied.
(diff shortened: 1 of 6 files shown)


docs/manpage.rst
```diff
@@ -752,6 +752,18 @@ If no node can be selected, the test will be marked as a failure with an appropr
 
       Slurm OR constraints and parenthesized expressions are supported in flexible node allocation.
 
+   .. versionchanged:: 4.7
+      The test is not failed if not enough nodes are available, but it is skipped instead.
```
teojgo, line 756:
```suggestion
      The test does not fail if not enough nodes are available, but it is skipped instead.
```

teojgo (changes requested):

docs/manpage.rst
vkarak: In this case it is the framework that "fails the test" (marks it as a failure) and not the test that fails.

vkarak: I will change it to "the test is not marked as a failure [...]" to make it clearer.

teojgo (approved):

PR #3185, opened 2024-05-06
[feat] Update `archspec` and topology file call for CPU model name

jack-morrison:
This follows https://github.com/reframe-hpc/reframe/pull/3107 and https://github.com/reframe-hpc/reframe/pull/3132 to adopt the new function from [archspec v0.2.4](https://github.com/archspec/archspec/releases/tag/v0.2.4) returning a CPU brand string.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ packages = find_namespace:
 python_requires = >=3.6
 scripts = bin/reframe
 install_requires =
-    archspec <= 0.2.4
+    archspec >= 0.2.4
     argcomplete
     argcomplete <= 3.1.2; python_version < '3.8'
     jsonschema
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ packages = find_namespace:
 python_requires = >=3.6
 scripts = bin/reframe
 install_requires =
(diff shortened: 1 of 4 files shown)


victorusu:
test this patch

teojgo (changes requested):
The `setup.cfg` should be updated as well so that the wheel validation test passes. and the correct pypi package is produced.

jack-morrison:
> The `setup.cfg` should be updated as well so that the wheel validation test passes. and the correct pypi package is produced.

Done -- thanks for the pointer @teojgo 😃

setup.cfg
```diff
@@ -26,7 +26,7 @@ packages = find_namespace:
 python_requires = >=3.6
 scripts = bin/reframe
 install_requires =
-    archspec <= 0.2.2
+    archspec <= 0.2.4
```
vkarak, line 29:
```suggestion
    archspec >= 0.2.4
```

vkarak (changes requested):

setup.cfg
jack-morrison: oops

vkarak (approved):

PR #3189, opened 2024-05-13
[feat] Add a perflog entry also when sanity fails

### LINKED ISSUE #3186: If a performance test fails in sanity or earlier no perflog entry is created
Performance logging happens after the test finishes since version 4.0 and if a test is a performance test. We need to understand the exact conditions when this happens and what triggers this behaviour.

This is also related to #2853.

vkarak:
The idea is the same as exposed in the issue.
In case of sanity failures, we execute also the performance stage but in dry-run mode suppressing any non-aborting exception. To achieve so, we also extended the dry-run mode inside the performance stage to also generate the `perfvalues` which are required by the logging handlers for emitting the performance record.

In case of sanity failures, the performance values will be logged as `None`.

### Todos

- [x] Add a `check_fail_reason` log format placeholder for allowing the fail reason to be also logged.

Closes #3186.

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -19,7 +19,7 @@ import urllib
 import reframe.utility.color as color
 import reframe.utility.jsonext as jsonext
 import reframe.utility.osext as osext
-from reframe.core.exceptions import ConfigError, LoggingError
+from reframe.core.exceptions import ConfigError, LoggingError, what
 from reframe.core.warnings import suppress_deprecations
 from reframe.utility import is_trivially_callable
 from reframe.utility.profile import TimeProfiler
@@ -855,6 +855,8 @@ class LoggerAdapter(logging.LoggerAdapter):
         self.extra['check_partition'] = task.testcase.partition.name
         self.extra['check_environ'] = task.testcase.environ.name
         self.extra['check_result'] = 'pass' if task.succeeded else 'fail'
+        fail_reason = what(*task.exc_info) if not task.succeeded else None
+        self.extra['check_fail_reason'] = fail_reason
         if msg is None:
             msg = 'sent by ' + self.extra['osuser']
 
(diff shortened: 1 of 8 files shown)


reframe/core/pipeline.py
```diff
@@ -2282,27 +2279,30 @@ def check_performance(self):
...
+            tag = key.split(':')[-1]
+            try:
+                sn.evaluate(
+                    sn.assert_reference(
+                        val, ref, low_thres, high_thres,
+                        msg=('failed to meet reference: %s={0}, '
+                                'expected {1} (l={2}, u={3})' % tag))
+                )
+            except SanityError as e:
+                raise PerformanceError(e) from None
```
vkarak, line 2305: This is simply indented out, as it does not need to be executed from the stage dir.

reframe/core/logging.py
```diff
@@ -128,14 +128,6 @@ def ignore_brokenpipe(hdlr, l):
...
 
 
-def _expand_params(check):
-    cls = type(check)
-    return {
-        name: getattr(check, name) for name, param in cls.param_space.items
-        if param.is_loggable()
-    }
-
-
```
vkarak, line 138: I removed this since it was unused.

unittests/test_policies.py
```diff
@@ -1400,3 +1400,28 @@ def test_perf_logging_param_test(make_runner, make_exec_ctx, perf_param_tests,
...
+    make_exec_ctx(config_perflog(
+        fmt='%(check_result)s|%(check_fail_reason)s|%(check_perfvalues)s',
+        perffmt='%(check_perf_value)s|'
+    ))
+    logging.configure_logging(rt.runtime().site_config)
+    runner = make_runner()
+    testcases = executors.generate_testcases([_X()])
+    _assert_no_logging_error(runner.runall, testcases)
+
+    logfile = (tmp_path / 'perflogs' / 'generic' / 'default' / '_X.log')
```
teojgo, line 1421: Remove the `()` here

teojgo (changes requested):

PR #3190, opened 2024-05-14
[feat] Index job reports in a local database

ekouts:


reframe/frontend/runreport.py
```diff
@@ -270,3 +265,94 @@ def junit_dump(xml, fp):
...
+    with sqlite3.connect(filename) as conn:
+        conn.execute(
+'''CREATE TABLE IF NOT EXISTS testcases(
+        name TEXT,
+        hash TEXT,
+        system TEXT,
+        partition TEXT,
+        environment TEXT,
+        time_start TEXT,
+        time_end TEXT,
```
victorusu, line 296: Can we make this a real DATETIME field?

reframe/frontend/runreport.py
```diff
@@ -270,3 +265,94 @@ def junit_dump(xml, fp):
...
+        run_index INTEGER,
+        test_index INTEGER,
+        report_file TEXT
```
victorusu, line 299: Can we add more fields like the number of cases and number of failures?

victorusu: I would love to see also the user, version, and time_elapsed fields. But this would require another table. I think we should do two tables. One with the test cases and another one with the runs. I would love to make fast reports using the later. This would have the num_cases and num_failures in the table. What do you think?

victorusu: In the second table ("experiments") we could add the "runs" fields, num_cases, num_failures, num_aborted, and num_skipped. What do you think?

vkarak:
I will close and reopen this PR as I have changed it a lot.

PR #3193, opened 2024-05-15
[bugfix] Constraint the lxml version on aarch64

teojgo:
There are no wheels available on PyPI for `python < 3.8` and `aarch64`


ekouts (approved):

vkarak (changes requested):
I think we should do the same in the `setup.cfg` too.

vkarak (approved):

PR #3195, opened 2024-05-15
[doc] Display Python imports in the first tutorial example

createyourpersonalaccount:
The tutorial was skipping the two imports

    import reframe as rfm

by using the :pyobject: option to only highlight stream_test. This is the first example shown in the tutorial and it is important that the entire file is displayed.

Presumably this technique is used to avoid the license notice; we can use :start-at: instead.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -85,7 +85,7 @@ Here is the full ReFrame test:
 
 .. literalinclude:: ../examples/tutorial/stream/stream_runonly.py
    :caption:
-   :pyobject: stream_test
+   :lines: 5-
 
 ReFrame tests are specially decorated classes that ultimately derive from the :class:`~reframe.core.pipeline.RegressionTest` class.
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -85,7 +85,7 @@ Here is the full ReFrame test:
 
 .. literalinclude:: ../examples/tutorial/stream/stream_runonly.py
    :caption:
-   :pyobject: stream_test
+   :start-at: "import reframe as rfm"
```
vkarak, line 88: That was an overlook; we don't need to use `pyobject` here, but rather simply `lines`.

```suggestion
   :lines: 5-
```

vkarak (changes requested):
Could you also rebase your PR over `master` (including your commit)?

docs/tutorial.rst
createyourpersonalaccount: I went with lines.

vkarak (approved):

PR #3197, opened 2024-05-15
[doc] Use long options in places

createyourpersonalaccount:
Long options are more self-documenting.


vkarak (changes requested):
I agree with this particular suggestion but not in general. In the command examples in the tutorial the short options are used, so if the option is mentioned in the text, it has to match that in the example.

If you rebase this on master, it'll be good to merge. Thanks!

createyourpersonalaccount:
@vkarak done!

vkarak (approved):

PR #3199, opened 2024-05-16
[bugfix] Handle `None` cpu model name returned from `archspec`

### LINKED ISSUE #3192: Processor model json validation fails if `cpu.brand_string` is `None`
`None` is not an accepted string type for the json schema

teojgo:
Fixes #3192


vkarak (changes requested):
~~Could you rebase your PR on master and target master instead, so as to release this in the next patch level release?~~

I thought that we had already bumped archspec in `master`. Sorry!

jack-morrison:
Thanks for this catch @teojgo!

vkarak (approved):

PR #3206, opened 2024-05-27
[bugfix] Do not use CLI defaults for options that are backed by a configuration option

### LINKED ISSUE #3200: Configuration option `purge_environment` not working in ReFrame 4.3.3
We are using in our systems ReFrame 4.3.3, installed with EasyBuild 4.9.1 as a module. We set the `purge_environment` general configuration option to `True` to ensure that any modules loaded when ReFrame is launched are removed during the compilation and run phases. However, we noticed that modules remain loaded. This does not happen if we set the command line option `--purge-env`.

Is this known behavior?

Our relevant configuration options:
```
site_configuration = {                                                                                                                                        
    'general' : [                                                                                                                                             
        {                                                                                                                                                     
            'use_login_shell': True,                                                                                                                          
            'purge_environment': True,                                                                                                                        
        },                                                                                                                                                    
    ],                                                                                                                                                        
}
```

vkarak:
Closes #3200.

### DIFF
--- a/reframe/frontend/cli.py
+++ b/reframe/frontend/cli.py
@@ -285,7 +285,7 @@ def main():
         envvar='RFM_STAGE_DIR', configvar='systems/stagedir'
     )
     output_options.add_argument(
-        '--save-log-files', action='store_true', default=False,
+        '--save-log-files', action='store_true',
         help='Save ReFrame log files to the output directory',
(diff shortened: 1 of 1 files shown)


vkarak:
@gkaf89 This should be fixing the problem you've encountered. Could you try it out?

ekouts (approved):
lgtm

PR #3232, opened 2024-07-11
[testlib] Fix compilation of gpu burn benchmark with hip

### LINKED ISSUE #3231: compilation gpu burn with hip
When I try to compile it with [makefile.hip](https://github.com/reframe-hpc/reframe/blob/develop/hpctestlib/microbenchmarks/gpu/src/gpu_burn/makefile.hip), i.e with the following compilation line `hipcc -O3 gpu_burn.cu -o gpu_burn.x -DTARGET_HIP  -std=c++14 -I/opt/rocm/rocm_smi/include -lnuma -lrocm_smi64 -lrocblas`, I get the following error :
```
gpu_burn.cu:182:24: error: __host__ function 'compute' cannot overload __host__ __device__ function 'compute'
void GemmTest<double>::compute()
                       ^
gpu_burn.cu:164:8: note: previous declaration is here
  void compute() = delete;
       ^
gpu_burn.cu:200:23: error: __host__ function 'compute' cannot overload __host__ __device__ function 'compute'
void GemmTest<float>::compute()
                      ^
gpu_burn.cu:164:8: note: previous declaration is here
  void compute() = delete;
       ^
3 warnings and 2 errors generated when compiling for gfx803.
make: *** [makefile.hip:4: gpu_burn] Error 1
```

after correcting the error I have an error in the link 
```
ok maintenant j'ai cette erreur :
ld.lld: error: undefined symbol: pthread_create
>>> referenced by gpu_burn.cu
>>>               /tmp/gpu_burn-92c6db.o:(void launch<double>(int, std::vector<int, std::allocator<int>> const&))
>>> referenced by gpu_burn.cu
>>>               /tmp/gpu_burn-92c6db.o:(void launch<float>(int, std::vector<int, std::allocator<int>> const&))
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [makefile.hip:4: gpu_burn] Error 1
```
On our AMD machine it would seem that adding pthread to the LDFLAGS is necessary.

CedricJour:
Fixes #3231


hpctestlib/microbenchmarks/gpu/gpu_burn.py
```diff
@@ -52,9 +52,9 @@ def setup_build(self):
 
         if self.gpu_build is None:
             # Try to set the build type from the partition features
-            if 'cuda' in curr_env.features:
+            if 'cuda' in curr_part.features:
                 self.gpu_build = 'cuda'
-            elif 'hip' in curr_env.features:
+            elif 'hip' in curr_part.features:
```
vkarak, line 57: I think here the comment must be updated instead.

vkarak (changes requested):

hpctestlib/microbenchmarks/gpu/gpu_burn.py
CedricJour: I think it mostly depends on how you want to define it in settings.py, 

from settings.py (here it's an AMD GPU)
```
'partitions': [
                {
                    'name': 'gpu',
                    'descr':'GPU nodes',
                    'scheduler': 'squeue',
                    'launcher': 'srun',
                    'access': [],
                    'environs': ['PrgEnv-cray'],
                    'max_jobs':100,
                    'access': ['-C MI250','-t 00:10:00'],
                    'features' : ['hip'],
```

HIP can be used to compile code that will run on an AMD GPU or nvidia GPU, so it's not necessarily hardware (and therefore partition) related. 

So does that mean you have to add feature hip in each environment? perhaps I did that in order to refactor the feature hip line but now that you've pointed it out, I've probably made a mistake in my thinking.

```
    'environments': [
        {
            'name': 'builtin',
            'cc': 'gcc',
            'cxx': 'g++',
            'ftn': 'gfortran',
            'target_systems': ['crusher','adastra']
        },
```

vkarak: Since in the test its use is related to compilation, I would rather use it as an environment feature.

CedricJour: done

vkarak (approved):

hpctestlib/microbenchmarks/gpu/gpu_burn.py
```diff
@@ -52,9 +52,9 @@ def setup_build(self):
 
         if self.gpu_build is None:
             # Try to set the build type from the partition features
-            if 'cuda' in curr_env.features:
+            if 'cuda' in curr_part.features:
```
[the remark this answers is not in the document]
CedricJour: that's make sense

PR #3247, opened 2024-08-30
[bugfix] Ignore records with unset performance when grouping test cases

vkarak:
Bug fix for #3227.

### DIFF
--- a/unittests/resources/checks/frontend_checks.py
+++ b/unittests/resources/checks/frontend_checks.py
@@ -63,9 +63,14 @@ class NoPrgEnvCheck(BaseFrontendCheck):
 
 @rfm.simple_test
 class SanityFailureCheck(BaseFrontendCheck):
+    '''A performance test that fails in sanity.'''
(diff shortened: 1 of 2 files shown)


unittests/resources/checks/frontend_checks.py
```diff
@@ -63,10 +63,15 @@ class NoPrgEnvCheck(BaseFrontendCheck):
...
 @rfm.simple_test
 class SanityFailureCheck(BaseFrontendCheck):
+    '''A performance test that fails in sanity.'''
     @sanity_function
     def validate_output(self):
         return sn.assert_found('foo', self.stdout)
 
+    @performance_function('Gflop/s')
+    def perf(self):
+        return sn.extractsingle(r'perf: (\d+)', self.stdout, 1, int)
```
vkarak, line 73: Converting this test to a performance test that fails in sanity triggers the bug with the current unit tests.

teojgo (approved):

PR #3263, opened 2024-09-20
[enhancement] Improve performance of module backend when detecting module conflicts

brandongc:
Running a typical subset of NERSC's test suite with ~500 tests, there are ~30 unique values passed to this routine and the result of call for a given input is static. `module` commands can take a surprisingly long amount of time to return.

This change reduces the time for

```
reframe -C ./nersc-config.py -c checks --dry-run --mode checkout
```

from ~735s to <600s.

### DIFF
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -1031,7 +1031,7 @@ class LModImpl(TMod4Impl):
 
         return ret
 
-    @functools.lru_cache
+    @functools.lru_cache(maxsize=128)
     def conflicted_modules(self, module):
         if module.collection:
             # Conflicts have no meaning in module collection. The modules
--- a/reframe/core/modules.py
+++ b/reframe/core/modules.py
@@ -8,6 +8,7 @@
 #
 
 import abc
+import functools
 import os
 import re
 from collections import OrderedDict
@@ -1030,6 +1031,7 @@ class LModImpl(TMod4Impl):
 
         return ret
 
+    @functools.lru_cache
     def conflicted_modules(self, module):
         if module.collection:
             # Conflicts have no meaning in module collection. The modules


brandongc:
It looks like some versions of python do not like the syntax.

Also, technically this introduces a potential memory leak as the cache is global and keyed on all of the arguments to the function including `self`. It doesn't look like this class gets instantiated more than once in normal use, but could be an issue.

A less concise but straightforward approach to achieve the same effect:

```
 import abc
 import os
 import re
-import functools
 from collections import OrderedDict

@@ -947,6 +946,7 @@ class LModImpl(TMod4Impl):
                               'this Lmod installation')

         self._extra_module_paths = []
+        self._conflicted_modules_cache = {}

     def name(self):
         return 'lmod'
@@ -978,18 +978,19 @@ class LModImpl(TMod4Impl):

         return ret

-#    @functools.lru_cache(maxsize=None)
     def conflicted_modules(self, module):
-        fp = "/global/u1/c/cookbg/reframe-nersc-tests/conflict_module_args.txt"
-        with open(fp, "a") as f:
-            f.write(str(module) + "\n")
         if module.collection:
             # Conflicts have no meaning in module collection. The modules
             # system will take care of these when restoring a module
             # collection
             return []

-        output = self.execute_with_path('show', str(module), path=module.path)
+        module_name = str(module)
+        if module_name in self._conflicted_modules_cache.keys():
+            output = self._conflicted_modules_cache[module_name]
+        else:
+            output = self.execute_with_path('show', module_name, path=module.path)
+            self._conflicted_modules_cache[module_name] = output

         # Lmod accepts both Lua and and Tcl syntax
         # The following test allows incorrect syntax, e.g., `conflict
```

reframe/core/modules.py
```diff
@@ -1030,6 +1031,7 @@ def available_modules(self, substr):
 
         return ret
 
+    @functools.lru_cache
```
vkarak, line 1034: I think the problem is that the variant of the decorator without arguments was added in Python 3.8. So I suggest to change it to a backward compatible syntax:

```suggestion
    @functools.lru_cache(maxsize=128)
```

vkarak (changes requested):

reframe/core/modules.py
brandongc: I made this change

vkarak (approved):

teojgo (approved):

PR #3268, opened 2024-09-24
[ci] Run scheduler unittests with GitHub Actions

### LINKED ISSUE #3261: Use `docker compose` to run the slurm related unit tests
The idea is to use a similar setup to the tutorial for [multi-node tests](https://reframe-hpc.readthedocs.io/en/stable/tutorial.html#running-the-multi-node-examples) to simulate a slurm virtual cluster and test the slurm related functionality there.

teojgo:
Fixes #3261


.github/pseudo-cluster/docker-compose.yml
vkarak, line 1: Are these copies of those in the tutorial examples or do they have any adaptations? Would symlinks just work for the files that are the same?

.github/pseudo-cluster/reframe/docker-entrypoint.sh
```diff
@@ -0,0 +1,24 @@
...
+cp -r /usr/local/share/reframe .
+cd reframe
+./bootstrap.sh
+
+echo "Running unittests with backend scheduler: ${BACKEND}"
+
+tempdir=$(mktemp -d -p /scratch)
+TMPDIR=$tempdir ./test_reframe.py -v \
+    --rfm-user-config=config/ci-cluster.py \
+    --rfm-user-system=pseudo-cluster:compute-${BACKEND:-squeue}
```
vkarak, line 24: I guess that's different from the examples.

config/ci-cluster.py
vkarak, line 1: I see that we had `cscs-ci` under the `config/` dir, but maybe it's better if we put this config under `ci-scripts/configs/` along with the rest of the configs used in CI.

vkarak (changes requested):
I think that we should also remove the `CSCS CI` now and the related machinery, shouldn't we?

.github/workflows/test-schedulers.yaml
```diff
@@ -0,0 +1,24 @@
+name: Test Schedulers
```
vkarak, line 1: Just to be in sync with the rest of workflow titles.

```suggestion
name: ReFrame CI / Scheduler backend tests
```

.github/pseudo-cluster/docker-compose.yml
teojgo: they have a small adaptation and also another adaptation in the entrypoint. I will make symlinks for the rest

.github/pseudo-cluster/reframe/docker-entrypoint.sh
teojgo: yes this and the dockerfile which actually copies the current source code

vkarak (approved):

PR #3274, opened 2024-10-01
[feat] Allow users define custom log handlers and attach them to the framework

### LINKED ISSUE #3270: Add support for a plugin log handler
I'd like to be able to define a custom log handler in a configuration file. This sort of capability seems similar to the idea of enabling custom launchers described in the [extending-the-framework](https://reframe-hpc.readthedocs.io/en/stable/howto.html#extending-the-framework) section of the docs.

I poked around the code base a bit and hacked together some minimal changes here: https://github.com/dmargala/reframe/tree/add-plugin-log-handler

With the changes on that branch, I can define add my custom log handler by including something like this in my config file:

```python
class PlugInLogHandler(logging.Handler):
    def emit(self, record):
        custom_logger.log(record)

def _create_plugin_handler(site_config, handlers_group):
    return PlugInLogHandler()

from reframe.core.logging import register_create_handler
register_create_handler('plugin', _create_plugin_handler)
```

Does this seem like a reasonable capability to add? Is there maybe a better way to go about implementing this capability?

dmargala:
This PR adds support for defining a custom log handler that can be specified in the reframe configuration file. 

For example, adding a plugin log handler would look something like the following:

```python
class PlugInLogHandler(logging.Handler):
    def emit(self, record):
        custom_logger.log(record)

def _create_plugin_handler(site_config, handlers_group):
    return PlugInLogHandler()

from reframe.core.logging import register_plugin_handler
register_plugin_handler(_create_plugin_handler)
```

There is some prior art in the reframe code base for a similar capability for custom launchers (see [extending-the-framework](https://reframe-hpc.readthedocs.io/en/stable/howto.html#extending-the-framework) section of the docs). There may be some other ways to implement this by extending the existing Stream or File log handlers to accept custom stream or file objects.

closes #3270


reframe/core/logging.py
```diff
@@ -678,39 +678,35 @@ def emit(self, record):
...
+    'filelog': _create_filelog_handler,
+    'syslog': _create_syslog_handler,
+    'stream': _create_stream_handler,
+    'graylog': _create_graylog_handler,
+    'httpjson': _create_httpjson_handler,
+}
+
+
+def register_plugin_handler(create_plugin_handler):
+    _create_handler_registry["plugin"] = create_plugin_handler
```
vkarak, line 692: I would make this a decorator taking the name of the handler to register that will populate the above registry upon definition of each handler. This registry could start empty here and every log handler creation function (even the builtin ones) could be defined as

```python
@register_log_handler('handler_name')
def _create_foo_handler(site_config, handlers_group):
    ...
```

The nice thing with the decorator is that we can also do some validation of the function signature if we want to.

```suggestion
def register_log_handler(name):
    def _deco(fn):
        _create_handler_registry[name] = fn
        return fn
     
     return deco
```

vkarak (changes requested):
Thanks for the PR! I have some suggestions on how to make this idea more in sync with the similar idea of custom launchers and schedulers.

We should also add a small documentation section about it in the "how to".

reframe/core/logging.py
```diff
@@ -678,39 +678,35 @@ def emit(self, record):
...
+            hdlr = create_handler(site_config, f'{handler_prefix}/{i}')
             if hdlr is None:
                 getlogger().warning('could not initialize the '
-                                    'httpjson handler; ignoring ...')
+                                    f'{handler_type} handler; ignoring ...')
                 continue
-        else:
+        except KeyError:
             # Should not enter here
             raise AssertionError(f'unknown handler type: {handler_type}')
```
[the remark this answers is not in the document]
dmargala: Done. I also changed `AssertionError` to `ConfigError` and dropped the "should not enter here" comment since we could enter there if a type is specified in the config that hasn't been registered.

reframe/core/logging.py
```diff
@@ -412,6 +412,19 @@ def stream_handler_kind(handler):
...
 
+# Registry for log handler creation functions
+_create_handlers = {}
+
+
+def register_log_handler(name):
+    '''Register the decorated log handler creation function'''
+    def _create_handler_wrapper(fn):
+        _create_handlers[name] = fn
+        return fn
```
vkarak, line 423: Just annoying code style nitpicking: leave a blank line after the return here.

vkarak (changes requested):
Lgtm, just some style nitpicking.

dmargala:
Thanks. I applied suggestions and also fixed indent spacing in the examples (the spacing in the code blocks had inherited the 3-space indents, I changed it to 4-spaces.

vkarak (approved):
Thanks!

PR #3280, opened 2024-10-04
[bugfix] Let Slurm decide partition in flexible node allocation if no default partition is defined

dmargala:
This PR fixes an issue with flexible node allocations when no partition is specified and there is no default partition in slurm. 

At NERSC, users specify a qos rather than a partition and Slurm chooses a partition for a job based on the combination of several options. A sys admin at our site suggested using `srun --test-only` to see which partition Slurm would chose for a job with a given set of options.

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -308,6 +308,21 @@ class SlurmJobScheduler(sched.JobScheduler):
 
         return None
 
+    def _get_actual_partition(self, options):
+        try:
+            completed = _run_strict(
+                ' '.join(['srun'] + options + ['--test-only', 'true'])
+            )
+            partition_match = re.search(r'partition (?P<partition>\S+)\s+',
+                                        completed.stderr)
+            if partition_match:
+                return partition_match.group('partition')
+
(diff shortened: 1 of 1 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -345,11 +360,13 @@ def filternodes(self, job, nodes):
             nodes &= self._get_reservation_nodes(reservation)
             self.log(f'[F] Filtering nodes by reservation {reservation}: '
                      f'available nodes now: {len(nodes)}')
-
         if partitions:
             partitions = set(partitions.strip().split(','))
         else:
-            default_partition = self._get_default_partition()
+            default_partition = (
```
vkarak, line 366: Maybe add a short comment inside the `else` branch explaining this concept.

vkarak (commented):
Lgtm, except some minor comment.

ekouts (approved):

reframe/core/schedulers/slurm.py
dmargala: Thanks, I've added a comment.

vkarak (approved):

PR #3281, opened 2024-10-07
[feat] New configuration parameter for specifying commands to install reframe remotely for topology detection

### LINKED ISSUE #2292: auto-detection, use existing module if told to
The auto-detection could perhaps be made to use an existing module on the remote system instead of bootstrapping itself.
I would make the process a lot faster.
A config option per site and/or partition would suffice

### LINKED ISSUE #2690: Skip bootstrap step at topology detect
Unfortunately, `RFM_REMOTE_DETECT=TRUE` calls `pip` via `bootstrap.sh`.
I am not able to use `pip` on my computer (usual proxy stuff).

@vkarak suggests to add an option to avoid the bootstrap and simply copy the dependencies over to the remote host.

I'll be glad to contribute but I don't know what to do exactly.

### LINKED ISSUE #2979: Enable running some commands before the bootstrap, similar to `prerun_cmds` in regular tests
The current job generated by the remote CPU detection script is:

```
#!/bin/bash
#SBATCH --job-name="rfm-detect-job"
#SBATCH --ntasks=1
#SBATCH --output=rfm-detect-job.out
#SBATCH --error=rfm-detect-job.err

_onerror()
{
    exitcode=$?
    echo "-reframe: command \`$BASH_COMMAND' failed (exit code: $exitcode)"
    exit $exitcode
}

trap _onerror ERR

./bootstrap.sh
mpirun -np 1 ./bin/reframe --detect-host-topology=topo.json
```
There is no option for the user to change this in any way currently, even though regular reframe tests have to option to define `prerun_cmds` to e.g. make some module environment available.

When I hit the issue in https://github.com/reframe-hpc/reframe/issues/2926 I found that I had quite few options for debugging. E.g. I wanted to provide a different pip through a virtualenv, to see if that was my problem. Something like `prerun_cmds` would enable one to do that. Admittedly, it's not a very concrete use case. Hence, @vkarak suggested to make this separate issue and wait for a more concrete use case before implementing it. 

So, if someone also runs into a situation where having a `prerun_cmds` in the CPU autodetect script would have been useful to them: please let us know here :)

Blanca-Fuentes:
This PR adds the option of specifying custom commands to install reframe in a remote partition for the topology auto-detection.

Previously, a clone of reframe was created running bootstrap or pip in order to launch reframe with auto-detect option. 
Now, a list of custom commands can be specified to install reframe through ```general:remote_install``` to avoid creating a reframe clone with ```pip``` or running ```.\bootstrap.sh```.

The ```reframe --detect-host-topology=topo.json``` is run after the commands for the remote reframe installation.

Closes #2292.
Closes #2690.
Closes #2979 .


reframe/frontend/autodetect.py
```diff
@@ -177,6 +185,26 @@ def _emit_script_for_pip(job, env):
                 _log_contents(job.stdout)
                 _log_contents(job.stderr)
                 topo_info = json.loads(_contents('topo.json'))
+        else:
+            with _copy_reframe(prefix) as (dirname, use_pip):
```
ekouts, line 189: Do we need we a copy of reframe for pip? And does it make sense to make this available also for custom commands?

docs/manpage.rst
```diff
@@ -1941,6 +1941,20 @@ Whenever an environment variable is associated with a configuration option, its
 
    .. versionadded:: 3.7.0
 
+.. envvar:: RFM_REMOTE_COMMAND
```
ekouts, line 1944: Add new envvars in alphabetical order to make it easier to find ;)

ekouts (commented):
For me this is a very good start. It would be good to have some unittests that check the basic functionality f the feature.
Some things that maybe we still need to discuss:
- Should we copy the reframe directory in the custom command case? I think not, as long as we don't have a use case for it.
- Do we give the option to the user to add commands pre/post in the current scripts (bootstrap/pip)?

reframe/frontend/autodetect.py
```diff
@@ -151,24 +151,32 @@ def _emit_script_for_pip(job, env):
...
-                    _emit_script_for_pip(job, [part.local_env])
+                if custom_command:
+                    _emit_custom_script(job, [part.local_env], custom_command)
                 else:
-                    _emit_script_for_source(job, [part.local_env])
-
+                    if use_pip:
+                        _emit_script_for_pip(job, [part.local_env])
+                    else:
+                        _emit_script_for_source(job, [part.local_env])
```
ekouts, line 179:
```suggestion
                elif use_pip:
                    _emit_script_for_pip(job, [part.local_env])
                else:
                    _emit_script_for_source(job, [part.local_env])
```
No need to indent this further I think

reframe/schemas/config.json
```diff
@@ -587,6 +591,7 @@
         "general/purge_environment": false,
         "general/remote_detect": false,
         "general/remote_workdir": ".",
+        "general/remote_command": [],
```
ekouts, line 594: Keep alphabetical order in the list

unittests/test_autodetect.py
```diff
@@ -119,3 +133,13 @@ def test_remote_autodetect(remote_exec_ctx):
         pytest.skip('job submission not supported')
 
     autodetect.detect_topology()
+
+
+def test_remote_custom_autodetect(remote_custom_exec_ctx):
+    # Test that a dummy topo.json file is created if a command to do
+    # so is specified as custom command for remote detection
```
ekouts, line 140: At the moment we are only triggering the autodetection, but cannot tell why the topo.json was created (since the reframe command will be added in the end). This is fine if it's not possible to check the script, but maybe update the comment to reflect that? 🤔

ekouts (commented):
Looks good overall. I have a few last comments

reframe/frontend/cli.py
```diff
@@ -1037,6 +1037,20 @@ def restrict_logging():
...
+        detect_topology_cmd = "reframe --detect-host-topology=topo.json"
+        detect_topology_found = [
+            s for s in remote_commands if detect_topology_cmd in s
+        ]
+        if not detect_topology_found:
+            remote_cmd_args = site_config.get('general/0/remote_command')
+            printer.debug(
+                "Adding ReFrame topology detection to the curstom command"
+            )
+            remote_cmd_args.append(detect_topology_cmd)
```
vkarak, line 1052: This is not needed. We should always add the reframe command to execute.

unittests/test_autodetect.py
```diff
@@ -65,6 +65,20 @@ def remote_exec_ctx(make_exec_ctx, temp_topo):
...
+    if test_util.USER_CONFIG_FILE is None:
+        pytest.skip('no user configuration file supplied')
+
+    ctx = make_exec_ctx(test_util.USER_CONFIG_FILE,
+                        test_util.USER_SYSTEM,
+                        {'general/remote_detect': True,
+                         'general/remote_command': [
+                             'echo \'{"dummy": "value"}\' > topo.json',
+                         ]})
+    yield ctx
```
vkarak, line 79: Just extend the existing `remote_exec_ctx` to accept additional configuration options.

unittests/test_autodetect.py
vkarak: With my suggestion of only requiring the reframe installation commands, this unit test would only need to check that the installation command was executed. We could for example have a dummy installation command `touch foo.txt` and here we could check that `foo.txt` exists.

vkarak (changes requested):
I like the idea of allowing users to specify the exact commands to run, but I would rather restrict that to the installation commands only. The `reframe --detect-host-topology=topo.json` should be added always by the framework.

As a result we should rather call the configuration variable as `remote_install`.

reframe/frontend/cli.py
ekouts: What if the command needs to be `./bin/reframe ...`, like it is the case for the bootstrap option?

vkarak: I guess that would be the case when sb wants to clone explicitly and bootstrap it, because by default we make a copy (NB: the method we discuss is something extra to what we do now). Well, in this case the user could set the `PATH` to the local `bin/`:

```python
remote_install = ['git clone ... && cd reframe && ./boostrap.sh && cd -',
                  'export PATH=$PWD/reframe/bin:$PATH']
```

unittests/test_autodetect.py
```diff
@@ -55,13 +55,23 @@ def default_exec_ctx(make_exec_ctx_g, temp_topo):
...
 @pytest.fixture
-def remote_exec_ctx(make_exec_ctx, temp_topo):
+def remote_exec_ctx(request, make_exec_ctx, temp_topo):
     if test_util.USER_CONFIG_FILE is None:
         pytest.skip('no user configuration file supplied')
 
-    ctx = make_exec_ctx(test_util.USER_CONFIG_FILE,
-                        test_util.USER_SYSTEM,
-                        {'general/remote_detect': True})
+    custom_install = getattr(request, 'param', False)
```
vkarak, line 62: Why this wouldn't work?
 
```suggestion
    custom_install = request.param
```

unittests/test_autodetect.py
```diff
@@ -55,13 +55,23 @@ def default_exec_ctx(make_exec_ctx_g, temp_topo):
...
+        ctx = make_exec_ctx(test_util.USER_CONFIG_FILE,
+                            test_util.USER_SYSTEM,
+                            {'general/remote_detect': True})
+    else:
+        ctx = make_exec_ctx(test_util.USER_CONFIG_FILE,
+                            test_util.USER_SYSTEM,
+                            {'general/remote_detect': True,
+                             'general/remote_install': [
+                                 'echo \'{"dummy": "value"}\' > topo.json',
+                             ]})
```
vkarak, line 74: Rewrite this as follows to avoid duplication:

```suggestion
    options = {'general/remote_detect': True}
    custom_install = getattr(request, 'param', False)
    if custom_install:
        options.update({'general/remote_install': ['echo install_reframe']})

    yield make_exec_ctx(test_util.USER_CONFIG_FILE,
                        test_util.USER_SYSTEM,
                        options)
```

unittests/test_autodetect.py
```diff
@@ -55,13 +55,23 @@ def default_exec_ctx(make_exec_ctx_g, temp_topo):
...
+    if not custom_install:
+        ctx = make_exec_ctx(test_util.USER_CONFIG_FILE,
+                            test_util.USER_SYSTEM,
+                            {'general/remote_detect': True})
+    else:
+        ctx = make_exec_ctx(test_util.USER_CONFIG_FILE,
+                            test_util.USER_SYSTEM,
+                            {'general/remote_detect': True,
+                             'general/remote_install': [
+                                 'echo \'{"dummy": "value"}\' > topo.json',
```
vkarak, line 73: I don't understand the purpose of this. Maybe it's a leftover from the previous implementation. I've proposed below a simple `echo`.

unittests/test_autodetect.py
```diff
@@ -110,10 +120,17 @@ def test_autotect_with_invalid_files(invalid_topo_exec_ctx):
...
 def test_remote_autodetect(remote_exec_ctx):
     # All we can do with this test is to trigger the remote auto-detection
     # path; since we don't know what the remote user system is, we cannot test
     # if the topology is right.
+    # When the parameter in remote_exec_ctx is set to True, custom remote
+    # installation commands are tested creating a dummy json like topo.json.
+    # This is done to avoid a warning (which would be raised as an error)
+    # when ReFrame tries to search for the topo.json after
+    # reframe --detect-host-topology=topo.json, thus checking that the custom
+    # command was executed.
```
vkarak, line 133: This seems to be leftover from the previous implementation.

unittests/test_autodetect.py
```diff
@@ -110,10 +120,17 @@ def test_autotect_with_invalid_files(invalid_topo_exec_ctx):
     assert part.devices == []
 
 
+@pytest.mark.parametrize('remote_exec_ctx', [False, True], indirect=True)
```
vkarak, line 123: When parametrizing boolean fixtures, I don't prefer using the raw true/false values because they don't convey any information in the test name. Instead of `pytest.mark.parametrize`, I would parameterize a fixture and pass this to `remote_exec_ctx`:

```python
@pytest.fixture(params=['default', 'custom_install'])
def custom_install(request):
    return request.param == 'custom_install'


@pytest.fixture
def remote_exec_ctx(make_exec_ctx, temp_topo, custom_install):
```

unittests/test_autodetect.py
```diff
@@ -55,13 +55,23 @@ def default_exec_ctx(make_exec_ctx_g, temp_topo):
 
 
 @pytest.fixture
-def remote_exec_ctx(make_exec_ctx, temp_topo):
+def remote_exec_ctx(request, make_exec_ctx, temp_topo):
```
vkarak, line 58: Check my suggestion below.

docs/config_reference.rst
```diff
@@ -618,13 +618,18 @@ System Partition Configuration
...
+        a. ReFrame creates a temporary directory created under ``.`` by default.
+           This temporary directory prefix can be changed by setting the :envvar:`RFM_REMOTE_WORKDIR` environment variable or the :attr:`general.remote_workdir` configuration option.
+
+        b. ReFrame changes to that directory and launches a job for the topology auto-detection ``reframe --detect-host-topology=topo.json``.
+           The :option:`--detect-host-topology` option causes ReFrame to detect the topology of the current host, which in this case would be one of the remote compute nodes.
+
+         - If a custom set of commands for the remote installation of ReFrame is specified through :attr:`general.remote_install`, ReFrame launches a job with the passed list of commands. The command ``reframe --detect-host-topology=topo.json`` is always added as a last command. Thus, the installation commands must make sure that the reframe installation is found.
+         - Otherwise, ReFrame creates a fresh clone of itself in the temporary directory and launches a job that will first bootstrap the fresh clone and then run that clone with ``{launcher} ./bin/reframe --detect-host-topology=topo.json``.
```
vkarak, line 632: Maybe we can simplify a bit this description and we'd need also to mention the `pip` path (which was not mentioned before, wrongly 😄 )

vkarak (changes requested):

vkarak (approved):
Lgtm, thanks @Blanca-Fuentes!

PR #3283, opened 2024-10-11
[enhancement] Various enhancements to the results DB feature

vkarak:
More specifically, this PR adds the following enhancements:

- Support for computed virtual test attributes. This mechanism allows us to extend the test queries in existing databases without having to migrate the DBs and update the test records.
  - `basename`: the test's name without any parameters (essentially the test class name). This is useful for parameter-based aggregations.
  - `sysenv`: the formatted system, partition, environment combination
  - `pdiff`: the performance difference when doing comparisons
- Support for completely custom groupings. The syntax os the `CMPSPEC` is extended to get an explicit list of attributes for grouping test cases.
- Support for arbitrary table column selection. The syntax of the `CMPSPEC` is extended to get an explicit list of attribute columns to show. As a result the `--table-hide-columns` option is removed as now the exact columns of the output can be specified.
- Extend `--list-stored-testcases` to show the performance summary of the selected cases. It uses the same `CMPSPEC` syntax.
- Format parameters when storing a report/session using the parameter's format function.
- Show A/B absolute values for non-aggregated attributes/columns.
- Enable `-E` for testcase filtering.
- Session extras keys are now stored unchanged in the DB (previously, they were prefixed with `$`). This allows them to be properly filtered with the session filtering expressions ~~passed to `--session-filter`~~.
- The `--session-extras` option can now be specified multiple times.
- Replace ``pretty`` table format with ``outline`` and add a new ``grid`` format. This is to allow nice printing of multiline cells.
- Remove `--session-filter` option and replace it with special syntax in the `CMPSPEC`. This allows us to select and compare different sets of sessions.
- Clean up and homogenize the internal storage API.
- All options dealing with sessions (e.g., `--list-stored-sessions`, `--delete-sessions`) are now accepting the extended query syntax (over time period, session uuid or session filter).
- Options that used to manipulate a single session have now their names converted to plural.

Docs and unit tests are updated accordingly.


vkarak:
As another improvement, what would you think of making `--session-extras` an "append" option so that it can be specified multiple times to augment any "default" extras passed for example in a pipeline?

teojgo (approved):

PR #3293, opened 2024-10-25
[enhancement] Report the undefined parameters if an instantiation of an abstract test or fixture is attempted

### LINKED ISSUE #3254: ReFrame detects the test has undefined parameters, but doesn't say which ones
I sometimes get 

```
WARNING: skipping test 'X': test has one or more undefined parameters
```

And recently got the same question from a user who was trying to run one of my tests. I found https://github.com/reframe-hpc/reframe/issues/2240 which is related, and read the answer, but... I don't fully understand the issue here. Wouldn't a piece of code like:

```
if (cls.is_abstract()):
    params = cls.param_space.params
    undefined_params = []
    for param in params:
        if not cls.param_space.defines(param):
            undefined_params.append(param)
    # Raise more detailed warning
    getlogger().warning(
        f'skipping test {cls.__qualname__!r}: '
        f'test has one or more undefined parameters:'
        ", ".join(undefined_params)
    )
```

Replacing [this segment](https://github.com/reframe-hpc/reframe/blob/1bef008fb2f561795aa2b64f56074f0baa80bda3/reframe/core/decorators.py#L173) of code do the trick?

I've seen the argumentation in #2240 that tracing it back to the line might be difficult, and that's fine by me. But naming _which_ parameter is undefined can already be a big help. Sure, users would still have to see where that parameter is defined, and why it is possible undefined, but they'd have a clear starting point for the search :)

Disclaimer: I haven't _actually_ tested the code above, but it seems like this is how it _should_ work (famous last words :))

Blanca-Fuentes:
This PR is to improve the error reporting when skipping tests or reporting ReframeSyntaxErrors.

- Tests skipped because of undefined parameters now print information regarding the undefined parameters.

Closes #3254

### DIFF
--- a/reframe/core/decorators.py
+++ b/reframe/core/decorators.py
@@ -8,7 +8,7 @@
 #
 
 __all__ = ['simple_test']
-
+_TREAT_WARNINGS_AS_ERRORS = False
 
 import inspect
 import sys
@@ -104,7 +104,7 @@ class TestRegistry:
                     kwargs['reset_sysenv'] = reset_sysenv
                     leaf_tests.append(test(*args, **kwargs))
                 except SkipTestError as e:
-                    getlogger().verbose(
+                    getlogger().warning(
                         f'skipping test {test.__qualname__!r}: {e}'
                     )
                 except Exception:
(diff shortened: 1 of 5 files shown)


vkarak (approved):
I updated the PR with my suggestions. Lgtm now, thanks!

PR #3295, opened 2024-10-29
[feat] Add a configuration option to enable/disable results storage

vkarak:
By default, results storage is enabled but users are allowed to disable this feature.


docs/config_reference.rst
```diff
@@ -1674,6 +1682,16 @@ Result storage configuration
    Permissions of an existing DB file have to be changed manually.
 
 
+.. py:attribute:: storage.target_systems
+
+   :required: No
+   :default: ``["*"]``
+
+   A list of systems *only* that this storage configuration is valid for.
```
teojgo, line 1690:
```suggestion
   A list of systems that this storage configuration is exclusively valid for.
```

vkarak: The text here is copy/paste from the other `target_systems` definitions. Practically, the message is that only systems can be specified (not partitions).

teojgo (approved):

PR #3313, opened 2024-11-08
[enhancement] Disable results storage by default

### LINKED ISSUE #3312: Disable results storage by default
Since this may cause disruption to users, we will require users to explicitly enable it. See also #3311.

vkarak:
Closes #3312.


ekouts (approved):
Lgtm! I also tested on eiger and it works fine.

PR #3314, opened 2024-11-11
[enhancement] Issue a more compact message when JSON schema validation fails

### LINKED ISSUE #3168: Do not print the whole configuration schema when validation of the configuration fails

vkarak:
We modify our core code for chaining exception message to only print the `message` of the validation error and not its verbose string representation. This is now printed in debug mode at the error site.

Closes #3168.


ekouts (approved):
lgtm

PR #3318, opened 2024-11-13
[bugfix] Add job options from the command-line to processor auto-detection remote job

### LINKED ISSUE #3218: Job options passed from the cli are not respected by the autodetection remote job

ekouts:
Closes #3218.


reframe/frontend/cli.py
```diff
@@ -1110,7 +1110,20 @@ def restrict_logging():
 
         sys.exit(0)
 
-    autodetect.detect_topology()
+    # Need to parse the cli options before autodection
```
vkarak, line 1113:
```suggestion
    # Need to parse the cli options before autodetection
```

reframe/frontend/autodetect.py
```diff
@@ -136,7 +136,8 @@ def _is_part_local(part):
             part.launcher_type.registered_name == 'local')
 
 
-def _remote_detect(part):
+def _remote_detect(part, job_options=None):
+    job_options = [] if job_options is None else job_options
```
vkarak, line 140: I would avoid having another default here. This is an internal function after all:

```suggestion
def _remote_detect(part, job_options):
```

vkarak: Maybe also call the argument `cli_job_options` to match the name we use in the scheduler backends.

reframe/frontend/autodetect.py
```diff
@@ -280,7 +282,8 @@ def detect_topology():
                 _save_info(topo_file, part.processor.info)
             elif detect_remote_systems:
                 with runtime.temp_environment(modules=modules, env_vars=vars):
-                    part._processor = ProcessorInfo(_remote_detect(part))
+                    part._processor = ProcessorInfo(_remote_detect(
+                                                    part, job_options))
```
vkarak, line 286: Better reformat as follows; it's cleaner:

```suggestion
                    part._processor = ProcessorInfo(
                        _remote_detect(part, job_options)
                    )
```

reframe/frontend/autodetect.py
ekouts: Ah, I didn't notice the second comment. Do you mean in `_remote_detection` only or in both `_remote_detection` and `detect_topology`?

vkarak (approved):
Lgtm

PR #3327, opened 2024-11-27
[bugfix] Treat correctly Slurm constrains with hyphens when filtering nodes

### LINKED ISSUE #3325: Flexible node allocation does not work when Slurm constraints have a hyphen in their name

vkarak:
When converting the constraint expression to a Python one we replace hyphens in constraints and node features with `_` and then evaluate the validity.

Ideally, we would need to properly support the Slurm constraint syntax, but that would an enhancements on its own.

Closes #3325.

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -351,6 +351,7 @@ class SlurmJobScheduler(sched.JobScheduler):
         option_parser.add_argument('-w', '--nodelist')
         option_parser.add_argument('-C', '--constraint')
         option_parser.add_argument('-x', '--exclude')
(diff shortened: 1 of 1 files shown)


victorusu (approved):
`lgmt`. I wonder how we haven't caught this before. Should we also add unit tests?

vkarak:
> I wonder how we haven't caught this before. Should we also add unit tests?

The problem is that we can't emulate the flexible allocation easily for a unit test. Generally our solution is suboptimal, since we don't recognize the full syntax of Slurm constraints. We simply provide a minimal meaningful subset, because I don't know if it makes sense to re-implement Slurm's node selection mechanism by constraints. Ideally, Slurm should be giving us a way to do the node selection outside submission.

PR #3328, opened 2024-11-27
[bugfix] Exit successfully if `--duration` limit is reached

### LINKED ISSUE #3324: When running with `--duration` do not exit with a non-zero exit code just because of the time out of the last run
Also, we should document what is the expected exit code in case of reruns, either through the `--duration` or the `--reruns` option. For example, for `--reruns` it seems that reframe exits with zero if the last rerun was successful, despite failures in previous reruns.

vkarak:
Also if `--reruns` or `--duration` is used, then failures are accounted from all runs.

Closes #3324.


victorusu (approved):
`lgtm`

PR #3330, opened 2024-11-29
[enhancement] Prefix special (generated) parameters with `.` instead of `$`

vkarak:
The problem with `$` is two-fold: it has a special meaning both for shells and for the regex pattern, so it makes it tricky to select a generated test's name with `-n`.

### DIFF
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -1043,7 +1043,7 @@ The way the tests are generated and how they interact with the test filtering op
 
    Parameterize a test on an existing variable.
 
-   This option will create a new test with a parameter named ``$VAR`` with the values given in the comma-separated list ``VAL0,VAL1,...``.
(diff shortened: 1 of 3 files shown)


docs/manpage.rst
```diff
@@ -1043,7 +1043,7 @@ The way the tests are generated and how they interact with the test filtering op
 
    Parameterize a test on an existing variable.
 
-   This option will create a new test with a parameter named ``$VAR`` with the values given in the comma-separated list ``VAL0,VAL1,...``.
+   The test will behave as if the variable ``VAR`` was a paramter taking the values ``VAL0,VAL1,...``.
```
teojgo, line 1046:
```suggestion
   The test will behave as if the variable ``VAR`` was a parameter taking the values ``VAL0,VAL1,...``.
```

teojgo (changes requested):

PR #3339, opened 2024-12-05
[bugfix] Fix nodelist abbreviation when last node number is a multiple of 10

### LINKED ISSUE #3338: Slurm node abbreviation fails for sequential nodes when the last node number is a power of 10
Using pin_nodes on a job where the requested nodes end with a power of 10 causes an invalid slurm node string to be generated in rfm_job.sh.  Example: vs-std-0009:vs-std-0010 results in #SBATCH --nodelist=vs-std-000[9-10]

I've tracked this down to reframe/utility/__init__.py:862 count_digits function.  I've fixed it locally by changing use of that function to len(str()) in nodelist_abbrev and nodelist_expand functions.  I also created a unittest in unittests/test_utility.py for this specific problem and verified the fix does not fail the existing tests.

vkarak:
Closes #3338.

### DIFF
--- a/reframe/utility/__init__.py
+++ b/reframe/utility/__init__.py
@@ -866,7 +866,7 @@ def count_digits(n):
     '''
 
     num_digits = 1
-    while n > 10:
+    while n >= 10:
         n /= 10
         num_digits += 1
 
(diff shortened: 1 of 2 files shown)


victorusu (approved):
`lgtm`. Great catch!

vkarak:
> `lgtm`. Great catch!

Kudos to @harrysorensennrel for catching this.

PR #3343, opened 2024-12-11
[enhancement] Optimize JSON decoding for sessions when applying filtering criteria

### LINKED ISSUE #3340: Consider indexing the sessions table on the session extras
Since `--session-extras` are primarily used to tag the various sessions, it makes sense to create an index based on the extras as this could potentially speed up queries based on extras significantly.

vkarak:
When decoding and filtering sessions based on filtering criteria (`?` syntax) we do not decode the whole session, but rather only the session information `session_info` sub-object. We then apply the filtering criteria and we only decode fully a session if it passes the filter. This way, the efficiency of queries based on the `?` filtering syntax is considerably improved as only the required subsets of sessions are fully decoded and not the whole DB.

Closes #3340.

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -234,7 +234,7 @@ class SlurmJobScheduler(sched.JobScheduler):
                 job._cli_options = other_cli_opts
 
             arg = '&'.join(f'({c.strip()})' for c in constraints)
-            job._sched_access = [f'--constraint={arg}']
+            job._sched_access = [f'--constraint={arg}'] + access_other
 
         if not self._sched_access_in_submit:
             for opt in job.sched_access:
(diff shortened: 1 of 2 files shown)


reframe/frontend/reporting/storage.py
```diff
@@ -233,26 +233,39 @@ def _decode_sessions(self, results, sess_filter):
...
+            r'\"session_info\":\s+(?P<sess_info>\{.*?\})'
+        )
+
+        def _extract_sess_info(s):
+            return sess_info_patt.search(s).group('sess_info')
+
+        @time_function
+        def _mass_json_decode(json_objs):
+            data = '[' + ','.join(json_objs) + ']'
+            getlogger().debug(f'decoding {len(data)} bytes')
```
teojgo, line 246:
```suggestion
            getlogger().debug(f'decoding {len(data)} code points')
```

teojgo (changes requested):

reframe/frontend/reporting/storage.py
vkarak: Yes, they are not bytes :-) But few will understand the code points seeing the logs. I would also add the byte count (assuming ASCII characters which is 99% of the cases). Something like:

```
decoding JSON raw data of length X (X bytes assuming ASCII chars)
```

teojgo (approved):

PR #3347, opened 2024-12-13
[feat] Asyncio execution policy

Blanca-Fuentes:
This PR changes the asynchronous execution policy to use `asyncio` (`AsyncioExecutionPolicy`).

The main changes introduced are the following:

- `KeyBoardInterrupt` management: introduction of a new type of exception `KeyboardError` to raise in case of keyboard interruption during the execution of the code.
- `attach_hooks()`: if the hook is the `run` / `wait` methods of a test then, they need to be treate as asyncio coroutines
- Logging context manager: switching the control within the logging context must be treated carefully. I created a `_global_logger` and  a `tasks_logger`.  The `_global_logger` is used when the asyncio loop has not been yet opened and the `tasks_logger` to retrieve the right logging context for each task.
- The `stage_dir` and the `output_dir` must be absolute paths for the right context switching with `with change_dir_global`. The working directory of reframe is now set at the beginning with `set_working_dir`, it is retrieved at any point in the code with `get_working_dir` because changing directory with asyncio does not guarantee that we will be in the right directory when doing `os.getcwd()`.
- The test methods `compile`, `run`, `compile_wait`, `run_wait` are now asynchronous.
- New `run_command_asyncio` method was added to create an asyncio subprocess, which has the same output as the `run_command_async`. This method waits (asynchronously) for the command to be executed. The method `run_command_async_alone` just starts the subprocess but does not wait for completion (this allows the local scheduler to poll for the job completion).
- The termination and killing (`_term_all()` and `_kill_all()`) of the processes in the local scheduler is more challenging because the communication of signals between children and parent processes is not handled in the same way. The children of a process are retrieved explicitly and signals are sent to each process individually.
- `_run_strict` and `_run_strict_s` are the asyncio and synchronous versions of `_run_strict`.
- The handling of the different cases (tests) takes place inside `execute()` not in `exit()` as in the previous `AsynchronousExecutionPolicy`.
- Adaptation of the tests for the new `asyncio` policy.

### TODOs

Extend the implementation to other schedulers (only slurm /local)

- [x] pbs
- [x] flux
- [x] sge
- [x] slurm
- [x] local
- [x] ssh
- [x] oar

Other tasks

- [ ] Test performance vs previous AsynchornousExecutionPolicy
- [ ] Create more tests for the unittests


reframe/frontend/cli.py
```diff
@@ -32,7 +32,7 @@
                                              getallnodes, repeat_tests,
                                              parameterize_tests)
 from reframe.frontend.executors.policies import (SerialExecutionPolicy,
-                                                 AsynchronousExecutionPolicy)
```
vkarak, line 35: Why the need to change the name if we are not keeping the original one?

reframe/frontend/executors/__init__.py
```diff
@@ -763,9 +815,42 @@ def timeout_expired(self):
...
         pass
 
     @abc.abstractmethod
-    def runcase(self, case):
-        '''Run a test case.'''
+    def execute(self, testcases):
+        '''Execute the policy for a given set of testcases and exit.'''
+
+
+def asyncio_run(coro):
```
vkarak, line 826: Once I have tried an async def of the `main()` and started the loop there with `asyncio.run(main())`. I haven't changed anything else in the code and that was enough to run as-is the existing code (ie. not asyncio-based). Why do we need all this complex event loop logic here?

vkarak:
Thanks @Blanca-Fuentes for the effort! It was a very good exercise to see what it would take such a transition.

PR #3354, opened 2024-12-20
[bugfix] Improve handling of requests in httpjson logger

### LINKED ISSUE #3353: Check response status code in httpjson logger
Sometimes we end up losing a few data from the daily runs, because the http request that sends the data actually fails with `429 Too Many Requests`. Reframe should at the very least print some error when it fails to send the data and ideally retry after a few seconds in the case of 429.

ekouts:
Fixes #3353 .

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -680,7 +680,6 @@ class HTTPJSONHandler(logging.Handler):
             return
 
         if self._debug:
-            import time
             ts = int(time.time() * 1_000)
             dump_file = f'httpjson_record_{ts}.json'
             with open(dump_file, 'w') as fp:
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -689,11 +689,18 @@ class HTTPJSONHandler(logging.Handler):
             return
 
         try:
-            response = requests.post(
-                self._url, data=json_record,
-                headers=self._headers
-            )
-            if response.status_code != 200:
+            while True:
+                response = requests.post(
+                    self._url, data=json_record,
+                    headers=self._headers
+                )
+                if response.status_code == 200:
+                    break
+
+                if response.status_code == 429:
+                    time.sleep(1)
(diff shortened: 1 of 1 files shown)


reframe/core/logging.py
```diff
@@ -689,10 +688,22 @@ def emit(self, record):
...
+                    break
+
+                if response.status_code == 429:
+                    time.sleep(1)
+                    continue
+
+                raise LoggingError(
+                    f'logging failed: HTTP response code '
+                    f'{response.status_code}'
+                )
```
ekouts, line 706: Should we crash in this case, or try to print a warning?

vkarak: How does the outpupt look like when we raise `LoggingError`? I see that currently we do raise it, but I'm not sure it's the right thing. I would rather go for a warning instead (remember to include the name of the handler in the message).

reframe/core/logging.py
```diff
@@ -689,10 +688,22 @@ def emit(self, record):
...
-            requests.post(
-                self._url, data=json_record,
-                headers=self._headers
-            )
+            while True:
+                response = requests.post(
+                    self._url, data=json_record,
+                    headers=self._headers
+                )
+                if response.status_code == 200:
```
vkarak, line 696: Maybe we should go for the more inclusive:

```suggestion
                if response.ok:
```

reframe/core/logging.py
```diff
@@ -689,10 +688,22 @@ def emit(self, record):
...
+                    break
+
+                if response.status_code == 429:
+                    time.sleep(1)
```
vkarak, line 700: Since this can slow down the pipeline what about spreading the wait with a cyclic interval list, say `[.1, .2, .4, .8, 1.6, 3.2]`?

vkarak (changes requested):
Consider also rebasing this PR onto `master`.

PR #3370, opened 2025-01-29
[feat] Enable YAML configuration files

### LINKED ISSUE #3322: Support for YAML configs
For static configuration, this would lead to much more compact files.

vkarak:
This PR introduces YAML configuration files and deprecates raw JSON configuration files.

Check the updated docs for more details:

- docs/config_reference.rst
- docs/manpage.rst

Closes #3322.

## Todos

- [x] Issue deprecation warning when a JSON configuration file is loaded.


teojgo (approved):

vkarak:
I'm converting this PR to draft, since I had some second thoughts on the `sh` binding. It's safer if we remove it. The same effect could be achieved through an environment variable, but ReFrame will not be executing arbitrary code:

```
FOO=$(echo "hello")
```

and then using `getenv("FOO")` in the config.

I'm adding some other useful bindings such as username, groupname, uid, and gid.

victorusu (approved):
This PR is awesome! `lgtm`.

PR #3389, opened 2025-02-11
[doc] Fix tutorial on `--performance-report` default behaviour

### LINKED ISSUE #3373: Outdated tutorial text about test result storage
The tutorial listings regarding the results storage are using an older output format (tabular without grid lines). Also the following [text](https://reframe-hpc.readthedocs.io/en/stable/tutorial.html#running-a-test) is inaccurate:

> Finally, you can use also the [--performance-report](https://reframe-hpc.readthedocs.io/en/stable/manpage.html#cmdoption-performance-report) option, which will print a summary of the results of the performance tests that have run in the current session and compare them (by default) to their last obtained performance.

The `--performance-report` option does not compare with past results by default.

vkarak:
Closes #3373.


ekouts (approved):
lgtm

PR #3393, opened 2025-02-13
[bugfix] Do not copy or `git clone` `sourcesdir` if `--dont-restage` is passed

### LINKED ISSUE #3305: `--keep-stage-files --dont-restage` fails when used together with `sourcesdir`
Something like

```
class MPIWrapper(rfm.CompileOnlyRegressionTest):
    sourcesdir = 'https://github.com/eschnett/MPIwrapper'
    build_system = 'CMake'
```

fails., when used with `--keep-stage-files --dont-restage`

```

Valentin Churavy
  Thursday at 4:38 PM
FAILURE INFO for MPIWrapper ~local:default+julia_1.10 (run: 1/1)
  * Description: 
  * System partition: local
  * Environment: julia_1.10
  * Stage directory: /home/vchuravy/src/trixi-parallel-performance/stage/local/default/julia_1.10/MPIWrapper_e9d9ef7f
  * Node list: 
  * Job type: local (id=None)
  * Dependencies (actual): []
  * Maintainers: []
  * Failing phase: compile
  * Reason: spawned process error: command 'git clone  https://github.com/eschnett/MPIwrapper /home/vchuravy/src/trixi-parallel-performance/stage/local/default/julia_1.10/MPIWrapper_e9d9ef7f' failed with exit code 128:
--- stdout ---
--- stdout ---
--- stderr ---
fatal: destination path '/home/vchuravy/src/trixi-parallel-performance/stage/local/default/julia_1.10/MPIWrapper_e9d9ef7f' already exists and is not an empty directory.
```

vkarak:
There was a more general problem with `--dont-restage` apart from trying to re-issue a `git clone` command if `sourcesdir` was a url. Even if it wasn't, it would still copy the contents of the `sourcesdir` to stagedir. Now we neither copy nor clone if `--dont-restage` is passed. 

However, since we now disallow a copy if `--dont-restage` is passed, that would mean that `--dont-restage` would cause a test to fail if it was run for the first time with `--dont-restage` as its resources wouldn't be copied. To overcome this, each test now write a mark in the stage directory. Before copying the resources to the stage directory, it checks if the mark exists. If it does, it means that there has been a previous run and that the stagedir contents haven't been wiped out, which means `--dont-restage` was passed. If the mark does not exists, the it's a fresh run (either the first ever run of a test, even with `--dont-restage`, or just a regular rerun without `--dont-restage`).

Fixes #3305.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -2645,7 +2645,7 @@ class RunOnlyRegressionTest(RegressionTest, special=True):
         The resources of the test are copied to the stage directory and the
         rest of execution is delegated to the :func:`RegressionTest.run()`.
         '''
-        if (self.sourcesdir and self._requires_stagedir_contents()):
+        if self.sourcesdir and self._requires_stagedir_contents():
             if osext.is_url(self.sourcesdir):
                 self._clone_to_stagedir(self.sourcesdir)
             else:
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -20,6 +20,7 @@ import itertools
 import numbers
 import os
 import shutil
+from pathlib import Path
 
 import reframe.core.fields as fields
 import reframe.core.hooks as hooks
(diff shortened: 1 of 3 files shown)


reframe/core/pipeline.py
```diff
@@ -2628,7 +2645,7 @@ def run(self):
         The resources of the test are copied to the stage directory and the
         rest of execution is delegated to the :func:`RegressionTest.run()`.
         '''
-        if self.sourcesdir:
+        if (self.sourcesdir and self._requires_stagedir_contents()):
```
teojgo, line 2648:
```suggestion
        if self.sourcesdir and self._requires_stagedir_contents():
```

teojgo (changes requested):

teojgo (approved):

PR #3395, opened 2025-02-19
[feat] Add support for configure directory in CMake builds

### LINKED ISSUE #3359: Add CMake.configuredir Property for Source Directory Specification
ReFrame provides `Autotools.configuredir` for specifying the source directory in Autotools builds. However, there is currently no equivalent property for CMake. Users must manually add `-S <source_dir>` to `build_system.config_opts`:

```shell
self.build_system.config_opts = [
    '-DOPTION=TRUE',
    '-S', 'src'
]
```

**Requested Feature:**
Introduce `CMake.configuredir` to specify the source directory, mirroring the existing Autotools behavior.

**Benefits:**
- Eliminates the need to manually pass the `-S` option in `build_system.config_opt`s.
- Aligns the CMake build workflow with how Autotools is handled in ReFrame.

mredenti:
Closes #3359.

### Changes:
- Move `configuredir` to `ConfigureBasedBuildSystem`, making it available to all configure-based build systems (Autotools, CMake).
- Update `CMake.emit_build_commands` to respect `configuredir`.
- Remove redundant `configuredir` definition from `Autotools`, as it now inherits this property from `ConfigureBasedBuildSystem`.

Now `CMake.configuredir` allows users to specify the source directory explicitly, aligning with `Autotools.configuredir`.

### Why We Don't Use `-S`
Although CMake provides the `-S` flag to specify the source directory, this option was **introduced in CMake 3.13**. To maintain compatibility with older versions, this implementation does not use `-S` explicitly. Instead, `configuredir` is passed as the last argument in the `cmake` command, which ensures compatibility with both older and newer CMake versions.

Indeed, CMake v3.12 does not support the `-S` flag https://cmake.org/cmake/help/v3.12/manual/cmake.1.html but v3.13 does https://cmake.org/cmake/help/v3.13/manual/cmake.1.html


reframe/core/buildsystems.py
```diff
@@ -481,6 +481,15 @@ class ConfigureBasedBuildSystem(BuildSystem):
...
     #: :default: :class:`None`
     srcdir = variable(str, type(None), value=None)
+    
+    #: The directory of the configure script.
+    #:
+    #: This can be changed to do an out of source build without copying the
+    #: entire source tree.
+    #:
+    #: :type: :class:`str`
+    #: :default: ``'.'``
```
vkarak, line 491: Add a `.. versionchanged:: 4.8` annotation stating that this variable is now available in all configure-based build systems (Autotools, CMake).

reframe/core/buildsystems.py
```diff
@@ -578,11 +587,13 @@ def emit_build_commands(self, environ):
 
         if self.config_opts:
             cmake_cmd += self.config_opts
-
+
```
vkarak, line 590: Remove this whitespace.

vkarak (changes requested):
I agree with the solution to move the `configuredir` to the base class. My comments are only related to the documentation.

As for the CI failure, I am having a look; I don't think your PR's fault.

vkarak (approved):

PR #3398, opened 2025-02-24
[bugfix] Mark the stagedir after the copy or clone of resources

### LINKED ISSUE #3397: Git repos cannot be cloned in stage directory
This was introduced in #3393 since the mark file is written before cloning the source code, therefore trying to clone fails after since the directory is not empty.

vkarak:
Closes #3397.

### DIFF
--- a/unittests/test_pipeline.py
+++ b/unittests/test_pipeline.py
@@ -763,6 +763,23 @@ def test_sourcesdir_build_system(local_exec_ctx):
     _run(MyTest(), *local_exec_ctx)
 
 
+def test_sourcesdir_git(local_exec_ctx):
+    @test_util.custom_prefix('unittests/resources/checks')
+    class MyTest(rfm.RunOnlyRegressionTest):
+        sourcesdir = 'https://github.com/octocat/hello-worId.git'
+        executable = 'true'
+        valid_systems = ['*']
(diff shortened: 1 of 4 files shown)


unittests/test_pipeline.py
```diff
@@ -763,6 +763,23 @@ def validate(self):
     _run(MyTest(), *local_exec_ctx)
 
 
+def test_sourcesdir_git(local_exec_ctx):
+    @test_util.custom_prefix('unittests/resources/checks')
+    class MyTest(rfm.RunOnlyRegressionTest):
+        sourcesdir = 'https://github.com/octocat/hello-worId.git'
```
vkarak, line 769: Do you think we should instead create a dummy repo under reframe-hpc?

ekouts, line 769:
```suggestion
        sourcesdir = 'https://github.com/octocat/hello-world.git'
```
I fine with keeping this or creating a repo under reframe-hpc. But is the capital i in the path a typo? 😆 Both repos exist but I assume the hello world is the original

vkarak: True! I think it's better to have a repo hosted in reframe-hpc so that we can control it.

teojgo (approved):

PR #3426, opened 2025-03-13
[doc] Add warning about combining `build_locally=0` and scheduler backends that cannot reliably retrieve the job's exit code

### LINKED ISSUE #3111: Test does not fail during the compilation stage if build fails when `build_locally=False`
This is confusing because (a) this behaviour is different when `build_locally=True` and (b) the test is successful if it's compile-only (see build fixtures).

@giordano I think you've hit this before but I'm not sure if you've opened an issue already.

vkarak:
Closes #3111.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -2220,7 +2220,7 @@ class RegressionTest(RegressionMixin, jsonext.JSONSerializable):
               more details.
 
            .. versionchanged:: 3.4
-              Overriding this method directly in no longer allowed. See `here
+              Overriding this method directly is no longer allowed. See `here
               <migration_2_to_3.html#force-override-a-pipeline-method>`__ for
               more details.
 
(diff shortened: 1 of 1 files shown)


reframe/core/pipeline.py
```diff
@@ -2702,6 +2702,14 @@ class CompileOnlyRegressionTest(RegressionTest, special=True):
     However, if a sanity function is defined, it will be used to validate the
     test.
 
+    .. warning::
+
+       ReFrame checks the exit code of the associated build job to mark the
+       test as a failure. If :attr:`build_locally` is set to :obj:`False` and
+       the scheduler backend cannot retrieve the exit code of the submitted
```
ekouts, line 2709: Small suggestion, but feel free to ignore. Does it make sense to add `squeue` as an example? Many users may not be aware of this limitation.

ekouts (approved):
Lgtm overall. I had a small suggestion, but feel free to resolve, if you don't agree.

reframe/core/pipeline.py
vkarak: @ekouts Do you think it's better to move this warning to the documentation of `build_locally`? I think it would have more visibility there, wdyt?

vkarak: We already mention `squeue`'s case there, so it's better if I move it.

ekouts: I agree, it's better in the build_locally documentation.

teojgo (approved):

PR #3431, opened 2025-03-16
[feat] Support selecting A/B column variants in performance comparisons

### LINKED ISSUE #3341: Add support for selecting the A or B column variant in performance comparisons
In performance comparisons, When listing columns not in the group-by list the amount of total columns of the final matrix can grow significantly, since for non-grouped-by attribute, two columns will be listed. While there are cases, that both columns are useful, it would be nice in the column spec of the comparison options to be able to select the A or B variant of the column.

vkarak:
When doing performance comparisons (`--performance-compare` or `--performance-report`), users can now explicitly select the A or B variant for extra columns, e.g., `+result_A` will only list the aggregated results of the first set.

Closes #3341.


victorusu (approved):
`lgtm`

PR #3432, opened 2025-03-16
[bugfix] Disallow command options from execution modes

### LINKED ISSUE #3424: using an action option via mode configuration does not work
A colleague noticed that some of our execution mode configurations are no longer working as expected after upgrading to reframe 4.7 and pinpointed the specific commit where the behavior seemed to change:

> After some manual bisection on git tag -l I found that this seems to have started breaking between 4.6.4 and 4.7.0. It looks like commit aa75c377 was the one where it started failing.

It seems like maybe an initial parsing of the cli arguments is failing before the options for the specified mode are added.

Here's a minimal reproducer:

```console
> cat test-config.py
site_configuration = {
    'modes': [
        {
            'name': 'run',
            'options': ['--run']
        }
    ]
}
> reframe --version
4.7.1
> reframe -C test-config.py --mode=run
usage: reframe [-h] [--compress-report] [--dont-restage] [--keep-stage-files]
               [-o DIR] [--perflogdir DIR] [--prefix DIR] [--report-file FILE]
               [--report-junit FILE] [-s DIR] [--save-log-files]
               [--timestamp [TIMEFMT]] [-c PATH] [-R] [--cpu-only] [--failed]
               [--gpu-only] [--maintainer PATTERN] [-n PATTERN] [-p PATTERN]
               [-T PATTERN] [-t PATTERN] [-x PATTERN] [-E EXPR]
               (--ci-generate FILE | --delete-stored-sessions QUERY | --describe | --describe-stored-sessions QUERY | --describe-stored-testcases QUERY | --detect-host-topology [FILE] | --dry-run | -L [{C,T}] | --list-stored-sessions [QUERY] | --list-stored-testcases QUERY | -l [{C,T}] | --list-tags | --performance-compare CMPSPEC | -r | --show-config [PARAM] | -V)
               [--disable-hook NAME] [--duration TIMEOUT] [--exec-order ORDER]
               [--exec-policy POLICY] [--flex-alloc-nodes {all|STATE|NUM}]
               [--flex-alloc-strict] [-J OPT] [--max-retries NUM]
               [--maxfail NUM] [--mode MODE] [--reruns N]
               [--restore-session [REPORT]] [--retries-threshold VALUE[%]]
               [-S [TEST.]VAR=VAL] [--skip-performance-check]
               [--skip-prgenv-check] [--skip-sanity-check]
               [--skip-system-check] [-M MAPPING] [-m MOD]
               [--module-mappings FILE] [--module-path PATH]
               [--non-default-craype] [--purge-env] [-u MOD]
               [--distribute [{all|avail|STATE}]] [-P VAR:VAL0,VAL1,...]
               [--repeat N] [--performance-report [PERFORMANCE_REPORT]]
               [--session-extras KV_DATA] [-C FILE] [--failure-stats]
               [--nocolor] [--system SYSTEM]
               [--table-format {csv,pretty,plain}] [-v] [-q]
reframe: error: one of the arguments --ci-generate --delete-stored-sessions --describe --describe-stored-sessions --describe-stored-testcases --detect-host-topology --dry-run -L/--list-detailed --list-stored-sessions --list-stored-testcases -l/--list --list-tags --performance-compare -r/--run --show-config -V/--version is required
```

vkarak:
Execution modes are not allowed to use command/action options. This was never the intent of execution modes and since 4.7, since `--mode` is not a command option, it was required to pass one in the command line. If the mode had already a command option, then which of the two would take effect was implementation defined. This PR disallows completely the use of command options in execution modes to avoid such abmiguities.

Closes #3424.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1698,6 +1698,14 @@ The options of an execution mode will be passed to ReFrame as if they were speci
 
    The command-line options associated with this execution mode.
 
+   .. warning::
+      :ref:`Command options <commands>` are *not* allowed in execution modes.
+      These should always be passed explicitly in command line.
+
+   .. versionchanged:: 4.8
+
+      Command options are explicitly disallowed in execution modes.
+
 
 .. py:attribute:: modes.target_systems
 
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -666,6 +666,14 @@ Options controlling ReFrame execution
    .. versionchanged:: 4.1
       Options that can be specified multiple times are now combined between execution modes and the command line.
 
+   .. versionchanged:: 4.7
+      The :option:`--mode` must always be combined with a :ref:`command option <commands>`.
+      If the mode contains a command option already, the command option that will finally take effect is implementation defined.
+
+   .. versionchanged:: 4.8
+      Command options are disallowed from execution modes.
+
+
 .. option:: --reruns=N
 
    Rerun the whole test session ``N`` times.
--- a/reframe/frontend/argparse.py
+++ b/reframe/frontend/argparse.py
@@ -160,6 +160,10 @@ class _ArgumentHolder:
         # options. Values are tuples of the form (envvar, configvar)
         self._option_map = shared_options if shared_options is not None else {}
 
+        # We store the options (without actions) to quickly decide if an option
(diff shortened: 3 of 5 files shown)


victorusu (approved):
`lgtm`

Issue #2857, opened 2023-04-10
Support absolute value thresholds in the reference tuple


vkarak:
Currently the thresholds are essentially percentages relative to the reference value. As a result, reference values cannot be 0, as also the bounds will be set to zero and if done so, reframe will ignore them. See #2833. We need a way to support absolute thresholds as well as relative.

jack-morrison:
I have an interest in `0` reference values for some tests, so I'll take a look at this one.

PR #3434, opened 2025-03-18
[feat] Support reference values of zero

jack-morrison:
Partially addresses #2857.

The spirit of #2857 is for fully supporting both relative and absolute value reference thresholds. That's not what's provided here. Rather, this PR specifically supports the case for reference values of 0. When a reference value of 0 is provided, lower and upper thresholds are treated as absolute values.


vkarak (changes requested):
Only docs are missing, which you can add here:

https://github.com/reframe-hpc/reframe/blob/20fed5da1569a4962bed12510a89f53083aea3ef/reframe/core/pipeline.py#L726

vkarak (approved):

PR #3438, opened 2025-03-24
[feat] Support discrete testcase filters when comparing performance

### LINKED ISSUE #3437: Support performance comparisons between testcases based on disjoint filtering criteria
Although we can currently compare the tests cases from sessions with disjoint extras (see the `?` syntax), this is not the case when filtering directly test cases. This is because the `-E` option applies to both A and B sets of comparison. We should be able to specify different testcase filters for selecting the A and B sets.

vkarak:
This PR allows user to specify the `-E` twice when combined with the `--performance-compare` option, in which each of the filter expression will be applied to the corresponding testcase set.

Closes #3437.


victorusu (approved):
`lgtm`

PR #3440, opened 2025-03-26
[enhancement] List parameterized tests in their instantiation order

vkarak:
Although the parameterized tests were instantiated in the order they appear in the test file and in the order of their parameter list, they were listed in reverse order, which was annoying, although technically not a bug (the default listing is implementation-defined).

The problem was that we were iterating over them in reverse order in order to instantiate any fixtures, so the final list of tests was reverted. This is fixed by using a `deque` list and popping from the left side.

### DIFF
--- a/reframe/core/decorators.py
+++ b/reframe/core/decorators.py
@@ -9,6 +9,7 @@
 
 __all__ = ['simple_test']
 
+import collections
 import inspect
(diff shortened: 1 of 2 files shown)


victorusu (approved):
`lgtm`

PR #3441, opened 2025-03-26
[bugfix] Update special parameter character in test name

vkarak:
The `name` test property was not updated when the special parameter character was changed from `$` to `.` in #3330.


victorusu (approved):
`lgtm`

ekouts (approved):

PR #3442, opened 2025-03-31
[feat] Allow changing the default mount location of stagedir inside containers

### LINKED ISSUE #3367: Allow Overriding the Default Mount Point of the Stage Directory in Containers
**Description**

Currently, ReFrame mounts the test’s stage directory inside the container to `/rfm_workdir` by default. This behavior is hard-coded and does not provide flexibility for users to specify an alternative mount point. - see https://reframe-hpc.readthedocs.io/en/stable/regression_test_api.html#reframe.core.containers.ContainerPlatform.workdir

**Proposes feature**

Introduce a configuration option that allows users to override the stage directory as the default mount point that points to `/rfm_workdir` directory inside the container. 

**Motivation**

In environments with non-shared filesystems, input files are not located on the login node where the stage directory resides. As a result, the default mounting behavior restricts the accessibility of input files within the container. In this case one has to specify and additional mount point and and also set the `self.container_platform.workdir`. It would be more efficient to allow the possibility of changing the default mount point in such situations

vkarak:
This introduces a new `mount_stagedir` variable that controls the mount point of the stagedir. If `None`, the stagedir will not be mounted.

This PR also refactors the `ContainerPlatform` so as to use the builtin variables instead of the raw fields.

Closes #3367.

### DIFF
--- a/reframe/core/containers.py
+++ b/reframe/core/containers.py
@@ -10,9 +10,6 @@ import reframe.utility.typecheck as typ
 from reframe.core.meta import RegressionTestMeta
 
 
-_STAGEDIR_MOUNT = '/rfm_workdir'
-
-
 class _ContainerMeta(typ.ConvertibleType, RegressionTestMeta, abc.ABCMeta):
     '''Metaclass for container platforms.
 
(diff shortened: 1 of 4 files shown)


reframe/core/containers.py
```diff
@@ -5,15 +5,20 @@
...
+class _ContainerMeta(typ.ConvertibleType, RegressionTestMeta, abc.ABCMeta):
+    '''Metaclass for container platforms.
+
+    This essentially combines the ABCMeta with our standard metaclass that
+    provides the variables.
+    '''
 
 
-class ContainerPlatform(abc.ABC):
+class ContainerPlatform(abc.ABC, metaclass=_ContainerMeta):
```
victorusu, line 21: I just thinking out loud... Whenever, I see the use if `RegressionTestMeta` I wonder wether it should be implemented using a `RegrssionMixin`...

victorusu (approved):
`lgtm`.  Is it just me, or the implementation via `variables` feels more consistent and correct? 😄

reframe/core/containers.py
vkarak: Maybe you are right, I'll try that out. The intent was to bring in the "variable" definitions.

vkarak:
> `lgtm`. Is it just me, or the implementation via `variables` feels more consistent and correct? 😄

Actually, we can't use `RegressionMixin` in this case, since we'd be introducing a circular import. I agree though that we should refactor this part in the future.

PR #3448, opened 2025-04-02
[bugfix] Fix constraint evaluation for Slurm features starting with numbers

yqshao:
Features starting with a number (e.g. `4xA40`) seem valid to slurm, but are not valid python identifiers. Currently this causes nodes getting filtered out when such feature is used as constraints. ~~This prefixes all variables during eval with underscore to get around it.~~ 

Edit: this now inserts boolean values directly into the constraints to support such Slurm features.

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -732,31 +732,22 @@ class _SlurmNode(sched.Node):
         return not self.is_avail()
 
     def satisfies(self, slurm_constraint):
-        def _replacemany(s, replacements):
-            for src, dst in replacements:
-                s = s.replace(src, dst)
-
-            return s
-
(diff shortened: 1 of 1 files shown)


yqshao:
:-1: this naively broke any expression,  will try fix.

vkarak:
You can also try the unit tests locally by running `./test_reframe.py`.

yqshao:
I scratched the original idea of prefixing all symbols, as inserting the booleans directly seems easier if the symbols needs to be replace anyway.

PS: Sorry for the force-push spam (I intended to rebase & tidy the history). I thinks it's ready to be reviewed now.

vkarak (changes requested):
I've tested it and it looks good. Just a couple of style comments.

yqshao:
Thanks! Committed the suggested changes.

vkarak (approved):

PR #3450, opened 2025-04-03
[bugfix] Fix lock order when storing reports to the database

vkarak:
We should acquire the file lock before attempting to connect to the database as we do for session removal already. This way we avoid connection timeouts when there are multiple writers.


victorusu (approved):
`lgtm`

PR #3452, opened 2025-04-03
[feat] Do not emit the `-lselect` option if `num_tasks=None`

### LINKED ISSUE #3315: Allow users to completely override the `-l select` line of PBS backend
We could omit the line if `num_tasks = None` similarly to how the Slurm backend treats `None` for `num_tasks` specially, by requiring the users to explicitly set the `self.job.options`.

vkarak:
Users can then specify it manually through `self.job.options`.

Closes #3315.


victorusu (approved):
`lgtm`

PR #3471, opened 2025-04-29
[bugfix] Use always `|` to join unique values in column aggregations except for `--table-format=pretty`

### LINKED ISSUE #3466: `job_nodelist` from different testcases should not be joined with `,` when listing performance results
This is problematic since we cannot differentiate which node list is from which test case. This does not happen when `--table-format=csv`, where the different node lists are joined with `|`.

vkarak:
Closes #3466.

### DIFF
--- a/reframe/frontend/reporting/__init__.py
+++ b/reframe/frontend/reporting/__init__.py
@@ -646,8 +646,8 @@ def _group_testcases(testcases, groups, columns):
 @time_function
 def _aggregate_perf(grouped_testcases, aggr_fn, cols):
     # Update delimiter for joining unique values based on the table format
-    table_foramt = runtime().get_option('general/0/table_format')
-    if table_foramt == 'pretty':
+    table_format = runtime().get_option('general/0/table_format')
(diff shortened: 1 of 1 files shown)


ekouts (approved):

reframe/frontend/reporting/__init__.py
```diff
@@ -647,12 +647,10 @@ def _group_testcases(testcases, groups, columns):
 def _aggregate_perf(grouped_testcases, aggr_fn, cols):
     # Update delimiter for joining unique values based on the table format
     table_foramt = runtime().get_option('general/0/table_format')
```
teojgo, line 649:
```suggestion
    table_format = runtime().get_option('general/0/table_format')
```

teojgo (changes requested):

reframe/frontend/reporting/__init__.py
vkarak: Done.

teojgo (approved):

PR #3472, opened 2025-04-29
[enhancement] Show up front if result storage is enabled

### LINKED ISSUE #3469: Show in the `[ReFrame Setup]` output immediately whether results storage is enabled or not
Currently, the results file is printed regardless, but no other information. This is confusing, because you might have to wait until the end to get the warning that results storage is not enabled.

vkarak:
Closes #3469.


victorusu (approved):
`lgtm`

PR #3523, opened 2025-07-17
[bugfix] Introduce optional file locking for appending to perflogs

### LINKED ISSUE #3521: Perflogs on NFS get corrupted when updated by multiple reframe processes
The scenario can happen when even two reframe processes on different nodes try to append to the same perflog file multiple hundrends of lines at the same time. This can be triggered easier when `perflog_compat=True` with multiple performance variables being logged in separate lines. The result is that multiple NULL characters are inserted in the log file and some lines are also corrupted.

The problem is that NFS does not guaranteed atomic appends, so to fix this we need to support locking.

vkarak:
This PR introduces two new configuration variables (`locking_enable` and `locking_file_mode`) for the `filelog` handler. Setting this will cause the `filelog` handler to use file locking when appending to the perflog files. 

This PR also introduces the following improvements:

- Introduces a new utility function `osext.flock()` to handle the creation of the lock file for the different Python versions; this function is now used also for the DB.
- Make sure the logging mechanism is shut down correctly when switching configurations and when the reframe exits.
- Use Python syntax for ``true`` and ``false`` in the configuration reference to be consistent with the use of ``None`` instead of the JSON's ``null``.

Closes #3521.

### DIFF
--- a/.github/workflows/test-flux.yaml
+++ b/.github/workflows/test-flux.yaml
@@ -1,6 +1,6 @@
 name: Test Flux Scheduler
-on:
-  pull_request: []
+
+on: [push, pull_request]
 
 jobs:
   build:
--- a/.github/workflows/test-schedulers.yaml
+++ b/.github/workflows/test-schedulers.yaml
@@ -1,6 +1,6 @@
 name: ReFrame CI / Scheduler backend tests
-on:
-  pull_request: []
+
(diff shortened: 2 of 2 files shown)


vkarak:
I'm converting this to draft since the fix has to be designed in accordance with #3532.

PR #3529, opened 2025-07-18
[bugfix] Fix progress numbering for XFAIL/XPASS tests

### LINKED ISSUE #3525: `XFAIL` status does not update and does not show the test case counter
Here is an example:

```
[     SKIP ] (63/66) TestA %.num_nodes=0 /8fa27e09 @system:part+builtin [could not satisfy the minimum task requirement: required 4, found 0]
[    XFAIL ] TestA %.num_nodes=0 /c41c74b2 @system:part+builtin [see bug XX]
[       OK ] (64/66) TestA %.num_nodes=0 /09cee7ee @system:part+builtin
[     FAIL ] (65/66) TestA %.num_nodes=0 /f1b36c78 @system:part+builtin
==> test failed during 'performance': test staged in '/home/user/stage/system/part/builtin/TestA_f1b36c78'
```

The `XFAIL` line should have been counted as 64/66 and the last `FAIL` line as 66/66. Note also that the counter is not printed in the `XFAIL` line. I believe the same happens with `XPASS` but I haven't tested it.

This is a follow up from #2378.

vkarak:
Closes #3525.

### DIFF
--- a/reframe/frontend/printer.py
+++ b/reframe/frontend/printer.py
@@ -70,7 +70,8 @@ class PrettyPrinter:
                 status = color.colorize(status, color.GREEN)
 
         final_msg = f'[ {status} ] '
-        if status_stripped in ('ABORT', 'OK', 'SKIP', 'FAIL'):
+        if status_stripped in ('ABORT', 'OK', 'SKIP', 'FAIL', 'XFAIL',
+                               'XPASS', 'ERROR'):
             if self._progress_count < self._progress_total:
                 self._progress_count += 1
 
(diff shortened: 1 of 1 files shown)


reframe/frontend/printer.py
```diff
@@ -70,7 +70,8 @@ def status(self, status, message='', just=None, level=logging.INFO):
                 status = color.colorize(status, color.GREEN)
 
         final_msg = f'[ {status} ] '
-        if status_stripped in ('ABORT', 'OK', 'SKIP', 'FAIL'):
+        if status_stripped in ('ABORT', 'OK', 'SKIP', 'FAIL', 'XFAIL',
```
teojgo, line 73:
```suggestion
        if status_stripped in {'ABORT', 'OK', 'SKIP', 'FAIL', 'XFAIL',
```

Make this a set.

teojgo (changes requested):

PR #3573, opened 2025-10-16
[feat] Treat nodes in `MAINTENANCE` state as available for flexible allocations on Slurm `MAINT`-flagged reservations

### LINKED ISSUE #3572: Improve slurm handling of maintenance reservations
Sometimes in CSCS we need to run `--distribute` together with a reservation that has enabled the flag `MAINT`. Even if we run with the `RFM_IGNORE_REQNODENOTAVAIL` set to `True`, we are getting failures for nodes that are not down and part of the selected maintenance reservation. Reframe always assumes that the node is down when its state is `MAINTENANCE`.

ekouts:
Add `MAINTENANCE` to the available states for nodes that belong in a reservation with the `MAINT` flag, when the user explicitly requests for this reservation.

Fixes #3572 .

### DIFF
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -1032,6 +1032,7 @@ The way the tests are generated and how they interact with the test filtering op
    - ``all``: Tests will run on all the nodes of their respective valid partitions regardless of the node state.
    - ``avail``: Tests will run on all the nodes of their respective valid partitions that are available for running jobs.
      Note that if a node is currently allocated to another job it is still considered as "available."
+     Also, for ReFrame partitions using the Slurm backends, if this option is used on a reservation with the ``MAINT`` flag set, then nodes in ``MAINTENANCE`` state will also be considered as available.
    - ``NODESTATE``: Tests will run on all the nodes of their respective valid partitions that are exclusively in state ``NODESTATE``.
      If ``NODESTATE`` is not specified, ``idle`` is assumed.
    - ``NODESTATE*``: Tests will run on all the nodes of their respective valid partitions that are at least in state ``NODESTATE``.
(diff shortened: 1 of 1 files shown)


reframe/core/schedulers/slurm.py
```diff
@@ -436,15 +437,23 @@ def _get_reservation_nodes(self, reservation):
             raise JobSchedulerError("could not extract the node names for "
                                     "reservation '%s'" % reservation)
 
+        flags_match = re.search(r'Flags=(\S+)', completed.stdout)
+        if flags_match:
+            if 'MAINT' in flags_match[1].split(','):
```
```suggestion
            if 'MAINT' in flags_match.group(1).split(','):
```

reframe/core/schedulers/slurm.py
```diff
@@ -691,19 +700,19 @@ def poll(self, *jobs):
...
+            nodes.add(_SlurmNode(descr, addl_avail_states=addl_avail_states))
 
     return nodes
 
 
 class _SlurmNode(sched.Node):
     '''Class representing a Slurm node.'''
 
-    def __init__(self, node_descr):
+    def __init__(self, node_descr, addl_avail_states=None):
```
vkarak: I would rather pass the flags and let the `__init__()` decide what to do with them, in which case you don't need the `addl_avail_states`. Simply, add the `MAINTENANCE` state in the `available_states` if the flag is set.

vkarak (changes requested):
I agree with the overall fix; I have just some suggestions on the implementation.

unittests/test_schedulers.py
```diff
@@ -975,22 +985,115 @@ def slurm_nodes():
...
+    monkeypatch.setattr(slurm, '_run_strict',
+                        functools.partial(_run_command_patched, check=True))
+
+
+@pytest.fixture
+def slurm_scheduler_patched(slurm_nodes, slurm_commands):
     ret = getscheduler('slurm')()
-    ret.allnodes = lambda: _create_nodes(slurm_nodes)
+    # ret.allnodes = lambda: _create_nodes(slurm_nodes)
     ret._get_default_partition = lambda: 'pdef'
```
vkarak: I need to remove this so as to test the exact scheduler logic.

PR #3575, opened 2025-10-30
[feat] Updates and improvements to the poll control mechanism

### LINKED ISSUE #211: Generalise the poll rate control mechanism
Currently, the poll rate control is only available for the async policy and it is hidden and non-configurable from the users. The purpose of this issue is twofold:

1. Create a mechanism that would control the poll rate centrally. This mechanism should be aware at any time of the actual poll rate of the backend scheduler and should be able to suggest sleep times between subsequent polls, so as to ensure that the actual poll rate does not exceed a user-defined maximum.

2. Provide the necessary knobs to the user for fine-tuning the polling behaviour.

vkarak:
This PR updates and improves the poll control mechanism in different ways:

1. The poll interval (sleep duration) is not controlled directly anymore but rather through the poll rate.
2. The poll rate starts at a maximum value and decays towards a predefined minimum. At every step the desired poll rate is calculated and from this the required poll interval is derived.
3. The desired poll rate is reset every time a monitored job finishes, so as to capture quickly other jobs that have also finished (this is the same as before).
4. Support is added for adding random noise to the poll interval (disabled by default), by allowing the random decrease or increase of the poll interval. This is useful to avoid aggregate poll rate spikes when multiple parallel reframe processes execute the same test suite.
5. All the polling control parameters are exposed to the users and can be tuned
6. A script is added to the `tools/` that allows to generate the plots for debugging polling issues.

## Todos

- [x] Add reference docs
- [x] Add article explaining the polling mechanism
- [x] Document the plot script
- [x] Add requirements file in the `tools/` directory
- [x] Add unit tests for checking the validation of poll control parameters

Closes #211.

### DIFF
--- a/unittests/test_policies.py
+++ b/unittests/test_policies.py
@@ -17,6 +17,7 @@ import reframe.utility.osext as osext
 import unittests.utility as test_util
 
 from reframe.core.exceptions import (AbortTaskError,
+                                     ConfigError,
                                      FailureLimitError,
                                      ForceExitError,
                                      RunSessionTimeout)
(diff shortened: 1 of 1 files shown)


ekouts (approved):

victorusu (approved):
`lgtm`

PR #3584, opened 2025-11-16
[feat] Add new `pbspro` scheduler backend

colleeneb:
As discussed in detail in https://github.com/reframe-hpc/reframe/issues/3579, this adds PBSPro as a backend due to differences between PBS and PBSPro.


colleeneb:
Thank you for catching the typo and the fixes!

> However, we'd need to update the docs.
I can try to update the docs, is there a pointer to what needs to change? I see references for PBS here: https://github.com/reframe-hpc/reframe/blob/2d83a21f20dcb5aff3fcca5178c64869511ab2f8/docs/config_reference.rst?plain=1#L301 , is this where to add a reference to `pbspro` as a scheduler type?

PR #3613, opened 2026-01-21
[feat] Allow combining explicit system partitions and features/extras in `valid_systems`

casparvl:
This was requested in https://github.com/reframe-hpc/reframe/issues/2820 , although that was eventually closed with a suggestion for a workaround. 

However, I hit it again when trying to use `reframe.utils.find_modules(substr, ...)` in combination with wanting to specify additional features. Example use cases (and there's many more one can come up with):

- I may have a software X for which want to _not_ run CPU tests on a GPU node. I normally achieve that by assigning the `CPU` feature to my CPU partitions (but not to my GPU partitions) and then requesting the `CPU` feature from my CPU-only tests.
- I may have two variants of a bandwidth test (which needs module Y): one sending small packets, one sending larger packets. I only want to run larger one only on partitions that have the `infiniband` feature.

In both cases, we want to test _all_ modules with `X` or `Y` respectively that are available on our system, hence we want to use the `find_modules` function. However, the first element of the tuple this returns is the `system:partition` string for which it found the module matching the requested substring. This is clearly meant to be used in `valid_systems` directly (as is confirmed by the example [in the docs](https://reframe-hpc.readthedocs.io/en/stable/utility_functions_reference.html#reframe.utility.find_modules)), but for the examples above, you may want to only run _on a subset of these_ `system:partition` combinations (based on the required features. It would be very natural to then do:

```
module_info = parameter(find_modules('X'))
...
self.valid_systems - [module_info[0] + ' +feat1']
```

To make sure that a partition is only valid if it provides the right module _and_ the right feature. 

It turned out that the code changes needed are very limited (it seems like a lot because the indentation changed, but the key part is:
- not splitting the system:partition case from the feature/extra specification case
- Handling the system:partition case with an extra `elif` on the subspec:

```
            elif ':' in subspec and not subspec.startswith(('+', '-', '%')):
                # If there is a system:partition specified, make sure it
                # matches one of the items in valid_matches
                syspart_match = True if subspec in valid_matches else False
```

And adding that `syspart_match` to the return logic:

```
        # If the partition has all the plus features, none of the minus
        # all of the properties and the system:partition spec (if any)
        # matched, this partition is valid
        if (
            have_plus_feats and not have_minus_feats and have_props
            and syspart_match
        ):
            return True
```

Note that this logic does not assume the `system:partition` to strictly come before or after the features/extras, so I've adapted the valid syntax specification to allow both:

```
_VALID_SYS_SYNTAX = rf'^(({_FKV}\s+)*{_S}(\s+{_FKV})*|{_FKV}(\s+{_FKV})*)$'
```

If anything, I feel the code has become cleaner by not handling `system:partition` as a special, separate case, but as 'just another item' that may occur as a subspec.

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -691,6 +691,22 @@ class RegressionTest(RegressionTestPlugin, jsonext.JSONSerializable):
     #:
     #:    valid_systems = [r'+feat1 +feat2 %foo=1']
     #:
+    #: Features and key/value pairs can also be combined with an explicit
+    #: system partition combination in a single :attr:`valid_systems` entry. In
+    #: this case, the resulting constraint means that the requested system
+    #: partition combination is valid only if it also satisfies the feature and
+    #: key/value pair specification. In the following example, the ``sys:part``
+    #: system will only be selected if it also defines the ``foo`` feature:
+    #:
+    #: .. code-block:: python
+    #:
+    #:    valid_systems = [r'sys:part +foo']
+    #:
+    #: This case is generally useful in cases of parameterization of tests
+    #: based on configuration settings as well as in cases where additional
+    #: constraints can only be determined during the test initialization after
+    #: its parameterization.
+    #:
     #: Any partition/environment extra or
     #: :ref:`partition resource <scheduler-resources>` can be specified as a
     #: feature constraint without having to explicitly state this in the
@@ -730,6 +746,10 @@ class RegressionTest(RegressionTestPlugin, jsonext.JSONSerializable):
     #:
     #:  .. versionchanged:: 3.11.0
     #:     Extend syntax to support features and key/value pairs.
+    #:
+    #:   .. versionchanged:: 4.10
+    #:      Support for combining an explicit system partition combination with
+    #:      features and extras.
     valid_systems = variable(typ.List[typ.Str[_VALID_SYS_SYNTAX]])
 
     #: A detailed description of the test.


casparvl:
Hm, I have to recheck how the current `if...elif...elif` handles `sysname` `*` and `*:partname` cases. My bet is, it handles them incorrectly.

reframe/core/pipeline.py
```diff
@@ -80,7 +80,7 @@ def launch_command(self, stagedir):
 _VALID_ENV_SYNTAX = rf'^({_NW}|{_FKV}(\s+{_FKV})*)$'
 
 _S = rf'({_NW}(:{_NW})?)'   # system/partition
-_VALID_SYS_SYNTAX = rf'^({_S}|{_FKV}(\s+{_FKV})*)$'
+_VALID_SYS_SYNTAX = rf'^(({_FKV}\s+)*{_S}(\s+{_FKV})*|{_FKV}(\s+{_FKV})*)$'
```
victorusu: The PR would benefit from a rewrite of this line.
The system definition should not rely on `space`s.
All of these definitions should be valid
```
1. *
2. daint:gpu
3. daint:mc
4. daint:mc +gpu
5. daint:mc +gpu +uenv
6. daint:mc+gpu
7. daint:mc+gpu +uenv
8. daint:mc +gpu+uenv
9. daint:mc +gpu +uenv+modules
10. daint:mc +gpu+uenv +modules
```
In the current form only the first four cases are valid. 
The fifth case is miss interpreted, where one cannot see the `+gpu` entry.
And the last five case are not valid systems.

This implies that:
1. Tests won't run because of a `space`, without any hint. Users may lose a lot of time because of a missing `space`.
2. Potential introduction of silent bugs. The addition of a new feature makes the previous invalid.
3. Hinders the ability to write tests with partitions that have multiple features.

We should make this feature more robust to the end user and to the future.

What do you think?

casparvl: > The system definition should not rely on spaces.

Just to be clear: unless I'm mistaken, the _current_ version of ReFrame also requires spaces between features, right?

> All of these definitions should be valid

And same here: definitions 6-10 are not valid in the current reframe either?

Don't get me wrong: I'm not against allowing for non-space separated features. But I'm also not sure it should be handled in _this_ PR: allowing non-space seperated features is orthogonal to being able to combine the `sysname:partname` syntax with the `+feat` syntax. And I wouldn't want this PR to be delayed because of discussions on whether or not non-space separated features are desirable.

> The fifth case is miss interpreted, where one cannot see the +gpu entry.

That's strange, so you're saying that if you define:

```
daint:mc +gpu +uenv
```

You're effectively getting

```
daint:mc +uenv
```

? 

I haven't observed that, but I might have not tested this particular case. I'll try it out...

casparvl:
> I was about to press the "Merge button" but I realised we need to update the docs.

Oof, good point. I'm so used that for things like EESSI/EasyBuild the docs are in separate repo's and are typically updated asynchronously, that I didn't think about changing the docstrings :) Thanks for doing it!

smoors:
thanks a lot @vkarak !!!

PR #3615, opened 2026-01-28
[bugfix] Honor the `systems.partitions.sched_options.max_sacct_failures` configuration option for Slurm backends

ekouts:
The issue was present since the original PR: https://github.com/reframe-hpc/reframe/pull/3418 .
Since the comparison was wrong, Reframe would actually crash from the first retry without any warning message.


vkarak:
@ekouts Could you add a description in the PR with the behaviour that this bug was incurring? This is for future reference and to be able to provide a more accurate description in the release notes.

PR #3624, opened 2026-02-04
[feat] Add support for dynamically setting the authorization header in the `httpjson` log handler

### LINKED ISSUE #3622: Auth token in extra_headers expires on long running tests
I am using Azure Log Analytics to push test results to. [Example here](https://github.com/JimPaine/reframe-azure-perflog-handler/blob/main/config.py)

Like many storage services it uses a client crendential flow to obtain a short lived bearer token. 

Currently the logging handler is initialized at the start of the test session, but then not called until the end of the test session. When running longer tests this causes  the token to not be valued by the time it is used.

JimPaine:
Enables the use of short lived tokens where http services use flows such as `client credentials` by allowing the authorization header to be set dynamically via a callable which is evaluated at the point the http request is made.

Closes #3622

### DIFF
--- a/reframe/core/logging.py
+++ b/reframe/core/logging.py
@@ -711,7 +711,7 @@ class HTTPJSONHandler(logging.Handler):
 
         if not is_trivially_callable(self._json_format, non_def_args=3):
             raise ConfigError(
-                "httpjson: 'json_formatter' has not the right signature: "
+                "httpjson: 'json_formatter' has the wrong signature: "
                 "it must be 'json_formatter(record, extras, ignore_keys)'"
             )
 
--- a/unittests/test_logging.py
+++ b/unittests/test_logging.py
@@ -560,7 +560,7 @@ def test_httpjson_handler_no_port(make_exec_ctx, config_file,
 
 def test_httpjson_auth_not_callable_error():
     with pytest.raises(ConfigError,
-                       match=r'authorization_header.* is not a callable'):
+                       match=r'authorization_header.*has the wrong signature'):
         rlog.HTTPJSONHandler(url='http://xyz/rfm',
                              authorization_header='NOT CALLABLE')
 


reframe/core/logging.py
```diff
@@ -711,6 +712,11 @@ def __init__(self, url, extras=None, ignore_keys=None,
                 "it must be 'json_formatter(record, extras, ignore_keys)'"
             )
 
+        if authorization_header is not None and not callable(authorization_header):
```
victorusu: I like this idea of having a dedicated `authorization_header` and I have the impression one needs it because  one cannot pass a callable to the `extra_headers`. Have you considered passing a callable to `extra_headers` instead of having the dedicated `authorization_header`?

docs/config_reference.rst
```diff
@@ -1719,6 +1722,27 @@ This handler transmits the whole log record, meaning that all the information wi
    .. versionadded:: 4.1
 
 
+.. py:attribute:: logging.handlers_perflog..httpjson..authorization_header
+
+   :required: No
+   :default: :obj:``None``
```
vkarak:
```suggestion
   :default: :obj:`None`
```

reframe/core/logging.py
```diff
@@ -596,6 +596,7 @@ def _create_httpjson_handler(site_config, config_prefix):
     extras = site_config.get(f'{config_prefix}/extras')
     ignore_keys = site_config.get(f'{config_prefix}/ignore_keys')
     json_formatter = site_config.get(f'{config_prefix}/json_formatter')
+    authorization_header = site_config.get(f'{config_prefix}/authorization_header')
```
vkarak:
```suggestion
    authorization_header = site_config.get(
        f'{config_prefix}/authorization_header'
    )
```

unittests/test_logging.py
```diff
@@ -556,3 +556,46 @@ def test_httpjson_handler_no_port(make_exec_ctx, config_file,
...
+
+    handler = rlog.HTTPJSONHandler(
+        url='http://xyz/rfm', authorization_header=mock_get_token)
+
+    yield handler
+    handler.close()
+
+
+def test_httpjson_auth_header_set(httpjson_handler,
+                                  record_with_check_tags, mock_requests_post_200):
```
vkarak:
```suggestion
def test_httpjson_auth_header_set(httpjson_handler,
                                  record_with_check_tags,
                                  mock_requests_post_200):
```

vkarak (changes requested):
Lgtm, I have just minor style comments.

PR #3628, opened 2026-02-13
[doc] Add import statement for xfail in tutorial

gppezzi:
I suggest to mention in the tutorial whether the `import xfail` is needed or not.

Question came while discussing this [PR](https://github.com/eth-cscs/cscs-reframe-tests/pull/519).

I read somewhere that builtins do not require import, but apparently it is needed for `xfail()` and it may cause confusion for such a trivial issue.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -1531,6 +1531,8 @@ Here is an example:
 
 .. code-block:: python
 
+   from reframe.core.builtins import xfail
+
    reference = {
       'tutorialsys': {
          'copy_bw':  xfail('demo fail', (100_000, -0.1, 0.1, 'MB/s')),


victorusu (approved):
`lgtm`

vkarak:
> It is strange that it requires import, I'll try to reproduce.

I've just tried adding an xfail to a reference in the tutorial and the import was not required as expected. It is required though whey you would like to decorate a test to mark an expected failure, which is also expected. But this "xfail" is not a builtin it's in the decorators and simply doing `@rfm.xfail` should work.

vkarak (changes requested):
Now I've seen the PR where this has happened. This was expected, as well as my example's behaviour (adding xfail in the reference definition in the class body). This is what is happening:

- Builtin names are "imported" in the test class' namespace automatically during the test class construction. That's why they are all available in the class body and an explicit import is not required.
- When you try to use them (unprefixed) inside a test method (aka hook), which was the case of your PR, then Python will think the name refers to a local variable and will fail. In this case, using `self.xfail` to refer to the already imported name at the class level or using an explicit import as you did will work.

This is a more general behaviour of the builtins, so I think we should mention this in the corresponding docs for the builtins, in general. There, I think we should mention that it is advisable to always import the builtins, since this helps IDEs too. When the test variables and parameters were initially introduced, their implementation was not so cleanly separated as it is now with the `builtins` package, so there was no way to actually import a builtin.

PR #3629, opened 2026-02-23
[feat] Allow git sparse checkouts and passing arbitrary git options when cloning a repo from `sourcesdir`

### LINKED ISSUE #3053: Consider using depth of 1 when automatically cloning a git repo
This would make cloning much faster

### LINKED ISSUE #3627: only clone a subdirectory of a git repo
Can ReFrame support sparse git clones ?
- https://stackoverflow.com/questions/600079/how-do-i-clone-a-subdirectory-only-of-a-git-repository

For now, I do it with a hook (I could also use a fixture but that's not the point):

```python
import os
import reframe as rfm


@rfm.simple_test
class git_clone_test(rfm.CompileOnlyRegressionTest):
    repo = variable(
        str, value='https://github.com/eth-cscs/alps-gh200-reproducers.git')
    sourcesdir = None
    build_system = 'SingleSource'
    sourcepath = 'hello.c'
    valid_systems = ['*']
    valid_prog_environs = ['*']

    @run_before('compile')
    def prepare_clone(self):
        self.prebuild_cmds = [
            f'git clone --no-checkout --depth=1 {self.repo}',
            'cd alps-gh200-reproducers',
            'git config core.sparseCheckout true',
            'echo intranode-pinned-host-comm > .git/info/sparse-checkout',
            'git checkout main',
            'cd ..'
        ]

    @sanity_function
    def validate_test(self):
        return sn.assert_not_found(r'error', self.stdout)
```

Ideally, I'd like to use:

```
    sourcesdir = 'https://github.com/eth-cscs/alps-gh200-reproducers.git:intranode-pinned-host-comm'
```

victorusu:
This expands the `sourcesdir` syntax to include a dictionary format. In this case, the git `url` is passed via the `url` key.
Additional command options can be added via the `opts` key. This capability was previously available in the `git_clone` function, but not accessible to the `sourcesdir` git clone interface.

The `git sparse-checkout` path is available when one defines the `files` keys with a list of files to checkout.
In this case, instead of emitting a `git clone ...`  command the framework will emit the following set of commands
```
git clone --no-checkout --depth=1 ...
git sparse-checkout set --no-cone ${file list}
git checkout
```
where, `${file list}` is the list of files defined in the `files` list.

An example test is this.
```python
@rfm.simple_test
class git_clone_test(rfm.CompileOnlyRegressionTest):
    sourcesdir = {
         'url': 'https://github.com/eth-cscs/alps-gh200-reproducers.git',
         'files' : ['intranode-pinned-host-comm'],
         'opts' : ['--no-optional-locks', '--no-advice', '--no-pager']
    }
    build_system = 'SingleSource'
    sourcepath = 'intranode-pinned-host-comm/intranode_pinned_host_comm.cpp'
    valid_systems = ['*']
    valid_prog_environs = ['*']
```

Please note that I haven't added any unit tests because the framework does not include any unit tests for the original `git_clone` function.

closes #3627
closes #3053

### DIFF
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -2565,11 +2565,9 @@ class RegressionTest(RegressionTestPlugin, jsonext.JSONSerializable):
 
             if self._requires_stagedir_contents():
                 srcdir = self.sourcesdir
-                if not srcdir:
-                    return
                 if isinstance(srcdir, dict):
                     if 'url' not in srcdir:
-                        raise ReframeError(f'The {srcdir} misses the url key')
+                        raise ReframeError(f'{srcdir} misses the url key')
 
                     url = srcdir['url']
                     if not osext.is_url(url):
--- a/reframe/core/pipeline.py
+++ b/reframe/core/pipeline.py
@@ -2573,8 +2573,9 @@ class RegressionTest(RegressionTestPlugin, jsonext.JSONSerializable):
 
                     url = srcdir['url']
                     if not osext.is_url(url):
-                        raise ReframeError(f'The {srcdir} syntax only supports '
+                        raise ReframeError(f'The dictionary syntax only supports '
                                            'git repositories')
+
                     self._clone_to_stagedir(url,
                                             files=srcdir[url]['files'] if 'files'
                                                   in srcdir[url] else None,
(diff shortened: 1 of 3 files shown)


vkarak (changes requested):

vkarak (approved):
@victorusu I have updated and rebased your PR. Can you give it a try? I will likely need to add another unit test to increase coverage, but other than that lgtm now.

vkarak:
Something was messed with the rebase.

PR #3642, opened 2026-03-25
[feat] Remove unnecessary Slurm polling when retrieving job pending reason

vkarak:
This PR improves the polling of jobs for the pending reason.

This is now done in a single command for all pending jobs. Two knobs are also exposed to users now as configuration options and environment variables:

1. `slurm_job_cancel_reasons`: This is a list of pending reasons that reframe will check and will cancel the job proactively.
2. `slurm_pending_job_reason_poll_freq`: This controls the frequency that pending jobs will be polled for their pending reasons (valid only for `slurm` backend).

Closes #3640.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -455,11 +455,13 @@ System Partition Configuration
 
    Ignore the ``ReqNodeNotAvail`` Slurm state.
 
-   If a job associated to a test is in pending state with the Slurm reason ``ReqNodeNotAvail`` and a list of unavailable nodes is also specified, ReFrame will check the status of the nodes and, if all of them are indeed down, it will cancel the job.
-   Sometimes, however, when Slurm's backfill algorithm takes too long to compute, Slurm will set the pending reason to ``ReqNodeNotAvail`` and mark all system nodes as unavailable, causing ReFrame to kill the job.
+   If a job associated to a test is in pending state with the Slurm reason ``ReqNodeNotAvail``, ReFrame will cancel the job.
+   Sometimes, however, when Slurm's backfill algorithm takes too long to compute, Slurm will set the pending reason to ``ReqNodeNotAvail`` and mark all system nodes as unavailable, causing ReFrame to its jobs.
    In such cases, you may set this parameter to :obj:`True` to avoid this.
 
    This option is relevant for the Slurm backends only.
+
+   .. seealso:: :attr:`~config.systems.partitions.sched_options.slurm_job_cancel_reasons`
 
 .. py:attribute:: systems.partitions.sched_options.job_submit_timeout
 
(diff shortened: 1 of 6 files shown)


docs/config_reference.rst
```diff
@@ -438,12 +438,14 @@ System Partition Configuration
 
    Ignore the ``ReqNodeNotAvail`` Slurm state.
 
-   If a job associated to a test is in pending state with the Slurm reason ``ReqNodeNotAvail`` and a list of unavailable nodes is also specified, ReFrame will check the status of the nodes and, if all of them are indeed down, it will cancel the job.
-   Sometimes, however, when Slurm's backfill algorithm takes too long to compute, Slurm will set the pending reason to ``ReqNodeNotAvail`` and mark all system nodes as unavailable, causing ReFrame to kill the job.
+   If a job associated to a test is in pending state with the Slurm reason ``ReqNodeNotAvail``, ReFrame will cancel the job.
+   Sometimes, however, when Slurm's backfill algorithm takes too long to compute, Slurm will set the pending reason to ``ReqNodeNotAvail`` and mark all system nodes as unavailable, causing ReFrame to its jobs.
    In such cases, you may set this parameter to :obj:`True` to avoid this.
```
jack-morrison, line 443: Typo "causing ReFrame to ___ its jobs"

```suggestion
   Sometimes Slurm’s backfill scheduling can take so long to compute that it temporarily sets the pending reason to ``ReqNodeNotAvail`` and reports all nodes as unavailable. ReFrame may interpret this condition and cancel its jobs. If you encounter this scenario, set this parameter to :obj:`True` to prevent ReFrame from killing jobs with the ``ReqNodeNotAvail`` reason.
```

docs/config_reference.rst
```diff
@@ -490,6 +492,45 @@ System Partition Configuration
...
 
 
+.. py:attribute:: systems.partitions.sched_options.slurm_job_cancel_reasons
+
+   :required: No
+   :default: ``["ReqNodeNotAvail"]``
+
+   Reasons to proactively cancel a pending Slurm job.
+
+   If a job associated to a test is in pending state with one of the reasons listed here, ReFrame will cancel the job.
```
jack-morrison, line 502:
```suggestion
   If a job associated to a test is in pending state with one of the reasons listed, ReFrame will cancel the job.
```

docs/config_reference.rst
```diff
@@ -490,6 +492,45 @@ System Partition Configuration
...
+
+   This option is relevant for the Slurm backends only.
```
jack-morrison, line 504:
```suggestion
   This option is relevant for the Slurm backend only.
```

reframe/core/schedulers/slurm.py
```diff
@@ -125,24 +117,19 @@ def __init__(self):
...
         # intervention is required)
-        self._cancel_reasons = ['FrontEndDown',
-                                'Licenses',         # May require sysadmin
-                                'NodeDown',
-                                'PartitionDown',
-                                'PartitionInactive',
-                                'PartitionNodeLimit',
-                                'QOSJobLimit',
-                                'QOSResourceLimit',
-                                'QOSUsageThreshold']
```
jack-morrison, line 136: Are we dropping these default cancel reasons? Should they be made a part of `slurm_job_cancel_reasons`?

reframe/core/schedulers/slurm.py
```diff
@@ -554,74 +538,72 @@ def poll(self, *jobs):
...
+            for jobid, reason in reasons.items():
+                try:
+                    pending_reasons[pending_jobs[jobid]] = reason
+                except KeyError:
+                    # Job listed in reasons is not pending
+                    continue
+        else:
+            job_ids = ",".join(pending_jobs.keys())
+            completed = osext.run_command(f'squeue -h -j {job_ids} -o "%A|%r"')
+            pending_reasons = {}
```
jack-morrison, line 580: Recommend moving assignment

```
pending_reasons = {}
```

above `if reasons:` block to avoid `NameError`s in the try block.

jack-morrison (changes requested):

docs/config_reference.rst
vkarak: I would keep the plural here, because there are two Slurm backends: `slurm` and `squeue`.

reframe/core/schedulers/slurm.py
vkarak: I thought about it, but this list is not so accurate and to maintain it, we would have to deal with detailed Slurm semantics. For example, the `Licenses` reason is not a cancel reason in reality. Previously, we were poking Slurm to check if the reported nodes in the reason were indeed unavailable, which was not so efficient. That's why, I'm only included `ReqNodeNotAvail` reason as a cancel, which is indeed one, and leave any additional reason at user discretion.

docs/config_reference.rst
vkarak: I applied these change exactly, except that I put every sentence in a separate line.

jack-morrison (approved):

PR #3651, opened 2026-04-08
[feat] Modernize ReFrame installation

### LINKED ISSUE #3588: Consider migrating to the `uv` package manager

vkarak:
ReFrame has migrated to `pyproject.toml` and is now PEP517/518 compliant.

- All the legacy installation files are removed: `setup.cfg`, `requirements.txt`, `docs/requirements.txt`.
- The `bin/reframe` is removed. This is managed by `pyproject.toml`.
- `./bootstrap.sh` is removed. Users can run ReFrame using ephemeral venv on multiple architectures using `uvx run reframe`.
- Remove use of `./boostrap.sh` from everywhere in the code and unit tests
- Update container images
- Update docs
- Extra denendencies and dependency groups, such as for running the unit tests and building the documentation, are now handled by `pyproject.toml`.
- Shell completions and man pages are stored in their OS standard locations and is handled by the `pyproject.toml`.
- README is updated.
- Remove support of Python 3.9 and add support of Python 3.14

Closes #3588.


victorusu (commented):
This PR also removes the python 3.9 installation

victorusu (approved):
`lgtm`

vkarak:
> This PR also removes the python 3.9 installation

Yes, because now, if you use `uv`, it will pull in the needed Python version, so we should not care so much about the default OS Python version. We should follow the Python lifecycle.

PR #3657, opened 2026-04-25
[ci] Update `.readthedocs.yaml` to use `uv` for building the docs

vkarak:
This is a follow up of #3651 as the docs fail to build right now.


victorusu (approved):
`lgtm`

PR #3658, opened 2026-04-26
[bugfix] Treat `.` properly in external references keys

vkarak:
Using `.` internal for denoting the key levels was problematic for keys that were regexes.
We no more use `.` as an internal separator of key levels; we represent them internally as a list.
We use `.` to concatenate them for error reporting only.

This is a follow up of #3630.


victorusu (approved):
`lgtm`

PR #3667, opened 2026-05-15
[bugfix] Treat properly Slurm constraints with `.`

Guy-Coates:
The reframe slurm node constraint validator rejects features with a "." in the name (eg nodes with `Feature=c5.xlarge` defined in slurm).

 This results in Reframe generating no tests when attempting to target nodes of  that type using `--constraint=c5.xlarge` in the `systems.partitions.access` config.

  This PR relaxes the validation/tokenisation regexes to allow ".".

### DIFF
--- a/reframe/core/schedulers/slurm.py
+++ b/reframe/core/schedulers/slurm.py
@@ -777,11 +777,11 @@ class _SlurmNode(sched.Node):
         # their combinations; since Slurm features are not valid Python
         # identifiers, we replace them with booleans before evaluating.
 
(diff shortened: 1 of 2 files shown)


vkarak (changes requested):
Thanks for the fix! Lgtm. Since we are not going to make another patch release for 4.9 (4.10 is around the corner), we'd need to rebase this PR on `develop`.

vkarak (approved):
Rebased PR on develop.

PR #3671, opened 2026-05-20
[feat] New option `--warn-as-error` for treating warnings as errors

### LINKED ISSUE #3598: Warnings-as-errors or similar
Sometimes tests get excluded with a warning because of issues with the tests. Some examples we've encountered recently:

- Name conflicts (`WARNING: skipping test file 'a.py': name conflict error: test 'Foo' from 'a.py' is already defined in 'b.py'`)
- Version mismatch (`WARNING: skipping test file 'a.py': attribute error: module 'reframe' has no attribute 'xfail'`)
- Other test test issues (`WARNING: skipping test file 'a.py': permission error: [Errno 13] Permission denied: '/a/b/c'`; the test setup makes assumptions about the system)

Reframe warns about these which is great. However, in automated testing (e.g. CI) none of these are actually noticed without manually looking at logs because the tests are silently ignored instead of causing a failure.

As far as I could tell, there's no way to ask for this currently when running tests. Did I miss it or would it be possible to add such a mode (`--warnings-as-errors`)?

What I'm thinking of is perhaps not failing immediately on warnings (i.e. still run the tests that are discovered) but exit at the end with an error, even if all tests were otherwise successful.

Similarly, if no tests are found it'd be great to have a way to ask reframe to make that a failure to avoid silently having no tests run.

vkarak:
Closes #3598.

### DIFF
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -1254,6 +1254,14 @@ Miscellaneous options
 
    This option can also be set using the :envvar:`RFM_VERBOSE` environment variable or the :attr:`~config.general.verbose` general configuration parameter.
 
+.. option:: --warn-as-error
+
+   Treat warnings as errors.
+
+   This option can also be set using the :envvar:`RFM_WARN_AS_ERROR` environment variable.
+
+   .. versionadded:: 4.10
+
 
 .. _test_naming_scheme:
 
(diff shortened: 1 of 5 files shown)


ekouts (approved):
LGTM!

PR #3675, opened 2026-05-28
[bugfix] Install uv before building dist packages in publish workflow

ekouts:


vkarak (approved):

vkarak (changes requested):
I think we should do something like this:

https://docs.astral.sh/uv/guides/integration/github/

vkarak:
> I think we should do something like this:
> 
> https://docs.astral.sh/uv/guides/integration/github/

https://docs.astral.sh/uv/guides/integration/github/#publishing-to-pypi

vkarak (approved):

vkarak:
Approving this PR as we do already most of the things in the uv example.

PR #3677, opened 2026-06-04
[doc] Fix example with `authorization_header` in the tutorial

sschmittgwdg:
The call to the authorization header function is done in the implementation at

https://github.com/reframe-hpc/reframe/blob/c6a4b9b4a9b48dec0953bbe4222269f293f1b7b2/reframe/core/logging.py#L754

Therefore, `()` must be absent from the configuration file.


vkarak:
Thanks for the fix; could you rebase this PR against `master`, so as to release this fix in 4.10.1?

vkarak (changes requested):
Simply changing the target branch is not enough. I've updated the contribution guide with instructions on how to deal with such scenarios:

https://github.com/reframe-hpc/reframe/wiki/Contributing-to-ReFrame#targeting-the-wrong-branch

vkarak (approved):

PR #3678, opened 2026-06-04
Split only once to allow for colons in partition name

sschmittgwdg:


vkarak:
I am bit reluctant accepting this PR: `:` is the separator of a system and partition throughout the framework. I would instead suggest that the framework should deny partitions with `:` and avoid crashes. In your case, why is `:` important in the name of the partition?

sschmittgwdg:
It's not important. I started having ReFrame's partitions named as our slurm partitions and the latter had a `:`. Denying `:` in ReFrame's partitions is fine for me.

vkarak:
Closing this and I opened an issue: https://github.com/reframe-hpc/reframe/issues/3681

PR #3679, opened 2026-06-05
[feat] Allow not using hashes in stage and output dirs

### LINKED ISSUE #2670: Add symlinks to the hashed dir with the test's human readable name
This will make much easier the navigation inside the stage, output and perflog dirs. Note that the hashed named will be used internally, as the special characters of the human readable name could make some tools unhappy.

vkarak:
- New configuration option `general.stagedir_hashes` and the associated environment variable `RFM_STAGEDIR_HASHES` are introduced.

Closes #2670.

NOTE:

Conversely to the suggestion of symlinks in #2670, this PR introduces an on/off toggle to switch on or off the hashes (default is on). The reason is that if we had symlinks additionally to the hashes, then a recursive directory listing would list twice the same files.


reframe/core/pipeline.py
```diff
@@ -2407,14 +2407,19 @@ def _setup_paths(self):
         self.logger.debug('Setting up test paths')
         try:
             runtime = rt.runtime()
+            if runtime.get_option('general/0/stagedir_hashes'):
+                test_name = self.short_name
+            else:
+                test_name = self.display_name.replace(' ', '')
```
victorusu: would you consider '_', instead of joining the words?
It seems to me that it would be a bit more readable.

victorusu (approved):
`lgtm`

reframe/core/pipeline.py
vkarak: No, because this is primarily meant to "compact" the parameters and it is consistent with the `-n` option behaviour where you select tests with `MyTest%param=1` instead of `MyTest %param=1`. But actually I could be more precise and do `replace(' %', '%')` instead.

vkarak: Updated it in https://github.com/reframe-hpc/reframe/pull/3679/commits/3cad064ee4dabe1df6410a64b82acd88cffc5f11.

PR #3680, opened 2026-06-10
[bugfix] Record the job nodelist in the SSH scheduler backend

0nebody:
Set the nodelist for a job using the SSH scheduler. This fixes an issue with job_nodelist being empty in the performance report for this scheduler.

### DIFF
--- a/reframe/core/schedulers/ssh.py
+++ b/reframe/core/schedulers/ssh.py
@@ -145,6 +145,7 @@ class SSHJobScheduler(JobScheduler):
         job._submit_time = time.time()
         job._ssh_options = stripped_opts
         job._host = self._reserve_host(host)
+        job._nodelist = [job._host]
 
         self._push_artefacts(job)
         self._do_submit(job)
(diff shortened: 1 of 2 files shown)


reframe/core/schedulers/ssh.py
```diff
@@ -145,6 +145,7 @@ def submit(self, job):
         job._submit_time = time.time()
         job._ssh_options = stripped_opts
         job._host = self._reserve_host(host)
+        job._nodelist = [job._host]
```
victorusu: I think this is very good approach! 
`lgtm`

victorusu (approved):

reframe/core/schedulers/ssh.py
vkarak: We will also need to respect the `unqualified_hostnames` configuration option as in the local scheduler:

https://github.com/reframe-hpc/reframe/blob/c6a4b9b4a9b48dec0953bbe4222269f293f1b7b2/reframe/core/schedulers/local.py#L75-L79

vkarak (changes requested):
Also could you rebase your PR against `master` so that we schedule this fix for 4.10.1? Check the contribution guide here:

https://github.com/reframe-hpc/reframe/wiki/contributing-to-reframe#targeting-the-wrong-branch

reframe/core/schedulers/ssh.py
0nebody: Thanks, I have updated with this logic as well as the configuration reference docs.

vkarak (approved):
Lgtm, thanks!

PR #3684, opened 2026-06-16
[packaging] Make polars installation optional

### LINKED ISSUE #3676: consider making polars an optional dependency
The size on disk of a python venv for reframe has ballooned from 54M to 264M between reframe 4.9.3 and 4.10.0, largely due to the addition of polars as a dependency. It'd be nice to keep the runtime requirements minimal so that reframe can be included in minimal base system images or containers without impacting overall size too significantly.

```console
> du -csh reframe-env/lib/python3.11/site-packages/* | sort -h | tail -n11
1.2M	reframe-env/lib/python3.11/site-packages/jsonschema
1.3M	reframe-env/lib/python3.11/site-packages/jinja2
2.2M	reframe-env/lib/python3.11/site-packages/ClusterShell
2.7M	reframe-env/lib/python3.11/site-packages/reframe
3.1M	reframe-env/lib/python3.11/site-packages/yaml
9.3M	reframe-env/lib/python3.11/site-packages/polars
11M     reframe-env/lib/python3.11/site-packages/setuptools
12M     reframe-env/lib/python3.11/site-packages/lxml
13M     reframe-env/lib/python3.11/site-packages/pip
202M	reframe-env/lib/python3.11/site-packages/_polars_runtime_32
264M	total
```

vkarak:
This PR introduces an extra called `no-analytics` that if passed, it will skip the `polars` dependency. However, this dependency is needed if `RFM_ENABLE_RESULTS_STORAGE` is set. We expand the validation check in the CLI for this. Also to avoid failures in lightweight installations, polars is only imported in the functions that is needed.

Finally, this PR introduces also documentation improvements.

Future note: `polars` will likely be completely removed from dependencies.

Closes #3676.

### DIFF
--- a/ci-scripts/dockerfiles/reframe-envmodules.dockerfile
+++ b/ci-scripts/dockerfiles/reframe-envmodules.dockerfile
@@ -29,4 +29,4 @@ RUN uv sync --group dev && \
     echo ". $BASH_ENV" >> /home/rfmuser/.profile
 ENV BASH_ENV=/home/rfmuser/.profile
 
-CMD ["/bin/bash", "-c", "uv run coverage run --source=reframe ./test_reframe.py --rfm-user-config=ci-scripts/configs/envmod.py; uv run coverage xml -o coverage.xml"]
+CMD ["/bin/bash", "-c", "uv run coverage run --source=reframe ./test_reframe.py --rfm-user-config=ci-scripts/configs/envmod.py; rv=$?; uv run coverage xml -o coverage.xml; exit $rv"]
--- a/README.md
+++ b/README.md
@@ -38,6 +38,27 @@ Please visit the project's documentation [page](https://reframe-hpc.readthedocs.
 
 ReFrame is fairly easy to install as PyPI package.
 
+### Installing using `pip`
+
+```bash
+pip install reframe-hpc
+```
+
+This will also install the manpages and the shell completions under your pip's virtual environment:
+
+| | |
+|---|---|
+| `reframe` executable | `${VIRTUAL_ENV}/bin` |
+| Man pages | `${VIRTUAL_ENV}/share/man` |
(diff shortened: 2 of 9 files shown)


vkarak:
@dmargala With this you will be able to `pip install reframe-hpc[no-analytics]` and it will not install `polars`.

vkarak:
Converting it back to draft; trying to understand why the CI doesn't pick `polars` as a dependency although it should do it by default.

vkarak:
It seems the `uv sync --group dev` done to run unit tests removes `polars`, thus the errors.

ci-scripts/dockerfiles/reframe-envmodules.dockerfile
```diff
@@ -29,4 +29,4 @@ RUN uv sync --group dev && \
     echo ". $BASH_ENV" >> /home/rfmuser/.profile
 ENV BASH_ENV=/home/rfmuser/.profile
 
-CMD ["/bin/bash", "-c", "uv run coverage run --source=reframe ./test_reframe.py --rfm-user-config=ci-scripts/configs/envmod.py; uv run coverage xml -o coverage.xml"]
+CMD ["/bin/bash", "-c", "uv run coverage run --source=reframe ./test_reframe.py --rfm-user-config=ci-scripts/configs/envmod.py; rv=$?; uv run coverage xml -o coverage.xml; exit $rv"]
```
victorusu, line 32: This is interesting

victorusu (approved):
`lgtm`

ci-scripts/dockerfiles/reframe-envmodules.dockerfile
vkarak: Yes, because, unit tests could fail and we could miss them :) The reason of not using `&&` is because we still want the coverage to be uploaded, even if a unit test has failed.

PR #3687, opened 2026-06-22
[packaging] Generate a PyPI-friendly README before publishing the package

### LINKED ISSUE #3683: Use (again) a PyPI-compliant README for the published dist package
This was removed with #3651 but it turns out the [PyPI does not like](https://pypi.org/project/reframe-hpc/) the current `README.md` which switches logos based on the background. That was the reason for the original `README_minimal.md`. However maintaining a copy of the `README.md` is a maintenance burden, so we should probably generate/fix the PyPI README on-the-fly before publishing.

vkarak:
Also in this PR:

- The links and references to Twitter are removed.
- The `README_minimal.md` is removed.

Closes #3683.


ekouts (approved):

victorusu (approved):
`lgtm`

PR #3688, opened 2026-06-23
Demonstrate how to list individual test results

sschmittgwdg:
Demonstrate how to list individual test results by artificially aggregating over a unique field. Method courtesy of @vkarak.

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -2275,7 +2275,8 @@ You can also combine :option:`--list-stored-testcases` and :option:`--describe-s
 
    reframe --list-stored-testcases=now-1d:now/mean:/ -n stream_test% -E 'num_threads == 2'
 
-To list individual results from past tests, artificially aggregate on a unique field, e.g., `uuid`:
+To list individual results from past tests, you can aggregate on a unique field, e.g., `uuid`.
+Since the value is unique, the aggregation will just return the value:
 
 .. code-block:: bash
    :caption: Run in the single-node container.
@@ -2292,7 +2293,6 @@ To list individual results from past tests, artificially aggregate on a unique f
    stream_test  generic:default+builtin  copy_bw|triad_bw  MB/s           28539.7
    stream_test  generic:default+builtin  copy_bw|triad_bw  MB/s           28755.4
 
-Note that although `pval` is annotated with the aggregation function `mean` in the output, it is not applied to the variable.
 
 Comparing performance of test cases
 -----------------------------------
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -2275,6 +2275,24 @@ You can also combine :option:`--list-stored-testcases` and :option:`--describe-s
...
+
+   name         sysenv                   pvar              punit      pval (mean)
+   -----------  -----------------------  ----------------  -------  -------------
+   stream_test  generic:default+builtin  copy_bw|triad_bw  MB/s           28583.8
+   stream_test  generic:default+builtin  copy_bw|triad_bw  MB/s           28281.3
+   ...
+   stream_test  generic:default+builtin  copy_bw|triad_bw  MB/s           28539.7
+   stream_test  generic:default+builtin  copy_bw|triad_bw  MB/s           28755.4
+
+Note that although `pval` is annotated with the aggregation function `mean` in the output, it is not applied to the variable.
```
vkarak: What do you mean with this? 🤔

vkarak (changes requested):
We've recently made DCO commits required. Would you mind re-doing your commit with `git commit -s` or adding the following line in your commit:

```
    Signed-off-by: Your Name <your@email.com>
```

docs/tutorial.rst
sschmittgwdg: The comment tried to explain that although "mean" is in the title of the results column, the value itself is from a single run. The mean of a single value is of course the value itself, however, I think it could confuse users as the goal of the exercise is to _not_ do any aggregation but list the result of a single run.

vkarak: I see, I figured you meant probably like this. Technically, the aggregation happens, but since it's only a single element, it will always be the value. Any aggregation function should work in this case. You could remove this note and maybe expand the sentence before it.

docs/tutorial.rst
```diff
@@ -2275,6 +2275,24 @@ You can also combine :option:`--list-stored-testcases` and :option:`--describe-s
 
    reframe --list-stored-testcases=now-1d:now/mean:/ -n stream_test% -E 'num_threads == 2'
 
+To list individual results from past tests, artificially aggregate on a unique field, e.g., `uuid`:
```
vkarak:
```suggestion
To list individual results from past tests, you can aggregate on a unique field, e.g., `uuid`.
Since the value is unique, the aggregation will just return the value:
```

vkarak:
Lgtm, only thing remaining is signing-off your commits (see instructions here: https://github.com/reframe-hpc/reframe/pull/3688/checks?check_run_id=83194318291)

vkarak (approved):

PR #3691, opened 2026-06-25
[feat] Add a new `count` aggregation for querying past results

sschmittgwdg:


unittests/test_reporting.py
```diff
@@ -289,6 +290,8 @@ def test_parse_cmp_spec_aggregations(aggregator):
         assert agg['pval (p95)'][0] == 10
     elif aggregator == 'p05':
         assert agg['pval (p99)'][0] == 10
+    elif aggregator == 'count':
+        assert 'pval (count)' in agg.columns
```
vkarak, line 294: You will need to assert also the value here as with the other aggregations above.

vkarak (changes requested):
We will also need to update the docs.

unittests/test_reporting.py
sschmittgwdg: Thanks and done.

sschmittgwdg:
I've added `count` to `manpage.rst`. Do you want to see it documented somewhere else in addition?

vkarak (approved):

vkarak:
@sschmittgwdg Could you please sign-off your last commit so that we can merge the PR?

PR #3697, opened 2026-07-08
[bugfix] Fix validation of partition/environment features in configuration

### LINKED ISSUE #3696: ReFrame 4.10.1 does not allow features to have `-`
This was introduced with #3685 that enforced strict validation of system/partition/environment names. Features were (wrongly) using the `alphanum_string` validator instead of the `alphanum_ext_string` and so now features with hyphens are rejected. The configuration schema must be updated.

vkarak:
Closes #3696. 

I have also adapted the unit tests so as to expose this bug.


ekouts (approved):
lgtm

PR #3702, opened 2026-07-16
[feat] Extend the JUnit report

toxa81:
This PR is adding extra information to the <failure/> element of JUnit report. In particular, it adds the following text information:
- working directory
- last 20 lines of stdout
- last 20 lines of stderr

Use case: get useful information quickly while browsing the test matrix in the gitlab UI.


reframe/frontend/reporting/__init__.py
```diff
@@ -172,6 +172,14 @@ def _do_restore(self, testcase):
                 f'could not restore testcase {testcase!r}') from e
 
 
+def _tail_file(filename, num_lines):
+    try:
+        with open(filename, encoding='utf-8', errors='replace') as fp:
+            return ''.join(fp.readlines()[-num_lines:])
+    except OSError:
+        return ''
```
vkarak, line 180: This is not needed: you can use the `osext.tail()` function.

vkarak (commented):
I'm not very much in favor of a JUnit report diverging from the JSON report in the information it provides. The JSON report should contain this information and JUnit report should simply use it. 

I'm working already on updating the information reported in the the JSON report so I could include this in an upcoming PR. Also it's safer to use a byte threshold for tailing instead of just the lines.

reframe/frontend/reporting/__init__.py
```diff
@@ -539,7 +547,28 @@ def generate_xml_report(self):
...
+                            stderr = _tail_file(
+                                os.path.join(workdir, job_stderr), 20
+                            )
+
+                    testcase_msg.text = '\n\n'.join([
+                        f"{tc['fail_phase']}: {fail_reason}",
+                        workdir,
+                        stdout,
+                        stderr
+                    ])
```
vkarak, line 571: @toxa81 I'm working right now on extending the JSON report and the last lines of stdout/stderr are now included in case of errors. The idea is that the JUnit report will use this info directly. If I understand correctly you want include the stdout/stderr in the error message, right? I was wondering if there is dedicated field in the JUnit schema for logging stdout/stderr 🤔

vkarak: I think the standard convention is to include it special elements, which is what we should do:

https://github.com/testmoapp/junitxml#properties-in-test-output

vkarak:
Closing this PR as it is superseded by #3722. @toxa81 That PR emits the stdout/stderr in case of failed tests and it should be interacting well with Gitlab. Please test it if you can.

PR #3704, opened 2026-07-22
schedulers: add queue-wait-free Job.start_time property

raffenet:
## Problem

`Job.submit_time` and `Job.completion_time` are currently the only in-memory
timestamps exposed on a `Job`. Both span the job's full lifetime including
any time spent `PENDING` in the scheduler queue, so
`completion_time - submit_time` conflates queue wait with actual execution
time.

This showed up as a real false alarm in our CI: a test that queued for
~5 hours on a busy partition but only *executed* for 2 seconds was flagged
as an 18000+ second "runtime" by monitoring code that (reasonably) assumed
`completion_time - submit_time` approximated execution duration.

## Fix

Adds a new `Job.start_time` property, marking when the job actually began
executing (falls back to `submit_time` for backends that don't support a
distinct start time, so this is backwards compatible for all schedulers).

For the Slurm backend, `start_time` is populated from `sacct`'s `Start`
field, added to the existing `poll()` query (previously
`jobid,state,exitcode,end,nodelist`, now
`jobid,state,exitcode,start,end,nodelist`). A new `_update_start_time()`
method mirrors the existing `_update_completion_time()`, using `min()`
instead of `max()` since we want the earliest reported start across
job-array/heterogeneous-job components.

`completion_time - start_time` is now a queue-wait-free execution duration.

Also exposes the new value as a loggable performance variable
(`%(check_job_start_time_unix)s`) via a new `_job_start_time` property in
`pipeline.py`, mirroring the existing `_job_submit_time`/
`_job_completion_time` properties, and documents it in
`docs/config_reference.rst`.

## Testing

- `./test_reframe.py unittests/test_schedulers.py -q` -> 154 passed, 199
  skipped, 0 failed (including all 27 Slurm-specific scheduler tests).
- Full unit test suite (`./test_reframe.py unittests/ -q`) -> 1441 passed,
  280 skipped, 0 failed.

I searched existing issues/PRs and didn't find prior work covering this;
happy to adjust naming/placement if there's a preferred convention I
missed.

### DIFF
--- a/docs/config_reference.rst
+++ b/docs/config_reference.rst
@@ -1326,6 +1326,7 @@ All logging handlers share the following set of common attributes:
       ``%(check_job_completion_time_unix)s``, The completion time of the associated run job (see :attr:`~reframe.core.schedulers.Job.completion_time`).
       ``%(check_job_exitcode)s``, The exit code of the associated run job.
       ``%(check_job_nodelist)s``, The list of nodes that the associated run job has run on.
+      ``%(check_job_start_time_unix)s``, The actual start time of the associated run job, excluding any time spent pending in the scheduler queue (see :attr:`~reframe.core.schedulers.Job.start_time`).
       ``%(check_job_submit_time)s``, The submission time of the associated run job (see :attr:`~reframe.core.schedulers.Job.submit_time`).
       ``%(check_jobid)s``, The ID of the associated run job.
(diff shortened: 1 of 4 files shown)


vkarak:
@raffenet Yes, this is a known issue and I have something already in the works. An updated report where also the start time of the jobs is logged, so the exact elapsed time can also be extracted easily.

PR #3705, opened 2026-07-23
[doc] Fix tutorial example on extracting the unique performance values

sschmittgwdg:
Example in the documentation on single test retrieval was not optimal as the stream test consists of two performance variables.
This PR extends the documentation and also shows an example on how to use the `count` aggregation to ensure that results from a single test run are returned.


vkarak (approved):
I will rebase this to `master` so as to make it to 4.10.2.

PR #3714, opened 2026-08-25
[bugfix] Fix possible DB deadlock when creating indexes

vkarak:
When creating the DB indexes upon every DB file connection, ReFrame did not acquire the write lock as it ought to. This could lead to deadlocks in cases of contentions.

This was introduced in #3546.


victorusu (approved):
`lgtm`

PR #3724, opened 2026-09-09
[bugfix] Enforce strict POSIX names for environment variable names in the configuration

vkarak:
This is a bug introduced in 3.11 by #2479 and it practically allows invalid names to be defined as environment variable names in the various `env_vars` configuration options.


victorusu (approved):
`lgtm`

PR #3727, opened 2026-09-10
[feat] Consume all special keys starting with `$` in the reference dict

### LINKED ISSUE #3721: Ignore any `$` prefixed keys in external references
Currently, ReFrame consumes only the special `$index` key which is used for indexing the remaining references. However, it would be useful if it consumed and ignored all other top-level keys starting with `$`. This would allow other tools that consume the reference file, to add extra keys that are meaningful to them.

vkarak:
As mentioned in the issue, this is useful both for allowing users to add custom annotations to references and for future compatibility.

Closes #3721.


victorusu (approved):
`lgtm`

PR #3732, opened 2026-09-11
[doc] Add a note on limitations of testcase serialization

### LINKED ISSUE #3730: Fail to convert from `None` to `null` when writing JSON report
Consider this situation. I have two tests. `TestA` has a variable, a dict containing `None` objects, both as keys and as values; it is used to define a parameter. `TestB` depends on `TestA`; it calls up the dict from `TestA`, and, as validation, looks for `None` in it. The issue is that if `TestA` _AND_ `TestB` are run _in the same session_, validation passes. If `TestA` is restored from a previous session, validation fails. Indeed, it seems that when restoring, the object of the dict are not converted to `None`, but are kept as `"null"`.

I would not state that's a bug, but indeed it was a bit unexpected. And I agree that using `None` as a key-dictionary might be awkward, still, it is something allowed by python syntax.

### Minimal Working Example

```python
import reframe as rfm
from reframe.core.builtins import run_after, sanity_function, parameter, variable

@rfm.simple_test
    valid_systems = ['*']
    valid_prog_environs = ['*']
    executable = 'true'

    a_dict = variable(dict, value={None: {"nested": None}})
    var = parameter(a_dict)

    @sanity_function
    def validate(self):
        return sn.assert_true(True)

@rfm.simple_test
class TestB(rfm.RunOnlyRegressionTest):
    valid_systems = ['*']
    valid_prog_environs = ['*']
    executable = 'true'

    dep_test_name = "TestA"

    @run_after("init")
    def setup_deps(self):
        self.depends_on(self.dep_test_name)

    @sanity_function
    def validate(self):
        dep_test = self.getdep(self.dep_test_name)
        self.logger.info(f"- Dependent test var: {dep_test.var}")
        self.logger.info(f"- Dependent test dict: {dep_test.a_dict}")
        return sn.assert_true(dep_test.var in dep_test.a_dict)
```

#### Tests run in the same session
```console
$ reframe --report-file dummy_test/report.json -r [...]
[==========] Started on Fri Sep 11 11:45:54 2026+0200

[ RUN      ] TestA %var=None /5f8f53fd @juno:login+fssuite-2026.02.1
[       OK ] (1/2) TestA %var=None /5f8f53fd @juno:login+fssuite-2026.02.1
[ RUN      ] TestB /cc291487 @juno:login+fssuite-2026.02.1
- Dependent test var: None
- Dependent test dict: {None: {'nested': None}}
[       OK ] (2/2) TestB /cc291487 @juno:login+fssuite-2026.02.1

[  PASSED  ] Ran 2/2 test case(s) from 2 check(s) (0 failure(s), 0 expected failure(s), 0 skipped, 0 aborted)
[==========] Finished on Fri Sep 11 11:47:39 2026+0200
```

Extract from report:
```json
{"runs": [ { "testcases": [ {
  "name": "TestA %var=None",
  "var": null,
  "a_dict": { "null": { "nested": null } }
} ] } ] }
```

Now that I have a clear look at it, it seems that the error is already here, since the key of the dict is the string `"null"` and not `null`.

#### Restoring `TestA` from previous session

```console
$ # Forcing only TestB
$ reframe --report-file dummy_test/report-2.json --restore-session dummy_test/report.json -n TestB -r [...]
===] Running 1 check(s)
[==========] Started on Fri Sep 11 11:49:12 2026+0200

[ RUN      ] TestB /cc291487 @juno:login+fssuite-2026.02.1
- Dependent test var: None
- Dependent test dict: {'null': {'nested': None}}
[     FAIL ] (1/1) TestB /cc291487 @juno:login+fssuite-2026.02.1
==> test failed during 'sanity': test staged in '/tmp_user/juno/rmilani/CODA_src/coda-e2e-tests-public/stage/juno/login/fssuite-2026.02.1/TestB'

[  FAILED  ] Ran 1/1 test case(s) from 1 check(s) (1 failure(s), 0 expected failure(s), 0 skipped, 0 aborted)
[==========] Finished on Fri Sep 11 11:50:01 2026+0200
```

Notice as the dict has string `"null"` as key, and not `None` as initially defined in the python.

vkarak:
Closes #3730

### DIFF
--- a/docs/manpage.rst
+++ b/docs/manpage.rst
@@ -731,6 +731,10 @@ Options controlling ReFrame execution
    .. note::
       This option will not work with the :ref:`test generation options <test-generators>`.
 
+   .. note::
+      ReFrame serializes test cases to JSON inside their respective stage directory and restores them from there.
+      As a result, any custom user attributes that are not natively representable in JSON, such as dictionaries with non-string keys, may not be restored correctly.
+
    .. versionadded:: 3.4
 
    .. versionchanged:: 3.6.1


victorusu (approved):
`lgtm`

PR #3733, opened 2026-09-11
[doc] Use the `slurm` backend instead of `squeue` in the tutorial

### LINKED ISSUE #3729: Multi-node tutorial examples could use the `slurm` backend instead of `squeue`
`sacct` is available and working in the container, so the tutorial examples should use the standard `slurm` backend instead of the stripped-down `squeue`.

vkarak:
Closes #3729.


victorusu (approved):
`lgtm`

PR #3736, opened 2026-09-14
[packaging] Build and include the man pages in the final PyPI package

### LINKED ISSUE #3731: ReFrame PyPI package does not contain the man page
```bash
# 1. Install from PyPI and check
uv tool install reframe-hpc
man reframe
# → No manual entry for reframe

find "$(uv tool dir)/reframe-hpc" -iname "*man*"
# → only share/man/man1/{clubak,cluset,clush,nodeset}.1 and
#   share/man/man5/{clush.conf,groups.conf}.5, all from the
#   ClusterShell dependency — nothing from reframe itself

# 2. Inspect the published wheel directly
curl -sLO https://files.pythonhosted.org/packages/f9/b2/ace4d98825d81dc2f5aec07f576dc3970ccc937265b7d01b4036fc8c341f/reframe_hpc-4.10.3-py3-none-any.whl
unzip -l reframe_hpc-4.10.3-py3-none-any.whl | grep man
# → no output

# compare: other shared-data entries whose source IS committed to git
# do make it into the wheel just fine
unzip -l reframe_hpc-4.10.3-py3-none-any.whl | grep -E "bash-completion|fish|tcsh"
# → present

# 3. Inspect the published sdist
curl -sLO https://files.pythonhosted.org/packages/9f/af/cc2e31c3335d235a3c5149a13b9b16fd8cb30014d1730adc9a964d96536d/reframe_hpc-4.10.3.tar.gz
tar tzf reframe_hpc-4.10.3.tar.gz | grep -i docs
# → no output; no docs/ directory at all despite pyproject.toml's
#   sdist include list referencing docs/man/...
```

This is due to the fact the `uv build` does not build the docs. We should likely wire man-page generation into the build itself, e.g. a hatchling build hook (hatch_build.py implementing BuildHookInterface.initialize) that runs sphinx-build -b man docs docs/man (or equivalent) before the sdist/wheel are assembled, so uv build alone is sufficient and the step can't be silently skipped. Also reconcile the reframe.8 vs reframe.settings.8 naming mismatch between the sdist include list and the wheel shared-data mapping.

vkarak:
Closes #3731.


victorusu (approved):
`lgtm`

victorusu (approved):
Much better compared to the previous version!
`lgtm`

